PR #8, opened 2017-10-27
image-copy: first commit

tamirkamara:
@derekbekoe , @coscor-ms please review


src/image-copy/azext_imagecopy/custom.py
```diff
@@ -0,0 +1,135 @@
...
+    logger.warn("Target location count: {0}".format(target_locations_count))
+
+    create_resource_group(target_resource_group_name, target_locations_list[0].strip())
+
+    if parallel_degree == -1:
+        pool = Pool(target_locations_count)
+    elif parallel_degree == 1:
+        pool = Pool(1)
+    else:
+        pool = Pool(min(parallel_degree, target_locations_count))
```
derekbekoe: I don't think you need `elif parallel_degree == 1:`.
min(1, target_locations_count) would be 1 anyway right since there'll be at least 1 location.

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # try to get a handle on arm 409s
+    azure_pool_frequency = 5
+    if target_locations_count >= 5:
+        azure_pool_frequency = 15
+    elif  target_locations_count >= 3:
+        azure_pool_frequency = 10
+
+    tasks = []
+    for location in target_location.split(','):
```
derekbekoe: If you use nargs like I describe elsewhere, this would already be a list so then no need for this.

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -0,0 +1,135 @@
...
+    logger.warn("Getting os disk id of the source vm/image")
+    cmd = prepare_cli_command([source_type, 'show', \
+        '--name', source_object_name, \
+        '--resource-group', source_resource_group_name])
+
+    json_cmd_output = run_cli_command(cmd, True)
+
+    source_os_disk_id = json_cmd_output['storageProfile']['osDisk']['managedDisk']['id']
+    source_os_type = json_cmd_output['storageProfile']['osDisk']['osType']
+    logger.debug("source_os_disk_id: " + source_os_disk_id + " source_os_type: " + source_os_type)
```
derekbekoe: Logger prefers lazy loading of arguments so:
```
logger.debug("source_os_disk_id: %s source_os_type: %s", source_os_disk_id, source_os_type)
```

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -0,0 +1,135 @@
...
+    logger.warn("Starting async process for all locations")
+
+    for task in tasks:
+        pool.apply_async(create_target_image, task)
+
+    try:
+        pool.close()
+        pool.join()
+    except KeyboardInterrupt:
+        print('xxx - parent')
```
derekbekoe: Can remove?

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -0,0 +1,135 @@
...
+        logger.warn('User cancelled the operation')
+        if 'true' in cleanup:
```
derekbekoe: If `cleanup` is defined as `action='store_true'` then this would simply be `if cleanup` with is a lot less likely to break e.g. with letter casing.

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -0,0 +1,135 @@
...
+            logger.warn('To cleanup temporary resources look for ones tagged with "image-copy-extension"')
```
derekbekoe: Would be good if you actually print the command to do this.
```
az resource list --tag ...
```

src/image-copy/setup.py
```diff
@@ -0,0 +1,42 @@
...
+
+setup(
+    name='imagecopyextension',
+    version=VERSION,
+    description='An Azure CLI Extension that copies images from region to region.',
+    long_description='An Azure CLI Extension that copies images from region to region.',
+    license='MIT',
+    author='Tamir Kamara',
+    author_email='tamir.kamara@microsoft.com',
+    url='https://github.com/ORG/REPO',
```
derekbekoe: Make this https://github.com/Azure/azure-cli-extensions

derekbekoe:
@tamirkamara Thanks for the PR!
I added some comments.

src/image-copy/azext_imagecopy/cli_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    full_cmd = [sys.executable, '-m', 'azure.cli'] + cmd
+
+    if output_as_json:
+        full_cmd += ['--output', 'json']
+    else:
+        full_cmd += ['--output', 'tsv']
+
+    # tag newly created resources
+    if 'create' in cmd and ('container' not in cmd):
+        full_cmd += ['--tags', 'created_by=image-copy-extension']
```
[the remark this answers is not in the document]
tamirkamara: I wanted to tag resources created by the extension. Since there're many, it made sense to centralize this process, and I thought it made sense to put it here. However, since blob containers don't have tags I needed to single them out.

src/image-copy/azext_imagecopy/azext_metadata.json
```diff
@@ -0,0 +1,3 @@
+{
+    "azext.minCliVersion": "2.0.12"
+}
```
derekbekoe: You can remove this.
I know it was in the example but not needed in this case.

src/image-copy/azext_imagecopy/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    type: command
+    short-summary: Allows to copy a managed image (or vm) to other regions. Keep in mind that it requires the source disk to be available.
+"""
+
+def load_params(_):
+    with ParametersContext('image copy') as c:
+        c.register('source_resource_group_name', '--source-resource-group', help='Name of the resource group of the source resource')
+        c.register('source_object_name', '--source-object-name', help='The name of the image or vm resource')
+        c.register('target_location', '--target-location', nargs='+', help='Space separated location list to create the image in (use location short codes like westeurope etc.)')
+        c.register('source_type', '--source-type', default='image', help='image or vm')
```
derekbekoe: Add `choices=['image', 'vm']` if those are the only two allowed values.

src/image-copy/azext_imagecopy/cli_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    try:
+        cmd_output = check_output(cmd, stderr=STDOUT, universal_newlines=True)
+        logger.debug('command: %s ended with output: %s', cmd, cmd_output)
+
+        if return_as_json is True:
+            if cmd_output:
+                json_output = json.loads(cmd_output)
+                return json_output
+            else:
+                raise CLIError("Command returned an unexpected empty string.")
```
derekbekoe: Would be good to add which command to this error message?

src/image-copy/azext_imagecopy/cli_utils.py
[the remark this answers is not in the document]
derekbekoe: Okay makes sense.
Consider adding one line comment for this to explain?

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -0,0 +1,131 @@
...
+def create_resource_group(resource_group_name, location):
+    # check if target resource group exists
+    cmd = prepare_cli_command(['group', 'exists', \
+        '--name', resource_group_name], output_as_json=False)
+
+    cmd_output = run_cli_command(cmd)
+
+    if 'false' in cmd_output:
+        # create the target resource group
+        logger.warn("Creating resource group: %s", )
```
derekbekoe: Missing the actual value here it seems

src/image-copy/setup.py
```diff
@@ -0,0 +1,42 @@
...
+    version=VERSION,
+    description='An Azure CLI Extension that copies images from region to region.',
+    long_description='An Azure CLI Extension that copies images from region to region.',
+    license='MIT',
+    author='Tamir Kamara',
+    author_email='tamir.kamara@microsoft.com',
+    url='https://github.com/Azure/azure-cli-extensions',
+    classifiers=CLASSIFIERS,
+    packages=find_packages(),
+    package_data={'azext_imagecopy': ['azext_metadata.json']},
```
derekbekoe: You can remove this.

src/image-copy/azext_imagecopy/cli_utils.py
tamirkamara: There's one later on - this raises an exception in the try-catch block and the catch prints the cmd

PR #28, opened 2018-01-07
Merge Knack-based branch

tjprescott:
Since azure-cli 2.0.24 has been released, the index should be updated for the knack-based extensions.

---

### General Guidelines

### DIFF
--- a/src/index.json
+++ b/src/index.json
@@ -92,6 +92,49 @@
         ],
         "image-copy-extension": [
             {
+                "filename": "image_copy_extension-0.0.4-py2.py3-none-any.whl",
+                "sha256Digest": "47d0f4293e833bb16bac1abbc63f7946dd87ac8db40378a230ec376d1f128fd3",
+                "downloadUrl": "https://files.pythonhosted.org/packages/c3/60/723ad5e968d042d9d4564af420626b661e8cde0820b6dc90f8933e6059bf/image_copy_extension-0.0.4-py2.py3-none-any.whl",
+                "metadata": {
+                    "classifiers": [
+                      "Development Status :: 4 - Beta",
+                      "Intended Audience :: Developers",
+                      "Intended Audience :: System Administrators",
+                      "Programming Language :: Python",
+                      "Programming Language :: Python :: 2",
+                      "Programming Language :: Python :: 2.7",
+                      "Programming Language :: Python :: 3",
+                      "Programming Language :: Python :: 3.4",
+                      "Programming Language :: Python :: 3.5",
+                      "Programming Language :: Python :: 3.6",
+                      "License :: OSI Approved :: MIT License"
+                    ],
+                    "extensions": {
+                      "python.details": {
+                        "contacts": [
+                          {
+                            "email": "tamir.kamara@microsoft.com",
+                            "name": "Tamir Kamara",
+                            "role": "author"
+                          }
+                        ],
+                        "document_names": {
+                          "description": "DESCRIPTION.rst"
+                        },
+                        "project_urls": {
+                          "Home": "https://github.com/Azure/azure-cli-extensions"
+                        }
+                      }
+                    },
+                    "generator": "bdist_wheel (0.30.0)",
+                    "license": "MIT",
+                    "metadata_version": "2.0",
+                    "name": "image-copy-extension",
+                    "summary": "An Azure CLI Extension that copies images from region to region.",
+                    "version": "0.0.4"
+                }
+            },
+            {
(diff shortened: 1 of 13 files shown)


src/index.json
```diff
@@ -37,11 +37,11 @@
         ],
         "azure-cli-iot-ext": [
             {
-                "filename": "azure_cli_iot_ext-0.1.1-py2.py3-none-any.whl",
-                "sha256Digest": "28f5565fa0367da4694223bb8dfacd68be5213e6a49b3a463a10c2379c39da19",
-                "downloadUrl": "https://files.pythonhosted.org/packages/8e/69/6a10aa6be2ad1a054d874d93bd43c09aaca26a137bd0f8f961f03a249a9f/azure_cli_iot_ext-0.1.1-py2.py3-none-any.whl",
+                "filename": "azure_cli_iot_ext-0.3.1-py2.py3-none-any.whl",
+                "sha256Digest": "6932b46e39b5506b460ddc58723d9b2b2223e2a8894d2adda3b0007b4cdef445",
+                "downloadUrl": "https://github.com/Azure/azure-iot-cli-extension/releases/download/v0.3.1/azure_cli_iot_ext-0.3.1-py2.py3-none-any.whl",
```
derekbekoe: @digimaun Do you want to append to the list rather than overwrite the prev. value?
This is a list of extensions then the CLI has logic to filter the list and only install what's supported.
As this stands, once merged, a user on an older version of the CLI would not be able to install the old version of this extension.
If that's intended, it's fine to leave this as is.

src/index.json
```diff
@@ -80,26 +80,23 @@
...
             }
         ],
         "image-copy-extension": [
             {
-                "filename": "image_copy_extension-0.0.4-py2.py3-none-any.whl",
-                "sha256Digest": "47d0f4293e833bb16bac1abbc63f7946dd87ac8db40378a230ec376d1f128fd3",
-                "downloadUrl": "https://files.pythonhosted.org/packages/c3/60/723ad5e968d042d9d4564af420626b661e8cde0820b6dc90f8933e6059bf/image_copy_extension-0.0.4-py2.py3-none-any.whl",
+                "filename": "image_copy_extension-0.0.5-py2.py3-none-any.whl",
+                "sha256Digest": "23d3e197bcf9d1017117c3125dd8b86f7b8357bcae831c17a9e0f45f32f9d62b",
+                "downloadUrl": "https://files.pythonhosted.org/packages/af/d0/7a2d861df60b5d93512ed9341918c614fe6a067f47d47da5063bef7326f6/image_copy_extension-0.0.5-py2.py3-none-any.whl",
```
derekbekoe: @tamirkamara Do you want to append to the list rather than overwrite the prev. value?
This is a list of extensions then the CLI has logic to filter the list and only install what's supported.
As this stands, once merged, a user on an older version of the CLI would not be able to install the old version of this extension.
If that's intended, it's fine to leave this as is.

tamirkamara: @derekbekoe Should I just add another extension with the name name but in the different version?
Also, the previous version (0.0.4) doesn't have a maxcliversion, will it automatically choose the latest compatible version?

derekbekoe: Can you add a new JSON object to the list that starts with `"image-copy-extension": [`?

** If I have the knack based version of the CLI**
The CLI would get the possible candidates for 'image-copy-extension' so in this case 0.0.4 and 0.0.5. Eventually the logic would choose the latest version (0.0.5) which is what we want.

** If you have the non knack based version **
You would still start of with the possible candidates but then 0.0.5 would be filtered out as it's not compatible with the current CLI version so then 0.0.4 would be installed.


source code here: https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-extension/azure/cli/command_modules/extension/_resolve.py#L59

derekbekoe: IF you can modify this PR with that change, please do it. It not, create a PR against knack-based branch with your change then we can resync this PR.

src/index.json
digimaun: It's intended. The original v0.1.1 package should not be used at all anymore.

derekbekoe: Okay thanks.

derekbekoe:
@tamirkamara Let me know which option you want to go with here then we can merge this PR.

src/index.json
tamirkamara: @derekbekoe this PR is to merge 2 branches, and my original one #26 is already closed and merged to the knack-based. 
Not sure if I can do anything other opening a new one?

derekbekoe: The only option would be to create a PR against knack-based that adds back the old entry for the 0.0.4 extension then I can merge it.

PR #41, opened 2018-02-01
AEM: initial change for Azure Enhanced Monitoring Extension for SAP

yugangw-msft:
@derekbekoe, i am doing more testing, at the same time could you please take an initial look and point out major issues if there are?

### General Guidelines

- [x] Have you run `python scripts/ci/test_integration.py -q` locally?


src/aem/azext_aem/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+            g.custom_command('delete', 'delete_aem')
+            g.custom_command('verify', 'verify_aem')
+
+        return self.command_table
+
+    def load_arguments(self, _):
+        from knack.arguments import CLIArgumentType
+        from azure.cli.core.commands.parameters import get_resource_name_completion_list
+        name_arg_type = CLIArgumentType(options_list=['--name', '-n'], metavar='NAME')
+        existing_vm_name = CLIArgumentType(overrides=name_arg_type,
```
yugangw-msft: Would be nice to pick up the [common param](https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/_params.py#L122) from azure-cli-vm, but can't get it work so far.

src/aem/azext_aem/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class AEMCommandsLoader(AzCommandsLoader):
+
+    def __init__(self, cli_ctx=None):
+        from azure.cli.core.commands import CliCommandType
+        aem_custom = CliCommandType(
+            operations_tmpl='azext_aem.custom#{}')
+        super(AEMCommandsLoader, self).__init__(cli_ctx=cli_ctx,
+                                                custom_command_type=aem_custom)
```
derekbekoe: Set min profile here like with other command modules so it doesn't appear on the stack profile.

src/aem/azext_aem/custom.py
```diff
@@ -0,0 +1,546 @@
...
+
+
+class EnhancedMonitoring(object):
+    def __init__(self, cmd, resource_group, vm_name, vm_client,
+                 storage_client, skip_storage_analytics=None):
+        self._vm_client = vm_client
+        self._storage_client = storage_client
+        self._resource_group = resource_group
+        self._cmd = cmd
+        self._vm = vm_client.virtual_machines.get(resource_group, vm_name, expand='instanceView')
```
derekbekoe: If the VM or RG doesn't exist, is this handled appropriately?

src/aem/azext_aem/custom.py
```diff
@@ -0,0 +1,546 @@
...
+                             vm_client=get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_COMPUTE),
+                             storage_client=None)
+    aem.delete()
+
+
+def verify_aem(cmd, resource_group_name, vm_name, wait_time_in_minutes=15, skip_storage_check=False):
+    aem = EnhancedMonitoring(cmd, resource_group_name, vm_name,
+                             vm_client=get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_COMPUTE),
+                             storage_client=get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_STORAGE))
+    aem.verify(skip_storage_check, wait_time_in_minutes)
```
derekbekoe: Should any of the commands return the output from `aem.verify/enable/delete`?
Right now, looks like the commands wouldn't return anything.

src/aem/azext_aem/custom.py
```diff
@@ -0,0 +1,546 @@
...
+        self._skip_storage_analytics = skip_storage_analytics
+
+    def enable(self):
+        pub_cfg, pri_cfg = self._build_extension_cfgs(self._get_disk_info())
+        VirtualMachineExtension = self._cmd.get_models('VirtualMachineExtension')
+        existing_ext = self._get_aem_extension()
+        if existing_ext:
+            extension_instance_name = existing_ext.name
+        else:
+            extension_instance_name = self._extension['name']
```
derekbekoe: nit could be: `extension_instance_name = existing_ext.name if existing_ext else self._extension['name']`?

src/aem/azext_aem/custom.py
```diff
@@ -0,0 +1,546 @@
...
+            'Small': 'Small (A1)',
+            'Medium': 'Medium (A2)',
+            'Large': 'Large (A3)',
+            'ExtraLarge': 'ExtraLarge (A4)'
+        }
+        pub_cfg.update({
+            'vmsize': vm_size_mapping.get(vm_size, vm_size),
+            'vm.role': 'IaaS',
+            'vm.memory.isovercommitted': 0,
+            'vm.cpu.isovercommitted': 1 if vm_size == 'ExtraSmall (A0)' else 0,
```
derekbekoe: Do you also need to check for `vm_size == 'ExtraSmall'`?
Looking at the use of `vm_size_mapping.get(vm_size, vm_size)`, looks like you do need to check for that?

src/aem/azext_aem/custom.py
```diff
@@ -0,0 +1,546 @@
...
+                        'disk.connhour' + suffix: disk['account_name'] + '.hour',
+                    })
+
+        if not managed_disk:
+            unmanaged_disks = [disk_info['os_disk']] + disk_info['data_disks']
+            for disk in unmanaged_disks:
+                account_name = disk['account_name']
+                if disk['is_premium']:
+                    logger.info('[INFO] %s is skipped - Storage Account Metrics are not available '
+                                'for Premium Type Storage.', disk['name'])
```
derekbekoe: Why the `[INFO]` prefix here?
The logger.info would handle whether to add that prefix or not.

src/aem/azext_aem/custom.py
```diff
@@ -0,0 +1,546 @@
...
+                    'caching': data_disk.caching.value,
+                    'lun': data_disk.lun
+                })
+        else:
+            storage_accounts = list(self._storage_client.storage_accounts.list())
+            blob_uri = self._vm.storage_profile.os_disk.vhd.uri
+            parts = list(filter(None, blob_uri.split('/')))
+            storage_account_name = parts[1].split('.')[0]
+            disk_name, container_name = parts[-1], parts[-2]
+            storage_account = [x for x in storage_accounts if x.name.lower() == storage_account_name.lower()][0]
```
derekbekoe: Use `next()` like you do for `existing_ext` above?
I think this would have IndexError if not storage account matched.

src/aem/azext_aem/custom.py
```diff
@@ -0,0 +1,546 @@
...
+                disks_info['os_disk']['size'] = self._get_blob_size(storage_account.name, container_name,
+                                                                    disk_name, key)
+
+            disks_info['data_disks'] = []
+            for data_disk in self._vm.storage_profile.data_disks:
+                blob_uri = data_disk.vhd.uri
+                parts = list(filter(None, blob_uri.split('/')))
+                storage_account_name = parts[1].split('.')[0]
+                disk_name, container_name = parts[-1], parts[-2]
+                storage_account = [x for x in storage_accounts if x.name.lower() == storage_account_name.lower()][0]
```
derekbekoe: Same thing here as well?

src/aem/azext_aem/custom.py
```diff
@@ -0,0 +1,546 @@
...
+                })
+                if is_premium:
+                    disks_info['data_disks'][-1]['size'] = self._get_blob_size(storage_account.name, container_name,
+                                                                               disk_name, key)
+
+        return disks_info
+
+    def _get_blob_size(self, storage_account_name, container, blob, key):
+        storage_client = self._get_storage_client(storage_account_name, key)
+        return int(storage_client.get_blob_properties(container, blob).properties.content_length / (1 << 30))
```
derekbekoe: Maybe just me but what is `(1 << 30)` for?

src/aem/azext_aem/custom.py
```diff
@@ -0,0 +1,546 @@
...
+        seconds = 60 * timeout_in_minutes
+        waited = 0
+        while waited < seconds:
+            entities = table_client.query_entities(table_name, filter_string)
+            if entities.items:
+                return True
+            logger.warning("\t\t\tWait %s seconds for table '%s' has date propagated ...",
+                           15, table_name)
+            time.sleep(15)
+            waited += 15
```
derekbekoe: nit: 15 should be a variable/const?

src/aem/setup.py
```diff
@@ -0,0 +1,42 @@
...
+    version=VERSION,
+    description='CLI Extension to managed Azure Enhanced Monitoring Extension for SAP',
+    long_description='N/A',
+    license='MIT',
+    author='Yugang Wang',
+    author_email='yugangw@microsoft.com',
+    url='https://github.com/Azure/azure-cli-extensions',
+    classifiers=CLASSIFIERS,
+    package_data={'azext_aem': ['azext_metadata.json']},
+    packages=find_packages(),
```
derekbekoe: Add exclude tests like https://github.com/Azure/azure-cli-extensions/blob/master/src/servicebus/setup.py#L39

derekbekoe:
Also, modify the CODEOWNERS file.

derekbekoe:
For this CI error, `Unable to import 'msrestazure.tools' (import-error)`, you may have to modify `ignored-modules` in pylintrc at root of repo and commit that change.

src/aem/azext_aem/custom.py
yugangw-msft: I made it like `all good unless you see an error emitted`, but let me know your reference

src/aem/azext_aem/custom.py
yugangw-msft: CLI is able to capture and output a good error when group or vm doesn't exist. So I don't do more client side validations

src/aem/azext_aem/__init__.py
yugangw-msft: Thanks. I put a comment in the code to do this, but looks like i still forgot it

src/aem/azext_aem/custom.py
derekbekoe: That's fine.

src/aem/azext_aem/custom.py
derekbekoe: cool

PR #48, opened 2018-02-06
Adding a work around  for calling kudu to warm up 

panchagnula:
---

### General Guidelines

### DIFF
--- a/src/webapp/azext_webapp/create_util.py
+++ b/src/webapp/azext_webapp/create_util.py
@@ -83,3 +83,11 @@ def check_if_asp_exists(cmd, rg_name, asp_name):
         if item.name == asp_name:
             return True
     return False
+
+
+def check_app_exists(cmd, rg_name, app_name):
+    client = web_client_factory(cmd.cli_ctx)
+    for item in list(client.web_apps.list_by_resource_group(rg_name)):
+        if item.name == app_name:
+            return True
+    return False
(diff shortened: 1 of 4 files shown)


src/webapp/azext_webapp/__init__.py
```diff
@@ -28,7 +28,7 @@ def load_command_table(self, _):
     def load_arguments(self, _):
         with self.argument_context('webapp new') as c:
             c.argument('name', options_list=['--name', '-n'], help='name of the new webapp')
-            c.argument('dryrun',
+            c.argument('dryrun', options_list=['--dryrun', '-dr'],
```
derekbekoe: The short option should only be a single character.
https://docs.python.org/3/library/argparse.html#option-value-syntax

src/webapp/azext_webapp/create_util.py
```diff
@@ -83,3 +83,11 @@ def check_if_asp_exists(cmd, rg_name, asp_name):
...
             return True
     return False
+
+
+def check_app_exists(cmd, rg_name, app_name):
+    client = web_client_factory(cmd.cli_ctx)
+    for item in list(client.web_apps.list_by_resource_group(rg_name)):
+        if item.name == app_name:
+            return True
+        return False
```
derekbekoe: Looks like the return False is at the wrong indentation level as it's in the for loop.

PR #52, opened 2018-02-08
aem:update index for releasing

yugangw-msft:
### General Guidelines

- [na] Have you run `./scripts/ci/test_static.sh` locally? (`pip install pylint flake8` required)
- [na] Have you run `python scripts/ci/test_integration.py -q` locally?


yugangw-msft:
> Once CI passes, you can merge but sha256Digest should be lower case.

Done. The upper case one was generated by windows ps command.

PR #66, opened 2018-02-21
Add alias as an official Azure CLI extension

chewong:
Add alias as an official Azure CLI extension. Also fixed `src/index.json` as it contains a mix of spaces and tabs, and converted to 4-space indented.

@troydai and @derekbekoe, could you review this PR? Thanks!

---

### General Guidelines

- [x] Have you run `python scripts/ci/test_integration.py -q` locally?


chewong:
Sorry for accidentally tagging @yugangw-msft as a code reviewer

derekbekoe:
@chewong When CI is green, let me know

src/alias-preview/setup.py
```diff
@@ -0,0 +1,45 @@
...
+    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.4',
+    'Programming Language :: Python :: 3.5',
+    'Programming Language :: Python :: 3.6',
+    'License :: OSI Approved :: MIT License',
+]
+
+DEPENDENCIES = [
+    'ddt==1.1.1',
+    'configparser==3.5.0'
```
troydai: use `~=`

src/alias-preview/azext_alias/telemetry/telemetry.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+EXTENSION_NAME = 'azure-cli-alias-extension'
+ALIAS_EXTENSION_PREFIX = 'Context.Default.Extension.Alias.'
+
+logger = get_logger(__name__)
+
+
+# pylint: disable=too-many-instance-attributes
+class AliasExtensionTelemetrySession(object):
```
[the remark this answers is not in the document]
chewong: This is a helper class similar `telemetry.py` in the core that collects telemetry data from the extension. A `AliasExtensionTelemetrySession` represents the duration of the alias transformation.`alias.py` rely on this class to temporarily store telemetry data and send it to the core once the alias transformation is done.

chewong:
@derekbekoe, the CI (finally) turns green. Basically, the dependency of alias (`ddt`) was installed in the temp folder, but we were running the tests against the extension folder, instead of the temp folder, that's why the tests cannot find the `ddt` module initially. I tried different methods of solving this problem and found that the simplest solution is to include `ddt.py` in the alias src folder. Please let me know if you have any concern about this approach.

src/alias/azext_alias/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+import timeit
+
+from knack.log import get_logger
+
+from azure.cli.core import AzCommandsLoader
+from azure.cli.core.commands import CliCommandType
+from azure.cli.core.commands.events import EVENT_INVOKER_PRE_CMD_TBL_TRUNCATE
+from azext_alias.alias import AliasManager
+from azext_alias._const import DEBUG_MSG_WITH_TIMING
+from azext_alias.telemetry import telemetry
```
derekbekoe: Move an many imports here to the actual methods that use them for perf. reasons.

I think only these ones would need to stay at the top.
```
from knack.log import get_logger
from azure.cli.core import AzCommandsLoader
```

src/alias/azext_alias/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+
+logger = get_logger(__name__)
+VERSION = '0.1.0'
```
derekbekoe: I don't like having the same version no. defined both here and in setup.py.

derekbekoe: Consider doing something like what Python SDK does at https://github.com/Azure/azure-sdk-for-python/blob/master/azure-mgmt-cdn/setup.py#L45.

Alternatively, call `from azure.cli.core.extension import get_extension`. Then call it with the extension name then you can get your version that way.

src/alias/azext_alias/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+        super(AliasExtensionLoader, self).__init__(cli_ctx=cli_ctx,
+                                                   min_profile='2017-03-10-profile',
+                                                   custom_command_type=CliCommandType())
+
+        self.cli_ctx.register_event(EVENT_INVOKER_PRE_CMD_TBL_TRUNCATE, alias_event_handler)
+
+    def load_command_table(self, _):  # pylint: disable=no-self-use
+        return {}
+
+    def load_arguments(self, _):
```
derekbekoe: You didn't need the pylint disable for no-self-use here? Interesting.

src/alias/azext_alias/_const.py
```diff
@@ -0,0 +1,27 @@
...
+GLOBAL_CONFIG_DIR = get_config_dir()
+ALIAS_FILE_NAME = 'alias'
+ALIAS_HASH_FILE_NAME = 'alias.sha1'
+COLLIDED_ALIAS_FILE_NAME = 'collided_alias'
+COLLISION_CHECK_LEVEL_DEPTH = 5
+
+PLACEHOLDER_REGEX = r'\s+{\d+}'
+
+INCONSISTENT_INDEXING_ERROR = (
+    'azure-cli-alias-extension: Placeholder indexing should be zero-indexed, '
```
derekbekoe: Hasn't the name of the extension changed?
Same below.

src/alias/azext_alias/alias.py
```diff
@@ -0,0 +1,322 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import os
+import re
+import hashlib
+import json
+from configparser import ConfigParser
```
derekbekoe: Would prefer you use `from six.moves import configparser`.
See https://pythonhosted.org/six/.

The configparser library you used hasn't been updated since 2016 and not sure you really need it either.

src/alias/azext_alias/alias.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def load_collided_alias(self):
+        """
+        Load (create, if not exist) the collided alias file.
+        """
+        # w+ creates the alias config file if it does not exist
+        open_mode = 'r+' if os.path.exists(GLOBAL_COLLIDED_ALIAS_PATH) else 'w+'
+        with open(GLOBAL_COLLIDED_ALIAS_PATH, open_mode) as collided_alias_file:
+            collided_alias_str = collided_alias_file.read()
+            self.collided_alias = json.loads(collided_alias_str if collided_alias_str else '{}')
```
derekbekoe: How do you handle invalid json in `collided_alias_str`?

src/alias/azext_alias/alias.py
```diff
@@ -0,0 +1,322 @@
...
+            # Only care about the first word in the alias because alias
+            # cannot have spaces (unless they have positional arguments)
+            word = alias.split()[0]
+            for level in range(1, levels + 1):
+                collision_regex = r'^{}{}($|\s)'.format(r'([a-z\-]*\s)' * (level - 1), word.lower())
+                if list(filter(re.compile(collision_regex).match, self.reserved_commands)):
+                    if word not in self.collided_alias:
+                        self.collided_alias[word] = []
+                    self.collided_alias[word].append(level)
+        telemetry.set_collided_aliases(list(self.collided_alias.keys()))
```
derekbekoe: How is this useful for telemetry?

src/alias/azext_alias/telemetry/telemetry.py
```diff
@@ -0,0 +1,127 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
```
derekbekoe: Why not just have `src/alias/azext_alias/telemetry.py`?
Seems like you only have one file under `src/alias/azext_alias/telemetry`.

src/alias/azext_alias/telemetry/telemetry.py
```diff
@@ -0,0 +1,127 @@
...
+
+import datetime
+
+from knack.log import get_logger
+
+from azure.cli.core import telemetry as telemetry_core
+from azure.cli.core.telemetry import _remove_cmd_chars, _remove_symbols, _get_stack_trace
```
derekbekoe: Why are we importing private methods?
@troydai will we make them public in the future?

src/alias/azext_alias/telemetry/telemetry.py
```diff
@@ -0,0 +1,127 @@
...
+from azure.cli.core._environment import get_config_dir
+import azext_alias
+
+EXTENSION_NAME = 'azure-cli-alias-extension'
```
derekbekoe: We changed the name right?

src/alias/azext_alias/telemetry/telemetry.py
```diff
@@ -0,0 +1,127 @@
...
+# --------------------------------------------------------------------------------------------
+
+import datetime
+
+from knack.log import get_logger
+
+from azure.cli.core import telemetry as telemetry_core
+from azure.cli.core.telemetry import _remove_cmd_chars, _remove_symbols, _get_stack_trace
+from azure.cli.core._environment import get_config_dir
+import azext_alias
```
derekbekoe: What's this import for?

src/alias/azext_alias/telemetry/telemetry.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @classmethod
+    def set_custom_properties(cls, prop, name, value):
+        if name and value is not None:
+            # 512 characters limit for strings
+            prop['{}{}'.format(ALIAS_EXTENSION_PREFIX, name)] = value[:512] if isinstance(value, str) else value
+
+    @classmethod
+    def get_exception_message(cls, exception):
+        exception_message = str(exception).replace(get_config_dir(), '.azure')
```
derekbekoe: You want to hide the actual users full path to .azure?

src/alias/setup.py
```diff
@@ -0,0 +1,45 @@
...
+    'Programming Language :: Python :: 2.7',
+    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.4',
+    'Programming Language :: Python :: 3.5',
+    'Programming Language :: Python :: 3.6',
+    'License :: OSI Approved :: MIT License',
+]
+
+DEPENDENCIES = [
+    'ddt~=1.1.1',
```
derekbekoe: ddt is only used for tests. Do we want to ship it in the product as well?

src/alias/setup.py
```diff
@@ -0,0 +1,45 @@
...
+
+setup(
+    name='alias',
+    version=VERSION,
+    description='Azure CLI Alias Extension',
+    long_description='An Azure CLI extension that provides command alias functionality',
+    license='MIT',
+    author='Ernest Wong',
+    author_email='t-chwong@microsoft.com',
+    url='https://github.com/chewong/azure-cli-alias-extension',
```
derekbekoe: Change this URL to this repo.

src/alias/azext_alias/telemetry/telemetry.py
chewong: I'll make a copy of these private methods in `telemetry.py` these private methods import.

src/alias/azext_alias/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+
+logger = get_logger(__name__)
+VERSION = '0.1.0'
+
+
+class AliasExtensionLoader(AzCommandsLoader):
+
+    def __init__(self, cli_ctx=None):
+        super(AliasExtensionLoader, self).__init__(cli_ctx=cli_ctx,
+                                                   min_profile='2017-03-10-profile',
```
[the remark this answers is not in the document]
derekbekoe: @chewong You would probably want to remove this.
`min_profile='2017-03-10-profile'` is used so the extension isn't shown in stack profile which wouldn't be a problem here.

.github/CODEOWNERS
```diff
@@ -15,3 +15,5 @@
 /src/subscription/ @wilcobmsft
 
 /src/managementgroups/ @rajshah11
+
+/src/alias/ @chewong
```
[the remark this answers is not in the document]
derekbekoe: I think so!
https://help.github.com/articles/about-codeowners/

src/alias/azext_alias/alias.py
chewong: Just remember why i didn't use `six.moves.configparser`: in python 2, `six.moves.configparser` is not able to detect duplicated sections and options in the config file. That means that users can create duplicated aliases or duplicated command section within the same alias without raising an error.

derekbekoe: What about this?
```
When strict is True (the default), the parser won’t allow for any section or option duplicates while reading from a single source (file, string or dictionary), raising DuplicateSectionError or DuplicateOptionError. 
https://docs.python.org/3/library/configparser.html#configparser.ConfigParser
```

It would only work on Python 3 since I couldn't find that same note on Python 2 but I don't think it's worth adding a 2 year old extension just for that.
If you have duplicate sections, then I assume just the last one would be registered. That seems ok to me.

chewong:
Here are some changes I made to the PR according to the comments:
- Added @troydai as a code owner
- Removed `ddt` and `configparser` dependencies
- Used `six.moves.configparser` instead of `configparser`
- Moved `ddt.py` to `azext_alias/tests` so it won't be included when installing the wheel file
- Moved `telemetry.py` to `azext_alias` folder
- Changed extension name from `azure-cli-alias-extension` to `alias`
- Added `version.py` to store the version of the extension
- Changed the URL of the extension to https://github.com/Azure/azure-cli-extensions

derekbekoe:
@chewong What was wrong with verify_codeowners.py?

derekbekoe:
@chewong Did you try not using ddt?

chewong:
@derekbekoe I made some changes to my [test file](https://github.com/chewong/azure-cli-alias-extension/blob/dev/azext_alias/tests/test_alias.py) without using ddt. Could you take a look and let me know what you think before moving it to this PR?

derekbekoe:
It looks good.
For these kinds of calls, what does 5  mean?
`alias_manager.build_collision_table(5)`
Calling the function with keyword arguments would be clearer. Also, why 5 and not 4?

src/alias/setup.py
```diff
@@ -0,0 +1,47 @@
...
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import os
+import re
+from setuptools import setup, find_packages
+
+# Inspired by https://github.com/Azure/azure-sdk-for-python/blob/master/azure-mgmt-cdn/setup.py#L45
+extension_path = os.path.dirname(os.path.realpath(__file__))
+with open(os.path.join(extension_path, 'azext_alias/version.py'), 'r') as version_file:
```
derekbekoe: Do not hard code the `/` here. Use os.path.join for compatibility with both Linux and Windows.
Something like `os.path.join(extension_path, 'azext_alias', 'version.py')`

PR #78, opened 2018-02-23
updating version on setup

panchagnula:
---

### General Guidelines


src/webapp/setup.py
```diff
@@ -8,7 +8,7 @@
 from codecs import open
 from setuptools import setup, find_packages
 
-VERSION = "0.0.9"
+VERSION = "0.1.0"
```
tjprescott: Should this be 0.1.0 or 0.0.10?

PR #91, opened 2018-02-28
Do not send telemetry if no alias is hit

chewong:
---

### General Guidelines


troydai:
@chewong could you rebase on the master and then we're good to go.

PR #97, opened 2018-03-08
Alias 0.2.0

chewong:
---

### General Guidelines


derekbekoe:
@chewong What are the changes to scripts/ci/test_source.py for?

chewong:
@derekbekoe ran into this issue when I wanted to integrate `ddt` with my tests in the last PR. Basically, when I run test_source.py, my tests were not able to find the dependencies (in this case jinja2). The dependencies are installed in `self.ext_dir/ext`, but we are running the tests against `ext_path`. So the change I made is to install dependencies in `ext_path`

src/alias/azext_alias/_const.py
```diff
@@ -11,11 +11,15 @@
 COLLIDED_ALIAS_FILE_NAME = 'collided_alias'
 COLLISION_CHECK_LEVEL_DEPTH = 4
 
-PLACEHOLDER_REGEX = r'\s+{\d+}'
+POS_ARG_NAME_REGEX = r'{{\s*([a-zA-Z0-9_]+)\s*}}'
+EXPRESSION_REGEX = r'{{\s*([a-zA-Z0-9_]+.*?)\s*}}'
+NUMBER_PLACEHOLDER_REGEX = r'{{\s*\d+\s*}}'
```
troydai: 3 times the regex, 3 times the fun.

src/alias/azext_alias/alias.py
```diff
@@ -210,14 +215,15 @@ def get_full_alias(self, query):
...
+
         return next((section for section in self.alias_table.sections() if section.split()[0] == query), '')
 
     def load_full_command_table(self):
         """
         Perform a full load of the command table to get all the reserved command words.
         """
         load_cmd_tbl_func = self.kwargs.get('load_cmd_tbl_func', None)
-        if load_cmd_tbl_func:
+        if callable(load_cmd_tbl_func):
```
troydai: This worries me a bit. In what situation this variable is not callable? If that's valid scenario it worths consider changing the variable name.

src/alias/azext_alias/argument.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    Get all the positional arguments' names in order.
+
+    Args:
+        arg: The word which this function performs searching on.
+
+    Returns:
+        A list of positional arguments' names.
+    """
+    pos_args_name = list(map(str.strip, re.findall(POS_ARG_NAME_REGEX, arg)))
```
troydai: Overall, the map is discouraged in the situations where a list comprehension can accomplish the same task. 
- https://stackoverflow.com/questions/1247486/python-list-comprehension-vs-map

src/alias/azext_alias/argument.py
```diff
@@ -0,0 +1,165 @@
...
+    Surround placeholders expressions with "" so we can preserve spaces inside the positional arguments.
+
+    Args:
+        arg: The string to process.
+
+    Returns:
+        A processed string where placeholders are surrounded by "" and
+        numbered placeholders are prepended with "_".
+    """
+    number_placeholders = list(map(str.strip, re.findall(NUMBER_PLACEHOLDER_REGEX, arg)))
```
troydai: Same as above, avoid `map` if `list comprehension` can achieve the same thing..

src/alias/azext_alias/argument.py
```diff
@@ -0,0 +1,165 @@
...
+    for number_placeholder in number_placeholders:
+        number = re.search(r'\d+', number_placeholder).group()
+        arg = arg.replace(number_placeholder, '{{_' + number + '}}')
```
troydai: Between the previous method and this, there are two places where the underscore is added to the prefix of a number to avoid the invalid name. I suggest it should be extracted to a method called `normalize_variable_name` or such.

src/alias/azext_alias/argument.py
```diff
@@ -0,0 +1,165 @@
...
+                                                      len(pos_args_placeholder),
+                                                      '' if len(pos_args_placeholder) == 1 else 's',
+                                                      len(pos_args))
+        raise CLIError(error_msg)
+
+    # Escape '"' because we are using "" to surround placeholder expressions
+    for i, pos_arg in enumerate(pos_args):
+        pos_args[i] = pos_arg.replace('"', '\\"')
+
+    return dict(zip(pos_args_placeholder, pos_args))
```
troydai: nice use of `zip`

src/alias/azext_alias/argument.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Args:
+        cmd_derived_from_alias: The string to be injected with positional arguemnts.
+        pos_args_table: The dictionary used to rendered.
+    Returns:
+        A processed string with positional arguments injected.
+    """
+    try:
+        cmd_derived_from_alias = stringify_placeholder_expr(cmd_derived_from_alias)
+        template = Template(cmd_derived_from_alias)
```
troydai: advice: to improve readability, I would have written this line as
```
template = jinja.Template(cmd_derived_from_alias)
```
This form makes it more obvious to the reader that a Jinja template is in play here.

src/alias/azext_alias/argument.py
```diff
@@ -0,0 +1,165 @@
...
+    that there is no runtime error (such as index out of range).
+
+    Args:
+        cmd_derived_from_alias:
+        pos_args_table:
+    """
+    for placeholder, value in pos_args_table.items():
+        exec('{} = "{}"'.format(placeholder, value))  # pylint: disable=exec-used
+
+    expressions = list(map(str.strip, re.findall(EXPRESSION_REGEX, cmd_derived_from_alias)))
```
troydai: Same as above.

src/alias/azext_alias/_const.py
chewong: I've removed all the complex regular expressions above and convert them into actual code/function. Though simple regular expressions are still used in some places to simplify implementation.

src/alias/azext_alias/alias.py
chewong: I don't think load_cmd_tbl_func will ever be non-callable. Converted the above to:
```python
...
load_cmd_tbl_func = self.kwargs.get('load_cmd_tbl_func', None)
if load_cmd_tbl_func is not None:
...
```

src/alias/azext_alias/alias.py
```diff
@@ -210,14 +215,15 @@ def get_full_alias(self, query):
...
+
         return next((section for section in self.alias_table.sections() if section.split()[0] == query), '')
 
     def load_full_command_table(self):
         """
         Perform a full load of the command table to get all the reserved command words.
         """
         load_cmd_tbl_func = self.kwargs.get('load_cmd_tbl_func', None)
-        if load_cmd_tbl_func:
+        if load_cmd_tbl_func is not None:
```
troydai: Between this an previous version, the change is subtle but substantial. Does it mean there is a change that `load_cmd_tbl_func` will be assigned to a value whose _truthiness_ is `False` but `not None`? And that is a positive case?

scripts/ci/test_source.py
```diff
@@ -37,11 +37,14 @@ def __new__(mcs, name, bases, _dict):
...
+                ext_install_dir = os.path.join(self.ext_dir, 'ext')
                 pip_args = [sys.executable, '-m', 'pip', 'install', '--upgrade', '--target',
-                            os.path.join(self.ext_dir, 'ext'), ext_path]
+                            ext_install_dir, ext_path]
                 check_call(pip_args)
                 unittest_args = [sys.executable, '-m', 'unittest', 'discover', '-v', ext_path]
-                check_call(unittest_args)
+                check_call(unittest_args, env={
+                    'PYTHONPATH': '{}{}:'.format(os.environ.get('PYTHONPATH', ''), ext_install_dir)
+                })
```
derekbekoe: I meant more:
```
env = copy.copy(os.environ)
env['PYTHONPATH'] = ext_install_dir
check_call(..., env=env)
```

PR #99, opened 2018-03-09
Automatically create PR to update the list of available extensions doc

derekbekoe:


derekbekoe:
Marked as do not merge until the doc is finalized.

PR #119, opened 2018-03-30
[Subscription RP] create subscription command

wilcobmsft:
---

### General Guidelines

For new extensions:

- [X] My extension description/summary conforms to the [Extension Summary Guidelines](https://github.com/Azure/azure-cli-extensions/blob/master/docs/extension_summary_guidelines.md).


src/subscription/azext_subscription/_params.py
```diff
@@ -8,10 +8,9 @@
...
         c.argument('offer_type', required=True, help='The offer type of the subscription. For example, MS-AZR-0017P (EnterpriseAgreement) and MS-AZR-0148P (EnterpriseAgreement devTest) are available.', arg_type=get_enum_type(['MS-AZR-0017P', 'MS-AZR-0148P']))
-        c.argument('subscription_display_name', help='The subscription display name of the subscription definition.')
-
-    for scope in ['account subscription-definition create', 'account subscription-definition show']:
-        with self.argument_context(scope) as c:
-            c.argument('subscription_definition_name', options_list=['--name', '-n'], help='Name of the subscription definition.')
+        c.argument('display_name', help='The display name of the subscription.')
+        c.argument('object_id', help='The object id(s) of owners which should be granted access to the new subscription.')
+        c.argument('spn', help='The service principal names of owners which should be granted access to the new subscription.')
+        c.argument('upn', help='The user principal names of owners who should be granted access to the new subscription.')
```
derekbekoe: There is no help text for `--enrollment-account-name`

src/subscription/setup.py
```diff
@@ -29,8 +29,8 @@
 setup(
     name='subscription',
     version=VERSION,
-    description='An Azure CLI Extension to manage Subscription Definitions',
-    long_description='An Azure CLI Extension to manage Subscription Definitions',
+    description='An Azure CLI Extension to manage Subscriptions',
+    long_description='An Azure CLI Extension to manage Subscriptions',
```
derekbekoe: The description does not conform to https://github.com/Azure/azure-cli-extensions/blob/master/docs/extension_summary_guidelines.md

Currently, this is the description we created, "Support for subscription definitions preview.". If you're okay with this, please change the description to that or something else.
This 'guideline' was added as now we list extensions at https://docs.microsoft.com/en-us/cli/azure/azure-cli-extensions-list

derekbekoe: Also, if the extension is in preview, follow the instructions at https://github.com/Azure/azure-cli/blob/dev/doc/extensions/metadata.md#azextispreview to mark it as such.

derekbekoe:
@yugangw-msft Can you also take a look since it has some things to do with rbac.

src/subscription/setup.py
wilcobmsft: Fixed

src/subscription/azext_subscription/_params.py
wilcobmsft: Fixed

yugangw-msft: It is fine to be like this since it is an extension command, but if you plan to convert the extension to a core command module, I would suggest fold all 3 into a single `--owners` argument and your command do the hard work by making the graph call to figure out what the real type it is. Per usage feedback, lots of users have a hard time to figure out whose graph objects, nor they have any passions to learn those terms. By just seeing the `spn`, `upn`, they might already get lost.

src/subscription/azext_subscription/custom.py
```diff
@@ -3,16 +3,86 @@
...
-        subscription_display_name=subscription_display_name,
-        offer_type=offer_type)
+def _get_object_id_by_spn(graph_client, spn):
+    accounts = list(graph_client.service_principals.list(
+        filter="servicePrincipalNames/any(c:c eq '{}')".format(spn)))
+    if not accounts:
+        logger.warning("Unable to find user with spn '%s'", spn)
+        return None
+    if len(accounts) > 1:
+        logger.warning("Multiple service principals found with spn '%s'. "
```
yugangw-msft: `spn`/`upn` is supposed to be unique, I would suggest you just error out here as such situation should not occur

wilcobmsft: This code is copied from KeyVault's policy handling. The parameter names (object_id/spn/upn) also reflect what both the KeyVault and AuthZ modules are doing. Ideally there'd be a separate change to address this feedback across all modules. What do you think?

yugangw-msft: Sounds good to me

derekbekoe:
@wilcobmsft To actually publish the new version, update index.json.

Issue #84, opened 2018-02-27
Can't use az image copy


Wolf7horne:
### 'NoneType' object is not subscriptable

Getting os disk id of the source vm/image
'NoneType' object is not subscriptable
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 194, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 342, in execute
    six.reraise(*sys.exc_info())
  File "/opt/az/lib/python3.6/site-packages/six.py", line 693, in reraise
    raise value
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 315, in execute
    result = cmd(params)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 175, in __call__
    return super(AzCliCommand, self).__call__(*args, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/knack/commands.py", line 109, in __call__
    return self.handler(*args, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/__init__.py", line 349, in default_command_handler
    result = op(**command_args)
  File "/home/Wolfthorne/.azure/cliextensions/image-copy-extension/azext_imagecopy/custom.py", line 27, in imagecopy
    source_os_disk_id = json_cmd_output['storageProfile']['osDisk']['managedDisk']['id']
TypeError: 'NoneType' object is not subscriptable

derekbekoe:
@Wolf7horne Are you attempting to use this against Public Azure?

Wolf7horne:
@derekbekoe Yes I am.

tamirkamara:
@Wolf7horne:
1. Can you share your full command?
2. Is it a managed image (an image created from a managed vm/disk)?
3. Does the source disk still exists in your subscription?

Wolf7horne:
@tamirkamara 
1. az image copy --source-resource-group WestUS --source-object-name Goldv9Image --target-resource-group EastUS --targe-location eastus2 --cleanup
2. osDisk and dataDisk both have managedDisk = null, another thing to note is that it is already generalized as well.
3. Yes it does

tamirkamara:
@Wolf7horne 
I think this means that the vm isn't based on managed disks. Can you rerun the command with _--debug_ and send the full output?

[...]

PR #122, opened 2018-04-02
Feature/blob disk support

PoisonousJohn:
---

### General Guidelines

Added support for images based on Blob disks.
I tested on an older version of extension, which is installed via az cli, and it worked.

Could you please review & test it?

This PR addresses #84.

### DIFF
--- a/src/image-copy/azext_imagecopy/custom.py
+++ b/src/image-copy/azext_imagecopy/custom.py
@@ -3,6 +3,8 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
+import datetime
+
 from multiprocessing import Pool
 
 from azext_imagecopy.cli_utils import run_cli_command, prepare_cli_command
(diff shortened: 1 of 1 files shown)


src/image-copy/azext_imagecopy/custom.py
```diff
@@ -127,13 +153,11 @@ def create_resource_group(resource_group_name, location):
...
 
-    if 'true' in cmd_output:
-        return
-
-    # create the target resource group
-    logger.warn("Creating resource group: %s", resource_group_name)
-    cli_cmd = prepare_cli_command(['group', 'create',
-                                   '--name', resource_group_name,
-                                   '--location', location])
+    if 'false' in cmd_output:
```
tamirkamara: Just being curios - why did you change this?

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -23,34 +25,57 @@ def imagecopy(source_resource_group_name, source_object_name, target_location,
...
 
-    source_os_disk_snapshot_url = json_output['accessSas']
-    logger.debug("source os disk snapshot url: %s",
-                 source_os_disk_snapshot_url)
+        json_output = run_cli_command(cli_cmd, return_as_json=True)
+
+        source_os_disk_snapshot_url = json_output['accessSas']
+        logger.debug("source os disk snapshot url: %s", source_os_disk_snapshot_url)
+    else:
+        logger.warn("blob OS disk detected")
```
tamirkamara: logger.warn writes to the console, so I think you should change that and let the user know what is the action performed, something like: Getting sas url for the source blob.

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -23,34 +25,57 @@ def imagecopy(source_resource_group_name, source_object_name, target_location,
...
+        blob_uri = json_cmd_output['storageProfile']['osDisk']['blobUri'].replace('https://', '')
+        logger.warn("blob uri: %s" % blob_uri)
```
tamirkamara: I believe this should be logger.debug or removed. If you decide to keep it I would also include the word "source" to distinct from the blob/snapshots used later on the "target" phase.

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -23,34 +25,57 @@ def imagecopy(source_resource_group_name, source_object_name, target_location,
...
+        blob_account = blob_uri.split('.', 1)[0]
+        blob_uri_components = blob_uri.split('/',)
+        blob_container = blob_uri_components[1]
+        blob_name = blob_uri_components[2]
+        cli_cmd = prepare_cli_command(['storage', 'blob', 'generate-sas',
+                                       '--https-only',
+                                       '--expiry',
+                                       (datetime.datetime.utcnow() +
```
tamirkamara: I'd move the date to an external var (easier to read)

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -23,34 +25,57 @@ def imagecopy(source_resource_group_name, source_object_name, target_location,
...
+                                        datetime.timedelta(days=1))
+                                       .isoformat().split('.', 1)[0]+'Z',
+                                       '--permissions=r',
+                                       '--account-name', blob_account,
+                                       '--container-name', blob_container,
+                                       '-n', blob_name])
```
tamirkamara: Please use full parameter names, i.e. --name (and not -n)

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -23,34 +25,57 @@ def imagecopy(source_resource_group_name, source_object_name, target_location,
...
+        sas = run_cli_command(cli_cmd, return_as_json=False).strip().replace('"', '')
+        full_sas_url = 'https://%s?%s' % (blob_uri, sas)
```
tamirkamara: I think this should be merged with the following line

PoisonousJohn:
@tamirkamara the job is failing due to `R: 18, 0: Too many local variables (29/25) (too-many-locals)`, which I didn't change.

src/image-copy/azext_imagecopy/custom.py
PoisonousJohn: That's my mistake. I applied a patch, based on older version of the script. So it's a regression I would say. I'll revert it back

PoisonousJohn:
@tamirkamara thx for the review. Addressed all the changes you requested

tamirkamara:
This PR addresses a use-case where an image is created from a vhd/blob disk = an un-managed disk. However, an image is managed by nature and it can create managed VMs only, so essentially this "converts" an un-managed disk VM to a managed one.

@PoisonousJohn and I agreed that it looks like this use case isn't very common and It's our believe that the extension should not support that.

@PoisonousJohn - Can you please close this PR?

PR #129, opened 2018-04-05
For tests, get CLI from edge

derekbekoe:


chewong:
I tried to use the edge build to run alias source tests but got the following error: 
image

PR #131, opened 2018-04-11
Alias 0.5.0

chewong:
---

Added `az alias import`, `az alias export`, `az alias remove-all` commands + minor fixes.
A large portion of the PR is just modification on test code.

### General Guidelines

For new extensions:


tjprescott:
With the current core and the current version of Alias, the following fails:
```
az alias create -n del-grp -c "group delete tjp-rg"
alias: Invalid Azure CLI command "group delete tjp-rg"
```

But this succeeds
```
az alias create -n del-grp -c "group delete -g tjp-rg"
```

chewong:
@tjprescott how can I try the positional argument feature? I pulled the latest change from `dev` and ran `python scripts/dev_setup.py`. When I ran `az group delete <resource_group_name>`, I got the following error: 
```az group delete: error: the following arguments are required: --name/-n/--resource-group/-g```

**Comment by chewong** [comment]
@tjprescott is it possible to have two positional arguments?

**Comment by tjprescott** [comment]
@chewong it depends what you mean.  You can only have one argument per command that is positional, but you COULD have that one argument be `nargs='+'` or `nargs='*'` where it could accept multiple things.  So whether you consider that "multiple positional" arguments are not is up to you.

**Comment by chewong** [comment]
@tjprescott I've incorporated alias with your positional argument change in core.

File: src/alias/azext_alias/__init__.py [hunk 1]
```diff
@@ -41,10 +50,17 @@ def __init__(self, cli_ctx=None):
         self.cli_ctx.register_event(EVENT_INTERACTIVE_POST_SUB_TREE_CREATE, enable_aliases_autocomplete_interactive)
 
     def load_command_table(self, _):
+
+        def remove_all_confirmation(_):
+            return bool(prompt_y_n('Are you sure you want to remove all registered aliases?'))
```
derekbekoe: It returns True/False already right?
https://github.com/Microsoft/knack/blob/master/knack/prompting.py#L80-L96

derekbekoe: Also, you don't need this. You can just pass the string as `confirmation='Are you sure you want to remove all registered aliases?'` right?

derekbekoe: https://github.com/Microsoft/knack/blob/master/knack/commands.py#L117

src/alias/azext_alias/custom.py
```diff
@@ -29,6 +36,41 @@ def create_alias(alias_name, alias_command):
...
+        exclusions: Space-separated aliases excluded from export.
+    """
+    alias_table = get_alias_table()
+    for exclusion in exclusions or []:
+        if exclusion not in alias_table.sections():
+            raise CLIError(ALIAS_NOT_FOUND_ERROR.format(exclusion))
+        alias_table.remove_section(exclusion)
+
+    _commit_change(alias_table, export_path=export_path, post_commit=False)
+    print(POST_EXPORT_ALIAS_MSG.format(export_path))  # pylint: disable=superfluous-parens
```
derekbekoe: Should print this to logger.warning instead of using `print()`

PR #142, opened 2018-04-19
Initial PR for SignalR Service preview CLI extension

zackliu:
---

### General Guidelines

For new extensions:


src/signalr/azext_signalr/_client_factory.py
```diff
@@ -0,0 +1,14 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+
+def _signalr_client_factory(cli_ctx, *_):
```
troydai: It seems unnecessary to have this function as the only reference is in `cf_signalr`. These two methods can be merged.

src/signalr/azext_signalr/_params.py
```diff
@@ -0,0 +1,44 @@
...
+                   completer=get_resource_name_completion_list(SIGNALR_RESOURCE_TYPE),
+                   help='Name of signalr service.')
+        c.argument('tags', arg_type=tags_type)
+
+    with self.argument_context('signalr create') as c:
+        c.argument('sku', help='The sku name of the signalr service', choices=SIGNALR_SKU)
+        c.argument('unit_count', help='The number of signalr service unit count', type=int)
+
+    with self.argument_context('signalr key renew') as c:
+        c.argument('key_type', help='The name of access key to regenerate', choices=['primary', 'secondary'])
```
troydai: The key type choices can be moved into `_constants.py` since the value of which is also used in other places.

src/signalr/azext_signalr/custom.py
```diff
@@ -0,0 +1,33 @@
...
+
+    return client.create_or_update(resource_group_name, signalr_name, parameter)
+
+
+def signalr_delete(client, signalr_name, resource_group_name):
+    return client.delete(resource_group_name, signalr_name)
+
+
+def signalr_list(client, resource_group_name=None):
+    if resource_group_name is None:
```
troydai: Can be changed to `if not resource_group_name`. [Truthiness in Python](https://www.udacity.com/wiki/cs258/truthiness-in-python)

src/signalr/azext_signalr/tests/latest/test_signalr_commands.py
```diff
@@ -0,0 +1,80 @@
...
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import time
+import unittest
+from azure.cli.testsdk import ScenarioTest, ResourceGroupPreparer
+
+
+class AzureSignalRServiceScenarioTest(ScenarioTest):
+    @ResourceGroupPreparer(random_name_length=20)
```
troydai: Can I know why is the length of the resource group name is limited to 20 instead of using the default value? Nothing wrong with configuring the value, but I'd like to know the motivation here.

src/signalr/azext_signalr/tests/latest/test_signalr_commands.py
```diff
@@ -0,0 +1,80 @@
...
+                     self.check('name', '{signalr_name}'),
+                     self.check('location', '{location}'),
+                     self.check('provisioningState', 'Succeeded'),
+                     self.check('signalrsku.name', '{sku}'),
+                     self.check('signalrsku.capacity', '{unit_count}'),
+                     self.check('tags.{}'.format(tags_key), tags_val),
+                     self.exists('hostName'),
+                     self.exists('publicPort'),
+                     self.exists('serverPort'),
+
```
troydai: Remove the blank line.

troydai:
@derekbekoe  is the CI failure here related?

src/signalr/azext_signalr/_client_factory.py
derekbekoe: It is the general pattern though used in other modules. Makes it easy to add new client factories later.

src/signalr/azext_signalr/key.py
```diff
@@ -0,0 +1,19 @@
...
+def signalr_key_list(client, resource_group_name, signalr_name):
+    return client.list_keys(resource_group_name, signalr_name)
+
+
+def signalr_key_renew(client, resource_group_name, signalr_name, key_type):
+    if key_type == 'primary':
+        key_type = KeyType.primary
+    else:
+        key_type = KeyType.secondary
+    return client.regenerate_key(resource_group_name, signalr_name, key_type, polling=False)
```
derekbekoe: Change this to:
```
sdk_no_wait(True, client.regenerate_key, resource_group_name, signalr_name, key_type)
```
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/util.py#L251

src/index.json
```diff
@@ -605,6 +605,51 @@
                     "version": "0.0.2"
                 }
             }
+        ],
+        "signalr": [
+            {
+                "filename": "signalr-0.1.0-py2.py3-none-any.whl",
+                "sha256Digest": "b902a43c7d5d07feffe313fec023e62437988f33e31eba933e5ef9e9bea21757",
+                "downloadUrl": "https://azureclisignalr.blob.core.windows.net/cli-extensions/signalr-0.1.0-py2.py3-none-any.whl",
+                "metadata": {
```
derekbekoe: You'll have to add the CLI specific metadata  here also otherwise CI will fail.
```
"azext.minCliCoreVersion": "2.0.31",
"azext.isPreview": true
```

src/signalr/azext_signalr/tests/latest/test_signalr_commands.py
zackliu: The default value (75) is so large that the deployment will fail because the RP seems can only accept the total length within 64.

src/signalr/azext_signalr/key.py
zackliu: If I use sdk_no_wait, the output will be raw. Thus, when I run `key renew`, it will raise error with `TypeError: <urllib3.response.HTTPResponse object at 0x0539B310> is not JSON serializable`

src/signalr/setup.py
```diff
@@ -0,0 +1,38 @@
...
+    name='signalr',
+    version=VERSION,
+    description='Support for signalr management preview.',
+    long_description='Microsoft Azure Command-Line Extensions for SignalR Service',
+    license='MIT',
+    author='Visual Studio China SignalR Team',
+    author_email='vscsignalr@microsoft.com',
+    url='https://github.com/Azure/azure-cli-extensions',
+    classifiers=CLASSIFIERS,
+    package_data={'azext_webapp': ['azext_metadata.json']},
```
derekbekoe: CI is failing because of this. It should be `azext_signalr`.

src/signalr/azext_signalr/tests/latest/test_signalr_commands.py
troydai: Cool.

PR #146, opened 2018-04-25
Fix alias default export path

chewong:
---

### General Guidelines

For new extensions:

### DIFF
--- a/src/alias/azext_alias/_help.py
+++ b/src/alias/azext_alias/_help.py
@@ -43,7 +43,7 @@ helps['alias create'] = """
 
 helps['alias export'] = """
     type: command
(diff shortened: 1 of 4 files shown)


src/alias/azext_alias/_help.py
```diff
@@ -41,7 +43,7 @@
 
 helps['alias export'] = """
     type: command
-    short-summary: Export all registered aliases to a given path, as an INI configuration file.
+    short-summary: Export all registered aliases to a given path, as an INI configuration file. If no export path is specified, the alias configuration file is export to the current working directory.
```
derekbekoe: Should be 'is exported' rather than 'is export'

PR #155, opened 2018-05-01
botservice cli extension

swagatmishra2007:
this changelist contains the following changes -
1) Bot service python sdk generated from swagger
2) Botservice cli function code.


src/botservice/azext_bot/.vscode/tags
```diff
@@ -0,0 +1,478 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
derekbekoe: Does this file need to be checked-in to the repo?
What's it for?

src/botservice/azext_bot/__init__.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+from azure.cli.core import AzCommandsLoader
+from azext_bot._help import helps #pylint: disable=unused-import
+import pdb
```
derekbekoe: Looks like a debug import that should be removed.

src/botservice/azext_bot/_help.py
```diff
@@ -0,0 +1,134 @@
...
+    type: command 
+    short-summary: delete an existing Bot Service.
+"""
+helps['bot update'] = """
+    type: command 
+    short-summary: update an existing Bot Service.
+    examples:
+        - name: Update description on a bot
+          text: |-
+            az bot update -n botname -g resource_group_name --set properties.description="some description"
```
derekbekoe: Prefer `MyResourceGroup` over `resource_group_name`

derekbekoe: Same elsewhere.

src/botservice/azext_bot/custom.py
```diff
@@ -0,0 +1,466 @@
...
+
+def create(cmd, client, resource_group_name, resource_name, kind, description=None, display_name=None,
+           endpoint=None, msa_app_id=None,password=None, tags=None, storageAccountName = None, 
+           location = 'Central US', sku_name = 'F0', appInsightsLocation = 'South Central US', bot_json = None, language='Csharp'):
+    display_name = display_name or resource_name
+    kind = kind.lower()
+
+    if not msa_app_id:
+        msa_app_id, password = provisionConvergedApp(resource_name)
+        print('obtained msa app id and password. Provisioning bot now.')
```
derekbekoe: Use logger.warning instead of print.

src/botservice/azext_bot/custom.py
```diff
@@ -0,0 +1,466 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+from knack.prompting import prompt_y_n
+from knack.util import CLIError
+from knack.log import get_logger
+from azext_bot.botservice import AzureBotService
+from azext_bot.botservice.models import Bot, BotProperties,sku, BotChannel
+from azure.cli.command_modules.appservice.custom import enable_zip_deploy, config_source_control, get_app_settings, _get_site_credential, _get_scm_url
+from azure.cli.command_modules.resource.custom import deploy_arm_template
```
derekbekoe: Depending on other command modules like this is potentially brittle.
The command modules you import have no idea you're using them in this way so this could break the extension in the future.

It's okay to leave it as is since you mentioned you plan to move in to the CLI so it will be less of an issue then but still not recommended.

src/botservice/azext_bot/custom.py
```diff
@@ -0,0 +1,466 @@
...
+    }
+    params = {k: {'value' : v} for k,v in paramsdict.items()}
+
+    import os
+    import json
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    deploy_result = deploy_arm_template(
+        cmd = cmd,
+        resource_group_name = resource_group_name,
+        template_file = '{0}\\{1}'.format(dir_path, template_name),
```
derekbekoe: Will this work on Linux/macOS?

src/botservice/azext_bot/custom.py
```diff
@@ -0,0 +1,466 @@
...
+            name = resource_name,
+            resource_group_name = resource_group_name,
+            repo_url = giturl,
+            branch = git_branch,
+            git_token = git_token
+        )
+
+    #since there is no git url it's definitely publish from local
+    if not code_dir:
+        import os
```
derekbekoe: You can really just import os at the very top of the file instead of importing it so many times.
Same with other standard python imports such as `shutil`

src/botservice/setup.py
```diff
@@ -0,0 +1,53 @@
...
+DEPENDENCIES = [
+]
+
+with open('README.rst', 'r', encoding='utf-8') as f:
+    README = f.read()
+with open('HISTORY.rst', 'r', encoding='utf-8') as f:
+    HISTORY = f.read()
+
+setup(
+    name='azurebotextension',
```
derekbekoe: Change name of extension.
e.g. to just `botservice` since that's the name of the directory in this PR.
Then user would install with `az extension add -n botservice`

src/botservice/azext_bot/.vscode/tags
swagatmishra2007: no it doesn't . i will add my .vscode folder to gitignore so that this doesnt get in by mistake

src/botservice/azext_bot/custom.py
swagatmishra2007: good catch! thanks. .updated the code to move os, shutil and json imports to the top of the file

src/botservice/azext_bot/custom.py
swagatmishra2007: changed to use os.path.join instead of string concatenation. That will make it os independent.

src/botservice/azext_bot/custom.py
swagatmishra2007: done.

src/botservice/azext_bot/custom.py
swagatmishra2007: that is true. longterm, should i talk to the other cli module teams to make sure that they continue to support these functions?

derekbekoe:
@swagatmishra2007 Thanks for the changes. Please take a look at the CI failure for style.

src/botservice/azext_bot/custom.py
tjprescott: In general, your extension/module should not rely on other modules so you shouldn't have these imports at all. You can either copy the functionality into your module or we could move them into the core so that a common implementation is available.

swagatmishra2007: either of the 2 funcitonalities are acceptable to me. i am thinking of making the changes in another PR though since i dont want to break anything this close to build conference. Is it ok to get this checked in and onto the index.json and then taking the dependency off (by copying the code over)?

derekbekoe:
@tjprescott Does it look good to you?

PR #173, opened 2018-05-09
Change codeowner for index.json

derekbekoe:


derekbekoe:
@williexu @tjprescott This is so you get notifications when someone changes the file

PR #188, opened 2018-05-21
iotext v0.4.5

digimaun:
---

### General Guidelines

For new extensions:

- [x] My extension description/summary conforms to the [Extension Summary Guidelines](https://github.com/Azure/azure-cli-extensions/blob/master/docs/extension_summary_guidelines.md).

### DIFF
--- a/src/index.json
+++ b/src/index.json
@@ -38,7 +38,7 @@
         "azure-cli-iot-ext": [
             {
                 "filename": "azure_cli_iot_ext-0.4.5-py2.py3-none-any.whl",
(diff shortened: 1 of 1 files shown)


williexu:
Hi @digimaun, change looks fine. Can I ask why there are two version entries kept for the iot extension?

digimaun:
Thanks @williexu. The initial idea for the two versions was the previous version is kept on the index in case some unforeseen breaking issue or similar event occurred such that down-revving the extension version would be the resolution. Users could see all the details about current 'live' IoT extension versions in one spot.

williexu:
@digimaun gotcha.
In that case, you may want to point your users to the index.json file in your team's docs if you aren't doing so already.
Users will generally only see the latest version from the CLI if they are using the `extension` commands.

PR #202, opened 2018-06-13
fix issue with publishing to bot

swagatmishra2007:


src/botservice/azext_bot/custom.py
```diff
@@ -233,7 +233,7 @@ def create_app(cmd, client, resource_group_name, resource_name, description, kin
         return create_bot_json(cmd, client, resource_group_name, resource_name, app_password=password)
 
 
-def publish_app(cmd, _, resource_group_name, resource_name, git_url=None, git_token=None, git_branch='master',  # pylint: disable=inconsistent-return-statements
+def publish_app(cmd, client, resource_group_name, resource_name, git_url=None, git_token=None, git_branch='master',  # pylint: disable=inconsistent-return-statements
```
williexu: This parameter is not being used in the function. You may want to remove it entirely.

swagatmishra2007: @williexu  i am going to be making change to this code path in my next iteration shortly when i will need the client. So instead of removing this now and readding (and changing the extension package at the source blob again), can we get this in for now?

swagatmishra2007: oh i see why you suggest this..the build is detecting hte unused parameter and failing...i will remove it

PR #207, opened 2018-06-18
DMS - Adding PG and MySQL extension 

temandr:
---

### General Guidelines

For new extensions:

- [x] My extension description/summary conforms to the [Extension Summary Guidelines](https://github.com/Azure/azure-cli-extensions/blob/master/docs/extension_summary_guidelines.md).


src/datamigration/setup.py
```diff
@@ -0,0 +1,40 @@
...
+    'Programming Language :: Python :: 3.4',
+    'Programming Language :: Python :: 3.5',
+    'Programming Language :: Python :: 3.6',
+    'License :: OSI Approved :: MIT License',
+]
+
+DEPENDENCIES = []
+
+setup(
+    name='dms-pg-mysql',
```
williexu: since this is the name that people will see once this is publicly released, you may want to change this to `data-migration` or something similar.

src/datamigration/setup.py
```diff
@@ -0,0 +1,40 @@
...
+    version=VERSION,
+    description='Microsoft Azure Command-Line Tools for the Data Migration Service (DMS) Command Module',
```
williexu: Summary should not include references such as `command-line`, `CLI`, `extension`, etc as per https://github.com/Azure/azure-cli-extensions/blob/master/docs/extension_summary_guidelines.md.
These are all implicitly implied.

src/datamigration/azext_dms/mgmt/__init__.py
```diff
@@ -0,0 +1 @@
+__import__('pkg_resources').declare_namespace(__name__)
```
williexu: Can you put all your sdk code into a `vendored_sdks` folder? i.e. your `mgmt` folder.
This will tell the CI static check to ignore those files for pylint/flake8.

src/datamigration/setup.py
```diff
@@ -0,0 +1,40 @@
...
+    license='MIT',
+    author='Artyom Pavlichenko',
+    author_email='arpavlic@microsoft.com',
+    url='https://github.com/Azure/azure-cli-extensions',
+    classifiers=CLASSIFIERS,
+    packages = find_packages(exclude=["test"]),
```
williexu: Do you have tests?

src/datamigration/setup.py
temandr: I want to specify in the name of the task the feature that we're shipping. This may help in the future when we add other extensions to dms to distinguish different feature sets.

PR #215, opened 2018-06-22
Updating output messages to be printed in default mode

saurabsa:
---

### General Guidelines

For new extensions:

### DIFF
--- a/src/dev-spaces-preview/azext_dev_spaces_preview/custom.py
+++ b/src/dev-spaces-preview/azext_dev_spaces_preview/custom.py
@@ -119,11 +119,10 @@ def _install_dev_spaces_cli(force_install):
 
     if should_install_azds:
         # Install AZDS
-        print('Installing Dev Spaces (Preview) commands...')
+        logger.warning('Installing Dev Spaces (Preview) commands...')
         if system == 'Windows':
-            print('A separate window will open to guide you through the installation process.')
-        else:
-            print('The following dependencies will be installed: curl, unzip, dotnet.')
+            logger.warning('A separate window will open to guide you through the installation process.')
+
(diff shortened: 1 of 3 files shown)


src/dev-spaces-preview/azext_dev_spaces_preview/custom.py
```diff
@@ -119,7 +119,11 @@ def _install_dev_spaces_cli(force_install):
 
     if should_install_azds:
         # Install AZDS
-        logger.info('Installing Dev Spaces (Preview) commands...')
+        print('Installing Dev Spaces (Preview) commands...')
+        if system == 'Windows':
+            print('A separate window will open to guide you through the installation process.')
+        else:
+            print('The following dependencies will be installed: curl, unzip, dotnet.')
```
williexu: Please use the logger instead of the print() method

saurabsa: Yes, that's what. logger.info outputs something when we have the --verbose flag. What if we want to output something by default (even w/o verbose)?

rakeshvanga: Also for the dependencies message, I have already added that line to linux install script. So no need to output that message here.

williexu: In the CLI, we try not to make commands act like scripts, where it is common for developers to output all information to the user and update them on progress.

If you really want to send a message during command execution, use `logger.warning()`.

saurabsa: logger.warning() help then

PR #280, opened 2018-08-31
update  botservice extension to fix bot creation on the native botservice cli

swagatmishra2007:
1) this overrides the az bot create command on the native cli to fix an issue in that code, until the native cli has a fix.


williexu:
@swagatmishra2007 could you make the new update a clone of the CLI module(with the fix) with all the commands? 
That way users using the older versions of the CLI will still be able to get all the functionality currently in the botservice module.

swagatmishra2007:
@williexu  the folks using an earlier version of cli can use the previous version of the extension though right? or is that not available anymore once we publish the new one?

swagatmishra2007:
i updated the index.json to have the previous version of the extension as well. from now on, any fixes and preview features wil be in the extension. the 0.3.0 version would be available for folks using an old cli version.

swagatmishra2007:
@williexu  validation barfs with my change. what key do i use for the different versions of botservice ? seems like the validation wants the key and name of the extension to match, but i want to maintain 2 botservices of the same name, if that's possible.

swagatmishra2007:
@williexu  validation barfs with my change. what key do i use for the different versions of botservice ? seems like the validation wants the key and name of the extension to match, but i want to maintain 2 botservices of the same name, if that's possible.

the main reason i dont want to be copying over the entire module from the native cli is that then i have to maintain the extension if it goes out of sync with the native module.So it's preferable to have 2 versions of the botservice module available.

williexu:
@swagatmishra2007 no, that will add a new extension. 
Do:
```
{
    "botservice":[
         {entry for old version},
         {entry for new version, min_cli_version=2.0.45}
    ],
    other extensions
}

williexu:
Please make CI pass

swagatmishra2007:
@williexu  all good now. i tested the new extension with the latest cli and it owrks as expected. please merge if it looks good

PR #288, opened 2018-09-13
Improved the presentation of find result

albaloo:
---

### General Guidelines

For new extensions:


src/find/azext_find/custom.py
```diff
@@ -25,23 +29,46 @@ def process_query(question):
         num_results_to_show = 1
```
williexu: Is this used anywhere?

src/find/azext_find/custom.py
```diff
@@ -25,23 +29,46 @@ def process_query(question):
...
         if answer_list:
```
williexu: Is it possible to not have an answer list?

src/find/azext_find/custom.py
```diff
@@ -25,23 +29,46 @@ def process_query(question):
...
 def call_aladdin_service(query):
+    context = {
+        'session_id': telemetry_core._session._get_base_properties()['Reserved.SessionId'],  # pylint: disable=W0212
+        'subscription_id': telemetry_core._get_azure_subscription_id(),  # pylint: disable=W0212
+        'client_request_id': telemetry_core._session.application.data['headers']['x-ms-client-request-id'],  # pylint: disable=W0212
+        'installation_id': telemetry_core._get_installation_id()  # pylint: disable=W0212
+    }
+
+    if query:
+        if query.startswith("az "):
```
williexu: combine the two conditionals

src/find/azext_find/custom.py
```diff
@@ -25,23 +29,46 @@ def process_query(question):
...
+            query = '"' + query + '"'
```
williexu: Can you explain this? why the quote if detecting a cli command?

src/index.json
```diff
@@ -972,7 +972,7 @@
                     "metadata_version": "2.0",
                     "name": "find",
                     "summary": "Intelligent querying for CLI information.",
-                    "version": "0.1.0"
+                    "version": "0.2.0"
```
williexu: How do you want to differentiate between a Minor version and a Patch version?

albaloo: I thought this is minor version as the experience has changed more than a patch.

src/find/azext_find/custom.py
albaloo: I can move this to the server side as well, but this is the way the query is treated in the server side. If it has the quotes the exact string will be matched, otherwise it will breakup into pieces. For commands the exact match gives better results.

src/find/azext_find/custom.py
albaloo: done.

src/find/azext_find/custom.py
albaloo: I haven't see it happening, but I had the if, just in case.

williexu: Show an error message in that case. Currently, that scenario will simply end the command with no explanation.

src/find/azext_find/custom.py
albaloo: yes! you are right, fixed it!

src/find/azext_find/custom.py
albaloo: done.

src/find/azext_find/custom.py
```diff
@@ -22,26 +26,50 @@ def process_query(question):
...
+                    current_snippet = current_snippet.replace('```', '').replace(current_title, '').strip()
+                    current_snippet = re.sub(r'\[.*\]', '', current_snippet).strip()
+                    print('\033[1m' + current_title + '\033[0m')
+                    print(current_snippet)
+
+                    if i + 1 < num_results_to_show:
+                        print("\n")
+        else:
+            print("Sorry I am not able to help with that. Try typing the beginning of a"
+                  "command e.g. 'az vm' or explain the task you want to accomplish e.g. 'create a vm' \n")
```
williexu: Please combine this with lines 31-32.
In addition, for error messages, please log a warning/error (Use the logger from knack: [example](https://github.com/Azure/azure-cli/blob/dev/src/command_modules/azure-cli-vm/azure/cli/command_modules/vm/custom.py#L19))

src/find/azext_find/custom.py
```diff
@@ -22,26 +26,50 @@ def process_query(question):
...
 
 
 def call_aladdin_service(query):
+    context = {
+        'session_id': telemetry_core._session._get_base_properties()['Reserved.SessionId'],  # pylint: disable=W0212
+        'subscription_id': telemetry_core._get_azure_subscription_id(),  # pylint: disable=W0212
+        'client_request_id': telemetry_core._session.application.data['headers']['x-ms-client-request-id'],  # pylint: disable=W0212
+        'installation_id': telemetry_core._get_installation_id()  # pylint: disable=W0212
```
williexu: Use `pylint: disable=protected-access` instead for clarity.

src/find/azext_find/custom.py
```diff
@@ -22,26 +26,50 @@ def process_query(question):
...
+    }
+
+    if (query and query.startswith("az ")):
+        query = '"' + query + '"'
+
     service_input = {
-        'paragraphText': "",
+        'paragraphText': "<div id='dummyHeader'></div>",
```
williexu: Quick question, why does the service require this?

albaloo: Because it was initially designed to consume html content.

src/find/azext_find/custom.py
albaloo: done.

adewaleo:
This looks good to me, do you want to approve this? @williexu

williexu:
According to @albaloo, the service is currently down. We will wait for the service to be back up and test the output of the extension before merging.

PR #301, opened 2018-09-19
[Network] Azure Firewall commands

tjprescott:
---

### General Guidelines

For new extensions:


src/azure-firewall/setup.py
```diff
@@ -0,0 +1,41 @@
...
+    'Programming Language :: Python :: 3.4',
+    'Programming Language :: Python :: 3.5',
+    'Programming Language :: Python :: 3.6',
+    'License :: OSI Approved :: MIT License',
+]
+
+DEPENDENCIES = []
+
+setup(
+    name='azure-firewall',
```
williexu: Can we rename to `network-firewall` or `firewall`?

tjprescott: Per our discuss, we will leave it azure-firewall because that is the name of the resource type in ARM.

PR #311, opened 2018-09-24
add botbuilder sdk v4 related features (publish, create). Also update index.json

swagatmishra2007:
add botbuilder sdk v4 related features (publish, create). Also update index.json

### DIFF
--- a/src/botservice/azext_bot/custom.py
+++ b/src/botservice/azext_bot/custom.py
@@ -24,7 +24,7 @@ from azure.cli.core._profile import Profile  # pylint: disable=unused-import
 logger = get_logger(__name__)
 
 
-def get_app_insights_location(key):
+def _get_app_insights_location(key):
     region_map = {
         'australiaeast': 'southeastasia',
         'australiacentral': 'southeastasia',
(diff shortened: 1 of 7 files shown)


src/index.json
```diff
@@ -812,7 +812,7 @@
                     "metadata_version": "2.0",
                     "name": "botservice",
                     "summary": "Bug fixes for issues in the native botservice cli command module.",
```
adewaleo: Shouldn't this summary also include that this extension is used for previews?

src/botservice/azext_bot/custom.py
```diff
@@ -3,15 +3,63 @@
...
+    enable_zip_deploy,
+    get_app_settings,
+    _get_site_credential,
+    _get_scm_url)
 from azure.cli.core._profile import Profile  # pylint: disable=unused-import
 
 logger = get_logger(__name__)
 
 
+def get_app_insights_location(key):
```
adewaleo: It might be good to start helper methods you do not want to export with _'s e.g. _get_app_insights_location

adewaleo:
We strongly recommend adding unittests, would you be able to add some?

src/index.json
swagatmishra2007: so far, we havent included any features that are preview only in this extension. most of the commits go to the native cli. We are making an update to the extension in this case for botbuilder v4 because we were coming in too hot to get onto the native cli release timeline

swagatmishra2007:
agreed on the unit tests...i will add some post the ignite conference

src/botservice/azext_bot/custom.py
swagatmishra2007: added in recent commit.

swagatmishra2007:
> We strongly recommend adding unittests, would you be able to add some?

will get on this post ignite. the current changes have been thoroughly tested. also they will be added into the native cli and removed from teh extension in the next cli release. there will be unit tests for these in the native cli

PR #316, opened 2018-09-25
[Log Analytics] Fix help + examples

alexeldeib:
---

### General Guidelines

For new extensions:

### DIFF
--- a/src/log-analytics/azext_loganalytics/_help.py
+++ b/src/log-analytics/azext_loganalytics/_help.py
@@ -7,6 +7,11 @@ from knack.help_files import helps
 
 # pylint: disable=line-too-long
 
+helps['monitor log-analytics'] = """
+    type: group
+    short-summary: Integrate with Log Analytics workspaces.
+"""
+
 helps['monitor log-analytics query'] = """
     type: command
     short-summary: Query a Log Analytics workspace.
@@ -24,5 +29,7 @@ helps['monitor log-analytics query'] = """
           type: array
           short-summary: Additional workspaces to union data for querying. Specify additional workspace IDs separated by commas.
     examples:
-        - name:
+        - name: Execute a simple query over past 3.5 days.
+          text: |
+            az monitor log-analytics query -w b8317023-66e4-4edc-8a5b-7c002b22f92f --analytics-query "AzureActivity | summarize count() by bin(timestamp, 1h)" -t P3DT12H
 """
--- a/src/log-analytics/azext_loganalytics/_help.py
+++ b/src/log-analytics/azext_loganalytics/_help.py
@@ -9,7 +9,7 @@ from knack.help_files import helps
 
 helps['monitor log-analytics'] = """
     type: group
-    short-summary: Integrate with Log Analytics workspaces.
+    short-summary: Commands for querying data in Log Analytics workspaces.
 """
 
 helps['monitor log-analytics query'] = """
--- a/src/log-analytics/README.rst
+++ b/src/log-analytics/README.rst
@@ -1,2 +1,12 @@
 Commands for working with Azure Log Analytics
 ==============================================
+
+# Basic Usage
+
+Query a Log Analytics workspace for data over the last 12 hours:
+.. code:: 
+    az monitor log-analytics query -w b8317023-66e4-4edc-8a5b-7c002b22f92f --analytics-query -t PT12H
+
+Query multiple workspaces at once:
+.. code::
+    az monitor log-analytics query -w b8317023-66e4-4edc-8a5b-7c002b22f92f f5b6ed22-4942-41c4-becd-1943e5a1b71f


alexeldeib:
@adewaleo Think it should be good. I bumped the package version and link in index. 

image

alexeldeib:
@mmacy FYI, and thanks for pointing this out.

src/log-analytics/azext_loganalytics/_help.py
```diff
@@ -7,22 +7,29 @@
...
-        - analytics-query: --analytics-query
-          type: string
-          short-summary: Query to execute over the Log Analytics data.
-        - timespan: --timespan -t
-          type: string
-          short-summary: Timespan over which to query data. Defaults to all data.
-        - workspaces: --workspaces
-          type: array
-          short-summary: Additional workspaces to union data for querying. Specify additional workspace IDs separated by commas.
+      - name: --workspace
```
adewaleo: No need to add help for the parameters here as they will be gotten from [_params.py](https://github.com/Azure/azure-cli-extensions/blob/90a71cac2d7b6915fa5d73d8f2d7e11699e6c9d7/src/log-analytics/azext_loganalytics/_params.py). It can be beneficial to add help for parameters here if you need to add more content like [long summaries or populator-commands](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_help.md).  

You don't need to remove the help you added here, just let them be consistent with the help in your _params.py file.

src/log-analytics/azext_loganalytics/_help.py
```diff
@@ -7,22 +7,29 @@
...
+        type: string
+        short-summary: GUID of the Log Analytics workspace.
+      - name: --analytics-query
+        type: string
+        short-summary: Query to execute over the Log Analytics data.
+      - name: --timespan
```
adewaleo: I believe the CI test you are failing is because you don't have the correct parameter help names for some of your parameters. In[ _params.py](https://github.com/Azure/azure-cli-extensions/blob/90a71cac2d7b6915fa5d73d8f2d7e11699e6c9d7/src/log-analytics/azext_loganalytics/_params.py), you have:

- your `workspace` parameter has the options `options_list=['--workspace', '-w']`. However, in your help.py your parameter name is `--workspace` instead of `--workspace -w`
- your `timespan` parameter has the options `options_list=['--timespan', '-t']`. However, in your help.py your parameter name is `--timespan` instead of `--timespan -t`

So to solve this problem, you can: 
- use the old help.py yaml parameter names before your change
- or remove the help entry for your **all** your parameters, so the help can be generated from _params.py.

alexeldeib: The old parameter names don't seem to parse any more, seems like some validation changed on the key fields?

Additionally, I [tried using the naming](https://github.com/Azure/azure-cli-extensions/pull/316/commits/65753c4ea35fd5c90585f5555a2b2fbc6d3bde04) you recommend after finding [this issue[(https://github.com/Microsoft/knack/issues/85). It didn't work for me, the validation still failed CI. 

Conveniently, if I remove the parameters which fail name validation from the help file, they autopopulate the parameter help correctly. The remaining parameters pass name validation but the help generation doesn't always work, so I added a string in the help file to populate the --analytics-query parameter properly.

src/log-analytics/azext_loganalytics/_help.py
alexeldeib: This did not work 100% for me. --analytics-query did not load any text from the parameter python file (or elsewhere). Did work for other parameters, though.

alexeldeib:
@adewaleo CI seems happy now 😄 Barring any more feedback, this looks good to me.

src/log-analytics/HISTORY.rst
```diff
@@ -1,12 +1,22 @@
+0.1.4
+++++++++++++++++++
```
adewaleo: Shouldn't there be only one history.rst change?

src/log-analytics/azext_loganalytics/_help.py
```diff
@@ -7,22 +7,20 @@
...
-        - analytics-query: --analytics-query
-          type: string
-          short-summary: Query to execute over the Log Analytics data.
-        - timespan: --timespan -t
-          type: string
-          short-summary: Timespan over which to query data. Defaults to all data.
-        - workspaces: --workspaces
-          type: array
-          short-summary: Additional workspaces to union data for querying. Specify additional workspace IDs separated by commas.
+      - name: --analytics-query
```
adewaleo: Is there a particular reason we have this parameter's help here? If you aren't adding any other attributes to this parameter's help like `long-summary` then it's best to remove this parameter's help as well.

src/log-analytics/setup.py
```diff
@@ -8,7 +8,7 @@
 from codecs import open
 from setuptools import setup, find_packages
 
-VERSION = "0.1.2"
+VERSION = "0.1.4"
```
adewaleo: Shouldn't this be 0.1.3?

src/log-analytics/azext_loganalytics/_help.py
adewaleo: Yes, if help for a parameter is missing in the Yaml file, it would be generated from the help argument in _params.py, as mentioned above.

Hmm that's strange that using the old naming didn't work. 

I am happy though that you were able to resolve this issue.

src/log-analytics/HISTORY.rst
alexeldeib: Yes, PyPi.

alexeldeib: https://pypi.org/manage/project/log-analytics/releases/ see past releases

src/log-analytics/azext_loganalytics/_help.py
adewaleo: Was this before or after adding the help information for this parameter in _params.py. This [line](https://github.com/Azure/azure-cli-extensions/blob/24f3fcd2c3aa81b8a157339fb998fc09c325fec7/src/log-analytics/azext_loganalytics/_params.py#L12) of code in _params.py should provide the help summary. If you add it, you shouldn't need to add this parameter's help information in _help.py. Just make sure the string you pass in for help is the short summary you need.

src/log-analytics/HISTORY.rst
adewaleo: Ah I see. Thanks for explaining. I think next time it is better to have the PR merged in before publishing the whl file on pypi.

src/log-analytics/azext_loganalytics/_help.py
adewaleo: Please try this out and let me know if it works.

alexeldeib: Using the following code (snippets): 

```python
def load_arguments(self, _):
    with self.argument_context('monitor log-analytics query') as c:
        c.argument('workspace', options_list=['--workspace', '-w'], help='GUID of the Log Analytics Workspace')
        c.argument('analytics-query', options_list=['--analytics-query'], help='Query to execute over Log Analytics data.')
        c.argument('timespan', options_list=['--timespan', '-t'], help='Timespan over which to query. Defaults to querying all available data.')
        c.argument('workspaces', nargs='+', help='Additional workspaces to union data for querying. Specify additional workspace IDs separated by space.')
```

```python
helps['monitor log-analytics query'] = """
    type: command
    short-summary: Query a Log Analytics workspace.
    examples:
      - name: Execute a simple query over past 3.5 days.
        text: |
          az monitor log-analytics query -w b8317023-66e4-4edc-8a5b-7c002b22f92f --analytics-query "AzureActivity | summarize count() by bin(timestamp, 1h)" -t P3DT12H
"""
```

I see the following results (note the missing text on the first line):

image

adewaleo: Did you rebuild the whl file and re-add your extension after making these changes?

alexeldeib: Yes, I rebuilt the wheel and added the extension. This is not from a previous version of code. I ran this right now. The commands are cut off in the screenshot, this is the exact sequence I ran.

```powershell
rmdir -Force ./dist
python setup.py bdist_wheel
az extension remove -n log-analytics
az extension add -s ./dist/log_analytics-0.1.4-py2.py3-none-any.whl
az monitor log-analytics query -h
```

alexeldeib: (to be clear, I haven't changed the version again locally which is why it matches the one in PyPi)

adewaleo: Okay, this is strange, let me look into this. I will checkout your pr.

adewaleo: @alexeldeib after a while of debugging. I found the issue. 

TL;DR

your destination command parameter, **analytics-query**, name in _params.py does not match your method parameter name, **analytics_query**, in custom.py.

P.s. you do not need `options_list` if you parameter name is the same as your method name with the hyphens swapped for underscores and without the first two hyphens. 

See [here](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#customize-arguments) for info on customizing command arguments.

======================================================

Long explanation:

so if you look in _params.py you have:

```
...
        c.argument('analytics-query', options_list=['--analytics-query'], help='Query to execute over Log Analytics data.')
                          ^
                          |
dest method parameter name

...
```
whereas in custom.py you have:

```
...
def execute_query(client, workspace, analytics_query, timespan=None, workspaces=None):
                                         ^
                                         |
               method parameter name
...
``` 
Do you see the difference between the variable your are registering the parameter with and the variable you are passing to your method?

When customizing your parameter, you are saying `--analytics-query` maps to some method parameter `analytics-query`, that you want to add help text to. However this function parameter does not exist in your method definition. 

On the other hand you did not explicitly customize the `analytics_query` method param. So it has no special help or options. The cli automatically generates a vanilla command parameter of `--analytics-query`. 

By the way if your destination method parameter name is correct, you do not need to add an option list. The cli would know to convert your method parameter names with underscores to a suitable command parameter name with hyphens.

alexeldeib: Totally missed the hyphen vs. underscore, thank you for helping with the debug! And my apologies for sending you on a lengthy debug. I can make the code change and bump version, or leave as is? Alternatively, I can move storage from PyPi to an Azure blob and stick with 0.1.3 version as it *should* be.

adewaleo: Oh no worries. I missed it too, lol. If you can move your storage to an azure blob, that would be nice. Would the whl file being on pypi affect users of your extension? Would you be able to remove or update the whl file on pypi so as not to confuse your users?

alexeldeib: After I upload to an Azure blob, I'll delete the PyPi project. Should have zero impact.

PR #345, opened 2018-10-17
Rename the find command to explain

albaloo:
---

### General Guidelines

For new extensions:


williexu:
Why are we renaming this?

albaloo:
> Why are we renaming this?

Jason did some in person experiments and concluded that find is confusing. He suggested to rename to explain.

williexu:
@albaloo the problem with renaming is that it's a huge breaking change(multiplied by the fact that the extension only brings about the one command). 
In addition, one of the main discussions for blessing this extension was based on it replacing the current "az find" within the CLI. 
Renaming the extension command nulls that discussion and opens the CLI up for multiple commands (`find` and `explain`) doing arguably similar tasks. This will likely lead to confusion among our users.

PR #369, opened 2018-10-24
mesh : remove bug for current deployment

jeffj6123:
---

### General Guidelines

For new extensions:


src/mesh/azext_mesh/custom.py
```diff
@@ -372,7 +372,7 @@ def _deploy_arm_template_core(cli_ctx, resource_group_name,  # pylint: disable=t
             sleep(timestep)
             wait_time += timestep
 
-        parsed_template = smc.deployments.validate(resource_group_name, deployment_name, properties).properties.additional_properties['validatedResources']
+        parsed_template = validation.properties.additional_properties['validatedResources']
```
williexu: Can you explain why this change is needed?

jeffj6123: Yes, so it appears there has been a change to the validate arm deployment endpoint which will now return an error about previous deployments in progress. The issue was this call could happen during a deployment, and thus return an error about a deployment in progress. Also since the template already gets validated before the deployment starts, this second check wasn't even necessary.

PR #375, opened 2018-10-26
EventGrid: Update index.json to reflect latest version.

kalyanaj:
---
EventGrid: Updated index.json to reflect latest version.

### General Guidelines

For new extensions:

### DIFF
--- a/src/eventgrid/azext_eventgrid/_help.py
+++ b/src/eventgrid/azext_eventgrid/_help.py
@@ -138,7 +138,7 @@ helps['eventgrid domain topic list'] = """
     short-summary: List available topics in a domain.
     examples:
         - name: List all topics in a domain.
-          text: az eventgrid domain list -g rg1 -n domain1
+          text: az eventgrid domain topic list -g rg1 -n domain1
     """
 helps['eventgrid domain topic show'] = """
     type: command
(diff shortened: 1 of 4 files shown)


kalyanaj:
@tjprescott , @williexu: Need your help with this sometime today if possible to review/merge, the only remaining issue is one CI issue which appears to be a false positive. We are looking to release this version out today if possible.

williexu:
@kalyanaj 
- `--ids` parameter is meant to be used for the resource indicated by the command-group. ex:
For the command, `az eventgrid topic delete`, the `--ids` parameter can be used to specify multiple topics(the command-group resource) ids. In `az eventgrid topic key list` the presence of `--ids` to refer to the topic can be confusing because it is not the rightmost "resource" indicated by the command-group. Similarly, `--name` should ideally be renamed to `--topic`.
- In addition, for `az eventgrid topic key list`, what would be the reason for a user to specify multiple ids as the parameter implies? If multiple ids were provided, would you expect all the keys for each topic to be combined and then returned? Again, confusing, hence the rule.

The way to accept ONE id, which i believe is the intention, is to combine the functionality of `--topic` to accept both the topic name and topic resource id.

kalyanaj:
Thanks for the info, that makes sense. What's the best way for me to remove the --ids parameter for these cases?  Taking the example of "domain key list", I have the following in commands.py and I didn't quite get how I can disable support for the --ids parameter?

    with self.command_group('eventgrid domain', domains_mgmt_util, client_factory=domains_factory) as g:
        g.command('key list', 'list_shared_access_keys')

williexu:
This is done in the _params.py file. The following line: https://github.com/Azure/azure-cli-extensions/blob/master/src/eventgrid/azext_eventgrid/_params.py#L66
Adds the `--ids` parameter for all commands under that group. To disable it for the specific command, specify `id_part=None`

kalyanaj:
Thanks @williexu for the information, I have made the changes to remove the --ids parameter from these commands.

PR #379, opened 2018-10-31
Change version to be compatible with Azure DevOps

samontgo:
Azure DevOps will no longer allow me to install this extension when I use the `Azure CLI` task on a hosted agent. The installed version on the hosted agent is `azure-cli (2.0.46)`. Can this file be edited to allow that to work?

---

### General Guidelines


williexu:
@kalyanaj can you comment on why the min cli version was upgraded?
If the goal is to still enable older versions of the CLI, you can consider adding a separate entry in the index.json (one to target newer versions, one to target older). 

@samontgo one solution in the meantime would be to update your version of the CLI.

kalyanaj:
The latest EventGrid extension has a dependency on a bugfix in Knack 0.4.4 and hence we had to update the min CLI core version to 2.0.49.

PR #384, opened 2018-11-02
App Insights data plane metrics, events, query 

alexeldeib:
---

### General Guidelines

For new extensions:


alexeldeib:
Hi! One thing I couldn't figure out, was how to actually run tests? I must have done it before for log-analytics but I'm struggling now? I created a venv after cloning the repo, ran dev_setup.py, and tried invoke `azdev test` with various file/class names corresponding to the test I included. Any insights to what I'm missing?

alexeldeib:
Seems like CI is upset about the metadata configuration containing `azext.isPreview: true`, but I see this in all of the other extensions? I copied the config from another cmdlet, any insights to what is wrong with the index here?

alexeldeib:
Thanks @williexu! I added it to the file, but seems like the same issue?

williexu:
Did you generate a new whl and reupload/update the index?

alexeldeib:
@williexu That was part of it, but I still can't seem to get the index to validate? I tried a few more times, I'm pretty confident I've built it with the metadata, uploaded the new whl file, and synced the hash.

src/application-insights/azext_metadata.json
```diff
@@ -0,0 +1,3 @@
+{
+    "azext.isPreview": true
+}
```
williexu: Your whl file is not in the `azext_applicationinsights` namespace as your setup.py indicates. Please move it.

alexeldeib:
@williexu just bumping this after the weekend -- let me know if I should fix the tests to pass in py 2.7 or if this is good to go. I guess I happened to use a 3+ specific type check.

williexu:
@alexeldeib Yes, please try to fix the CI to pass for 2.7.

src/application-insights/azext_applicationinsights/_params.py
```diff
@@ -0,0 +1,14 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+# pylint: disable=line-too-long
+
+
+def load_arguments(self, _):
+    with self.argument_context('monitor app-insights query') as c:
+        c.argument('application', options_list=['--app-id'], help='GUID of the Application Insights applications')
```
williexu: Are app-insights applications ARM resources? Asking to see if a user can refer to them through a ARM id.

alexeldeib: Yes, it accepts both. Should I accept the fully qualified ARM id here? Or accept the IDs parameter on each command and exclusively accept GUID here?

I also realized the default generation yields `app-id` as parameter name which goes against CLI guidelines; I can change this to `app` for all commands and accept ARM resources.

williexu: Yes, please change to `app` or `application`. Also can you show me what the arm resource path/id for an application would like? 
I think ideally, this parameter should accept both the guid or the resource name (which would be similar to most other commands in the CLI).

alexeldeib: The GUID is accessible via an "API Access" blade from App Insights resource in Azure Portal. I'll include this info. 

Example resource: `/subscriptions/b98981de-4152-480f-a515-59b099299283/resourceGroups/ace-test/providers/microsoft.insights/components/ace-test`

williexu: Yes, please include the information in the help and allow users to specify the application through resource-name/resource-group, etc. 
Take a look at: https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-the-ids-parameter

alexeldeib:
@williexu is there a way to override help generation for only a single parameter? I want to change app-id, event-id, metric-id to app, event, and metric, but these are all non-customized commands. When I change from `g.command()` to `g.custom_command()`, I lose all of the automatic help generation from the SDK. Otherwise, I can copy the help from the existing params into _help.py

alexeldeib:
And also, is there a way to access the cli_ctx inside a custom_command? I'm working around the GUID issue by doing the below on invocation, where cf_components is the management plane SDK. 

I am not sure how to access cli_ctx from the custom.py (the chain between user executing commands and invocation of custom functions is not clear to me). 

```python
def get_id_from_azure_resource(cli_ctx, app):
    if is_valid_resource_id(app):
        resource_group, name = parse_resource_id["resource_group"], parse_resource_id["name"] 
        return cf_components(cli_ctx, _).get(resource_group, name)["app_id"]
    return app
```

alexeldeib:
One more tangential question: `azdev test azext_applicationinsights --live` seems to always output the ouput to my `~/.azure/cliextension`? No related env vars set as far as I can tell. Is this a matter of `pip install`ing the extension elsewhere when I run the tests?

williexu:
@alexeldeib glad you were able to figure it out :)
As for the question about `azdev test`, I'll have to look into that, ideally, the recordings should output in the same folder as where you set `AZURE_EXTENSION_DIR `

src/application-insights/azext_applicationinsights/util.py
```diff
@@ -0,0 +1,15 @@
...
+
+from msrestazure.tools import is_valid_resource_id, parse_resource_id
+from azext_applicationinsights._client_factory import cf_components
+
+
+def get_id_from_azure_resource(cli_ctx, app):
+    if is_valid_resource_id(app):
+        parsed = parse_resource_id(app)
+        resource_group, name = parsed["resource_group"], parsed["name"]
+        return cf_components(cli_ctx, None).get(resource_group, name).app_id
```
williexu: Would customers also want to run CRUD commands on a application?

src/application-insights/azext_applicationinsights/_params.py
```diff
@@ -0,0 +1,39 @@
...
+        c.argument('event_type', options_list=['--type'], help='The type of events to retrieve.')
+        c.argument('event', options_list=['--event'], help='GUID of the event to retrieve. This could be obtained by first listing and filtering events, then selecting an event of interest.')
+        c.argument('timespan', options_list=['--timespan', '-t'], help='The timespan over which to retrieve metric values. This is an ISO8601 time period value. If timespan is omitted, a default time range of `PT12H` ("last 12 hours") is used. The actual timespan that is queried may be adjusted by the server based. In all cases, the actual time span used for the query is included in the response.')
+
+    with self.argument_context('monitor app-insights query') as c:
+        c.argument('application', options_list=['--app, -a'], help='GUID or fully-qualified Azure resource name of Application Insights component. The application GUID may be acquired from the API Access menu item on any Application Insights resource in the Azure portal')
```
williexu: I believe this only accepts the full ARM resource ID(as well as the GUID). 
Please make it so that users can specify the resource-name and group using `--app` and `--resource-group`, which will be consistent with other commands.

Look here: https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-the-ids-parameter
Using `id_part=` will also expose the `--ids` parameter for users to specify the full arm ID.

src/application-insights/azext_applicationinsights/_params.py
```diff
@@ -0,0 +1,39 @@
...
+        c.argument('analytics_query', help='Query to execute over Application Insights data.')
+        c.argument('timespan', options_list=['--timespan', '-t'], help='Timespan over which to query. Defaults to querying all available data.')
+        c.argument('applications', options_list=['--extra-apps'], nargs='+', help='Additional applications over which to union data for querying. Specify additional workspace IDs separated by space.')
```
williexu: I just thought of this, instead of telling users to specify both a single main application as well as a separate list of applications, would it be better to just have one `applications` parameter since the end result is simply an union of all the applications specified?

src/application-insights/azext_applicationinsights/custom.py
```diff
@@ -0,0 +1,33 @@
...
+from knack.log import get_logger
+from .util import get_id_from_azure_resource
+
+logger = get_logger(__name__)
+
+
+def execute_query(cmd, client, application, analytics_query, timespan=None, applications=None):
+    """Executes a query against the provided Application Insights application."""
+    from .vendored_sdks.applicationinsights.models import QueryBody
+    return client.query.execute(get_id_from_azure_resource(cmd.cli_ctx, application), QueryBody(query=analytics_query, timespan=timespan, applications=applications))
```
williexu: The extra applications should also allow for either ARM id or GUID.

src/application-insights/azext_applicationinsights/util.py
alexeldeib: Yes, there's an open issue for this on azure-cli. It's out of scope for this PR however, i'm only covering data plane right now. My team *just* took ownership of the corresponding management plane as part of internal reshuffling. We are committed to fleshing out our support for management plane too, but one step at a time. 

I'll fix remaining comments.

src/application-insights/azext_applicationinsights/_params.py
alexeldeib: How do you want to reconcile this with specifying app + resource group multiple times? I've actually run into a similar issue elsewhere. We support an operation that I would probably call `metrics list` in the CLI, basically a POST version of the GET request behind `metrics show`. The REST API looks like: 
`POST https://api.applicationinsights.io/v1/apps/APP_GUID/metrics`
```json
[
  {
    "id": "failed",
    "parameters": {
      "metricId": "requests/failed",
      "timespan": "P2D",
      "interval": "P1D"
    }
  },
  {
    "id": "count",
    "parameters": {
      "metricId": "requests/count",
      "timespan": "P2D",
      "interval": "P1D"
    }
  }
]
```
In the CLI, how should this look? My difficulty is unpacking the array of objects from a flat command:
```
az monitor app-insights metrics list (--metric requests/failed --label failed -t P2D --interval P1d) (--metric requests/count --label count -t P2D --interval P1d) 
```

We're pushing people away from this and potentially deprecating so I'm not including it, but it's the same kind of issue.

alexeldeib:
For `--app` on query command, here's what I did. A user can provide any of the following:

Single application:
`-a 578f0e27-12e9-4631-bc02-50b965da2633`
`--app 578f0e27-12e9-4631-bc02-50b965da2633`
`--apps ace-test -g ace-test`
`--ids /subscriptions/b98981de-4152-480f-a515-59b099299283/resourceGroups/ace-test/providers/microsoft.insights/components/ace-test`

A list of mixed GUID/Azure resource ID:
`--app 578f0e27-12e9-4631-bc02-50b965da2633 /subscriptions/b98981de-4152-480f-a515-59b099299283/resourceGroups/ace-test/providers/microsoft.insights/components/ace-test`

A list of only Azure resource IDs, via the --ids parameter:
`--ids /subscriptions/b98981de-4152-480f-a515-59b099299283/resourceGroups/ace-test/providers/microsoft.insights/components/ace-test /subscriptions/b98981de-4152-480f-a515-59b099299283/resourceGroups/ace-test/providers/microsoft.insights/components/ace-test`

And lastly, using an RG with a component name restricts you to querying a single resource (no lists):
`--apps ace-test 578f0e27-12e9-4631-bc02-50b965da2633 -g ace-test` --> doesn't work

I removed the separate parameter for extra apps.

alexeldeib:
Ran into a few more minor issues 🙁 This subscription can't be found in CI build, but copy-pasting the failed command works for me. Live tests too: https://travis-ci.org/Azure/azure-cli-extensions/jobs/451772284#L926 
Any tips?
image


And for `events list`, can I disable the `no_ids_for_list_commands` lint rule? I think the use case makes sense? If disabling it is fine, how can I do so? I saw examples in the main CLI repo but couldn't find an extension with the same.

src/application-insights/azext_applicationinsights/_params.py
```diff
@@ -37,7 +38,7 @@ def load_arguments(self, _):
         c.argument('resource_group_name')
 
     with self.argument_context('monitor app-insights query') as c:
-        c.argument('application', options_list=['--app', '--apps', '-a'], nargs='+', id_part='name', help='GUID, app name, or fully-qualified Azure resource name of Application Insights component. The application GUID may be acquired from the API Access menu item on any Application Insights resource in the Azure portal. If using an application name, please specify resource group.')
```
tjprescott: This should just be `--apps`. Argpase will accept `--app` anyways based on how it parses options.

src/application-insights/azext_applicationinsights/_validators.py
```diff
@@ -4,11 +4,10 @@
...
 # pylint: disable=len-as-condition
-from azure.cli.core.commands.client_factory import get_subscription_id
 from knack.util import CLIError
 
 
 def validate_applications(namespace):
-    if namespace.resource_group:
-        if len(apps) != 1:
+    if namespace.resource_group_name:
+        if isinstance(namespace.application, list) and len(namespace.application) != 1:
```
tjprescott: This is a weird convention. The pattern in CLI is that any things referred to only by name are considered from `resource_group`. For GUIDs or ARM IDs, it would not matter. I wouldn't recommend throwing an error for this case.

src/application-insights/azext_applicationinsights/_params.py
```diff
@@ -0,0 +1,44 @@
...
+        c.argument('metric', options_list=['--metrics', '-m'], help='The metric to retrieve. May be either a standard AI metric or an application-specific custom metric.')
+        c.argument('timespan', options_list=['--timespan', '-t'], help='The timespan over which to retrieve metric values. This is an ISO8601 time period value. If timespan is omitted, a default time range of `PT12H` ("last 12 hours") is used. The actual timespan that is queried may be adjusted by the server based. In all cases, the actual time span used for the query is included in the response.')
+        c.argument('aggregation', help='The aggregation to use when computing the metric values. To retrieve more than one aggregation at a time, separate them with a comma. If no aggregation is specified, then the default aggregation for the metric is used.')
```
tjprescott: The CLI only accepts Space-separated lists. If the service requires comma-separated anything, on the CLI-side you need to accept the space-separated list (nargs='+') and join them with commas.

src/application-insights/azext_applicationinsights/_params.py
```diff
@@ -0,0 +1,44 @@
...
+        c.argument('interval', help='The time interval to use when retrieving metric values. This is an ISO8601 duration. If interval is omitted, the metric value is aggregated across the entire timespan. If interval is supplied, the server may adjust the interval to a more appropriate size based on the timespan used for the query. In all cases, the actual interval used for the query is included in the response.')
```
tjprescott: You should look to other `monitor` commands here. Requiring ISO8601 is generally frowned upon as being user-unfriendly.

src/application-insights/azext_applicationinsights/_params.py
```diff
@@ -0,0 +1,44 @@
...
+        c.argument('top', help='The number of segments to return.  This value is only valid when segment is specified.')
+        c.argument('filter_arg', options_list=['--filter'], help=' An expression used to filter the results.  This value should be a valid OData filter expression where the keys of each clause should be applicable dimensions for the metric you are retrieving.')
+        c.argument('resource_group_name')
+
+    with self.argument_context('monitor app-insights metrics get-metadata') as c:
+        c.argument('application', options_list=['--app, -a'], id_part='name', help='GUID, app name, or fully-qualified Azure resource name of Application Insights component. The application GUID may be acquired from the API Access menu item on any Application Insights resource in the Azure portal. If using an application name, please specify resource group.')
+        c.argument('resource_group_name')
```
tjprescott: You don't need to keep duplicating the registration of 'application' for each command. You can simply register them at a single context (probably `monitor app-insights`). Also, I'm not sure you even need to register `resource_group_name` as there's already a global registration for it and you aren't changing it.

src/application-insights/azext_applicationinsights/_params.py
```diff
@@ -0,0 +1,44 @@
...
+
+    with self.argument_context('monitor app-insights events list') as c:
+        c.argument('application', options_list=['--app, -a'], id_part='name', help='GUID, app name, or fully-qualified Azure resource name of Application Insights component. The application GUID may be acquired from the API Access menu item on any Application Insights resource in the Azure portal. If using an application name, please specify resource group.')
+        c.argument('event_type', options_list=['--type'], help='The type of events to retrieve.')
```
tjprescott: What values can this take?

src/application-insights/azext_applicationinsights/tests/latest/test_applicationinsights_commands.py
```diff
@@ -0,0 +1,53 @@
...
+        query_guid = self.cmd('az monitor app-insights query --app 578f0e27-12e9-4631-bc02-50b965da2633 --analytics-query "requests | getschema"').get_output_in_json()
+        query_name_rg = self.cmd('az monitor app-insights query --apps ace-test -g ace-test --analytics-query "requests | getschema"').get_output_in_json()
+        query_azure_id = self.cmd('az monitor app-insights query --analytics-query "requests | getschema" --ids /subscriptions/b98981de-4152-480f-a515-59b099299283/resourceGroups/ace-test/providers/microsoft.insights/components/ace-test').get_output_in_json()
+        assert query_guid == query_name_rg
+        assert query_name_rg == query_azure_id
+        assert len(query_guid['tables'][0]['rows']) == 37
+        assert isinstance(query_guid['tables'][0]['rows'][0][1], (int, float, complex))
+
+    def test_metrics_show(self):
+        result = self.cmd('az monitor app-insights metrics show --app /subscriptions/b98981de-4152-480f-a515-59b099299283/resourceGroups/ace-test/providers/microsoft.insights/components/ace-test -m availabilityResults/count').get_output_in_json()
```
williexu: This will fail the CI as we scrub the subscription id for the recordings. 
Please make this subscription agnostic.

src/application-insights/azext_applicationinsights/_params.py
```diff
@@ -0,0 +1,44 @@
...
+        c.argument('event', options_list=['--event'], help='GUID of the event to retrieve. This could be obtained by first listing and filtering events, then selecting an event of interest.')
+        c.argument('timespan', options_list=['--timespan', '-t'], help='The timespan over which to retrieve metric values. This is an ISO8601 time period value. If timespan is omitted, a default time range of `PT12H` ("last 12 hours") is used. The actual timespan that is queried may be adjusted by the server based. In all cases, the actual time span used for the query is included in the response.')
+        c.argument('resource_group_name')
+
+    with self.argument_context('monitor app-insights query') as c:
+        c.argument('application', validator=validate_applications, options_list=['--app', '--apps', '-a'], nargs='+', id_part='name', help='GUID, app name, or fully-qualified Azure resource name of Application Insights component. The application GUID may be acquired from the API Access menu item on any Application Insights resource in the Azure portal. If using an application name, please specify resource group.')
```
williexu: You can safely ignore the singular application in favor of `apps`. This will make it clear to users that this accepts multiple arguments, whereas the other commands only accept one.

src/application-insights/azext_applicationinsights/_validators.py
alexeldeib: Makes sense. I was trying to avoid users mixing GUID with name/rg. If I understand correctly, querying over app1, app2 and GUID in the same resource group should be possible like this: `az monitor app-insights query -a app1 app2 GUID -g group`  but the GUID would be interpreted as a component name which will naturally fail to resolve as a resource.

src/application-insights/azext_applicationinsights/_params.py
alexeldeib: The functionality as implemented is correct, I'll update the help text.

src/application-insights/azext_applicationinsights/tests/latest/test_applicationinsights_commands.py
alexeldeib: Makes sense. I saw the docs on preparers. Is that the route I should go/would I need to write an application preparer? It seems there were only a handful of preparers other applications were using.

src/application-insights/azext_applicationinsights/_params.py
alexeldeib: I see that get_datetime_type returns the ISO8601 version of the input. What exactly is the input that users can provide for this? Does it support the same range of options, e.g. doing durations? I'm thinking of what the equivalent of this would be using your suggestion. 

`az monitor app-insights metrics show -a ace-test -g ace-test -m requests/count --timespan P7D --interval PT1D --segment resultCode`

src/application-insights/azext_applicationinsights/tests/latest/test_applicationinsights_commands.py
williexu: I don't think so. It seems like your scenarios require app-insights applications, which won't be visible when not logged in with your credentials on your subscriptions.

Usually, for most test scenarios, the tests themselves would create the resources needed. However, since the mgmt-plane commands won't be available yet to create those resources, this won't be possible.

I'd recommend referring to the applications by GUID for now, or, after recording, changing the subscriptions in those arm-ids, to be zeros in your test.

alexeldeib: @harshdeepgupta for recorded tests, we should strip sensitive information manually for now

src/application-insights/azext_applicationinsights/_params.py
alexeldeib: @harshdeepgupta removing "--app" as a recognized flag

williexu:
@alexeldeib what is the status of this PR?

alexeldeib:
@williexu pinged you offline with further details. I'll fix outstanding comments here and merge from master. Also will be working to complete control plane CLI -- let me know if you'd like me to roll that in as an extension or add it to the core module.

src/application-insights/azext_applicationinsights/_params.py
alexeldeib: @williexu could you clarify here?

williexu: @alexeldeib if you take a look at ` az monitor autoscale profile create`, I believe the general pattern for monitor commands is to provide a start/end.

@tjprescott for clarification

src/application-insights/azext_applicationinsights/_params.py
alexeldeib: per another comment, I moved the definition of application parameter to top level and only redefine it on this command to include`--apps`.

src/application-insights/azext_applicationinsights/tests/latest/test_applicationinsights_commands.py
alexeldeib: I've scrubbed these manually, and left the application GUIDs.

src/application-insights/azext_applicationinsights/_params.py
alexeldeib: Added.

src/application-insights/azext_applicationinsights/_params.py
alexeldeib: Fixed.

src/application-insights/azext_applicationinsights/_params.py
alexeldeib: Fixed.

src/application-insights/azext_applicationinsights/_params.py
alexeldeib: @williexu @tjprescott I can do this, but it doesn't actually cover all of the scenarios? We would still need some method of describing intervals for segmentation operations. 

I saw autoscale actually has `format: duration` in the swaggers, which I think is almost sufficient. But I was under the impression this was not well-supported in the SDKs and it didn't seem like it uses the full range in the CLI? It's start, end, and recurrence? 

For `--timespan`, we can say period is not necessary and go with start/end. But for `--interval` i'm not sure the same applies? Could you clarify the desired solution there?

alexeldeib:
@williexu One remaining CI error and pending question about start/end/interval time format.

This CI error seems erroneous? The quotation is inded closed.
```
-  FAIL: faulty_help_example_parameters_rule
    Help-Entry: `monitor app-insights query` - 
	There is a violation:
	"az monitor app-insights query --app e292531c-eb03-4079-9bb0-fe6b56b99f8b --analytics-query 'requests"
	has no closing quotation. Tip: to continue an example command on the next line, use a "\" followed by a newline.
	If needed, you can escape the "\", like so "\\".
```

williexu:
@adewaleo I believe your linter rule caught something it shouldn't have.

adewaleo:
@alexeldeib I should have this fixed by the end of today. Hopefully earlier.

alexeldeib:
@williexu I also played a bit further with get_datetime_type() -- I think I understand how this works, except it seems like % characters are double encoded? Is this bug on our end or in SDK? Not sure what the expected result is here:

```
timespan=2018-02-28T10:30:00-08:00/2018-02-28T10:30:00-08:0 # plaintext
timespan=2018-02-28T10%253A30%253A00-08%253A00%252F2018-02-28T11%253A30%253A00-08%253A00 # cli produces (accepted by other monitor commands)
timespan=2018-02-28T10%3A30%3A00-08%3A00%2F2018-02-28T10%3A35%3A00-08%3A00 (what we accept)
```

I did some testing via other tooling for urlencode/decode to test -- e.g. `:` should encode to `%3A` but it encodes to `%253A` which would be urlencoded `%3A`? Seems like python agrees:

```
>>> urllib.parse.quote(':',safe='')
'%3A'
```

And here's the code that does that: https://github.com/Azure/msrest-for-python/blob/master/msrest/serialization.py#L628

Am I missing something? Other commands seem to have used this without issue

adewaleo:
@alexeldeib I am still on the linter issue, but should be done soon (one or two working days). Here is the [PR](https://github.com/Azure/azure-cli/pull/8608).

adewaleo:
@alexeldeib, @williexu, the issue has been fixed.

williexu:
@alexeldeib in response to your datetime question:

Unless I'm missing something, `get_datetime_type()` doesn't do any url query serialization, it simply takes your datetime string and returns a iso-formatted string: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/commands/parameters.py#L37-L85

In regards to @tjprescott 's comment before, it would be nice if your commands followed:
```
Time Arguments
    --end-time            : End time of the query. Defaults to the current time. Format: date (yyyy-
                            mm-dd) time (hh:mm:ss.xxxxx) timezone (+/-hh:mm).
    --interval            : The interval over which to aggregate metrics, in ##h##m format.
                            Default: 1m.
    --offset              : Time offset of the query range, in ##d##h format.  Default: 1h.
        Can be used with either --start-time or --end-time. If used with --start-time, then the end
        time will be calculated by adding the offset. If used with --end-time (default), then the
        start time will be calculated by subtracting the offset. If --start-time and --end-time are
        provided, then --offset will be ignored.
    --start-time          : Start time of the query. Format: date (yyyy-mm-dd) time (hh:mm:ss.xxxxx)
                            timezone (+/-hh:mm).
```

(Taken from `az monitor metrics list`)

alexeldeib:
@williexu Found the issue, I copied monitor metrics a [little too closely (quote_plus)](https://github.com/Azure/azure-cli/blob/2138265fdf5a6c08bc53030270e8654eaa858b12/src/command_modules/azure-cli-monitor/azure/cli/command_modules/monitor/custom.py#L127).

That solved my issue. I did also notice, the time parsing does not seem to accept negative TZ offsets, e.g. I think these should be the same, but aren't:
```
--start-time 2019-03-05 15:00:00        --end-time 2019-03-05 15:35:00        # when executed from local TZ -08:00 (Redmond)
--start-time 2019-03-05 23:00:00 -08:00 --end-time 2019-03-05 23:35:00 -08:00 # manually provide offset from UTC
```
I was thinking [this line](https://github.com/Azure/azure-cli/blob/2138265fdf5a6c08bc53030270e8654eaa858b12/src/azure-cli-core/azure/cli/core/commands/parameters.py#L61) should be `parser.isoparse` rather than `parser.parse`, but not 100%. 

Anyway...I think that's everything

williexu:
@alexeldeib 
the timezone offsets look fine to me. the code essentially does:
```
>>> dateutil.parser.parse("2019-03-05 15:00:00 -8:00").isoformat()
'2019-03-05T15:00:00-08:00'
>>> dateutil.parser.parse("2019-03-05 15:00:00 +8:00").isoformat()
'2019-03-05T15:00:00+08:00'
```

williexu:
@tjprescott can you re-review?

tjprescott:
Thanks for the hard work @alexeldeib!

PR #422, opened 2018-11-15
 Implemented MongoDB to Cosmos DB commands

vchske:
---

### General Guidelines

For new extensions:


vchske:
I have a unit test file I would like to include as soon as I get a problem with the recording file worked out. Until then, however, I'd like the team to start reviewing the new commands.

vchske:
Can someone help me identify the problem with the 'IndexRefDocVerify' step in Travis? I'm not sure what the log is telling me but it looks worse than just the code I'm trying to check in.

williexu:
@vchske I've managed to fix the CI by updating your branch.

src/dms-preview/azext_dms/_help.py
```diff
@@ -77,6 +79,10 @@
           type: bool
           short-summary: >
             (For SQL only) Whether to compare the schema information between source and target.
+        - name: --validate-only
+          type: bool
+          short-summary: >
+            (For MongoDB to Cosmos DB only) Whether to run validation only and NOT run migration. This is usefull for verifying connections, shard keys, RU settings, etc.
```
williexu: `usefull` -> `useful`

src/dms-preview/azext_dms/_help.py
```diff
@@ -138,6 +201,32 @@
...
+            {
+                "userName": null,   // if this is missing or null, you will be prompted
+                "password": null,   // if this is missing or null (highly recommended) you will be prompted
+                "connectionString": "mongodb://hostOrIp:port"
+            }
+"""
+
+helps['dms project task cancel'] = """
+    type: command
+    short-summary: This command is being deprecated. Use the stop command instead.
```
williexu: Isn't this command being added with this PR? If it's deprecated, why include it?

src/dms-preview/azext_dms/commands.py
```diff
@@ -25,5 +25,8 @@ def load_command_table(self, _):
         g.custom_command('create', 'create_or_update_project')
 
     with self.command_group('dms project task', dms_tasks_sdk, client_factory=dms_cf_tasks) as g:
+        g.custom_command('cancel', 'stop_task')
```
williexu: Seems like `cancel` is unnecesary

src/dms-preview/azext_dms/_help.py
vchske: The "cancel" command exists in the main CLI module. We've had a change of nomenclature to bring a more unified experience among UI, PowerShell, and CLI. I thought until this set of commands goes GA (i.e. copied out of the extension and into the main module) I would offer this message for anyone who might see it.

src/dms-preview/azext_dms/commands.py
vchske: See my response to the other comment about the "cancel" command.

src/dms-preview/azext_dms/_help.py
vchske: Thanks

src/dms-preview/azext_dms/_help.py
```diff
@@ -39,6 +40,7 @@
             1) SQL -> SQLDB :: OfflineMigration
             2) MySQL -> AzureDbForMySql :: OnlineMigration
             3) PostgreSQL -> AzureDbForPostgreSQL :: OnlineMigration
+            4) MongoDB -> Cosmos DB (via MongoDB API) :: OfflineMigration
```
temandr: Can you replace "Cosmos DB" with the target type? Eg, PostgreSQL -> AzureDbForPostgreSQL means that the source type is PostgreSQL and target type is AzureDbForPostgreSQL.

src/dms-preview/azext_dms/_help.py
```diff
@@ -48,23 +50,23 @@
         - name: --source-platform
           type: string
           short-summary: >
-            The type of server for the source database. The supported types are: SQL, MySQL, PostgreSQL.
+            The type of server for the source database. The supported types are: SQL, MySQL, PostgreSQL, MongoDB.
         - name: --target-platform
           type: string
           short-summary: >
-            The type of server for the target database. The supported types are: SQLDB, AzureDbForMySql, AzureDbForPostgreSQL.
+            The type of server for the target database. The supported types are: SQLDB, AzureDbForMySql, AzureDbForPostgreSQL, MongoDB.
```
temandr: Should "MongoDB" be a "CosmosDB"?

src/dms-preview/azext_dms/custom.py
```diff
@@ -33,15 +45,16 @@ def create_or_update_project(
     # Set inputs to lowercase
     source_platform = source_platform.lower()
     target_platform = target_platform.lower()
+    scenario_handled_in_core = core_handles_scenario(source_platform, target_platform)
 
     # Validation: Test scenario eligibility
-    if not determine_source_target_eligibility(source_platform, target_platform):
+    if scenario_handled_in_core:
```
temandr: Instead of checking what scenarios core supports, can we check for what scenarios the extension supports? This way we don't have to keep the two connected but purely keep any scenario supported in extension independent from the ones in core.

src/dms-preview/azext_dms/custom.py
```diff
@@ -162,42 +170,150 @@ def create_task(
...
+        task_name,
+        task_type,
+        object_name=None):
+    # For scenarios that support it, restart the entire migration if object name is empty,
+    # otherwise restart the specified object.
+    src, tgt = get_project_platforms(cmd,
+                                     project_name=project_name,
+                                     service_name=service_name,
+                                     resource_group_name=resource_group_name)
+    st = get_scenario_type(src, tgt, task_type)
```
temandr: Do we need to prompt the user the task type? Can we just fail if the user uses a scenario that isn't supported?

src/dms-preview/azext_dms/_help.py
```diff
@@ -111,6 +117,63 @@
...
+                // Set to null to use maximum resources available.
+                "throttling": {
+                    // percentage of the CPU to try to avoid using
+                    "minFreeCpu": 25,
+                    // amount of RAM (in MBs) to try to avoid using
+                    "minFreeMemoryMb": 1024,
+                    // max number of collections to migrate at a time
+                    "maxParallelism": 2
+                },
+                "databases": {
```
temandr: Should this be an array?

src/dms-preview/azext_dms/_help.py
vchske: Ok. But some extra text should be added somehow. The scenario is to enable migrating from MongoDB to Cosmos DB. But the actual type of both the source and the target is MongoDB, since we connect to a Comsos DB instance via their MongoDB API. I was trying to make that clear with this. If a user just saw "MongoDB -> MongoDB" they might not understand what's going on and what the intention of the scenario is. Your point is valid, I'll fiddle with the wording.

src/dms-preview/azext_dms/_help.py
vchske: Nope. See my other reply to your comment. The type for both source and target are MongoDB.

src/dms-preview/azext_dms/custom.py
vchske: MongoDB migrations will allow restarting both online and offline migrations. This might not be the case with other scenarios. Also, right now, the extension doesn't support online MongoDB migrations. So I think this needs to be left there. Unless we can come up with a work-around to using task type in general.

vchske: Thinking about this more, you might be right. The task is already created and the different combinations are filtered out there already. I think I will remove task type here as well. If it ends up that we need it later it should be easy to add back.

src/dms-preview/azext_dms/custom.py
```diff
@@ -162,42 +170,150 @@ def create_task(
...
+
 
+def stop_task(
+        cmd,
+        client,
+        resource_group_name,
+        service_name,
+        project_name,
+        task_name,
+        task_type,
```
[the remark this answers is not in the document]
vchske: You're right about this.

src/dms-preview/setup.py
```diff
@@ -8,7 +8,7 @@
 from codecs import open
 from setuptools import setup, find_packages
 
-VERSION = "0.5.0"
+VERSION = "0.5.1"
```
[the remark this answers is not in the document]
vchske: Ok. Though I'm not sure I would know the difference between major and minor version bumps in this case. It's all just preview.

src/dms-preview/azext_dms/_help.py
vchske: No. These are implemented as key/value pairs.

src/dms-preview/azext_dms/custom.py
```diff
@@ -69,90 +82,85 @@ def create_or_update_project(
                                    group_name=resource_group_name,
                                    service_name=service_name,
                                    project_name=project_name)
+
+
+def get_project_platforms(cmd, project_name, service_name, resource_group_name):
```
temandr: Can you move get_project_platforms down to where the Help methods region is?

src/dms-preview/azext_dms/custom.py
```diff
@@ -69,90 +82,85 @@ def create_or_update_project(
...
+                                        source_connection_json,
+                                        target_connection_json,
+                                        database_options_json,
+                                        enable_schema_validation,
+                                        enable_data_integrity_validation,
+                                        enable_query_analysis_validation)
         except:
+            # TODO: We currently don't have any CLI core code to perform any validations
+            # because of this we need to raise the error here.
             raise ValueError("The provided source-platform, target-platform, and task-type \
```
temandr: Since we removed the source/target platforms, I think we can remove this line. (Since the user didn't provide them)

src/dms-preview/azext_dms/custom.py
```diff
@@ -69,90 +82,85 @@ def create_or_update_project(
...
         enable_schema_validation=False,
         enable_data_integrity_validation=False,
-        enable_query_analysis_validation=False):
-
-    # Set inputs to lowercase
+        enable_query_analysis_validation=False,
+        validate_only=False):
+
+    # Get source and target platform abd set inputs to lowercase
+    src, tgt = get_project_platforms(cmd,
```
temandr: To be consistent with other naming conventions present, can you rename src and tgt to source_platform and target_platform, respectively?

src/dms-preview/azext_dms/custom.py
temandr: Do the same for projects as well.

vchske: Well the user does provide source and target platforms for a project so I'm going to leave that alone. As for here, I think it would be appropriate to change the wording to something like "The source and target platforms of the parent project are not valid..." something.

src/dms-preview/azext_dms/custom.py
vchske: Done.

src/dms-preview/azext_dms/custom.py
vchske: Done.

src/dms-preview/azext_dms/_help.py
williexu: I see, in that case, please remember that when these commands go GA and into the CLI module, you will still need to deprecate the commands from the CLI side for some time to not break users who have not used the extension.

PR #471, opened 2019-01-14
Removing sfmergeutility src and adding dependency version on mesh extension

vshan:
- Sfmergeutility src has been moved to Microsoft/sfmergeutility and hence is being removed from this repo.
- Adding version to mesh dependency to prevent breakage when there is API breaking change

---

### General Guidelines

For new extensions:


src/mesh/setup.py
```diff
@@ -26,7 +26,7 @@
 ]
 
 DEPENDENCIES = [
-    'sfmergeutility'
+    'sfmergeutility==0.1.5'
```
jeffj6123: Add a line in the the readme file about locking the version of this and bump the patch version.

PR #507, opened 2019-02-07
DMS - Fixing exceptions and a couple of bugs

temandr:
---

### General Guidelines

For new extensions:


src/dms-preview/azext_dms/custom.py
```diff
@@ -54,8 +55,9 @@ def create_or_update_project(
         # because of this we need to raise the error here.
         try:
             # TODO: Remove this check after validations are added to core
-            if scenario_handled_in_extension:
-                raise ValueError
+            if source_platform != "sql" or target_platform != "sqldb":
```
vchske: I think this should be a method as it's used in more than one place. If possible, maybe extension_handles_scenario could be modified?

src/dms-preview/azext_dms/custom.py
```diff
@@ -438,11 +454,10 @@ def run_command(client,
 
 
 def get_scenario_type(source_platform, target_platform, task_type=""):
-    if source_platform == "sql":
-        if target_platform == "sqldb":
-            scenario_type = ScenarioType.sql_sqldb_online if "online" in task_type else ScenarioType.sql_sqldb_offline
-        elif target_platform == "sqlmi":
-            scenario_type = ScenarioType.sql_sqlmi_online if "online" in task_type else ScenarioType.sql_sqlmi_offline
+    if source_platform == "sql" and target_platform == "sqldb":
```
vchske: Why this change? They are functionally identical and I split it out so that if there was branching based on target it would be easier to read. If you feel strongly the change makes it easier to read I shan't push back.

src/dms-preview/azext_dms/custom.py
temandr: The method scenario_handled_in_extension should only filter out scenarios that are in the extension. The validation here should only validate that the only scenario we're pushing to the core method is only the sql->sqldb scenario. This way, if a user puts in a scenario that isn't overridden in the extension, we don't send any non supported scenarios to the core. For example, say the user selected sql -> mongosql: since this is not a scenario supported by the extension, this will be sent to the core. 

Ideally, this should be in the scenario handled in the core. When we add this method to the core, we can remove this line completely from the extension.

src/dms-preview/azext_dms/custom.py
temandr: Adressed in the comment above.

vchske: So it forces it to fall through to type Unknown?

src/dms-preview/azext_dms/custom.py
vchske: I see what you're saying. We should probably get this check into the core as it confuses me every time I look at it with fresh eyes :)

src/dms-preview/azext_dms/custom.py
temandr: Sorry, I thought this was referring to ln 148. 

For this change, if the source is 'sql' but the target doesn't match "sqlmi" or "sqldb", it doesn't end up setting scenario_type.

src/dms-preview/azext_dms/custom.py
temandr: Agreed :)

PR #510, opened 2019-02-08
Update README.md

tjprescott:
Add references to `azdev extension update-index` command.

---

### General Guidelines

For new extensions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ Add your extension to the index to make it available in these CLI commands:
 
 ### How to generate sha256digest for an index.json entry?
 
+The `azdev extension update-index` command will calculate the SHA256 digest for you and add your entry to the index.json file. For more information visit https://github.com/Azure/azure-cli-dev-tools.
+
+As a fallback:
+
 MacOS
 ```
 shasum -a 256 path_to_whl.whl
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,6 +55,10 @@ Note: Hash should be in lowercase in index.json otherwise CI will fail.
 
 ### How to fill in the metadata for an index.json entry?
 
+The `azdev extension update-index` command will add the appropriate metadata to the index.json file. For more information visit https://github.com/Azure/azure-cli-dev-tools.
```
williexu: Can you add some info about how `update-index` will actually add a new entry? Right now this sounds like, if I had the entry for the new version of an extension, a `metadata` field will be added with the right info to that entry.

tjprescott: Updated the wording. Let me know what you think. I'm trying to avoid putting too many implementation details here to avoid the readme going stale as the command evolves.

README.md
```diff
@@ -51,6 +55,10 @@ Note: Hash should be in lowercase in index.json otherwise CI will fail.
 
 ### How to fill in the metadata for an index.json entry?
 
+The `azdev extension update-index` command can be used to simplify the process of updating the index file. Run `azdev extension update-index <URL>` where URL is the fully-qualified URL to your published extension WHL. This will gather the appropriate metadata and add an entry to your extension's metadata entries. For more information visit https://github.com/Azure/azure-cli-dev-tools.
```
williexu: Instead of `add an entry to your extension's metadata entries.`, I recommend `add an entry to your extension in the index.`

README.md
```diff
@@ -37,6 +37,10 @@ Add your extension to the index to make it available in these CLI commands:
 
 ### How to generate sha256digest for an index.json entry?
 
+The `azdev extension update-index` command will calculate the SHA256 digest for you and add your entry to the index.json file. For more information visit https://github.com/Azure/azure-cli-dev-tools.
+
```
williexu: It might be good to make this message more consistent with the message for the metadata below.

tjprescott: I updated it to only refence that it calculates the SHA256 digest. So it's actually less similar but I think it makes more sense.

README.md
tjprescott: Reworded.

PR #521, opened 2019-02-14
Update Index file for latest batch extension

bgklein:
---

### General Guidelines

For new extensions:

### DIFF
--- a/src/index.json
+++ b/src/index.json
@@ -203,10 +203,10 @@
         ],
         "azure-batch-cli-extensions": [
             {
-                "downloadUrl": "https://github.com/Azure/azure-batch-cli-extensions/releases/download/azure-batch-cli-extensions-2.5.1/azure_batch_cli_extensions-2.5.1-py2.py3-none-any.whl",
-                "filename": "azure_batch_cli_extensions-2.5.1-py2.py3-none-any.whl",
+                "downloadUrl": "https://github.com/Azure/azure-batch-cli-extensions/releases/download/azure-batch-cli-extensions-2.5.2/azure_batch_cli_extensions-2.5.2-py2.py3-none-any.whl",
+                "filename": "azure_batch_cli_extensions-2.5.2-py2.py3-none-any.whl",
                 "metadata": {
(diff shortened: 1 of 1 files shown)


bgklein:
@xingwu1

src/index.json
```diff
@@ -239,9 +239,50 @@
...
                     ],
                     "summary": "Additional commands for working with Azure Batch service",
-                    "version": "2.5.1"
+                    "version": "2.5.2"
+                },
+                "sha256Digest": "58dc9f5830bf0f3e75e5d4c65bc41b4f7cce47743678685d61fb3d7c64e2384c"
+            },
+            {
+                "downloadUrl": "https://github.com/Azure/azure-batch-cli-extensions/releases/download/azure-batch-cli-extensions-3.0.1/azure_batch_cli_extensions-3.0.1-py2.py3-none-any.whl",
+                "filename": "azure_batch_cli_extensions-3.0.1-py2.py3-none-any.whl",
```
tjprescott: Is there some feature in 2.0.59 of the CLI that this extension relies upon? Just curious why the two entries.

bgklein: It has to do with dependencies.  Both the extension and the normal CLI commandlets  use the Batch Python SDK. We made a breaking change in the SDK which means 3.0.1 is not compatable with 2.0.58.

tjprescott: Does your extension not vendor the SDK?  If it did, it shouldn't matter what version of SDK the azure-cli-batch module uses. If you are relying on the SDK from the module, then I can see it. Typically, you should only need to break on the CLI version number if there is some change to azure-cli-core that you rely on.

bgklein: Currently we do not vendor the SDK.  @xingwu1 might be able to add historical context on why we don't or if it was just an oversight on our part.

If I had to guess it was that we didn't want different SDK versions on the CLI and CLI extension, as this meant commands that are overwritten in the extension could be submitting to a different REST API as the other commands, which could be confusing to user if they were referencing documentation.

tjprescott: Fair enough. cc/ @williexu

PR #558, opened 2019-03-06
[AKS] Improve messages for upgrade command

lalo:
- Inform users of specific behaviours when they try to upgrade a cluster
to the current version of that cluster.

Mirror PR of this https://github.com/Azure/azure-cli/pull/8625

---

### General Guidelines

For new extensions:


williexu:
@zqingqing1 can you review as codeowner?

zqingqing1:
@lalo , what about the cluster is at updating/Deleting status, should we message customer?

lalo:
@zqingqing1 this PR is mainly to keep it in sync with the non-extension code (https://github.com/Azure/azure-cli/pull/8625). I don't think we need to notify the customer on those cases, what makes upgrading special is that there is certain logic that gets triggered only if a user upgrades a failed cluster (mainly a cluster that failed previously while upgrading) and nothing happens if its in succeeded state. In the other cases, everything should be handled correctly already.

zqingqing1:
Got you, thanks :)

PR #574, opened 2019-03-15
Add ability to enable/disable HTTPS for custom domains in frontdoor. 

pichandwork:
---

### General Guidelines

- [Yes ] Have you run `./scripts/ci/test_static.sh` locally? (`pip install pylint flake8` required)

For new extensions:


src/front-door/azext_front_door/custom.py
```diff
@@ -235,6 +235,16 @@ def list_front_doors(cmd, resource_group_name=None):
...
 
 
+def list_fd_frontend_endpoints(cmd, resource_group_name, resource_name):
+    client = cf_fd_frontend_endpoints(cmd.cli_ctx, None)
+    return client.list_by_front_door(resource_group_name, resource_name)
+
+
+def get_fd_frontend_endpoints(cmd, resource_group_name, resource_name, item_name):
+    client = cf_fd_frontend_endpoints(cmd.cli_ctx, None)
+    return client.get(resource_group_name, resource_name, item_name)
```
tjprescott: It seems these two you should be able to register without needing a custom command. 

```Python
g.command('list', 'list_by_front_door')
g.command('show', 'get')
```

src/front-door/azext_front_door/custom.py
```diff
@@ -261,24 +271,40 @@ def update_fd_frontend_endpoints(instance, host_name=None, session_affinity_enab
...
+        secret_name=None,
+        secret_version=None,
+        certificate_type="Dedicated"
+    )
+    cf_fd_frontend_endpoints(cmd.cli_ctx, None).enable_https(resource_group_name, front_door_name,
+                                                             item_name, config)
+    return get_fd_frontend_endpoints(cmd, resource_group_name, front_door_name, item_name)
+
+
+def configure_fd_frontend_endpoint_https_keyvault(cmd, resource_group_name, front_door_name, item_name,
```
tjprescott: If you plan on supporting other HTTPS configurations, you'll want to consider how your command surface area will expand over time. You could have something like `az network front-door frontend-endpoint configure-https` that would allow you to configure whichever type or HTTPS you want or disable it all with a single command.

src/front-door/azext_front_door/custom.py
```diff
@@ -261,24 +271,40 @@ def update_fd_frontend_endpoints(instance, host_name=None, session_affinity_enab
...
 
-def configure_fd_frontend_endpoint_https(cmd, resource_group_name, front_door_name, frontend_endpoint_name,
-                                         disable=None, protocol=None, secret_name=None, secret_version=None,
-                                         certificate_type=None, certificate_source=None, vault=None):
-    if disable:
-        return cf_fd_frontend_endpoints(cmd.cli_ctx, None).disable_https(resource_group_name, front_door_name,
-                                                                         frontend_endpoint_name)
-    # if not being disabled, then must be enabled
+def configure_fd_frontend_endpoint_https_disable(cmd, resource_group_name, front_door_name, item_name):
+    return cf_fd_frontend_endpoints(cmd.cli_ctx, None).disable_https(resource_group_name, front_door_name,
```
tjprescott: Seems this could also be done with 
`g.command('disable-https', 'disable_https')`

src/front-door/azext_front_door/_params.py
```diff
@@ -49,6 +49,11 @@ def load_arguments(self, _):
     with self.argument_context('network front-door', arg_group='Frontend Endpoint') as c:
         c.argument('frontend_host_name', help='Domain name of the frontend endpoint.')
 
+    with self.argument_context('network front-door', arg_group='HTTPS') as c:
+        c.argument('secret_name', help='The name of the Key Vault secret representing the full certificate PFX')
+        c.argument('secret_version', help='The version of the Key Vault secret representing the full certificate PFX')
+        c.argument('vault_id', help='The resource id of the Key Vault containing the SSL certificate')
```
tjprescott: Some opportunities for QOL improvements here. If the user omits version, does it use latest? Could the user provide the URL for the secret (which can include the version, no?).  Could you use that secret's URL to figure out the KeyVault ID?  

Nothing critical or blocking, but some things that I think other command modules do to try and make working with KeyVault easier.

src/front-door/azext_front_door/custom.py
pichandwork: The generic command get the front door model, which doesnt have all the information(customHttpsConfiguration) of frontend endpoint. Thats why used the custom command to directly get the frontend endpoint which has all the information.

src/front-door/azext_front_door/custom.py
pichandwork: The problem was with required parameters. HTTPS can be enabled using frontdoor managed certificate or keyvault certificate. Both of them have seperate required parameters. Thats why i had different commands to enable and disable.

src/front-door/azext_front_door/custom.py
pichandwork: Tried it. Failing with below error
The operation 'azext_front_door.vendored_sdks.operations.front_doors_operations#FrontDoorsOperations.disable_https' is invalid.
  File "E:\azure-cli-extensions\env\lib\site-packages\azure\cli\core\__init__.py", line 498, in get_op_handler
    op = getattr(op, part)
AttributeError: type object 'FrontDoorsOperations' has no attribute 'disable_https'

src/front-door/azext_front_door/_params.py
pichandwork: Version, secret name and keyvault id are mandatory parameters. All these are in line with our portal experience and powershell. Thanks for the feedback.

src/front-door/azext_front_door/custom.py
pichandwork: Group
    az network front-door frontend-endpoint : Manage Front Door frontend endpoints.

Commands:
    create                 : Create a Front Door frontend endpoint.
    delete                 : Delete a Front Door frontend endpoint.
    disable-https          : Disable HTTPS protocol for a custom domain.
    enable-https-frontdoor : Enable HTTPS protocol for a custom domain using Front Door managed
                             certificate.
    enable-https-keyvault  : Enable HTTPS protocol for a custom domain using your own certificate in
                             Azure Key Vault. You need to setup the right permissions for Front Door
                             to access your Key vault. Learn more at
                             https://aka.ms/FrontDoorCustomDomain.
    list                   : List Front Door frontend endpoints.
    show                   : Get the details of a Front Door frontend endpoint.

pichandwork:
@tjprescott - Replied to all your comments. Please review and let me know your feedback.

src/front-door/azext_front_door/custom.py
tjprescott: Ah, you would need to add `client=cf_fd_frontend_endpoints` in the command definition.  Otherwise it defaults to the standard frontdoor one.

src/front-door/azext_front_door/custom.py
tjprescott: Many commands have this problem and use validators to ensure the user provides the correct set of arguments. The preference in CLI is overloading. The help would look something like:

```
Arguments
    --type             : HTTPS configuration type. Allowed values: FrontDoor, KeyVault, Disabled
KeyVault Arguments
    --keyvault         : Name or ID of the KeyVault containing the certificate.
    --secret-name      : Name of the KeyVault secret.
    --secret-version:  : KeyVault secret version. Omit to use latest.
```

The validator would look something like:
```Python
def process_http_configure_namespace(namespace):
  # keyvault, secret_name, secret_version retrieved from namespace
  keyvault_usage = 'usage error: --type KeyVault --keyvault NAME_OR_ID --secret-name NAME [--secret-version VERSION]'
  if type != 'KeyVault' and any([keyvault, secret_name, secret_version]):
    raise CLIError(keyvault_usage)
  if type == 'KeyVault' and not all([keyvault, secret_name]):
    raise CLIError(keyvault_usage)
```

This would allow you to accomplish all three scenarios with a single command (and support additional ones without creating new commands).

pichandwork:
@tjprescott - Combined enable-https-frontdoor and enable-https-keyvault to a single command enable-https. Kept disable-https command as is. Also fixed the delete command for frontend endpoint, as the previous one was acting on frontdoor instead of frontend endpoint resource.

tjprescott:
Looks great @pichandwork! Let me know if you run into issues updating the index. Will merge once CI is green.

PR #576, opened 2019-03-15
Add support for available zones

cliffwoodave14:
As "Available Zones" feature is under private review state for AKS, so this PR is added to support "Available Zones" with Azure CLI extensions.
Customer could use following command to specify available zones while they create AKS cluster:
$az aks create --node-zones 1 2 3


jluk:
@cliffwoodave14 could you update this to "--node-zones" to stay more consistent with k8s nomenclature? we have the agent pool property moving to be node-zones as well.

cliffwoodave14:
@jluk Sure, will modify the agent-zones to node-zones.

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -66,6 +66,7 @@ def load_arguments(self, _):
         c.argument('min_count', type=int, validator=validate_nodes_count)
         c.argument('max_count', type=int, validator=validate_nodes_count)
         c.argument('enable_vmss', action='store_true')
+        c.argument('node_zones')
```
tjprescott: This should probably leverage the global alias for zones.
```
from azure.cli.core.commands.parameters import zones_type
c.argument('node_zones', zones_type)
```

If you still want to customize on top that I would suggest:
```
c.argument('node_zones', zones_type, options_list='--node-zones', help='Space-separated list of availability zones where agent nodes will be placed.')
```

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -142,6 +142,9 @@
         - name: --enable-vmss
           type: bool
           short-summary: (PREVIEW) Enable VMSS agent type.
+        - name: --node-zones
+          type: string array
+          short-summary: (PREVIEW) Availability zones where agent nodes will be placed.
```
tjprescott: This will override the text you put in the help kwarg. You should remove this and, if desired, add "(PREVIEW)" to that help text.

PR #583, opened 2019-03-19
Adding feature to export snapshots instead of images. 

bebattis:
This allows for non-generalized (or non-sysprepped) osStates to be copied and deployed across regions. Will work as approved work around for copying VMs across regions which need to maintain identical runtimes.

---

Spoke with @tamirkamara on this previously offline. I was able to get the extension configured locally with the Azdev tools and ran a quick test to replicate a single region. This was successful. As far as development best practices, there may be a better way to do this, but for now I've essentially added two if/else statements and the switch parameter `--export-as-snapshot` which will 1) skip the image creation, and 2) place the temp snapshots into the `target_resource_group` vs the `transient_resource-group` depending on what the desired output is.

LMK if there is anything else that needs testing or improvement. Feel free to reach out again directly.


bebattis:
Not too sure on this error... should I un-comment "pylint: disable=too-many-statements" ?

PR #604, opened 2019-03-27
Add pod security policy support

norshtein:
---

### General Guidelines

For new extensions:

This PR enables configuring pod security policy for an AKS cluster.


src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -67,6 +67,7 @@ def load_arguments(self, _):
         c.argument('max_count', type=int, validator=validate_nodes_count)
         c.argument('enable_vmss', action='store_true')
         c.argument('node_zones', zones_type, options_list='--node-zones', help='(PREVIEW) Space-separated list of availability zones where agent nodes will be placed.')
+        c.argument('enable_pod_security_policy', action='store_true', options_list=['--enable-pod-security-policy'])
```
yugangw-msft: i believe `options_list=['--enable-pod-security-policy']` is redundant, the `options_list` is only usefully when you want to provide short ones

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -75,6 +76,8 @@ def load_arguments(self, _):
         c.argument('min_count', type=int, validator=validate_nodes_count)
         c.argument('max_count', type=int, validator=validate_nodes_count)
         c.argument('api_server_authorized_ip_ranges', type=str, validator=validate_ip_ranges)
+        c.argument('enable_pod_security_policy', action='store_true', options_list=['--enable-pod-security-policy'])
```
yugangw-msft: same here.

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -483,7 +484,8 @@ def aks_create(cmd, client, resource_group_name, name, ssh_key_value,  # pylint:
         service_principal_profile=service_principal_profile,
         network_profile=network_profile,
         addon_profiles=addon_profiles,
-        aad_profile=aad_profile)
+        aad_profile=aad_profile,
+        enable_pod_security_policy=True if enable_pod_security_policy else False)
```
yugangw-msft: this can be simplified as `bool(enable_pod_security_policy)`

norshtein:
Hi @yugangw-msft , thanks for reviewing! I have updated the code, PTAL.

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -76,8 +76,8 @@ def load_arguments(self, _):
         c.argument('min_count', type=int, validator=validate_nodes_count)
         c.argument('max_count', type=int, validator=validate_nodes_count)
         c.argument('api_server_authorized_ip_ranges', type=str, validator=validate_ip_ranges)
-        c.argument('enable_pod_security_policy', action='store_true', options_list=['--enable-pod-security-policy'])
-        c.argument('disable_pod_security_policy', action='store_true', options_list=['--disable-pod-security-policy'])
+        c.argument('enable_pod_security_policy', action='store_true')
+        c.argument('disable_pod_security_policy', action='store_true')
```
tjprescott: It would likely make more sense to have one parameter `--enable-pod-security-policy` that is a three-state flag and can be turned off by using `--enable-pod-security-policy false`. Otherwise, what do you do if the user supplies both flags?

norshtein: If user specifies both flags, the CLI with [return an error](https://github.com/Azure/azure-cli-extensions/pull/604/files#diff-071fc9e293b8a43c36af8e578fe122e4R564). It is designed to keep aligned with other flags: `enable-rbac`&`disable-rbac`, `enable-cluster-autoscaler`&`disable-cluster-authscaler`, does it make sense?

tjprescott: If the command already followed this anti-pattern, then it would make sense to stay consistent with itself.

norshtein:
Thanks everyone for reviewing! @yugangw-msft @tjprescott Could you merge the PR?

norshtein:
Hey @yugangw-msft @tjprescott, any blocker for merging the PR?

PR #611, opened 2019-03-30
Add ability to add redirect route and backend pools setting

pichandwork:
---

### General Guidelines

- [ Yes] Have you run `./scripts/ci/test_static.sh` locally? (`pip install pylint flake8` required)

For new extensions:


src/front-door/azext_front_door/_params.py
```diff
@@ -16,11 +18,18 @@
     validate_frontend_endpoints, validate_backend_pool, MatchConditionAction)
 
 
+class RouteType(str, Enum):
+    forward = "Forward"
+    redirect = "Redirect"
```
tjprescott: Is this enum not present in your SDK?

src/front-door/azext_front_door/_params.py
```diff
@@ -50,11 +59,14 @@ def load_arguments(self, _):
         c.argument('frontend_host_name', help='Domain name of the frontend endpoint.')
 
     with self.argument_context('network front-door', arg_group='HTTPS') as c:
-        c.argument('certificate_source', arg_type=get_enum_type(FrontDoorCertificateSource), help='Certificate source to enable HTTPS. Allowed values: FrontDoor, AzureKeyVault. For Azure Key Vault, right permissions need to be set for Front Door to to access the Key vault. Learn more at https://aka.ms/FrontDoorCustomDomain.')
+        c.argument('certificate_source', arg_type=get_enum_type(FrontDoorCertificateSource), help='Certificate source to enable HTTPS. Allowed values: FrontDoor, AzureKeyVault. For Azure Key Vault, right permissions need to be set for Front Door to access the Key vault. Learn more at https://aka.ms/FrontDoorCustomDomain.')
```
tjprescott: This is going to display the allowed values twice, once because it is in your help string, and once because of the `get_enum_type`. I would remove the "Allowed values" annotation, and move the extended description "For Azure Key vault..." to a long description using the `help.py` file.

src/front-door/azext_front_door/_params.py
```diff
@@ -50,11 +59,14 @@ def load_arguments(self, _):
...
         c.argument('secret_name', help='The name of the Key Vault secret representing the full certificate PFX')
         c.argument('secret_version', help='The version of the Key Vault secret representing the full certificate PFX')
         c.argument('vault_id', help='The resource id of the Key Vault containing the SSL certificate')
 
+    with self.argument_context('network front-door', arg_group='BackendPools Settings') as c:
```
tjprescott: Recommend: `arg_group="Backend Pool"` which will show as "Backend Pool Arguments" instead of "BackendPools Settings Arguments".

src/front-door/azext_front_door/_params.py
```diff
@@ -50,11 +59,14 @@ def load_arguments(self, _):
...
+        c.argument('enforce_certificate_name_check', arg_type=get_enum_type(EnforceCertificateNameCheckEnabledState), help='Whether to enforce certificate name check on HTTPS requests to all backend pools. No effect on non-HTTPS requests.')
```
tjprescott: Are the values here just true/false or enabled/disabled?  If so, I'd recommend using `arg_type=get_three_state_flag(...)`

src/front-door/azext_front_door/_params.py
```diff
@@ -114,6 +127,12 @@ def load_arguments(self, _):
         c.argument('custom_forwarding_path', help='Custom path used to rewrite resource paths matched by this rule. Leave empty to use incoming path.')
         c.argument('dynamic_compression', arg_type=get_three_state_flag(positive_label='Enabled', negative_label='Disabled', return_label=True), help='Use dynamic compression for cached content.')
         c.argument('query_parameter_strip_directive', arg_type=get_enum_type(FrontDoorQuery), help='Treatment of URL query terms when forming the cache key.')
+        c.argument('redirect_type', arg_type=get_enum_type(FrontDoorRedirectType), help='The redirect type the rule will use when redirecting traffic. Default: Moved.')
```
tjprescott: You shouldn't need to spell out the default here. The CLI will automatically pick up the default from the method signature if it is available. If you are calling out the service default (i.e. if None is sent to the service) then this may become inaccurate.

If you actually want to *specify* the default, then use the `default=...` kwarg, which will automatically populate this help text and ensure there's never a difference between what the help says and the CLI sends.

src/front-door/azext_front_door/_params.py
```diff
@@ -114,6 +127,12 @@ def load_arguments(self, _):
...
+        c.argument('redirect_protocol', arg_type=get_enum_type(FrontDoorRedirectProtocol), help='The protocol of the destination to where the traffic is redirected.  Default: MatchRequest.')
+        c.argument('custom_host', help='Host to redirect. Leave empty to use use the incoming host as the destination host.')
+        c.argument('custom_path', help='The full path to redirect. Path cannot be empty and must start with /. Leave empty to use the incoming path as destination path.')
+        c.argument('custom_fragment', help='Fragment to add to the redirect URL. Fragment is the part of the URL that comes after #. Do not include the #.')
+        c.argument('custom_query_string', help='The set of query strings to be placed in the redirect URL. Setting this value would replace any existing query string; leave empty to preserve the incoming query string. Query string must be in <key>=<value> format. The first ? and & will be added automatically so do not include them in the front, but do separate multiple query strings with &.')
```
tjprescott: Would be very wise to add an example for the commands that have this parameter to help clarify.

src/front-door/azext_front_door/_params.py
pichandwork: No, its not present in SDK. SDK implementation is polymorphism based and there is no route type defined.

src/front-door/azext_front_door/_params.py
pichandwork: These are backend pools settings which apply to all backends. So, 'BackendPools Settings' is opt here.

src/front-door/azext_front_door/_params.py
pichandwork: Fixed

src/front-door/azext_front_door/_params.py
pichandwork: Fixed. Moved the long summary to command 'enable-https' help as that seems the appropriate place

src/front-door/azext_front_door/_params.py
pichandwork: Fixed

PR #633, opened 2019-04-09
Azure Event Grid 0.4.1

ahamad-MS:
---

### General Guidelines

For new extensions:


src/eventgrid/azext_eventgrid/_help.py
```diff
@@ -188,7 +202,8 @@
           long-summary: |
             Example: --deadletter-endpoint /subscriptions/{SubID}/resourceGroups/rg1/providers/Microsoft.Storage/storageAccounts/sa1/blobServices/default/containers/containerName
         - name: --endpoint-type
-          short-summary: The type of the destination endpoint.
+          short-summary: The type of the destination endpoint. It is expected that the destination endpoint to be created and available for use before executing any Event Grid command.
```
kalyanaj: Nit: Consider changing "the destination endpoint to be created" to "the destination endpoint be created"

ahamad-MS: done

src/eventgrid/azext_eventgrid/_params.py
```diff
@@ -64,25 +69,58 @@ def load_arguments(self, _):
...
+        c.argument('topic_name', arg_type=name_type, help='Name of the topic.', id_part=None, completer=get_resource_name_completion_list('Microsoft.EventGrid/topics'))
+
+    with self.argument_context('eventgrid domain create') as c:
+        c.argument('domain_name', arg_type=name_type, help='Name of the domain.', id_part='name', completer=get_resource_name_completion_list('Microsoft.EventGrid/domains'))
+
+    with self.argument_context('eventgrid domain delete') as c:
+        c.argument('domain_name', arg_type=name_type, help='Name of the domain.', id_part='name', completer=get_resource_name_completion_list('Microsoft.EventGrid/domains'))
+
+    with self.argument_context('eventgrid domain update') as c:
+        c.argument('domain_name', arg_type=name_type, help='Name of the domain.', id_part='name', completer=get_resource_name_completion_list('Microsoft.EventGrid/domains'))
```
tjprescott: Instead of repeating this code three times, just use a scope of `eventgrid domain`, which will apply it to all commands and subgroups under the domain subgroup.

ahamad-MS: Per our discussion, we needed to do this for domain to allow us to use different arguments and related names in domain topic scope. In specific, using --name for domain-topic-name (rather than domain-name which is used at domain scope) and domain-name for specifying the domain name (which is --name is domain scope).

tjprescott: So a couple techniques to eliminate the duplication:

```Python
domain_name_type = CLIArgumentType(help='Name of the domain.', options_list='--domain-name', completer=get_resource_name_completion_list('Microsoft.EventGrid/domains'))

with self.argument_context('eventgrid domain') as c:
  c.argument('domain_name', domain_name_type, options_list=['--name', '-n'], id_part=name)

with self.argument_context('eventgrid domain topic') as c:
  c.argument('domain_name', domain_name_type, id_part=None)
```

And if you really needed to override per command (which is usually not the case):

```Python
for scope in ['create', 'update', 'delete']:
   with self.argument_context('eventgrid domain {}'.format(scope)) as c:
      c.argument('domain_name', …)
```

ahamad-MS: Thx for the suggestion. Makes sense. Fixed.

PR #636, opened 2019-04-09
[AKS]customizing node resource group

zqingqing1:
---

### General Guidelines

For new extensions:


zqingqing1:
Since backend release is not rolled out, so please don't merge it yet.

jluk:
@zqingqing1 since we're still waiting on this deployment (I think) could we make the parameter "--node-resource-group-name"?

zqingqing1:
we can, but won't the param name be a little bit longer?

tjprescott:
@jluk @zqingqing1 `--node-resource-group` would be more consistent with the rest of the CLI than `--node-resource-group-name`

jluk:
Sounds good to me then - j/w if it wold be clear what the parameter would mean but if thats consistent let's keep it!

jluk:
@zqingqing1 are we good to merge this? looks like the AKS 0408 release is finally out in prod.

zqingqing1:
Not really,  till now, Asian regions haven't been released yet, but should be done today. FYI, I am oncall this week. So the ETA could be next week.

jluk:
ack, sounds good to me thank you!

zqingqing1:
NOTE: when generating 2019-04-01 SDK, make sure the node-resource-group is not read only. @xizhamsft

zqingqing1:
Just confirmed, the backend release has been everywhere, should be good to merge this one.

src/aks-preview/azext_aks_preview/__init__.py
```diff
@@ -31,7 +31,12 @@ def load_command_table(self, args):
         return self.command_table
 
     def load_arguments(self, command):
-        super(ContainerServiceCommandsLoader, self).load_arguments(command)
+        from sys import version_info
+        if version_info[0] < 3:
+            super(ContainerServiceCommandsLoader, self).load_arguments(command)
```
tjprescott: This syntax works on Python 3 as well. There's no need for this if statement.

zqingqing1: Are you sure about that? Since my teammate and I have problem to use `super(ContainerServiceCommandsLoader, self).load_arguments(command)`, It failed with 
```
The command failed with an unexpected error. Here is the traceback:

super(type, obj): obj must be an instance or subtype of type
  File "/usr/local/Cellar/azure-cli/2.0.63/libexec/lib/python3.7/site-packages/knack/cli.py", line 206, in invoke
    cmd_result = self.invocation.execute(args)
  File "/usr/local/Cellar/azure-cli/2.0.63/libexec/lib/python3.7/site-packages/azure/cli/core/commands/__init__.py", line 445, in execute
    self.commands_loader.load_arguments(command)
  File "/usr/local/Cellar/azure-cli/2.0.63/libexec/lib/python3.7/site-packages/azure/cli/core/__init__.py", line 268, in load_arguments
    loader.load_arguments(command)  # this adds entries to the argument registries
  File "/Users/qizhe/cli-extension/azure-cli-extensions/src/aks-preview/azext_aks_preview/__init__.py", line 34, in load_arguments
    super(ContainerServiceCommandsLoader, self).load_arguments(command)
TypeError: super(type, obj): obj must be an instance or subtype of type
```, so I decide to use the if else statement to unblock us.

tjprescott: Nowhere else in the CLI or any extension does anyone do this, so the true problem is likely something else. That being said, I have no problem merging this, but it shouldn't be necessary.

PR #641, opened 2019-04-11
[aks-preview]: fix python3-only super() syntax

mboersma:
Restores the old, Python2-compatible syntax for calling `super()`.

Fixes Azure/AKS#908

cc: @zqingqing1 @xizhamsft

---

### General Guidelines

For new extensions:

### DIFF
--- a/src/aks-preview/azext_aks_preview/__init__.py
+++ b/src/aks-preview/azext_aks_preview/__init__.py
@@ -31,8 +31,7 @@ class ContainerServiceCommandsLoader(AzCommandsLoader):
         return self.command_table
 
     def load_arguments(self, command):
-        # super(ContainerServiceCommandsLoader, self).load_arguments(command)
-        super().load_arguments(command)
+        super(ContainerServiceCommandsLoader, self).load_arguments(command)
         from ._params import load_arguments
         load_arguments(self, command)
 


zqingqing1:
@mboersma, seems like it will block customer which is using python 3.7

super(type, obj): obj must be an instance or subtype of type
  File "/usr/local/Cellar/azure-cli/2.0.49/libexec/lib/python3.7/site-packages/knack/cli.py", line 197, in invoke
    cmd_result = self.invocation.execute(args)
  File "/usr/local/Cellar/azure-cli/2.0.49/libexec/lib/python3.7/site-packages/azure/cli/core/commands/__init__.py", line 262, in execute
    self.commands_loader.load_arguments(command)
  File "/usr/local/Cellar/azure-cli/2.0.49/libexec/lib/python3.7/site-packages/azure/cli/core/__init__.py", line 254, in load_arguments
    loader.load_arguments(command)  # this adds entries to the argument registries
  File "/Users/qizhe/cli-extension/azure-cli-extensions/src/aks-preview/azext_aks_preview/__init__.py", line 34, in load_arguments
    super(ContainerServiceCommandsLoader, self).load_arguments(command)
TypeError: super(type, obj): obj must be an instance or subtype of type

mboersma:
@zqingqing1 in python 3, `super()` made its two arguments optional, but the old syntax is compatible with Python 2 and Python 3. The previous code wouldn't work in Python2, which `az` must stay compatible with.

We use the two-argument `super()` syntax in the methods above and below this one. What's different about this function?

zqingqing1:
I don't know why, but looks like old syntax is not compatible with python 3.
So I will make change to check the version of python, and then inherit the super() for different version.

This should work.
```
def load_arguments(self, command):
        from sys import version_info
        if version_info[0] < 3:
            super(ContainerServiceCommandsLoader, self).load_arguments(command)
        else:
            super().load_arguments(command)

        from ._params import load_arguments
        load_arguments(self, command)
```

mboersma:
I don't know why either, but I agree that should work. :-)

PR #644, opened 2019-04-17
SQL: make "up" command work end to end

yugangw-msft:
1. Use `pymssql` as data plane for 2 reasons
   a. It doesn't need ODBC, which requires non trivial manual configuration in the machine, not cool
   b. The Linux and Windows package bundles freetds so install is smooth; for OSX, db-up extension will 
       handle the brew update to update  "freetds", so the install is smooth as well
2. Finish up the firewall rule and default db provision

### General Guidelines

For new extensions:

### DIFF
--- a/src/db-up/azext_db_up/commands.py
+++ b/src/db-up/azext_db_up/commands.py
@@ -47,5 +47,3 @@ def load_command_table(self, _):  # pylint: disable=too-many-locals, too-many-st
                          table_transformer=table_transform_connection_string)
         g.custom_command('down', 'server_down', validator=db_down_namespace_processor('sql'),
                          supports_no_wait=True, confirmation=True)
-        # Core SQL command "az sql db show-connection-string" does the similar thing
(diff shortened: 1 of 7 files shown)


src/db-up/azext_db_up/commands.py
```diff
@@ -47,4 +47,5 @@ def load_command_table(self, _):  # pylint: disable=too-many-locals, too-many-st
                          table_transformer=table_transform_connection_string)
         g.custom_command('down', 'server_down', validator=db_down_namespace_processor('sql'),
                          supports_no_wait=True, confirmation=True)
-        g.custom_command('show-connection-string', 'create_sql_connection_string')
+        # Core SQL command "az sql db show-connection-string" does the similar thing
+        # g.custom_command('show-connection-string', 'create_sql_connection_string')
```
tjprescott: Any reason not to simply remove this then instead of comment out?

PR #651, opened 2019-04-24
[AKS] Add Azure policy addon

robbiezhang:
---

### General Guidelines

For new extensions:


src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -642,7 +642,9 @@ def _remove_nulls(managed_clusters):
 
 ADDONS = {
     'http_application_routing': 'httpApplicationRouting',
-    'monitoring': 'omsagent'
+    'monitoring': 'omsagent',
+    'virtual-node': 'aciConnector',
+    'azure-policy': 'azurepolicy'
```
robbiezhang: this is the only change. other code are just copied from the azure-cli

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -300,3 +302,25 @@
...
+    These addons are available:
+        http_application_routing - configure ingress with automatic public DNS name creation.
+        monitoring - turn on Log Analytics monitoring. Requires "--workspace-resource-id".
+        virtual-node - enable AKS Virtual Node (PREVIEW). Requires --subnet-name to provide the name of an existing subnet for the Virtual Node to use.
+        azure-policy - enable Azure policy (PREVIEW).
+parameters:
+  - name: --addons -a
+    type: string
+    short-summary: Enable the Kubernetes addons in a comma-separated list.
+  - name: --workspace-resource-id
```
tjprescott: If this is an ARM resource ID, then this should just be `--workspace` and accept a name or ID. See:
https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-name-or-id-parameters

Ending an argument with `-id` is only allowed when the ID is a guid, not an ARM resource ID.

robbiezhang: it's copied from the azure-cli.
https://github.com/Azure/azure-cli/blob/0589ce7107beb79d3939630f5dd8a78a105a5c42/src/command_modules/azure-cli-acs/azure/cli/command_modules/acs/_help.py#L280-L282

robbiezhang: @tjprescott, are you fine with this? I don't want to change the existing command parameters.

robbiezhang: @tjprescott ping...

PR #683, opened 2019-05-08
VM Repair Extension

swbae31:
### Overview ###
VM repair scripts will enable Azure users to self-repair non-connectable VMs by copying the source VM’s OS disk and attaching it to a newly created repair VM.

### Commands Added ###

Create a new rescue VM and attach the faulty VM's copied OS disk as a data disk.
```
az vm repair create
```
Replace faulty VM's OS disk with data disk from rescue VM.
```
az vm repair restore
```

### How to try out on azure cloud shell ###

From the azure portal cloud shell run the command:
```
az extension add --source https://azurecomputeaidrepair.blob.core.windows.net/prod/vm_repair-0.1.0-py2.py3-none-any.whl
az vm repair create -n <vm-name> -g <resource-group-name> --verbose
az vm repair restore -n <vm-name> -g <resource-group-name> --verbose
```


azure-cli-extensions.pyproj
```diff
@@ -4793,6 +4812,17 @@
     <Content Include="src\virtual-network-tap\readme.md" />
     <Content Include="src\virtual-wan\azext_vwan\azext_metadata.json" />
     <Content Include="src\virtual-wan\readme.md" />
+    <Content Include="src\vm-repair\azext_vm_repair\__pycache__\commands.cpython-37.pyc" />
+    <Content Include="src\vm-repair\azext_vm_repair\__pycache__\custom.cpython-37.pyc" />
+    <Content Include="src\vm-repair\azext_vm_repair\__pycache__\_help.cpython-37.pyc" />
+    <Content Include="src\vm-repair\azext_vm_repair\__pycache__\_params.cpython-37.pyc" />
+    <Content Include="src\vm-repair\azext_vm_repair\__pycache__\__init__.cpython-37.pyc" />
```
tjprescott: *.pyc files should not be included

swbae31:
@tjprescott

src/index.json
```diff
@@ -2181,6 +2181,52 @@
                 },
                 "sha256Digest": "33e87daa9ef75664e9c49f619985a0869ba6606c2a76cd5c6329f309f0e828b1"
             }
+        ],
+		"vm-repair": [
+            {
+                "downloadUrl": "https://azurecomputeaidrepair.blob.core.windows.net/prod/vm_repair-0.0.1-py2.py3-none-any.whl",
+                "filename": "vm_repair-0.0.1-py2.py3-none-any.whl",
```
tjprescott: The initial version should be 0.1.0 instead of 0.0.1

src/index.json
```diff
@@ -2181,6 +2181,52 @@
...
+                                "Home": "https://github.com/Azure/azure-cli-extensions/tree/master/src/vm-repair"
+                            }
+                        }
+                    },
+                    "generator": "bdist_wheel (0.30.0)",
+                    "license": "MIT",
+                    "metadata_version": "2.0",
+                    "name": "vm-repair",
+                    "summary": "Auto repair commands to fix faulty VMs.",
+                    "version": "0.0.1"
```
tjprescott: Same here. 0.1.0

src/vm-repair/azext_vm_repair/_params.py
```diff
@@ -0,0 +1,31 @@
...
+        c.argument('vm_name', existing_vm_name)
+
+    with self.argument_context('vm repair swap-disk') as c:
+        c.argument('rescue_username', help='Admin username for rescue VM.')
+        c.argument('rescue_password', help='Admin password for the rescue VM.')
+
+    with self.argument_context('vm repair restore-swap') as c:
+        c.argument('rescue_vm_id', help='Rescue VM resource id.')
+        c.argument('disk_name', help='Name of fixed data disk. Defaults to the first data disk in the rescue vm.')
+        c.argument('delete', help='Deletes the rescue resources without confirmation.')
```
tjprescott: This should use `--yes` to be consistent with other CLI commands that allow you to skip confirmation messages.

src/vm-repair/azext_vm_repair/_params.py
```diff
@@ -0,0 +1,31 @@
...
+
+    with self.argument_context('vm repair') as c:
+        c.argument('vm_name', existing_vm_name)
+
+    with self.argument_context('vm repair swap-disk') as c:
+        c.argument('rescue_username', help='Admin username for rescue VM.')
+        c.argument('rescue_password', help='Admin password for the rescue VM.')
+
+    with self.argument_context('vm repair restore-swap') as c:
+        c.argument('rescue_vm_id', help='Rescue VM resource id.')
```
tjprescott: Parameters ending with `-id` refer to GUIDs. Since this is an ARM resource, it should be `--rescue-vm` and should accept a name or ID.

src/vm-repair/azext_vm_repair/repair_utils.py
```diff
@@ -0,0 +1,148 @@
...
+
+def _uses_managed_disk(vm):
+    if vm.storage_profile.os_disk.managed_disk is None:
+        return False
+    return True
+
+def _call_az_command(command_string, run_async=False, secure_params=None):
+    """
+    Uses subprocess to run a command string. To hide sensitive parameters from logs, add the
+    parameter in secure_params. If run_async is True then function returns the stdout,
```
swbae31: *If run_async is **False**

src/vm-repair/azext_vm_repair/_validators.py
```diff
@@ -0,0 +1,124 @@
...
+
+    resource_client = _resource_client_factory(cmd.cli_ctx).resources
+    vm = None
+    try:
+        vm = resource_client.get(resource_group_name, classic_vm_provider, '', vm_resource_type, vm_name, api_version)
+    except CloudError as cloudError:
+        # Resource does not exist or the API failed
+        logger.debug(cloudError)
+        return False
+    print('sd')
```
tjprescott: What is this?

src/vm-repair/azext_vm_repair/custom.py
```diff
@@ -0,0 +1,252 @@
...
+           and target_vm.storage_profile.image_reference.version == '2016.127.20190214':
+            os_image_name = "MicrosoftWindowsServer:WindowsServer:2016-Datacenter:2016.127.20190115"
+        else:
+            os_image_name = "MicrosoftWindowsServer:WindowsServer:2016-Datacenter:2016.127.20190214"
+
+    timestamp = datetime.now().strftime('%Y%m%d%H%M%S')
+    copied_os_disk_name = vm_name + '-DiskCopy-' + timestamp
+    copied_disk_uri = None
+    rescue_vm_name = ('rescue-' + vm_name)[:15]
+    rescue_rg_name = 'rescue-' + vm_name + '-' + timestamp
```
tjprescott: Creating a non-deterministic resource group name will force users to supply rescue VM by ID. Crafting a deterministic resource group would avoid this.

src/vm-repair/azext_vm_repair/_validators.py
swbae31: oops, forgot to remove that after recent test. Thanks for catching that.

azure-cli-extensions.pyproj
swbae31: removed

src/index.json
swbae31: changed to 0.1.0

src/index.json
swbae31: changed to 0.1.0

src/vm-repair/azext_vm_repair/_params.py
swbae31: The rescue VM is within the difference resource group from the target faulty vm. So only name would not work. What should I do here?

src/vm-repair/azext_vm_repair/custom.py
swbae31: I made the resource group non-deterministic because I wanted it to be unique for all cases. Since I can figure out that resource group using tags, I think I can make users supply the rescue VM by ID or name. Thanks for the feedback.

swbae31:
> This extension has implicit dependencies on network, compute, storage and ARM that should be explicit. Calling into CLI commands like that is a bad practice that will likely break your extension without warning. @tjprescott 

Before I started the project I consulted with your team to ask how I would reuse the infrastructure already built in the CLI commands such as creating VMs, and copying disks. I thought it would be inefficient to rewrite all the code which validated the VM templates, called the API, handled errors, and etc. which were all already heavily tested and used on the current CLI release. I was leaned on using the internal functions in other command modules but your team recommended that I call the CLI commands directly, since it would be cleaner and more durable.

@adewaleo 
Tosin added to this that while the repair command is an extension, I could run auto run scenario tests before the bimonthly CLI command rollout on the new release. And after the commands are joined in as core commands in some point in the future, the integration tests will ensure that if other module changes break the vm-repair code, the dev who made those changes will know and make the appropriate changes to fix the breaks.

src/vm-repair/azext_vm_repair/_params.py
swbae31: Got it. Done.

src/vm-repair/azext_vm_repair/_params.py
swbae31: Never mind the previous comment. Actually now that I think of it, because rescue VM lives in another resource group, the group and name has to be both specified and ID is the best way to do it. This is because the rescue-vm name might be the same in different resource groups. Please advise. @tjprescott

src/vm-repair/azext_vm_repair/custom.py
swbae31: Some users may want to make multiple rescue VMs for the same faulty VM. For these cases, I need a way to create a unique resource group name which causes it to be non-deterministic. Is there any methods you have in mind to fix this problem?

src/vm-repair/azext_vm_repair/_params.py
tjprescott: Even if you only accept an ARM ID, you need to drop the `-id` suffix.

swbae31: Wouldn't that confuse the user thinking that they could enter the rescue vm name?

adewaleo: I wonder what @tjprescott would respond with. But in the worst case, you can clarify this in the help text and with examples.

src/vm-repair/azext_vm_repair/custom.py
adewaleo: You can check for existence of a previous group and increment the suffix by one

`rescue-my-vm-1`, `rescue-my-vm-2`, etc.

swbae31: Then it wouldn't be deterministic at that point  right? Since it could be 1,2,3...

adewaleo: Yes, it is is not deteministic with respect to the vm name, but if the user knows the index then the behavior of the command can be deterministic, in the case where a user wants to make multiple rescue vms per target vm.

given the target vm name and the "rescue vm id" (1,2,3...) you can figure out the resource group of the rescue vm. In this way, this id is actually a guid. In the case where this guid is not provided. You can default to the first resource group.

adewaleo: This is just a suggestion though feel free to take this as you wish...

src/vm-repair/azext_vm_repair/_params.py
swbae31: So we should burden the users to look through the help text and examples rather than being clear on the parameter naming?

src/vm-repair/azext_vm_repair/_params.py
```diff
@@ -0,0 +1,34 @@
...
+                                       help="The name of the Virtual Machine. You can configure the default using `az configure --defaults vm=<name>`",
+                                       completer=get_resource_name_completion_list('Microsoft.Compute/virtualMachines'), id_part='name')
+
+    with self.argument_context('vm repair') as c:
+        c.argument('vm_name', existing_vm_name)
+
+    with self.argument_context('vm repair create') as c:
+        c.argument('repair_username', help='Admin username for repair VM.')
+        c.argument('repair_password', help='Admin password for the repair VM.')
+        c.argument('repair_vm_name', help='Name of repair vm.')
```
ansahdev: Consistency in help text, either VM or vm

src/vm-repair/azext_vm_repair/_validators.py
```diff
@@ -0,0 +1,199 @@
...
+    if namespace.repair_group_name:
+        if namespace.repair_group_name == namespace.resource_group_name:
+            raise CLIError('The repair resource group name cannot be the same as the source VM resource group.')
+        _validate_resource_group_name(namespace.repair_group_name)
+    else:
+        namespace.repair_group_name = 'repair-' + namespace.vm_name + '-' + timestamp
+
+    # Check encrypted disk
+    if _uses_encrypted_disk(source_vm):
+        # TODO, validate this with encrypted VMs
```
ansahdev: Is there any pending action item on this TODO?

src/vm-repair/azext_vm_repair/_validators.py
```diff
@@ -0,0 +1,199 @@
...
+        logger.warning('The source VM\'s OS disk is encrypted.')
+
+    # Validate Auth Params
+    if is_linux and namespace.repair_username:
+        logger.warning("Chaging admin username property is not allowed for Linux VMs. Ignoring the given repair-username parameter.")
```
ansahdev: "Changing"

src/vm-repair/azext_vm_repair/_validators.py
```diff
@@ -0,0 +1,199 @@
...
+    is_linux = _is_linux_os(source_vm)
+
+    # Check repair vm name
+    if namespace.repair_vm_name:
+        _validate_vm_name(namespace.repair_vm_name, is_linux)
+    else:
+        namespace.repair_vm_name = ('repair-' + namespace.vm_name)[:15]
+
+    # Check copy disk name
+    timestamp = datetime.now().strftime('%Y%m%d%H%M%S')
```
ansahdev: expected local datetime of server or can this be UTC datetime?

src/vm-repair/azext_vm_repair/_validators.py
```diff
@@ -0,0 +1,199 @@
...
+
+def _prompt_repair_password(namespace):
+    from knack.prompting import prompt_pass, NoTTYException
+    try:
+        namespace.repair_password = prompt_pass('Repair VM Admin Password: ', confirm=True)
+    except NoTTYException:
+        raise CLIError('Please specify password in non-interactive mode.')
+
+def _classic_vm_exists(cmd, resource_group_name, vm_name):
+    api_version = '2017-04-01'
```
waynekuo: Consider moving these constants to a more visible place if there's a need to update these in the future.

src/vm-repair/azext_vm_repair/_params.py
swbae31: Thanks for the catch. Changed all to VM.

src/vm-repair/azext_vm_repair/_validators.py
swbae31: Updated to 'Setting admin username property...'

src/vm-repair/azext_vm_repair/custom.py
```diff
@@ -0,0 +1,256 @@
...
+        logger.error("Repair swap-disk failed. Cleaning up created resources.")
+    except SkuNotAvailableError as skuNotAvailableError:
+        logger.error(skuNotAvailableError)
+        logger.error("Please check if the current subscription can create more VM resources. Cleaning up created resources.")
+    except UnmanagedDiskCopyError as unmanagedDiskCopyError:
+        logger.error(unmanagedDiskCopyError)
+        logger.error("Repair swap-disk failed. Please try again at another time. Cleaning up created resources.")
+    except WindowsOsNotAvailableError:
+        logger.error('A compatible Windows OS image is not available at this time, please check subscription.')
+    finally:
```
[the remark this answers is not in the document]
swbae31: Good call. I was thinking at first that it wasn't good to have one broad general except but I guess in this case it's inevitable. Added:

except Exception as exception:
        logger.error('An unexpected error occurred. Try running again with the --debug flag to debug.')
        logger.debug(exception)

src/vm-repair/azext_vm_repair/_validators.py
swbae31: Yeah I should do that or have a more reliable method to get the supported api_version. Will look into this. Thanks.

src/vm-repair/azext_vm_repair/_validators.py
swbae31: I tested on a encrypted VM created by Zihan but this didn't work. I'll look more into this.

src/vm-repair/azext_vm_repair/_validators.py
swbae31: Fixed to datetime.utcnow().strftime('%Y%m%d%H%M%S'). Good catch.

src/vm-repair/azext_vm_repair/_validators.py
swbae31: I'll work and finish this for the next iteration after I get feedbacks.

adewaleo:
Hey Sungwoo, can you rebase with master first before merging?

src/vm-repair/azext_vm_repair/__init__.py
```diff
@@ -0,0 +1,32 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from azure.cli.core import AzCommandsLoader
+import azext_vm_repair._help  # pylint: disable=unused-import
+
+
+class VmRepairCommandsLoader(AzCommandsLoader):
```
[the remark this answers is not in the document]
swbae31: Done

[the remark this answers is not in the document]
swbae31: Leaving as is for now

diegoalb:
The repair-restore script will fail if the VM´s OS Disk is Encrypted (as the OS Disk swap API does not support encrypted disks). This exception is handled on the restore phase only.

Can we add some warning for this aspect right after the repair-create command execution!? That should save time and prepare the script users for this scenario.

Also, on this same VM encrypted scenario, the attached "broken OS disk" on the repair VM will be unavailable (because its encrypted) - Can we consider adding the reference on how to unlock the drive on the output?

https://blogs.technet.microsoft.com/mckittrick/unlock-azure-drive-locked-with-bitlocker-bek-encryption-arm/
https://blogs.msdn.microsoft.com/mast/2016/11/27/azure-disk-encryption-how-to-recover-bek-file-from-azure-key-vault/

Example of my test:

_diego@Azure:~$ az vm repair restore -n EncryptedVM1 -g SRxxxxx --verbose
Searching for repair-vm within subscription...
Repair-vm-id not given, restoring from only matching repair VM: /subscriptions/xxxxxxxxxx/resourceGroups/repair-EncryptedVM1-20190709152458/providers/Microsoft.Compute/virtualMachines/repair-Encrypte
Disk-name not given. Defaulting to the first data disk attached to the repair VM: EncryptedVM1-DiskCopy-20190709152458.vhd
Detaching repaired data disk from repair VM...
Attaching repaired data disk to source VM as an OS disk...
ERROR: Swapping OS Disk is not supported for VMs using disk encryption.

Repair swap-disk failed. If the restore command fails at retry, please rerun the repair process from 'az vm repair create'.
command ran in 44.696 seconds._

PR #715, opened 2019-05-29
Mesh removeminprofile

jeffj6123:
---

### General Guidelines

For new extensions:


src/mesh/azext_mesh/__init__.py
```diff
@@ -14,7 +14,6 @@ def __init__(self, cli_ctx=None):
         from azure.cli.core.commands import CliCommandType
         custom = CliCommandType(operations_tmpl='azext_mesh.custom#{}')
         super(SbzCommandsLoader, self).__init__(cli_ctx=cli_ctx,
-                                                min_profile='2017-03-10-profile',
```
limingu: There are 4 profiles in CLI. See details: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/profiles/__init__.py#L89

The min_profile is used to tell which profile is supported or not. For example, '2017-03-10-profile' that means mesh is supporting all the profiles after 2017-03-10. So it did't support 2017-03-09-profile and support others.

If just removing this line, the mesh would support all profiles. Is that what you want? If yes, it can be removed. If not, I made the change in PR: https://github.com/Azure/azure-cli/pull/9469 to use resource type to tell which resource it can support.

src/mesh/azext_mesh/azext_metadata.json
```diff
@@ -1,5 +1,6 @@
 {
     "azext.maxCliCoreVersion": "2.0.66",
     "azext.minCliCoreVersion": "2.0.30",
+    "azext.maxCliCoreVersion": "2.0.66",
```
limingu: This is dup. You can remove it.

PR #791, opened 2019-06-17
DMS changing versions

temandr:
---

### General Guidelines

For new extensions:


src/index.json
```diff
@@ -737,12 +737,11 @@
         ],
         "dms-preview": [
             {
-                "downloadUrl": "https://azcliorcas.blob.core.windows.net/azclipath/dms_preview-0.8.1-py2.py3-none-any.whl",
-                "filename": "dms_preview-0.8.1-py2.py3-none-any.whl",
```
tjprescott: You should keep the old entry so there is a version that can be installed on older versions of the CLI.

limingu: Either you can keep the old entry. Or you can keep "azext.minCliCoreVersion": "2.0.43" and remove max cli core version. As the new code with min_profile can work with all core versions

vchske: I would go with removing max and keeping the original min.

temandr: Updated the version the original version.

tjprescott:
Please resolve the conflicts and get the CI passing.

PR #802, opened 2019-06-21
Resource Graph: Use ObjectArray format option (includes SDK update)

OlhaTkachenko:
Resource-graph extension consumes objectArray response format. For this it was required to use latest resource-graph sdk.
Tested locally.
Version will be updated in index, when more changes added

---

### General Guidelines

For new extensions:


src/resource-graph/azext_resourcegraph/vendored_sdks/resourcegraph/version.py
```diff
@@ -9,5 +9,5 @@
 # regenerated.
 # --------------------------------------------------------------------------
 
-VERSION = "0.6.0"
+VERSION = "2.0.0"
```
chiragg4u: We should release 1.0.0 version for PS/CLI as there was no major version before. Also, we are not doing any change which breaks backward compat.

chiragg4u:
Suggest differentiating the files which are changed by you vs. the one which was generated. It will help in review.

src/resource-graph/azext_resourcegraph/vendored_sdks/resourcegraph/version.py
OlhaTkachenko: It is python sdk

OlhaTkachenko:
> Suggest differentiating the files which are changed by you vs. the one which was generated. It will help in review.

Commit "SDK update" includes all not my changes, two other are my changes. Please let me know if it isn't good enough

chiragg4u:
Changes look good. Thanks, Olha.

PR #807, opened 2019-06-24
CLI Webapp extension for scan functionality

ppvasude:
This function is basically dependent on the scan functionality being implemented in KuduLite for webapps on Linux. More details can be found in below PRs.

1)	https://github.com/Azure-App-Service/KuduLite/pull/55  : PR on KuduLite for the core functionality
2)	https://github.com/Azure-App-Service/KuduLiteBuild/pull/11  : PR on KuduLiteBuild for installing clamAV and copying over the script onto the docker container


src/webapp/azext_webapp/__init__.py
```diff
@@ -29,6 +36,9 @@ def load_arguments(self, _):
             c.argument('port', options_list=['--port', '-p'],
                        help='Port for the remote connection. Default: Random available port', type=int)
             c.argument('name', options_list=['--name', '-n'], help='Name of the webapp to connect to')
+        with self.argument_context('webapp scan') as c:
+            c.argument('name', options_list=['--name', '-n'], help='Name of the webapp to connect to')
```
panchagnula: nit: Name of the webapp to scan?

panchagnula:
Add help for the new commands here https://github.com/Azure/azure-cli-extensions/blob/master/src/webapp/azext_webapp/_help.py

src/webapp/azext_webapp/__init__.py
ppvasude: Yes

src/webapp/azext_webapp/_help.py
```diff
@@ -14,3 +14,28 @@
...
 """
+
+helps['webapp scan'] = """
+    type: group
+    short-summary: Holds group of commands which cater to webapp scans
+"""
+
+helps['webapp scan start-scan'] = """
+    type: command
+    short-summary: Starts the scan on the specified webapp files
```
panchagnula: Can you give more clarity here - i.e how is this command going to be useful? Say Scans a specific web app - & what does it return?

ppvasude: It will basically scan the wwwroot directory and will return a json containing the scan ID, tracking and result url.
Will add this in the short-summary.

panchagnula: Thanks - basically we want the  customers to use this command so any information we can give them to tell them that this command  is useful - is going to be good to add.

panchagnula:
Update the version, SHA & the link to who file here https://github.com/Azure/azure-cli-extensions/blob/master/src/index.json#L1916

panchagnula:
Update the version here as well https://github.com/Azure/azure-cli-extensions/blob/master/src/webapp/setup.py#L11

src/webapp/azext_webapp/__init__.py
panchagnula: Also looks like this is supported for Linux only - at least for now, so can you update the help to reflect this?

ppvasude: That's a good point! I'll update the help section. Is there any other way to make sure users use this only on Linux based webapps?

panchagnula: you can also get the site object & check if is reserved it true -> if so throw an exception saying its supported for windows only. Add this as a pre-validation before making any other API calls

panchagnula:
@tjprescott the changes look good from our perspective - can you do a quick review & merge please?

src/webapp/azext_webapp/__init__.py
```diff
@@ -20,6 +20,13 @@ def __init__(self, cli_ctx=None):
     def load_command_table(self, _):
         with self.command_group('webapp') as g:
             g.custom_command('remote-connection create', 'create_tunnel')
+
+        with self.command_group('webapp scan') as g:
+            g.custom_command('start-scan', 'start_scan')
+            g.custom_command('get-scan-result', 'get_scan_result')
+            g.custom_command('track-scan', 'track_scan')
+            g.custom_command('get-all-scan-result', 'get_all_scan_result')
```
tjprescott: These command names are wildly inconsistent with the CLI. Recommended names:

```
webapp scan start
webapp scan show (or show-result)
webapp scan track
webapp scan list (or list-result)
```

src/webapp/azext_webapp/custom.py
```diff
@@ -28,6 +29,89 @@ def _ping_scm_site(cmd, resource_group, name):
...
+    scan_result_url = scm_url + '/api/scan/results'
+
+    import urllib3
+    authorization = urllib3.util.make_headers(basic_auth='{0}:{1}'.format(user_name, password))
+    headers = authorization
+    headers['content-type'] = 'application/octet-stream'
+
+    response = requests.get(scan_result_url, headers=authorization)
+
+    return response.json()
```
tjprescott: Why are these raw API calls instead of SDK calls?

sanchitmehta: We plan to add this feature in Preview, and it is still under development. That is why these are direct API calls, and was added to CLI extensions. We plan to add SDK calls later.

src/webapp/azext_webapp/__init__.py
ppvasude: Renamed to : 
webapp scan start
webapp scan show-result
webapp scan track
webapp scan list-result

src/webapp/azext_webapp/_help.py
```diff
@@ -14,3 +14,28 @@
     type: command
     short-summary: Creates a remote connection using a tcp tunnel to your web app
 """
+
+helps['webapp scan'] = """
+    type: group
+    short-summary: Holds group of commands which cater to webapp scans. Currently available only for Linux based webapps.
+"""
+
+helps['webapp scan start-scan'] = """
```
panchagnula: Shouldn't these be updated to reflect the command changes?

ppvasude:
@tjprescott can you please take a look at the updated names and let me know?

PR #828, opened 2019-07-01
Convert Travis pipeline to ADO

tjprescott:
Converts the key jobs in the old TravisCI file to use ADO.

TODO:
- [ ] convert the final publish step


tjprescott:
> I'll be curious to see if there are any changes requited to the scripts that you reference here. When I've migrated these build scripts in the past, I uncovered quite a few TravisCI specific assumptions.

Good point. I'll keep the Travis CI checks in place for now.

PR #829, opened 2019-07-01
Fix up scripts for ADO.

tjprescott:
Get CI working on ADO.
Also, start moving stages to use azdev.


azure-pipelines.yml
```diff
@@ -35,6 +36,7 @@ jobs:
   - bash: |
       #!/usr/bin/env bash
       set -ev
+      pip install wheel==0.30.0
```
marstr: I know I've asked before, so please forgive me: the shape of this is that the extensions themselves must be built with `wheel==0.30.0`, not that the CLI must have 0.30.0 correct? I'm trying to decide if we need to add `wheel>=0.29.0,<0.31.0` to our setup.py and `wheel==0.30.0` to each of our requirements.txt files. If possible, I'd like to skip this, and instead add it as a criteria to the extensions themselves, or even better: build criteria.

tjprescott: The extensions must be built with 0.29.0 or 0.30.0.

marstr: Got it, so the best course of action is to leave wheel constraints out of the Azure CLI's `setup.py`, and instead have that criteria baked into the `dev` requirements of individual extensions.

PR #852, opened 2019-07-15
Making 'FrontDoor' Default for certificate source

pichandwork:
---

### General Guidelines

- [Yes ] Have you run `azdev style <YOUR_EXT>` locally? (`pip install azdev` required)

For new extensions:

### DIFF
--- a/src/front-door/azext_front_door/custom.py
+++ b/src/front-door/azext_front_door/custom.py
@@ -285,7 +285,7 @@ def configure_fd_frontend_endpoint_disable_https(cmd, resource_group_name, front
 
 def configure_fd_frontend_endpoint_enable_https(cmd, resource_group_name, front_door_name, item_name,
                                                 secret_name=None, secret_version=None,
-                                                certificate_source=None, vault_id=None):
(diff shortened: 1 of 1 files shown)


src/front-door/azext_front_door/custom.py
```diff
@@ -285,7 +285,7 @@ def configure_fd_frontend_endpoint_disable_https(cmd, resource_group_name, front
 
 def configure_fd_frontend_endpoint_enable_https(cmd, resource_group_name, front_door_name, item_name,
                                                 secret_name=None, secret_version=None,
-                                                certificate_source=None, vault_id=None):
+                                                certificate_source='FrontDoor', vault_id=None):
```
tjprescott: Is this a breaking change?

pichandwork: Not really, previously the default value was None and command was not doing anything(Ideally if should enable https on custom domain). This change would fix that.

pichandwork: @tjprescott

PR #873, opened 2019-07-30
Add hack extension to streamline creation of resources commonly used when hacking

GeekTrainer:
`az hack` is designed to solve one of the largest gaps we see with students at hackathons, which is creating the various resources commonly used on Azure. Specifically, this extension creates an App Services website, a database of the user's choosing, and optionally a universal key for Cognitive Services.

---

### General Guidelines

For new extensions:


src/hack/README.rst
```diff
@@ -0,0 +1,30 @@
...
+
+::
+    az hack up -n demoname -d MySQL -r python -ai
+
+--------
+Switches
+--------
+
+**--name -n**
+Name of the application. This will be used to name the Resource Group, App Service Plan, App Service, database server and database. Name must contain only letters and numbers, and start with a letter.
```
zikalino: I have one suggestion here, while resource group must be unique in a subscription, database name for instance must be globally unique, so probably when creating database it would be good to add random prefix.
but this is something that could be done later.

src/hack/azext_hack/custom.py
```diff
@@ -0,0 +1,53 @@
...
+    create_website,
+    set_website_settings,
+    create_cogsvcs_key
+)
+
+logger = get_logger(__name__)
+
+
+def hack_up(cmd, name, runtime, database, ai=None):
+    location = 'westus'
```
zikalino: we could have default location, but it would be nice to be able to change it. it may happen for some reason that particular resource is unavailable lin "westus" location.

src/hack/README.rst
GeekTrainer: Thanks for the suggestion! I'm going to update the code to have the user provide the first 10 characters as a base, and then i'll append 6 random (or something like that)

PR #884, opened 2019-08-05
Add support for connecting ACR registry during AKS create

seanmck:
This PR adds the ability to connect an ACR registry to an AKS cluster at cluster creation time by performing the service principal role assignment implicitly. Two new optional properties are added to the `create` command:

- **acrName**: the name of the registry to be attached
- **acrResourceGroup**: the name of the resource group where the ACR can be found. If not specified, it is assumed to be the same as the RG where the AKS cluster is being created.

---


seanmck:
This functionality has now been added in #891

PR #891, opened 2019-08-08
Add support for ACR role assignment for AKS update

feiskyer:
This PR adds support for ACR role assignment during AKS creation and update. 

Usage: 

* az aks create –n <name> -g <resource-group> --enable-acr --acr <name-or-id>
* az aks update –n <name> -g <resource-group> --enable-acr --acr <name-or-id>

### DIFF
--- a/src/aks-preview/azext_aks_preview/_help.py
+++ b/src/aks-preview/azext_aks_preview/_help.py
@@ -162,10 +162,10 @@ helps['aks create'] = """
           short-summary: The node resource group is the resource group where all customer's resources will be created in, such as virtual machines.
         - name: --enable-acr
           type: bool
-          short-summary: Grant the 'acrpull' role assignment for the ACR set by --acr-name.
-        - name: --acr-name
+          short-summary: (PREVIEW) Grant the 'acrpull' role assignment for the ACR set by --acr.
+        - name: --acr
           type: string
-          short-summary: The name of the ACR in AKS resource group. If it's empty and --enable-acr is true, then a new ACR with name 'aks<resource-group>acr' would be created.
+          short-summary: (PREVIEW) ACR name in AKS resource group or ACR resource ID. If it's empty and --enable-acr is true, then a new ACR with name 'aks<resource-group>acr' would be created.
     examples:
         - name: Create a Kubernetes cluster with an existing SSH public key.
           text: az aks create -g MyResourceGroup -n MyManagedCluster --ssh-key-value /path/to/publickey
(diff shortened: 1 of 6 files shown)


src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -72,8 +72,9 @@ def load_arguments(self, _):
         c.argument('node_zones', zones_type, options_list='--node-zones', help='(PREVIEW) Space-separated list of availability zones where agent nodes will be placed.')
         c.argument('enable_pod_security_policy', action='store_true')
         c.argument('node_resource_group')
-        c.argument('enable_acr', is_preview=True)
-        c.argument('acr_name', is_preview=True)
```
feiskyer: @yugangw-msft I removed the `is_preview=True` because its documentation is in different formats compared to other preview arguments.

yugangw-msft: I see.
On other option is can you fix the other preview related arguments, instead of continue the inconsistency with the rest command modules?

feiskyer: I'd prefer to do it in a separate PR if it's something we want to do.

feiskyer: For current changes, I think keeping it in the same style as others is better.

feiskyer:
@yugangw-msft @zqingqing1 Ready for review. Could you help to take a look?

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -227,6 +230,15 @@
         - name: --disable-pod-security-policy
           type: bool
           short-summary: (PREVIEW) Disable pod security policy.
+        - name: --enable-acr
+          type: bool
+          short-summary: (PREVIEW) Grant the 'acrpull' role assignment for the ACR set by --acr-name or --acr-resource-id.
+        - name: --acr-name
+          type: string
+          short-summary: (PREVIEW) The name of the ACR in AKS resource group.
+        - name: --acr-resource-id
```
yugangw-msft: the convention is to combine `acr_name` and `azr_resource_id` into `acr`. It is very easy to detect if it is not a resource id then it is a name. Check out how VM is doing [here](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/vm/_validators.py#L45)

zqingqing1: agree, It will be concise.

feiskyer: The user interface design is from PM side. I think using different options is also more clear

cc @sauryadas

sauryadas: works for me .. this is concise and we of course should follow conventions. I did not know this existed.
@feiskyer in the CLI help , the help text will suggest that --acr can take either acr-name or acr-id .. please update the description

feiskyer: Thanks, would merge these two options into `--acr`.

feiskyer:
@yugangw-msft @zqingqing1 Addressed comments. PTAL

sauryadas:
LGTM

@yugangw-msft anything pending to merge this request?

sauryadas:
@tjprescott  Can you please help merge this while Yugang is out ? It is high priority work and this change needs to go into the CMU fall semester lab docs.

feiskyer:
@zikalino thanks

zikalino:
all the tests passed, however the status was not reported here for some reason. I am going to merge this pr.

PR #904, opened 2019-08-15
Remove hardcoded en-us locale from docs site links.

sptramer:


MyronFanQiu:
@yugangw-msft @tjprescott @zikalino Hello. There is a failure in the CI which is to check the equality between the metadata of .whl package and the metadata of the index.json. I downloaded the [.whl](https://azurecliaks.blob.core.windows.net/azure-cli-extension/aks_preview-0.4.8-py2.py3-none-any.whl) and manually checked them. They are same. I saw some words about the cache. Do we need to clear the cache and how?

MyronFanQiu:
@sptramer Hello. Please fix the content in this url. https://azurecliext.blob.core.windows.net/release/azure_cli_ml-1.0.55-py2.py3-none-any.whl

src/azure-firewall/azext_firewall/vendored_sdks/v2019_04_01/models/_models.py
```diff
@@ -3182,7 +3182,7 @@ class BGPCommunity(Model):
     :type community_name: str
     :param community_value: The value of the bgp community. For more
      information:
```
yugangw-msft: I suggest we back out changes under folders of `verndored_sdks` as those are generated files from swagger specs, and the next round of code-gen will wipe out the change from this PR. If we like to address the root cause, we should update [specs](https://github.com/Azure/azure-rest-api-specs/tree/master/specification).

yugangw-msft:
> https://azurecliext.blob.core.windows.net/release/azure_cli_ml-1.0.55-py2.py3-none-any.whl

@MyronFanQiu, this extension's source code us hosted at external repository. CC: @gogowings

src/azure-firewall/azext_firewall/vendored_sdks/v2019_04_01/models/_models.py
sptramer: @yugangw-msft If everything is generated from the Swagger specs, then this PR should be closed and we can address the root problem (which is the Swagger.) Otherwise feel free to merge it - I've taken an item to look into hardcoded `en-us` in Swagger.

sptramer:
@yugangw-msft I noticed this hadn't been merged and looked into the error happening on CI. I found that it's actually because the test is expecting `en-us` to be in the link:

```
AssertionError: {'aze[553 chars].com/en-us/azure/machine-learning/service/'}}}[526 chars].57'} != {'aze[553 chars].com/azure/machine-learning/service/'}}}, 'ext[520 chars].57'}
```

It looks like this is because it's comparing the non-`en-us` link against the currently published information, which is `en-us`. That means it looks like this test doesn't take into account the possibility that reference links were incorrect and need to be fixed.

I'd like to request that this PR be merged as a result, since I'm not sure of a good way to fix this test.

yugangw-msft:
CC @zikalino owner of CLI extension repository, and @gogowings who owns the ML extension.
The ML extension's source code is hosted in an external repository. I would suggest you back out the specific line change on the ML, open an issue for @gogowings to follow up. 
The general suggestion is we should try to avoid merging PRs which would break CI as that would incur maintenance cost. 
Of course, if @gogowings feels it easy to produce a new extension, he can submit a PR, we merge and you can rebase.

sptramer:
If I can get a URL for the ML extension, I can make a similar modification there. But I'm not sure why part of this repository's test infrastructure would be checking something from a different, disconnected repo.

gogowings:
Sorry for the delayed response as I am just back from vacation. As @yugangw-msft mentioned, azure-cli-ml is from an external repo and released separately. We usually release every two weeks and should have a new release this week. We can make the change for this week's CLI release so before end of the week, this will be fixed for the azure-cli-ml's project_urls, and then you can rebase.

Sounds like a plan?

sptramer:
@gogowings Sure, that's fine. Ping me in this thread again when that's fixed so I can close and reopen to force a rebuild.

gogowings:
The azure-cli-ml fix is in https://github.com/Azure/azure-cli-extensions/pull/937, waiting from approval now. FYI: @sptramer

gogowings:
#937 is merged.

sptramer:
@zikalino This should now be ready to merge. Thanks!

PR #910, opened 2019-08-19
support for target name and multiple locations

d3vil-st:
---

### General Guidelines

### DIFF
--- a/src/image-copy/azext_imagecopy/custom.py
+++ b/src/image-copy/azext_imagecopy/custom.py
@@ -131,11 +131,16 @@ def imagecopy(source_resource_group_name, source_object_name, target_location,
 
         tasks = []
         for location in target_location:
+            # Appending location to target name if multiple locations
+            final_target_name = target_name
+            if target_name and target_locations_count > 1:
+                final_target_name = target_name + location
+
             location = location.strip()
             tasks.append((location, transient_resource_group_name, source_type,
                           source_object_name, source_os_disk_snapshot_name, source_os_disk_snapshot_url,
                           source_os_type, target_resource_group_name, azure_pool_frequency,
-                          tags, target_name, target_subscription, export_as_snapshot, timeout))
+                          tags, final_target_name, target_subscription, export_as_snapshot, timeout))
 
         logger.warn("Starting async process for all locations")
 
--- a/src/image-copy/azext_imagecopy/custom.py
+++ b/src/image-copy/azext_imagecopy/custom.py
@@ -134,7 +134,7 @@ def imagecopy(source_resource_group_name, source_object_name, target_location,
             # Appending location to target name if multiple locations
             final_target_name = target_name
             if target_name and target_locations_count > 1:
-                final_target_name = target_name + location
+                final_target_name = target_name + '-' + location
 
             location = location.strip()
             tasks.append((location, transient_resource_group_name, source_type,
--- a/src/image-copy/setup.py
+++ b/src/image-copy/setup.py
@@ -8,7 +8,7 @@
 from codecs import open
 from setuptools import setup, find_packages
 
-VERSION = "0.2.1"
+VERSION = "0.2.2"
 
 CLASSIFIERS = [
     'Development Status :: 4 - Beta',


tamirkamara:
@d3vil-st 
what do you think about adding an option to enable this feature instead of relying on the number of targets?

d3vil-st:
> 
> 
> @d3vil-st
> what do you think about adding an option to enable this feature instead of relying on the number of targets?

Hey, @tamirkamara I don't mind, but the default behavior if we specifying multiple locations and do not specifying name of target, is appending a suffix.
In my opinion it should be the same behavior if we does, because now it throwing an error as image name already exists with the same name.

tamirkamara:
Ok, I understand the scenario better now. I don't think an option is required... Let's leave it as you have now.

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -131,11 +131,16 @@ def imagecopy(source_resource_group_name, source_object_name, target_location,
 
         tasks = []
         for location in target_location:
+            # Appending location to target name if multiple locations
+            final_target_name = target_name
+            if target_name and target_locations_count > 1:
+                final_target_name = target_name + location
```
tamirkamara: Let's be consistent with how it's done in the create_target method

```suggestion
                final_target_name = target_name + '-' + location
```

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -131,11 +131,16 @@ def imagecopy(source_resource_group_name, source_object_name, target_location,
 
         tasks = []
         for location in target_location:
+            # Appending location to target name if multiple locations
+            final_target_name = target_name
```
tamirkamara: target_name is optional. I think this new code block should only work when the value isn't None. 
What do you think?

src/image-copy/azext_imagecopy/custom.py
d3vil-st: I've do in that way in first time, but what if user would like to name image like that ImageNameeastus2 instead of appending '-'
Just providing more flexibility.

src/image-copy/azext_imagecopy/custom.py
d3vil-st: Do you mean to add another 'if' block to that?

tamirkamara: yes

src/image-copy/azext_imagecopy/custom.py
tamirkamara: Since we're adding the location to the name automatically, it's up to us to decide how we want the name to be formatted, it's not the user's decision... 
In the other method, that decision has been made already, and I think we should be consistent.

tamirkamara: Also consider that if the user doesn't provide a target_name, then even before your change the name of the image in each local will be something like:
xyz123-useast2

src/image-copy/azext_imagecopy/custom.py
d3vil-st: > 
> 
> yes

But next tasks.append function calling with final_target_name as argument and it could be undefined, so that why I unconditionally assign it.

src/image-copy/azext_imagecopy/custom.py
d3vil-st: Ok, that seems legit to me.

src/image-copy/azext_imagecopy/custom.py
tamirkamara: If target_name wasn't set by the user, meaning it's None, something will break - either on this method or in the create_target one.

You have 2 option:
1. start by setting final_target_name to target_name and then build the if around final_target_name
2. remove final_target_name and just overwrite target_name when appropriate

If it were me, I'd go with the second option. But both options should work IMO.

d3vil-st: Modifying of original target_name will have effect on next iteration how you suggest to resolve that?

d3vil-st: @tamirkamara so I do not understand what it breaking now,
If target_name is None, then final_target_name will also None and passed to array.
If target_name is set and count of target location is one, then it will pass original name to array as well.
If target_name and count of target location more then one, if will append suffix.
final_target_name - is temporary variable just to not modify target_name over iterations.

PR #948, opened 2019-09-11
add support for ccp only upgrade and default to upgrade all

xizhamsft:
With the new 0801 api version, we now supports control plane only cluster upgrade. Added the control-plane-only argument for this purpose and change the default to upgrade all (by setting kubernetes version of all agentpool profiles explicitly).

---

### General Guidelines

For new extensions:

### DIFF
--- a/src/aks-preview/azext_aks_preview/custom.py
+++ b/src/aks-preview/azext_aks_preview/custom.py
@@ -1096,7 +1096,8 @@ def aks_scale(cmd, client, resource_group_name, name, node_count, nodepool_name=
     raise CLIError('The nodepool "{}" was not found.'.format(nodepool_name))
 
 
-def aks_upgrade(cmd, client, resource_group_name, name, kubernetes_version, no_wait=False, **kwargs):  # pylint: disable=unused-argument
+def aks_upgrade(cmd, client, resource_group_name, name, kubernetes_version,
+                control_plane_only=False, no_wait=False, **kwargs):  # pylint: disable=unused-argument
     instance = client.get(resource_group_name, name)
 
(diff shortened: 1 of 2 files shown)


src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1112,8 +1113,43 @@ def aks_upgrade(cmd, client, resource_group_name, name, kubernetes_version, no_w
...
                            "attempt resolution of failed cluster state.", instance.kubernetes_version)
 
+    from knack.prompting import prompt_y_n
+
+    upgrade_all = False
     instance.kubernetes_version = kubernetes_version
 
+    vmas_cluster = False
+    for agent_profile in instance.agent_pool_profiles:
+        if agent_profile.type == "AvailabilitySet":
```
xuto2: case insensitive compare

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1112,8 +1113,43 @@ def aks_upgrade(cmd, client, resource_group_name, name, kubernetes_version, no_w
...
+            vmas_cluster = True
+            print("VMAS cluster")
```
xuto2: remove

xuto2:
check https://github.com/Azure/azure-cli/pull/10513/files#diff-fe2d1dd3c9fc1d841b3437f1a9aac146 for main CLI sync

zikalino:
ok, decided to merge it:
actually I was wondering if we could handle following scenarios, when user specified:

--control-plane-only

and
--control-plane-only False

specifically to avoid displaying prompt.

But this can be considered / checked later.

PR #963, opened 2019-09-24
Update Azure CLI to support new version of frontdoor (2019-05-01)

pichandwork:
---

### General Guidelines

- [ Yes] Have you run `azdev style <YOUR_EXT>` locally? (`pip install azdev` required)

For new extensions:


src/front-door/azext_front_door/commands.py
```diff
@@ -89,6 +89,9 @@ def load_command_table(self, _):
     #     g.show_command('show', 'get')
     #     g.generic_update_command('update', custom_func_name='update_fd_backend_pool')
 
+    with self.command_group('network front-door probe', frontdoor_sdk) as g:
+        g.custom_command('update', 'update_fd_health_probe_settings')
```
MyronFanQiu: Could we use `generic_update_command`

MyronFanQiu: Lot's of commands are commented out like `network front-door probe create\delete`. I'm the new maintainer for network part. Could you give me some context about these commands?

src/front-door/azext_front_door/_params.py
```diff
@@ -65,9 +65,11 @@ def load_arguments(self, _):
         c.argument('secret_name', help='The name of the Key Vault secret representing the full certificate PFX')
         c.argument('secret_version', help='The version of the Key Vault secret representing the full certificate PFX')
         c.argument('vault_id', help='The resource id of the Key Vault containing the SSL certificate')
+        c.argument('minimum_tls_version', arg_type=get_enum_type(MinimumTLSVersion), help='The minimum TLS version required from the clients to establish an SSL handshake with Front Door.')
 
     with self.argument_context('network front-door', arg_group='BackendPools Settings') as c:
         c.argument('enforce_certificate_name_check', arg_type=get_three_state_flag(positive_label='Enabled', negative_label='Disabled', return_label=True), help='Whether to disable certificate name check on HTTPS requests to all backend pools. No effect on non-HTTPS requests.')
+        c.argument('send_recv_timeout_seconds', type=int, help='Send and receive timeout on forwarding request to the backend. When timeout is reached, the request fails and returns.')
```
MyronFanQiu: In general, DO NOT put units in argument names. ALWAYS put the expected units in the help text. Example: --duration-in-minutes should simply be --duration. This prevents the need to add more arguments later if more units are supported.
https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#argument-naming-conventions

src/front-door/azext_front_door/custom.py
```diff
@@ -423,25 +430,44 @@ def remove_fd_backend(cmd, resource_group_name, front_door_name, backend_pool_na
     client.create_or_update(resource_group_name, front_door_name, frontdoor).result()
 
 
-def create_fd_health_probe_settings(cmd, resource_group_name, front_door_name, item_name, path, interval,
-                                    protocol=None):
+def create_fd_health_probe_settings(cmd, resource_group_name, front_door_name, item_name, probe_path, probe_interval,
```
MyronFanQiu: Is this command usable before this PR?

src/front-door/azext_front_door/commands.py
pichandwork: We dont support create/delete/update on probe, routingrules, frontend-endpoint. Initial swagger incorrectly stated that these are supported, which we fixed recently for all versions.

src/front-door/azext_front_door/custom.py
pichandwork: Yes. interval and path were options_list in _params.py
c.argument('probe_protocol', options_list='--protocol', arg_type=get_enum_type(FrontDoorProtocol), help='Protocol to use for sending probes.')
        c.argument('probe_interval', options_list='--interval', type=int, help='Interval in seconds between probes.')

src/front-door/azext_front_door/commands.py
MyronFanQiu: @pichandwork Hello. Could we use generic_update_command instead of custom_command?

pichandwork: @MyronFanQiu  My understanding is generic_update_command calls a PUT on the sub resource(in this case HealthProbeSettings). But PUT is not supported on HealthProbeSettings REST API. To update any settings in frontdoor, GET all settings from frontdoor, change the required values(HealthProbeSettings etc), then call PUT on frontdoor resource. So, thats why i had to use a custom command.

yonzhan:
@MyronFanQiu  will help you review.

src/front-door/azext_front_door/commands.py
MyronFanQiu: @pichandwork generic_update_command will call a GET method by design to get an instance. Then you can change the values. Then it will use a PUT method to return the instance to the server. The logic is almost same with what you write. So I think it's better to use `generic_update_command `. https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#generic-update-commands.
`generic_update_command` will expose some general arguments for user. User can set some fields by `--set` in case you don't expose them to user.

pichandwork: @MyronFanQiu - Which instance do you refer to when you say GET method to get an **instance**? Frontdoor instance or HealthProbeSettings instance? if its later(HealthProbeSettings), then there is no GET and PUT supported on this.

Issue #979, opened 2019-10-03
az image copy fails with 'command ended with an error'


RickKra:
**Command Name**
`az image copy
Extension Name: image-copy-extension. Version: 0.2.1.`

**Errors:**
```
Starting async process for all locations
westus2 - Creating target storage account (can be slow sometimes)
command ended with an error: ['C:\\Program Files (x86)\\Microsoft SDKs\\Azure\\CLI2\\python.exe', '-m', 'azure.cli', 'storage', 'account', 'create', '--name', 'westus2c7eafby7ltzpp1393', '--resource-group', 'image-copy-Nxs2019100102', '--location', 'westus2', '--sku', 'Standard_LRS', '--output', 'json', '--subscription', 'vsclk-core-dev', '--tags', 'created_by=image-copy-extension']
Deleting transient resources
```
> I tried it with --debug and that didn't give any additional info about the error.

## To Reproduce:
```
az image copy --target-subscription vsclk-core-dev --target-resource-group vsclk-online-dev --target-name Nxs2019100102 --target-location westus2 --source-resource-group VSNexusImages --source-object-name Nxs2019100102 --cleanup --timeout 36000 --temporary_resource_group_name image-copy-Nxs2019100102
```

## Expected Behavior

Copies the image to the target sub/rg.

## Environment Summary
```
Windows-10-10.0.18362-SP0
Python 3.6.6
Shell: cmd.exe

azure-cli 2.0.74

Extensions:
azure-devops 0.12.0
image-copy-extension 0.2.1

```
## Additional Context

I was getting this failure in an Azure DevOps release pipeline using the Azure CLI task.

I thought at first it might be related to using an Azure service principal or an Azure pipeline to do this, because I've been able to successfully do this myself locally.

However, one of the last mitigations I attempted was to update the Azure CLI locally and try that. I went from 2.0.70 to 2.0.74, and now I'm getting the same error locally.

The Azure CLI task was using 2.0.46 and also failing, so I tried updating the CLI using the automation suggested at https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-windows?view=azure-cli-latest#install-or-update. That updated the CLI on the build agent to 2.0.74 and it also failed there again.

Hoping you've got a way to install the 2.0.70 CLI via automation so I can work around this until you fix it?

Thanks,
Rick

yonzhan:
@qwordy  please take a look and comment.

qwordy:
[...]

PR #1006, opened 2019-10-18
added temp cleanup for warning string

jmhardison:
---

### General Guidelines

For new extensions:

This PR adds a temporary fix for resolving image-copy's failures related to the az cli command now outputting `WARNING: The default kind for created storage account will change to 'StorageV2' from 'Storage' in future`, which is causes the command output to have invalid JSON. See issue #979 for details, and https://github.com/Azure/azure-cli/issues/10687 for the azure-cli side issue.

In the future, if azure-cli removes the warning, or alters it to insert the warning as a value in the JSON command output, then this temporary fix could be reverted.

This can also be manually inserted by a user into their local install rather that pulling it into the project to remove it later. I'll leave that to you all as to the desire approach.


yonzhan:
@zikalino will help review.

src/image-copy/azext_imagecopy/cli_utils.py
```diff
@@ -23,6 +23,11 @@ def run_cli_command(cmd, return_as_json=False):
 
         if return_as_json:
             if cmd_output:
+                # temp cleanup to deal with new warning string  - issue #979
+                # Related to: https://github.com/Azure/azure-cli/issues/10687
+                warn_string = "WARNING: The default kind for created storage account will change to 'StorageV2' from 'Storage' in future"
+                cmd_output = cmd_output.replace(warn_string, "")
+                # end of temp cleanup
```
tamirkamara: Although this is perfectly fine, IMO you can make the fix more robust by removing all chars before the first "{" char. What do you think?

jmhardison: you would be correct, this is overly explicit to the problem. I'll adjust this and provide an update.

Thanks!

jmhardison: I've modified it a bit to handle any text before the payload that is valid. However, we have to handle [ or { in that case, as the account keys for example returns an array.
Also added in debug logging for when it fires.

And as for that force-push, I just rebuilt my desktop and didn't set my email prior to commit. So I amended the author on the last commit.

tamirkamara: What do you think about something like this?

```phthon
        if return_as_json:
            if cmd_output:
                json_start_chars = {"{", "["}
                json_start_at = next((i for i, ch in enumerate(str(cmd_output)) if ch in json_start_chars), -1)
                logger.debug('json starts at position: %i ', json_start_at)
                if json_start_at > 0:
                    logger.debug("json output did not start at position 0, stripping the prefix.")
                    cmd_output = str(cmd_output)[json_start_at:]
                    logger.debug("after fix:")
                    logger.debug(cmd_output)
                json_output = json.loads(cmd_output)
                return json_output
```

jmhardison: I like your approach there. I had thought about doing it off of position, wound up going the other way.

After chewing on mine a bit, there is a case it won't handle that yours will. I haven't seen that case returned by anything used by this extension yet. But, better to not assume it won't ever happen. :)

(Case in point, this code block can handle JSON objects, arrays of objects, and arrays of values where mine cannot handle arrays of values.)

I've pulled in the changes, with one minor change to the debug message to make them begin uniformly. Just in case anyone wants to grep that output for 'json output' simply.

Thanks for the help and great feedback!

jmhardison:
I can fire an update  branch to kick it off again / push a new commit empty commit / etc, but since travis doesn't publish to my knowledge a command, and there isn't a bot to use to rerun the travis-ci jobs. Someone with write access could go rerun it. 
Looks like they all failed for the same reason:
```
remote: Invalid username or password.
fatal: Authentication failed for 'https://github.com/Azure/azure-cli.git/'
```

jmhardison:
@tamirkamara CI successful now.

tamirkamara:
@zikalino can you merge?

PR #1024, opened 2019-10-28
Add extension ip-group for feature requested in 10726

haroldrandom:
Add extension ip-group for azure-cli feature requested in [10726](https://github.com/Azure/azure-cli/issues/10726)

---

### General Guidelines

For new extensions:


src/ip-group/azext_ip_group/_params.py
```diff
@@ -0,0 +1,21 @@
...
+def load_arguments(self, _):
+
+    # region IpGroup
+    with self.argument_context('network ip-group', min_api='2019-09-01') as c:
+        c.argument('location', get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
+        c.argument('tags', tags_type)
+        c.argument('ip_groups_name', options_list=['--name', '-n'], help='Name of the IpGroup')
+
+    with self.argument_context('network ip-group create', min_api='2019-09-01') as c:
+        c.argument('ip_addresses', nargs='+', help='Space-separated list of IpAddress or IpAddressPrefix.')
```
MyronFanQiu: so, does ip_address can be updated or not?

src/ip-group/setup.py
```diff
@@ -0,0 +1,60 @@
...
+    'Programming Language :: Python :: 3.5',
+    'Programming Language :: Python :: 3.6',
+    'Programming Language :: Python :: 3.7'
+    'License :: OSI Approved :: MIT License',
+]
+
+
+DEPENDENCIES = [
+    'azure-cli-core'
+    'azure-mgmt-network~=7.0.0'
```
MyronFanQiu: I think we don't need this dependency right?

src/ip-group/setup.py
```diff
@@ -0,0 +1,60 @@
...
+    'Programming Language :: Python :: 3.4',
+    'Programming Language :: Python :: 3.5',
+    'Programming Language :: Python :: 3.6',
+    'Programming Language :: Python :: 3.7'
+    'License :: OSI Approved :: MIT License',
+]
+
+
+DEPENDENCIES = [
+    'azure-cli-core'
```
MyronFanQiu: actually, this one is also unnecessary. But it's ok to keep it

zikalino: I think it's better to remove now. it always causes some problems later when releasing....

src/ip-group/azext_ip_group/_params.py
haroldrandom: According to the examples, https://github.com/Azure/azure-rest-api-specs/tree/network-october-release/specification/network/resource-manager/Microsoft.Network/stable/2019-09-01/examples. It only provides an IpGroupsUpdateByTags example.

src/ip-group/azext_ip_group/custom.py
```diff
@@ -25,8 +25,12 @@ def list_ip_groups(cmd, resource_group_name=None):
     return client.list()
 
 
-def update_ip_groups(cmd, instance, tags=None):
+def update_ip_groups(cmd, instance, ip_addresses=None, tags=None):
     with cmd.update_context(instance) as c:
         c.set_param('tags', tags)
+
+        if ip_addresses is not None:
+            c.set_param('ip_addresses', ip_addresses)
```
MyronFanQiu: c.set_param will check `if ip_addresses is not None`. We don't need `if ip_addresses is not None`

PR #1030, opened 2019-10-29
aks-preview extension release v0.4.19

bingosummer:
---

### General Guidelines

For new extensions:

### DIFF
--- a/src/aks-preview/HISTORY.md
+++ b/src/aks-preview/HISTORY.md
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+0.4.19
++++++
+* Refactor and remove a custom way of getting subscriptions
+
 0.4.18
 +++++
 * Update to use 2019-10-01 api-version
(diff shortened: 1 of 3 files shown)


bingosummer:
@zikalino @zqingqing1 please have a look.

zikalino:
@bingosummer no need to have 2 entries in index.json, you can remove the old one

bingosummer:
> @bingosummer no need to have 2 entries in index.json, you can remove the old one
> just to make sure you can compare them to make sure there's nothing specific the new entry is using.
> 
> we usually can have 2 entries, if the new release doesn't support some older version of Azure CLI core, and I think it's not the case here.

Got your point. Have removed the old one.

bingosummer:
@zikalino Thanks for your kindly review!
Just out of curiosity, if we keep only the latest version in index.json, can customers install the old versions of extension?

PR #1040, opened 2019-10-30
update ipwl to ga for update and create and fix linter warnings

gtracer:
---

### General Guidelines

For new extensions:


src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -358,12 +368,9 @@
           type: int
           short-summary: The maximum number of pods deployable to a node.
           long-summary: If not specified, defaults to 110, or 30 for advanced networking configurations.
-        - name: --node-zones
```
robbiezhang: --node-zones is removed from the azure-cli. please also remove it here.

robbiezhang:
Just 1 comment about removing the --node-zones. otherwise, LGTM

src/aks-preview/azext_aks_preview/_help.py
adilliadil: We agreed with @jluk that we would do the removal in two steps.
1. add --zones -z as possible parameters also indicate in the help that --node-zones would be deprecated. (PR already merged)
2. remove --node-zones (will send out another PR)
this is for the purpose that the customers are aware of the changes we make to extension and have time to make the adjustments.

ruchikaguptaa:
LGTM.

PR #1096, opened 2019-11-14
DevSpaces: Passing do_not_prompt option to Mac/Linux install scripts

DrEsteban:
Our Mac/Linux scripts need to know if `-y` was passed, so they can perform unattended installs.


src/dev-spaces/azext_dev_spaces/custom.py
```diff
@@ -115,6 +115,8 @@ def _install_dev_spaces_cli(force_install):
     else:
         raise CLIError('Platform not supported: {}.'.format(system))
 
+    if do_not_prompt:
+        setup_args.append('-y')
```
rakeshvanga: Pass in /quiet for windows to install silently.
`"C:\Users\rareddyv\Downloads\Azure Dev Spaces CLI.exe" /quiet` . I think this process runs as an admin when used on windows machine, if not windows OS will show the UAC to confirm the installation to proceed forward.

DrEsteban: 👍

PR #1108, opened 2019-11-22
[image-copy] version 0.2.2

tamirkamara:
Image copy 0.2.2:
- Disregard storage module warning even with json outputs
- When a single target region is requested or parallel_degree is set to 1, then directly call the method to create the target image without using an async Pool.
- log.warn -> log.warning

---

### General Guidelines

For new extensions:


fengzhou-msft:
Please fix travis error:
> -  FAIL: missing_parameter_help
>     Parameter: image copy, `tags` - Missing help

tamirkamara:
@fengzhou-msft I don't declare an argument call tags, but only have that in the method signature to be able to use the default-builtin parameter.

haroldrandom:
There is a lock of helo message about the new command

PR #1112, opened 2019-11-26
[Storage-preview] Bug fixes for ADLS Gen2

Juliehzl:
---
- `az storage blob list` list with directory flag
- `az storage blob move` add blob check for source
- `az storage blob directory move` add directory check for source and add --move-mode
- `az storage blob directory create` refine arguments name and help message

### General Guidelines

For new extensions:


jsntcy:
Which bugs does this PR fix? Could you please add more details in PR description?

src/storage-preview/azext_storage_preview/_help.py
```diff
@@ -207,6 +207,8 @@
     examples:
         - name: Create a storage blob directory in a storage container.
           text: az storage blob directory create -c MyContainer -d MyDirectoryPath --account-name MyStorageAccount
+        - name: Create a storage blob directory with permissions and umask.
+          text: az storage blob directory create -c MyContainer -d MyDirectoryPath --account-name MyStorageAccount --permissions rwxrwxrwx --umask 0000
```
haroldrandom: Is it dup? --permision and --umask. 

I think they do the same job but just differentiate in format.

src/storage-preview/azext_storage_preview/_validators.py
```diff
@@ -173,11 +174,25 @@ def validate_blob_directory_download_source_url(cmd, namespace):
...
+    destination = destination[1:] if destination.startswith('/') else destination
+    from azure.common import AzureException
+    try:
+        props = client.get_blob_properties(container, destination)
+        if not is_directory(props):
+            raise ValueError('usage error: You are specifying --destination-path with a blob name, not directory name. '
+                             'Please change to a valid blob directory name. If you want to upload to a blob file, '
+                             'please use `az storage blob upload` command.')
+    except AzureException:
+        pass
```
haroldrandom: 1. Can AzureException catch ValueError?
2. Why let the exception pass?

src/storage-preview/azext_storage_preview/_help.py
Juliehzl: No. They will cooperate. Th final permission setting will be permissions & ^umask.

src/storage-preview/azext_storage_preview/_validators.py
Juliehzl: 1. NO
2. When source doesn't exist, AzureException will be thrown but we can continue to create a new one for user. Therefore I let it pass.

src/storage-preview/azext_storage_preview/_help.py
haroldrandom: emmm... Interesting...

PR #1137, opened 2019-12-09
vm-repair extension URGENT bug fix for os_image_reference.version None case 

swbae31:
- Fixes errors when source VM's os image version is None.
=> Clients waiting for this issue to be resolved.

- Now logging stack trace app insights
- fixed minor scenario testing bugs


swbae31:
Travis is failing with the error "CLIError: Please run 'az login' to setup account." on the scenario tests. They run fine locally. What is the issue here? @zikalino

yonzhan:
@haroldrandom  please help with the PR.

haroldrandom:
src/vm-repair/azext_vm_repair/tests/latest/test_repair_commands.py has been modified but without updating recoding YAML files.

src/index.json
```diff
@@ -2582,9 +2582,9 @@
                     "metadata_version": "2.0",
                     "name": "vm-repair",
                     "summary": "Auto repair commands to fix VMs.",
-                    "version": "0.2.3"
+                    "version": "0.2.4"
                 },
-                "sha256Digest": "1dadfc622001eab35a7f8e0ecd8c02b75808e1db823fdc44b1ee5c02e9bbdf10"
```
qwordy: Is the bug fix for os_image_reference.version None case in this version of  vm_repair?

haroldrandom:
> > src/vm-repair/azext_vm_repair/tests/latest/test_repair_commands.py has been modified but without updating recoding YAML files.
> 
> The recording files were removed a few PRs back because the recorded tests will not work for this extension's case because it uses subprocess to run other CLI commands.

I believe your midification on  _test_repair_commands.py_ has effected tests.

One way to test is that try to modify _scripts/ci/test_source.py_ to test only your extensions to see if the output like CI or not.
The _scripts/ci/test_source.py_ is try to run `python -m unittest discover -v {YOUR_EXTENSION_PATH}`

BTW, if test is not working, why not comment them out or delete the _test_repair_commands.py_

src/index.json
swbae31: yes. all the changes within repair_utils.py

swbae31:
> src/vm-repair/azext_vm_repair/tests/latest/test_repair_commands.py has been modified but without updating recoding YAML files.
> 
> The recording files were removed a few PRs back because the recorded tests will not work for this extension's case because it uses subprocess to run other CLI commands.
> 
> I believe your midification on  test_repair_commands.py has effected tests.
> One way to test is that try to modify scripts/ci/test_source.py to test only your extensions to see if the output like CI or not.
> The scripts/ci/test_source.py is try to run python -m unittest discover -v {YOUR_EXTENSION_PATH} on two situations: when you login and when you are not login.
> Please share your results.
> BTW, if test is not working, why not comment them out or delete the test_repair_commands.py

Test Results

1) When Logged in: Works as expected without failure

2) When logged out: I get the error:
**knack.util.CLIError: Please run 'az login' to setup account.**
for all tests.
-> Why am i running the tests while logged out? Isn't the CI set up so that this is taken care of while running the tests live? 

3) I am keeping these tests because I need to run them live before pushing out a new version.

haroldrandom:
CI won't be able to run live without credentias after you login. 

I would suggest you those uncessary tests our or add recoding files together since they are are passed without failure now.

swbae31:
> CI won't be able to run live without credentias after you login.
> I would suggest you those uncessary tests our or add recoding files together since they are are passed without failure now.

Harold, why do you say these tests are unnecessary. We need to run these scenario tests to run live before merging to validate them. These tests have been in the repository from the first version of this extension. It is not like I just added them in this PR. Why is it failing all of a sudden? 

Also it is indicated https://github.com/Azure/azure-cli/blob/dev/doc/authoring_tests.md#nightly-live-test-run here that **tests are run live through CI.** 

Please a client is still waiting for this PR to go in from last week. Is there any way to speed up this process? We are waiting 10 hours for each other's response due to the time zone difference.

haroldrandom:
2 Solutions:
1. First live-run test on your local machine. If all test pass, run it again in non-live mode. If it pass too, upload your recording YAML files to see CI pass or not
2. Subclass LiveScenarioTest instead of ScenarioTest. let CI skip your tests.

swbae31:
Updated the test class to inherit LiveScenarioTest not ScenarioTest. This will make the CI skip recorded testing and pass the tests. Thanks for all the help @haroldrandom

PR #1160, opened 2019-12-21
vm-repair: Quick fix for users who does not have 'json' as default output

swbae31:
- Fixes bug when the user's default CLI output is not json. The bug led to json parsing errors.


qwordy:
Can you share which command's output is not in JSON format?

swbae31:
@qwordy all the command outputs follow the user's option of output. The bug happens for internal CLI calls where the code forces json formats. These are not shown to the user.

PR #1164, opened 2019-12-26
[Breadth Coverage] blueprint

fengzhou-msft:
---

### General Guidelines

For new extensions:


src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+template_type = CLIArgumentType(
+    options_list=['--template'],
+    help='ARM template in JSON format or file path to Json file with "@" prefix.'
+)
+
+
+def load_arguments(self, _):
+
+    with self.argument_context('blueprint', validator=scope_validator) as c:
+        c.argument('management_group', help='The management group for the scope of the resource. The scope could either be a subscription or a management group. For blueprint assignments management group scope is reserved for future use. If management-group is not specified, will use subscription as scope')
```
Juliehzl: For management_group parameter, should it be manage group name or ID? And the help message is a little confusing with "The scope could be a subscription or a management group...".

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+        c.ignore('scope')
+
+    with self.argument_context('blueprint create') as c:
+        c.argument('blueprint_name', options_list=['--name', '-n'], help='Name of the blueprint definition.')
+        c.argument('display_name', help='One-liner string explain this resource.')
+        c.argument('description', help='Multi-line explain this resource.')
+        c.argument('target_scope', arg_type=get_enum_type(['subscription', 'managementGroup']), help='The scope where this blueprint definition can be assigned.')
```
Juliehzl: For the enum values, you can consider utilizing the BlueprintTargetScope in src/blueprint/azext_blueprint/vendored_sdks/blueprint/models/_blueprint_management_client_enums.py.

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+        c.argument('parameters', arg_type=parameter_type, help='Parameters required by this blueprint definition.')
```
Juliehzl: May I know the reason why not flatten `parameters` here? Is it too complicated?

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+
+    with self.argument_context('blueprint show') as c:
+        c.argument('blueprint_name', options_list=['--name', '-n'], help='Name of the blueprint definition.')
+
+    with self.argument_context('blueprint list') as c:
+        pass
+
+    with self.argument_context('blueprint artifact delete') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+        c.argument('artifact_name', help='Name of the blueprint artifact.')
```
Juliehzl: Here we can give an options_list=['--name','-n'] for artifact name because it is the target resource now.

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+
+    with self.argument_context('blueprint resource-group create') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+        c.argument('rg_name', help='Name of this resource group. Leave empty if the resource group name will be specified during the blueprint assignment.')
+        c.argument('rg_location', help='Location of this resource group. Leave empty if the resource group location will be specified during the blueprint assignment.')
+        c.argument('artifact_name', help='A unique name of this resource group artifact.')
+        c.argument('display_name', help='Display name of this resource group artifact.')
+        c.argument('description', help='Description of the blueprint artifact.')
+        c.argument('depends_on', nargs='+', help='Artifacts which need to be deployed before the specified artifact.')
+        c.argument('tags', tags_type, help='Tags to be assigned to this resource group.')
```
Juliehzl: this tag is for resource group not blueprint?

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+    with self.argument_context('blueprint artifact show') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+        c.argument('artifact_name', help='Name of the blueprint artifact.')
+
+    with self.argument_context('blueprint artifact list') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+
+    with self.argument_context('blueprint resource-group create') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+        c.argument('rg_name', help='Name of this resource group. Leave empty if the resource group name will be specified during the blueprint assignment.')
```
Juliehzl: is it possible to utilize resource_group_type for rg_name?

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+        c.argument('rg_location', help='Location of this resource group. Leave empty if the resource group location will be specified during the blueprint assignment.')
+        c.argument('artifact_name', help='A unique name of this resource group artifact.')
+        c.argument('display_name', help='Display name of this resource group artifact.')
+        c.argument('description', help='Description of the blueprint artifact.')
+        c.argument('depends_on', nargs='+', help='Artifacts which need to be deployed before the specified artifact.')
```
Juliehzl: the parameter name is a little strange. what about dependencies?

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+
+    with self.argument_context('blueprint artifact role create') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+        c.argument('artifact_name', help='Name of the blueprint artifact.')
+        c.argument('display_name', help='DisplayName of this artifact.')
+        c.argument('description', help='Description of the blueprint artifact.')
+        c.argument('depends_on', nargs='+', help='Artifacts which need to be deployed before the specified artifact.')
+        c.argument('role_definition_id', help='Azure resource ID of the RoleDefinition.')
+        c.argument('resource_group_art', help='Name of the resource group artifact to which the policy will be assigned.')
+        c.argument('principal_ids', help='Array of user or group identities in Azure Active Directory. The roleDefinition will apply to each identity.')
```
Juliehzl: We need to  use `nargs='+'` for this parameter as a array.

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+
+    with self.argument_context('blueprint artifact role update') as c:
```
Juliehzl: what about role_definition_id here?

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+    with self.argument_context('blueprint resource-group list') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+
+    with self.argument_context('blueprint artifact policy create') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+        c.argument('artifact_name', help='Name of the blueprint artifact.')
+        c.argument('display_name', help='DisplayName of this artifact.')
+        c.argument('description', help='Description of the blueprint artifact.')
+        c.argument('depends_on', nargs='+', help='Artifacts which need to be deployed before the specified artifact.')
+        c.argument('policy_definition_id', help='Azure resource ID of the policy definition.')
```
Juliehzl: As we discussed before, _id is kept for guid. And for policy definition parameter, is it possible to support both resource id and name?

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+        c.argument('resource_group_art', help='Name of the resource group artifact to which the policy will be assigned.')
+        c.argument('parameters', arg_type=parameter_type)
+
+    with self.argument_context('blueprint artifact policy update') as c:
```
Juliehzl: update command doesn't need to specify policy_definition_id when doing the operation?

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+        c.argument('resource_group_art', help='Name of the resource group artifact to which the policy will be assigned.')
+        c.argument('parameters', arg_type=parameter_type)
+
+    with self.argument_context('blueprint artifact role create') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+        c.argument('artifact_name', help='Name of the blueprint artifact.')
+        c.argument('display_name', help='DisplayName of this artifact.')
+        c.argument('description', help='Description of the blueprint artifact.')
+        c.argument('depends_on', nargs='+', help='Artifacts which need to be deployed before the specified artifact.')
+        c.argument('role_definition_id', help='Azure resource ID of the RoleDefinition.')
```
Juliehzl: The same as policy definition id.

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+        c.argument('artifact_name', help='Name of the blueprint artifact.')
+        c.argument('display_name', help='DisplayName of this artifact.')
+        c.argument('description', help='Description of the blueprint artifact.')
+        c.argument('depends_on', nargs='+', help='Artifacts which need to be deployed before the specified artifact.')
+        c.argument('resource_group_art', help='Name of the resource group artifact to which the policy will be assigned.')
+        c.argument('parameters', arg_type=parameter_type)
+        c.argument('template', arg_type=template_type)
+
+    with self.argument_context('blueprint published create') as c:
```
Juliehzl: `blueprint publish create`? we usually use noun for subgroup name.

src/blueprint/README.md
```diff
@@ -0,0 +1,153 @@
...
+      "displayName": "Members",
+      "allowedValues": []
+    }
+}
+```
+The paramters are for artifacts which will be added in below commands.
+
+##### Add a Resource Group in the Blueprint
+```
+az blueprint resource-group create \
```
Juliehzl: why not use add as command name?

src/blueprint/azext_blueprint/_help.py
```diff
@@ -0,0 +1,451 @@
...
+    type: command
+    short-summary: Create a blueprint definition.
+    examples:
+      - name: SubscriptionBlueprint
+        text: |-
+               az blueprint create --subscription "00000000-0000-0000-0000-000000000000" --name \\
+               "simpleBlueprint" --description \\
+               "blueprint contains all artifact" --target-scope \\
+               "subscription"
+      - name: ManagementGroupBlueprint
```
Juliehzl: Examples description is too simple.

src/blueprint/azext_blueprint/_help.py
```diff
@@ -0,0 +1,451 @@
...
+        text: |-
+               az blueprint create --management-group myManagementGroup --name \\
+               "simpleBlueprint" --description \\
+               "blueprint contains all artifact" --target-scope \\
```
Juliehzl: Do you have a default value for target scope?

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+            c.argument('identity_type', arg_type=get_enum_type(['None', 'SystemAssigned', 'UserAssigned']), arg_group='identity', help='Type of the managed identity.')
+            c.argument('identity_principal_id', arg_group='identity', help='Azure Active Directory principal ID associated with this Identity.')
+            c.argument('identity_tenant_id', arg_group='identity', help='ID of the Azure Active Directory.')
+            c.argument('identity_user_assigned_identities', arg_group='identity', help='The list of user-assigned managed identities associated with the resource. Key is the Azure resource Id of the managed identity.')
+            c.argument('display_name', help='One-liner string explain this resource.')
+            c.argument('description', help='Multi-line explain this resource.')
+            c.argument('blueprint_id', help='ID of the published version of a blueprint definition.')
+            c.argument('parameters', arg_type=parameter_type, help='Blueprint assignment parameter values.')
+            c.argument('resource_groups', help='Names and locations of resource group placeholders.')
```
Juliehzl: "Names and locations" means it is array?

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: Not sure if it's the same management group, but the management group id in blueprint is not supported in parse_resource_id. I've submitted an issue for msrestazure: https://github.com/Azure/msrestazure-for-python/issues/139

For scope, blueprint swagger defines a `scope` parameter, it could either be a management group or a subscription. Powershell splited it into two parameters, and I followed their design since we can make use of the global argument --subscription. The logic to handle these 2 parameters is that if --management-group is provided, we ignore the --subscription, juse use use management-group for scope value, otherwise we fill the scope value from --subscription or the default subscription if both parameters are not provided.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: It's dynamic in terms of the number and also its subproperties, based on the parameters and type of the artifacts that are added to the blueprint.

src/blueprint/README.md
fengzhou-msft: good suggestion.

src/blueprint/README.md
```diff
@@ -0,0 +1,153 @@
...
+    --blueprint-name blueprintName \
+    --artifact-name myRgArt
+```
+
+##### Add a Role Assignment Artifact
+```
+az blueprint artifact role create \
+    --blueprint-name blueprintName \
```
yungezz: From CLI general guideline, for the referencing resource, `--blueprint` is enough to support both name and id, right?

src/blueprint/README.md
```diff
@@ -0,0 +1,153 @@
...
+    "Members":{
+        "value":"[parameters('MyPolicyName_Members')]"
+    }
+}
+```
+
+
+##### Publish a Blueprint
+```
+az blueprint published create \
```
yungezz: just to confirm, published will be renamed to publish

src/blueprint/azext_blueprint/_help.py
```diff
@@ -0,0 +1,451 @@
...
+"""
+
+helps['blueprint delete'] = """
+    type: command
+    short-summary: Delete a blueprint definition.
+    examples:
+      - name: ManagementGroupBlueprint
+        text: |-
+               az blueprint delete --management-group myManagementGroup --name \\
+               "simpleBlueprint"
```
yungezz: is there consisntent resource naming convention in help? Would suggest to have one. such as myResource etc

src/blueprint/README.md
```diff
@@ -0,0 +1,153 @@
...
+##### Import Blueprint Definition and Artifacts Settings
+
+```
+az blueprint import \
+--name blueprintName \
+--input-path "/path/to/blueprint/directory"
+
+```
+
+In the input directory, there should be a file named "blueprint.json" that has blueprint definition and parameters for artifacts. There should be a subdirectory named "artifacts" and it should contain files for artifact definition. Examples can be found [here]().
```
bquantump: All singular nouns should have an article in front of them. So, "...that has blueprint definition" should be either: "...that has **a** blueprint definition" or if multiple definitions are allowed, plural, then it should be something like: "...that has blueprint definition**s**". The same grammatical error occurs in the next sentence as well: "...files for artifact definition" should have an article in front of the noun as well: "files for **an** artifact definition"

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,215 @@
...
+    with self.argument_context('blueprint create') as c:
+        c.argument('blueprint_name', options_list=['--name', '-n'], help='Name of the blueprint definition.')
+        c.argument('display_name', help='One-liner string explain this resource.')
+        c.argument('description', help='Multi-line explain this resource.')
+        c.argument('target_scope', arg_type=get_enum_type(['subscription', 'managementGroup']), help='The scope where this blueprint definition can be assigned.')
+        c.argument('parameters', arg_type=parameter_type, help='Parameters required by this blueprint definition.')
+
+    with self.argument_context('blueprint import') as c:
+        c.argument('blueprint_name', options_list=['--name', '-n'], help='Name of the blueprint definition.')
+        c.argument('input_path', help='The directory path for json definitions of blueprint and artifacts. Artifacts json files should be in input-path/artifacts/.', completer=FilesCompleter())
```
bquantump: If this command only imports one blueprint, then it would be clearer if the help said: "...json definitions of **the** blueprint and artifacts," instead of "...json definitions of blueprint and artifacts." If the command can import many blueprints, just add an "s" to blueprint.

bquantump: When testing this command, it is not clear that input_path should NOT be an absolute path. For example, while running in my python virtual environment, attempting to give an absolute path to a json file to the import command fails:  az blueprint import --name test --input-path " C:\Users\stevens\Projects\azure-cli-extensions\src\blueprint\azext_blueprint\tests\latest\input\import_with_arm"
OSError: [Errno 22] Invalid argument: 'C:\\Users\\stevens\\Projects\\env\\ C:\\Users\\stevens\\Projects\\azure-cli-extensions\\src\\blueprint\\azext_blueprint\\tests\\latest\\input\\import_with_arm\\blueprint.json'

It will work if I give it the correct path relative to the environment of course: az blueprint import --name test --input-path "..\azure-cli-extensions\src\blueprint\azext_blueprint\tests\latest\input\import_with_arm"

src/blueprint/azext_blueprint/custom.py
```diff
@@ -0,0 +1,577 @@
...
+    from ._client_factory import cf_artifacts
+
+    artifact_client = cf_artifacts(cmd.cli_ctx)
+    body = {}
+    blueprint_path = os.path.abspath(os.path.join(input_path, 'blueprint.json'))
+
+    with open(blueprint_path) as blueprint_file:
+        blueprint = json.load(blueprint_file)
+        if blueprint['properties'] is None:
+            raise CLIError('Blueprint file does not contain properties field')
```
bquantump: This error is confusing. If the field 'properties' does not exist, python will immediately throw an KeyError exception, and hence not hit this line which seems to want to inform the user that the field properties does not exist in the file. Again, though, it would not get hit if the key 'properties' does not exist. I think you should use the python dict get method if you want this to work properly. This will return None if the key doesn't exist, thus hitting the raise CLIError. You can test this easily by modifying one of your test .json files an see the KeyError:
  File "C:\Users\stevens\Projects\env\lib\site-packages\azext_blueprint\custom.py", line 30, in import_blueprint_with_artifacts
    if blueprint['properties'] is None:
KeyError: 'properties'

src/blueprint/azext_blueprint/custom.py
```diff
@@ -0,0 +1,577 @@
...
+        blueprint_properties = blueprint['properties']
+        if 'displayName' in blueprint_properties:
+            body['display_name'] = blueprint_properties['displayName']  # str
+        if 'description' in blueprint_properties:
+            body['description'] = blueprint_properties['description']  # str
+        if 'targetScope' in blueprint_properties:
```
bquantump: Not a big deal, but this if else could be made into a one liner: body['target_scope'] = blueprint_properties.get('targetScope', 'subscription')

src/blueprint/README.md
```diff
@@ -0,0 +1,153 @@
...
+The import command will overwrite defnitions for blueprint and artifacts if a blueprint with the same name already exists.
+
+##### Create a Blueprint with Parameters
+
+```
+az blueprint create \
+    --name blueprintName \
+    --description "An example blueprint." \
+    --target-scope "subscription" \
+    --parameters @/path/to/blueprint_params.json
```
YalinLi0312: Just curious, does this only available from command line? Because I didn't find the place to create a blueprint with params in the Azure portal, or maybe I missed something.

fengzhou-msft: The parameters come from artifacts. In Azure portal, you create all artifacts and a blueprint together. So the portal adds the paramters from artifacts for you.

src/blueprint/README.md
fengzhou-msft: blueprint resource id can be under management group, currently parse_resource_id does not support it. I've submitted an [issue](https://github.com/Azure/msrestazure-for-python/issues/139). So now it's name only.

src/blueprint/README.md
fengzhou-msft: will just use `az blueprint publish` for this command.

src/blueprint/azext_blueprint/_help.py
fengzhou-msft: Yes, default is `subscription`

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: Which shell did you use? I cannot reproduce with cmd or powershell for the absolute path issue in a virtual env.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: No, this is an artifact in blueprint, so it's like a skeleton of a resource group, the name and location values could both be empty here.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: This is from the swagger. I think for arguments we can use verbs like `enable-xx`, so `depends-on` should be ok?

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: yes

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: No, it's also an artifact. So it uses `artifact_name` to identify the resource to update.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: It's not a real value, it's a placeholder string, an example value is "[parameters('owners')]". So there will be a parameter called `owners` in blueprint, and real values will be filled when the blueprint is assigned to a subscription. The artifact will use the real values filled in `owners` when creating the real role assignment.

src/blueprint/azext_blueprint/_params.py
bquantump: I'm using PS 6.2.4. I see what a happened here, though. If there is any leading white space in the input path, you get the error were it attempts to treat the input string as a relative path, which I assume is the expected behavior then.

src/blueprint/azext_blueprint/_help.py
```diff
@@ -0,0 +1,452 @@
...
+        text: |-
+               az blueprint artifact show --management-group myManagementGroup --blueprint-name \\
+               "simpleBlueprint" --name "storageTemplate"
+"""
+
+helps['blueprint artifact list'] = """
+    type: command
+    short-summary: List artifacts for a given blueprint definition.
+    examples:
+      - name: MG-ArtifactList
```
yungezz: name format not consistent with others

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,222 @@
...
+
+    with self.argument_context('blueprint update') as c:
+        c.argument('blueprint_name', options_list=['--name', '-n'], help='Name of the blueprint definition.')
+        c.argument('description', help='Multi-line explain this resource.')
+        c.argument('parameters', arg_type=parameter_type, help='Parameters required by this blueprint definition.')
+
+    with self.argument_context('blueprint delete') as c:
+        c.argument('blueprint_name', options_list=['--name', '-n'], help='Name of the blueprint definition.')
+
+    with self.argument_context('blueprint show') as c:
```
yungezz: in example, `subscription` is also need when show.

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,222 @@
...
+        c.argument('artifact_name', help='A unique name of this resource group artifact.')
+        c.argument('display_name', help='Display name of this resource group artifact.')
+        c.argument('description', help='Description of the blueprint artifact.')
+        c.argument('depends_on', nargs='+', help='Artifacts which need to be deployed before the specified artifact.')
+        c.argument('tags', tags_type, help='Tags to be assigned to this resource group.')
+
+    with self.argument_context('blueprint resource-group update') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+        c.argument('rg_name', help='Name of this resource group. Leave empty if the resource group name will be specified during the blueprint assignment.')
+        c.argument('rg_location', help='Location of this resource group. Leave empty if the resource group location will be specified during the blueprint assignment.')
```
yungezz: just to confirm, location is updatable?

src/blueprint/azext_blueprint/custom.py
```diff
@@ -0,0 +1,580 @@
...
+    # delete old artifacts
+    artifacts = artifact_client.list(scope=scope, blueprint_name=blueprint_name)
+    for artifact in artifacts:
+        artifact_client.delete(scope=scope,
+                               blueprint_name=blueprint_name,
+                               artifact_name=artifact.name)
+
+    for filename in os.listdir(os.path.join(os.path.expanduser(input_path), 'artifacts')):
+        artifact_name = filename.split('.')[0]
+        filepath = os.path.join(os.path.expanduser(input_path), 'artifacts', filename)
```
yungezz: filePath exists check?

src/blueprint/azext_blueprint/custom.py
```diff
@@ -0,0 +1,580 @@
...
+                     blueprint_name,
+                     management_group=None,
+                     scope=None,
+                     description=None,
+                     parameters=None):
+    body = client.get(scope=scope, blueprint_name=blueprint_name).as_dict()
+    if description is not None:
+        body['description'] = description  # str
+    if parameters is not None:
+        body['parameters'] = json.loads(parameters)  # dictionary
```
yungezz: json load exception catch/rethrow?

src/blueprint/azext_blueprint/custom.py
```diff
@@ -0,0 +1,580 @@
...
+                                  blueprint=body).resource_groups
+    return {k: v for k, v in rgs.items() if k == artifact_name}
+
+
+# todo test
+def remove_blueprint_resource_group(cmd, client, blueprint_name,
+                                    artifact_name, management_group=None, scope=None):
+    body = client.get(scope=scope, blueprint_name=blueprint_name).as_dict()
+    if artifact_name not in body.setdefault('resource_groups', {}):
+        raise CLIError('The specified artifact name can not be found.')
```
yungezz: with name?

src/blueprint/azext_blueprint/custom.py
```diff
@@ -0,0 +1,580 @@
...
+    body.setdefault('identity',
+                    {})['principal_id'] = identity_principal_id  # str
+    body.setdefault('identity', {})['tenant_id'] = identity_tenant_id  # str
+    body.setdefault(
+        'identity', {}
+    )['user_assigned_identities'] = identity_user_assigned_identities  # dictionary
+    body['display_name'] = display_name  # str
+    body['description'] = description  # str
+    body['blueprint_id'] = blueprint_id  # str
+    body['parameters'] = json.loads(
```
yungezz: same as above, json load exception catch/rethrow

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: I think the help message from swagger is wrong. An example value is "/providers/Microsoft.Authorization/policySetDefinitions/1f3afdf9-d0c9-4c3d-847f-89da613e70a8", the corresponding ARM resource id is "/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/providers/Microsoft.Authorization/policyAssignments/SecurityCenterBuiltIn". 

There's existing [code](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/resource/custom.py#L843) in azure-cli to use name or id for both policy set definition and the resource, but it needs to use the policy SDK. As this command is beyond the cutline of breadth coverage, we may not want to spend extra effort to go that far.

I'll name the option to be 'policy-definition', modify help message and apply similar changes to 'role-definition-id'.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: Actually, it could be real values or a placeholder. I will add `nargs='+'` and update the help message.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: updated

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: updated

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: The same as policy since it's an artifact, only artifact_name is needed.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: As required by PM, use `blueprint publish` directly.

src/blueprint/azext_blueprint/_help.py
fengzhou-msft: updated

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: `resource_groups` is a dict, so I used json here. But as we should limit the usage of json, I have changed it to actions and it can accept multiple resource group key=value pairs. Then I turned it into dictionary in the custom method before calling SDK.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: good to know about this.

src/blueprint/azext_blueprint/custom.py
fengzhou-msft: Thanks for spotting this. I've fixed it.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: Yes, you can update name and location of the resource group artifact in a blueprint.

src/blueprint/azext_blueprint/_help.py
fengzhou-msft: For common resource names, we can do that. For specific resource names, they may have naming constraints, I usually do not modify such values from swagger example.

src/blueprint/README.md
fengzhou-msft: updated

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: It can import one blueprint. updated.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: `subscription` is a global argument, we don't modify it, so it's usually not in `_params.py`, only in `custom.py`, the same for `resource-group`.

src/blueprint/azext_blueprint/custom.py
fengzhou-msft: added exception catch and rethrow

src/blueprint/azext_blueprint/custom.py
fengzhou-msft: Moved the json.loads() logic to validator, parameters is of CLIArgumentType, this type already handles json parse error. It will not reach the code here in case of json format error.

src/blueprint/azext_blueprint/custom.py
fengzhou-msft: updated

src/blueprint/azext_blueprint/custom.py
fengzhou-msft: Moved the json.loads() logic to validator.

src/blueprint/azext_blueprint/_help.py
fengzhou-msft: updated

yonzhan:
blueprint

src/blueprint/azext_blueprint/_help.py
yungezz: got it. While Swaggers themselves aren't consistent at all. Let's align in CLI

src/blueprint/azext_blueprint/custom.py
fengzhou-msft: updated

src/blueprint/azext_blueprint/_params.py
Juliehzl: For management group value, it is something related to tanant, not subscription. CLI authentication is based on default subscription information. I know our target here is bluprint client. But I am thinking if we need to confirm we have the access permission for management group or the subscription. If we don;t have the permission, should the specified scope be invalid?

In additon, should we consider grouping `--management-scope` and `--subscription` as `scope` argument group? Ad you know, currently --subscription is global argument part and it is hard for users to associate the --subcription to scope parameters.

I am just trying to come up with some potential issues. If it's fine to your service, feel free to let me know.

src/blueprint/README.md
Juliehzl: May I know how blueprint resource id looks like ? and is it different for management scope and subscription scope?

src/blueprint/azext_blueprint/_help.py
Juliehzl: The same as what I have mentioned in https://github.com/Azure/azure-cli-extensions/pull/1164/files#r392013317.
should the management group be in the same tenant with login account subscription?

src/blueprint/azext_blueprint/_help.py
```diff
@@ -0,0 +1,452 @@
...
+    short-summary: Delete a blueprint definition.
+    examples:
+      - name: Delete a management group blueprint
+        text: |-
+               az blueprint delete --management-group MyManagementGroup --name \\
+               MyBlueprint
+      - name: Delete a subscription blueprint
+        text: |-
+               az blueprint delete --subscription MySubscription --name \\
+               MyBlueprint
```
Juliehzl: For blueprint name, is it be unique in the world? 
The combination of scope info and name is to get the blueprint id, right? In this way, can we allow blueprint id input and add a custom validator if parse_resource_id doesn't work for your scenario?

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: We can add more sanity checks in depth coverage. Let's keep it simple in breadth coverage. The command could be used as a downstream to commands to first get one' management groups(we do have such commands already, right?), then pass that group here. 

As for scope, I splited it because first Powershell designed it this way and PMs want us to be consistent with Powershell, second if it's scope, you need to pass the full resource id of subscription or management group, with msrestazure not able to parse management group id now, it may not be convenient for users.

src/blueprint/README.md
fengzhou-msft: They are different:
`/providers/Microsoft.Management/managementGroups/CliMgmtGroup/providers/Microsoft.Blueprint/blueprints/group-bp`
or
`/subscriptions/0b1f6471-1bf0-4dda-aec3-cb9272f09590/providers/Microsoft.Blueprint/blueprints/first-blueprint`

src/blueprint/azext_blueprint/_params.py
```diff
@@ -0,0 +1,226 @@
...
+        c.argument('resource_group_art', help='Name of the resource group artifact to which the policy will be assigned.')
+        c.argument('parameters', arg_type=parameter_type, help='Parameters for policy assignment artifact. It can be a JSON string or JSON file path with "@" prefix.')
+
+    with self.argument_context('blueprint artifact role create') as c:
+        c.argument('blueprint_name', help='Name of the blueprint definition.')
+        c.argument('artifact_name', help='Name of the blueprint artifact.')
+        c.argument('display_name', help='DisplayName of this artifact.')
+        c.argument('description', help='Description of the blueprint artifact.')
+        c.argument('depends_on', nargs='+', help='Artifacts which need to be deployed before the specified artifact.')
+        c.argument('role_definition_id', options_list=['--role-definition'], help='The full role definition id.')
```
Juliehzl: I have two more questions here:
1. Can we support custom role definition?
2. Why not use role name to specify a role definition?

src/blueprint/azext_blueprint/_help.py
fengzhou-msft: We can use some patch to parse_resource_id to identify management group blueprint id. But we may need further patch for artifact id, published blueprint id, published blueprint artifact id, blueprint assignment id. It's not worth the extra effort for breadth coverage. Let's wait for msrestazure fix. And also Blueprint is not GAed and may be replaced by a new RP. Let's just roll it out quickly and add convenient feature if we need to go deeper. I appreciate your suggestions and it definitely make sense in the long run.

src/blueprint/azext_blueprint/_params.py
fengzhou-msft: 1. Blueprint does not support custom role definition.
2. The scope of blueprint and the scope of role are different, without knowing the scope of role, let's avoid using a simple role name although it seems now only roles in "/" scope are supported for blueprint.

PR #1172, opened 2020-01-03
[Image Copy] Fix snapshot create issue

qwordy:
This issue is due to a breaking change of snapshot service. The service changed a parameter (source storage account ID) from optional to mandatory recently. This PR infers source storage account ID automatically to fix this issue. Breaking change is catastrophic.

Fix issue https://github.com/Azure/azure-cli/issues/11625

---

### General Guidelines

For new extensions:


src/image-copy/azext_imagecopy/create_target.py
```diff
@@ -113,11 +116,21 @@ def create_target_image(location, transient_resource_group_name, source_type, so
     else:
         snapshot_resource_group_name = transient_resource_group_name
 
+    storage_account_name = target_blob_path.split('.')[0].split('/')[-1]
+    if target_subscription:
+        subscription_id = target_subscription
+    else:
+        subscription_id = get_subscription_id(cmd.cli_ctx)
```
yungezz: is this original service behavior? otherwise, pls help to document the behavior.

src/image-copy/azext_imagecopy/create_target.py
```diff
@@ -113,11 +116,21 @@ def create_target_image(location, transient_resource_group_name, source_type, so
...
+        subscription=subscription_id, resource_group=transient_resource_group_name,
+        namespace='Microsoft.Storage', type='storageAccounts', name=storage_account_name)
+
     cli_cmd = prepare_cli_command(['snapshot', 'create',
                                    '--resource-group', snapshot_resource_group_name,
                                    '--name', target_snapshot_name,
                                    '--location', location,
-                                   '--source', target_blob_path],
+                                   '--source', target_blob_path,
+                                   '--source-storage-account-id', source_storage_account_id],
```
yungezz: no need to specify api version? is `source-storage-account-id` a new property since specific version or always there?

yungezz:
the scenario: no specifying target_subscription is covered by test, right? and it doesn't fail because run only in recorded way, but not live run, right?

src/image-copy/azext_imagecopy/create_target.py
qwordy: It's an existing property. It was optional formerly, but it's mandatory now.

yungezz: got it.

qwordy:
> 
> 
> the scenario: no specifying target_subscription is covered by test, right? and it doesn't fail because run only in recorded way, but not live run, right?

This issue is caused by a breaking change in service. There is no automatic testing in this module. I test manually. I made the minimal change to fix issue 11625. "target_subscription" is not the focus of this issue and PR.

src/image-copy/azext_imagecopy/create_target.py
qwordy: The goal of this PR is to infer storage account ID. If target_subscription is None, use the default subscription.
Code that creates storage account.
```
cli_cmd = prepare_cli_command(['storage', 'account', 'create',
                                   '--name', target_storage_account_name,
                                   '--resource-group', transient_resource_group_name,
                                   '--location', location,
                                   '--sku', 'Standard_LRS'],
                                  subscription=target_subscription)
```

yungezz:
> > the scenario: no specifying target_subscription is covered by test, right? and it doesn't fail because run only in recorded way, but not live run, right?
> 
> This issue is caused by a breaking change in service. There is no automatic testing in this module. I test manually. I made the minimal change to fix issue 11625. "target_subscription" is not the focus of this issue and PR.

fix typo. `source-storage-account-id` :) . there's test to cover source-storage-account-id none?

qwordy:
OK. I can add a test case for this PR although there is no testing in this module now.

PR #1177, opened 2020-01-06
[Network][Feature]az network firewall create: support threat intel whitelist and private ranges

MyronFanQiu:
fix Azure/azure-cli#11647

---

### General Guidelines

For new extensions:

### DIFF
--- a/src/azure-firewall/azext_firewall/_validators.py
+++ b/src/azure-firewall/azext_firewall/_validators.py
@@ -143,3 +143,6 @@ def validate_rule_group_collection(namespace):
     if namespace.target_fqdns is not None and namespace.fqdn_tags is not None:
         raise CLIError('usage error: --target-fqdns | --fqdn-tags')
     return namespace
+
+def validate_private_ranges(namespace):
+    namespace.private_ranges = ', '.join(namespace.private_ranges)
--- a/src/azure-firewall/azext_firewall/_validators.py
+++ b/src/azure-firewall/azext_firewall/_validators.py
@@ -145,13 +145,16 @@ def validate_rule_group_collection(namespace):
     return namespace
 
 
-def validate_private_ranges(namespace):
-    namespace.private_ranges = ', '.join(namespace.private_ranges)
+def process_private_ranges(namespace):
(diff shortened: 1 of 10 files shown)


src/azure-firewall/azext_firewall/_validators.py
```diff
@@ -143,3 +143,15 @@ def validate_rule_group_collection(namespace):
     if namespace.target_fqdns is not None and namespace.fqdn_tags is not None:
         raise CLIError('usage error: --target-fqdns | --fqdn-tags')
     return namespace
+
+
+def validate_private_ranges(namespace):
+    namespace.private_ranges = ', '.join(namespace.private_ranges)
```
haroldrandom: No return?

src/azure-firewall/azext_firewall/_validators.py
```diff
@@ -143,3 +143,15 @@ def validate_rule_group_collection(namespace):
...
+
+
+def validate_threat_intel_whitelist_ip_addresses(namespace):
+    namespace.ip_addresses = ', '.join(namespace.ip_addresses)
```
haroldrandom: It that what validator targets to?

src/azure-firewall/azext_firewall/_validators.py
MyronFanQiu: no return. If you don't need parameter in validator, you don't need return anything. Usually you need to return a function if this validator need parameters.

src/azure-firewall/azext_firewall/_validators.py
MyronFanQiu: yes. It's a strange server. it accepts string with `,` to represent a list instead of an array.....

haroldrandom: No, I mean validator should be used to validate user input values not re-format them.
Is validator designed for this?

MyronFanQiu: Well, good question. From a bigger view, making the arguments validate also sounds reasonable. I'm not sure the original design spec. At least, in network module, it's common to re-format them.

PR #1208, opened 2020-01-17
Adding service-level tasks for public preview. Fixing some bugs.

vchske:
---

### General Guidelines

For new extensions:


src/dms-preview/azext_dms/custom.py
```diff
@@ -59,27 +65,21 @@ def create_or_update_project(
         # If not an extension scenario, run CLI core method
         # TODO: We currently don't have any CLI core code to perform any validations
         # because of this we need to raise the error here.
-        try:
```
vchske: @temandr I took this out as it was causing a pylint issue with immediately raising in the except block. I think this should be ok because an exception will get raised to the client in the same way and we're still performing our validations needed in the extension.

src/dms-preview/azext_dms/custom.py
```diff
@@ -125,38 +125,32 @@ def create_task(
     # Validation: Test scenario eligibility
     if not scenario_handled_in_extension:
         # If not an extension scenario, run CLI core method
-        try:
```
vchske: @temandr Same here

vchske:
@zikalino I think we're good to merge and publish when you get a moment. Thanks!

vchske:
@zikalino @jsntcy Can we get some eyes on this? It's been almost two weeks.

yonzhan:
@Juliehzl  please help with this PR.

temandr:
@yonzhan @Juliehzl Is there any update on this? We're currently releasing a new version and I need to add a change that conflicts with this change - so we need to get this change in so I can create a PR for the new change.

src/dms-preview/azext_dms/_help.py
```diff
@@ -5,6 +5,100 @@
...
 from knack.help_files import helps
 
+helps['dms task'] = """
+    type: group
+    short-summary: Manage service-level tasks for a Database Migration Service instance.
+    long-summary: |
+        Service-level tasks are non-migration tasks that are performed at the service level. This is usually for administrative and maintenance work which applies to the service's agent and VM.
+"""
+
+helps['dms task cancel'] = """
```
Juliehzl: can you also add examples for cancel command?

src/dms-preview/azext_dms/_help.py
```diff
@@ -5,6 +5,100 @@
...
+        The following tasks are currently supported:
+            1) Check OCI driver against Oracle source to check for compatibility
+            2) Upload OCI driver install package
+            3) Install an already uploaded OCI driver package
+
+    parameters:
+        - name: --task-type
+          type: string
+          short-summary: >
+            The type of task to create. The supported types are CheckOciDriver, UploadOciDriver, and InstallOciDriver.
```
Juliehzl: can you add parameter definition in _params.py file and use get_enum_type() for allowed values?

src/dms-preview/azext_dms/_help.py
```diff
@@ -5,6 +5,100 @@
...
+        - name: --task-options-json
```
Juliehzl: json format is not a recommended in cli command because it is hard to validate and show allowed type. Is it possible to flatten the json object to several parameters?

src/dms-preview/azext_dms/_params.py
```diff
@@ -5,12 +5,22 @@
...
-from azure.cli.core.commands.parameters import tags_type
+from azure.cli.core.commands.parameters import resource_group_name_type, tags_type
 
 
 def load_arguments(self, _):
     name_arg_type = CLIArgumentType(options_list=['--name', '-n'], metavar='NAME')
 
+    with self.argument_context('dms task') as c:
+        c.argument('group_name', resource_group_name_type)
+        c.argument('service_name', options_list=['--service-name'], help="The name of the Service.")
```
Juliehzl: Is there a list with all allowed service name? Which kind of service name should be provided here?

src/dms-preview/azext_dms/_help.py
```diff
@@ -5,6 +5,100 @@
...
+        - name: Create a task which checks if the currently installed OCI driver supports
+          text: >
+            az dms task create -g myresourcegroup --service-name mydms -n checkocitask1 --task-options-json "{\"serverVersion\": \"12.2.0.1\"}" --task-type CheckOciDriver
+"""
+
+helps['dms task delete'] = """
+    type: command
+    short-summary: Delete a service-level Task.
+    parameters:
+        - name: --delete-running-tasks
```
Juliehzl: We will recommend to use get_three_state_flag() for boolean parameter.

src/dms-preview/azext_dms/_help.py
```diff
@@ -5,6 +5,100 @@
...
+"""
+
+helps['dms task list'] = """
+    type: command
+    short-summary: List the service-level Tasks within a DMS instance. Some tasks may have a status of Unknown, which indicates that an error occurred while querying the status of that task.
+    parameters:
+        - name: --task-type
+          type: string
+          short-summary: >
+            Filters the list by the type of task. For the list of possible types see "az dms check-status".
```
Juliehzl: For the parameter message, can we use noun value here, e.g. "The type of task to be listed"?

src/dms-preview/azext_dms/custom.py
```diff
@@ -59,27 +65,21 @@ def create_or_update_project(
...
-            core_res = core_create_or_update_project(
-                client,
-                project_name,
-                service_name,
-                resource_group_name,
-                location,
-                source_platform,
-                target_platform,
-                tags)
-        except:
```
Juliehzl: In this way, how do you validate your command now?

src/dms-preview/azext_dms/custom.py
```diff
@@ -125,38 +125,32 @@ def create_task(
...
-                                        service_name,
-                                        project_name,
-                                        task_name,
-                                        source_connection_json,
-                                        target_connection_json,
-                                        database_options_json,
-                                        enable_schema_validation,
-                                        enable_data_integrity_validation,
-                                        enable_query_analysis_validation)
-        except:
```
Juliehzl: The same

src/dms-preview/azext_dms/custom.py
```diff
@@ -169,7 +163,8 @@ def create_task(
         if validate_only is False and validated_task_name is None:
             raise CLIError(
                 "When not validating a MongoDB task, you must supply a previously run 'validate_only' task name.")
-        elif validate_only is False and validated_task_name is not None:
+
+        if validate_only is False and validated_task_name is not None:
```
Juliehzl: May I know the reason to change from elif to if here?

src/dms-preview/azext_dms/custom.py
```diff
@@ -332,14 +350,13 @@ def extension_handles_scenario(
...
         task_type=""):
-    # Add scenario types to this list when moving them out of this extension (preview) and into the core CLI (GA)
+    # Remove scenario types from this list when moving them out of this extension (preview) and into the core CLI (GA)
     ExtensionScenarioTypes = [
         ScenarioType.sql_sqldb_online,
         ScenarioType.mysql_azuremysql_online,
         ScenarioType.postgres_azurepostgres_online,
         ScenarioType.mongo_mongo_offline,
         ScenarioType.mongo_mongo_online,
-        ScenarioType.oracle_azurepostgres_offline,
```
Juliehzl: Is there any command will be affected by the change here?

src/index.json
```diff
@@ -1046,8 +1046,8 @@
         ],
         "dms-preview": [
             {
-                "downloadUrl": "https://azcliorcas.blob.core.windows.net/azclipath/dms_preview-0.9.0-py2.py3-none-any.whl",
-                "filename": "dms_preview-0.9.0-py2.py3-none-any.whl",
+                "downloadUrl": "https://azcliorcas.blob.core.windows.net/azclipath/dms_preview-0.10.0-py2.py3-none-any.whl",
```
Juliehzl: You can append the new version information to dms-preview list.

src/dms-preview/azext_dms/_help.py
vchske: We had gone through our initial command review with Derek Bekoe (probably over a year ago at this point). Given the free-form nature of our inputs, it was decided back then that json was the best option. If you look at our base CLI as well as the extension here, json inputs are used throughout.

src/dms-preview/azext_dms/_help.py
vchske: May I ask why? The user will either want to delete the running tasks or not. Additionally, this param is used elsewhere in our cli commands. I don't like the idea of having it one way here and another way everywhere else. If it is a requirement for some reason, I think a more prudent way forward is to create a backlog item for fixing this throughout.

src/dms-preview/azext_dms/_params.py
vchske: There is no list. The name of the user's service is user-defined. They would know what service(s) they own.

src/dms-preview/azext_dms/custom.py
vchske: I believe you are referring the fact that the core doesn't validate the type of project created. The core only handles the sql to sql scenario, this extension handles all the rest. What needs to happen, since these other scenarios are coming off of public preview, is we will migrate these scenarios out of the extension and over to the core.

But right now, the type of project allowed is handled in the extension_handles_scenario method.

src/dms-preview/azext_dms/custom.py
vchske: Same answer

src/dms-preview/azext_dms/custom.py
vchske: azdev's style checkers prompted this change.

src/dms-preview/azext_dms/custom.py
vchske: No. That combination is not valid. This was a fix so that it would fail validation and not allow it.

src/index.json
vchske: We were explicitly told to overwrite. Additionally, at a glance I don't see any other extension with multiple version entries  in here.

src/dms-preview/azext_dms/_help.py
vchske: The check for allowed values is handled in get_service_task_properties on line 493. It is being checked there because other things have to happen depending on task type so it made sense to validate it there as well.

vchske:
@Juliehzl Both Artyom and I have attempted to get tests working in this extension over several calls and   meetings and we've never been able to make them work. Our next big rock in our CLI is to migrate the scenarios out of the extension and into the core where tests are working just fine. I would like to ask for another reprieve given our continued technical difficulties in this area and given how close we are to migrating over to the core cli project.

temandr:
@Juliehzl, @vchske and I have updated the PR with responses to your feedback as well as a passing build.

temandr:
@Juliehzl is there any update on this?

src/dms-preview/azext_dms/_help.py
Juliehzl: But get_enum_type() will give many convenience to users because it will make the parameters value case-insensitive.

src/dms-preview/azext_dms/_help.py
Juliehzl: get_three_state() is cli build-in argument type. It also can provide convenience to users. For more information, please refer to https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#registering-flags.

Juliehzl:
> @Juliehzl is there any update on this?

I have another concern here is that it looks that you don't update index.json here.

temandr:
> > @Juliehzl is there any update on this?
> 
> I have another concern here is that it looks that you don't update index.json here.

I have removed the update here since I will be releasing another version soon after this one gets checked in. I will include the index.json change in that PR.

src/dms-preview/azext_dms/_help.py
temandr: I agree with @vchske in that would be a more extensive update since we use this param in our core cli implementation. It makes sense to make the change throughout as opposed to in bits and pieces.

src/dms-preview/azext_dms/_help.py
vchske: I'm going to push back on this. The task type is already case-insensitive. Also, as with the three state ask, we have enum-like things sprinkled elsewhere in the core and extension without using enums. So if this is important, I again would like to make the recommendation that we do any work like this in bulk rather than have the code base be inconsistent.

temandr:
@Juliehzl Please take a look and let us know if there is anything else. Otherwise, please merge the pull request.

src/dms-preview/azext_dms/_help.py
Juliehzl: I really agree that we should keep consistent. From CLI command guideline, we recommend to use get_enum_types() for enum models. For more information, you can refer to https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#registering-enums.
Besides case insensitive, get_enum_type will also help format your help message, e.g. 
 
image

And when you have enum models, you can get enum values directly from SDK difinition.

Because your PR is in extension now, I am fine for current behavior. But when you want to make it in azure-cli repo, we hope you can consider using get_enum_type() for your enum type.

src/dms-preview/azext_dms/_help.py
Juliehzl: json format is accepted in some scenarios but not recommended. If you can flatten it, it will be better to flatten json fromat to several parameters. In this way, we can give users more description for parameter usage and figure out which parameter is not correct.
But in some very complicated scenarios, we also ca accept json file.

PR #1214, opened 2020-01-20
Shiqiu/sprint166 - Refine the descriptions and logic of app log tail command.

ShichaoQiu:
Refine the descriptions and logic of app log tail command.


yonzhan:
add to S164.

PR #1240, opened 2020-02-04
add outboundType param for userDefinedRouting

gtracer:
---

### General Guidelines

For new extensions:


jluk:
@yonzhan or @zikalino could we get your help to review this PR? we're looking to unblock some customer networking issues with these new commands. some of these changes are porting code already in the core build, as customers need to use it with preview features as well.

yonzhan:
@qwordy  and @MyronFanQiu  to help with this PR.

jluk:
@qwordy or @MyronFanQiu could you possibly look at this soon? this is the last mile of active networking issues with AKS, our customers are having lots of trouble completing this situation with ARM templates alone - a cli extension would resolve many active issues.

Also PR looks to be impacted by this issue opened by @gtracer 
https://github.com/Azure/azure-cli-extensions/issues/1244

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -60,6 +61,9 @@ def load_arguments(self, _):
         c.argument('load_balancer_managed_outbound_ip_count', type=int)
         c.argument('load_balancer_outbound_ips', type=str, validator=validate_load_balancer_outbound_ips)
         c.argument('load_balancer_outbound_ip_prefixes', type=str, validator=validate_load_balancer_outbound_ip_prefixes)
+        c.argument('load_balancer_outbound_ports', type=int, validator=validate_load_balancer_outbound_ports)
+        c.argument('load_balancer_idle_timeout', type=int, validator=validate_load_balancer_idle_timeout)
+        c.argument('outbound_type', type=str, validator=validate_outbound_type)
```
MyronFanQiu: Hello. You don't need to do it by yourself. You can put `arg_type=get_enum_type([CONST_OUTBOUND_TYPE_LOAD_BALANCER, CONST_OUTBOUND_TYPE_USER_DEFINED_ROUTING])`

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -108,6 +108,18 @@
           type: string
           short-summary: Load balancer outbound IP prefix resource IDs.
           long-summary: Comma-separated public IP prefix resource IDs for load balancer outbound connection. Valid for Standard SKU load balancer cluster only.
+        - name: --load-balancer-outbound-ports
```
qwordy: This parameter receives one integer. Plural form may cause ambiguity.

gtracer: This property name comes from slb and we want to stay consistent with how they have named it there

src/aks-preview/azext_aks_preview/_params.py
qwordy: you can use enum type and provide valid values

src/aks-preview/azext_aks_preview/_validators.py
```diff
@@ -219,3 +223,29 @@ def validate_user(namespace):
     if namespace.user.lower() != "clusteruser" and \
             namespace.user.lower() != "clustermonitoringuser":
         raise CLIError("--user can only be clusterUser or clusterMonitoringUser")
+
+
+def validate_load_balancer_outbound_ports(namespace):
```
qwordy: Add these rules to help. Otherwise, users need to retry many times to select a valid value.

src/aks-preview/azext_aks_preview/_validators.py
```diff
@@ -219,3 +223,29 @@ def validate_user(namespace):
...
+    if namespace.load_balancer_idle_timeout is not None:
+        if namespace.load_balancer_idle_timeout < 4 or namespace.load_balancer_idle_timeout > 120:
+            raise CLIError("--load-balancer-idle-timeout must be in the range [4,120]")
+
+
+def validate_outbound_type(namespace):
+    if not namespace.outbound_type:
+        return
+
+    if namespace.outbound_type not in [CONST_OUTBOUND_TYPE_LOAD_BALANCER, CONST_OUTBOUND_TYPE_USER_DEFINED_ROUTING]:
```
qwordy: Tell users valid values in help

qwordy:
There is a similar PR recently. It also involves --load-balancer-outbound-ports. Can you find the PR? What's the difference of these two PRs?

jluk:
@qwordy the previous PR you mention is to push these changes into the core-cli build. We need these exact changes in the preview extension because of the CLI extension model design - customers cannot use the core-cli changes we did if the preview-extension is installed.

It is a requirement to have preview customers using preview features be able to set these parameters as well.

src/aks-preview/azext_aks_preview/_params.py
gtracer: Good thinking, thanks for bringing that to my attention - updated

src/aks-preview/azext_aks_preview/_validators.py
gtracer: Good thinking, thanks for bringing that to my attention - updated

src/aks-preview/azext_aks_preview/_validators.py
gtracer: Added

src/aks-preview/azext_aks_preview/_params.py
gtracer: Good thinking, updated for outboundtype

gtracer:
@qwordy @MyronFanQiu feedback has been incorporated. Not sure what the linter is complaining about - any idea? It runs just fine locally:

Modules: aks-preview

Initializing linter with command table and help files...

 Results
=========

-  pass: faulty_help_example_parameters_rule
-  pass: faulty_help_example_rule
-  pass: faulty_help_type_rule
-  pass: unrecognized_help_entry_rule
-  pass: unrecognized_help_parameter_rule
-  pass: expired_command_group
-  pass: missing_group_help
-  pass: expired_command
-  pass: missing_command_help
-  pass: no_ids_for_list_commands
-  pass: bad_short_option
-  pass: expired_option
-  pass: expired_parameter
-  pass: missing_parameter_help

qwordy:
strange unicode error in linter
`- name: --load-balancer-outbound-ports`
maybe delete and retype this line?

gtracer:
> strange unicode error in linter
> `- name: --load-balancer-outbound-ports`
> maybe delete and retype this line?

So I think I figured it out. It looks to me like the CI is caching the preview CLI entry. So for example, if you make a change to version x.y.z and then even if you replace the whl in the container, provided that it has the same name it won't be re-pulled. 

When I updated the version to be x.y.z+1, even though it is the exact same container image, it now passed the CI linter

Let me know if that makes sense. If it does then I think one of two fixes is required:
1) When re-building a whl we need a naming strategy for future builds on the same version (we only want to bump the version once we actually publish)
2) easier solution is just don't cache the whl

qwordy:
> 
> 
> > strange unicode error in linter
> > `- name: --load-balancer-outbound-ports`
> > maybe delete and retype this line?
> 
> So I think I figured it out. It looks to me like the CI is caching the preview CLI entry. So for example, if you make a change to version x.y.z and then even if you replace the whl in the container, provided that it has the same name it won't be re-pulled.
> 
> When I updated the version to be x.y.z+1, even though it is the exact same container image, it now passed the CI linter
> 
> Let me know if that makes sense. If it does then I think one of two fixes is required:
> 
>     1. When re-building a whl we need a naming strategy for future builds on the same version (we only want to bump the version once we actually publish)
> 
>     2. easier solution is just don't cache the whl

linter is a source code analysis tool. The source code is pulled from github repo. I am not sure the cause is whl file. Pipeline jobs are executed in identical, independent environment.

qwordy:
Related azure pipelines definition:
https://github.com/Azure/azure-cli-extensions/blob/master/azure-pipelines.yml#L85
https://github.com/Azure/azure-cli-extensions/blob/master/scripts/ci/verify_modified_index.sh#L35
Only modified (index change) extensions will be verified.

PR #1258, opened 2020-02-08
{omsagent} role assignment changes for msi (#1)

rashmichandrashekar:
Role assignment changes to assign role for msi if it is created for omsagent addon


yonzhan:
add to S165.

rashmichandrashekar:
> Hello. Like we talked offline, could you change the help message in extension too? Better to have same code with the main azure cli

I have added the changes in the help too

PR #1263, opened 2020-02-12
[AKS] add nodepool tag for cli extension

zqingqing1:
---

### General Guidelines

For new extensions:

### DIFF
--- a/src/aks-preview/azext_aks_preview/custom.py
+++ b/src/aks-preview/azext_aks_preview/custom.py
@@ -645,6 +645,7 @@ def aks_create(cmd,     # pylint: disable=too-many-locals,too-many-statements,to
                node_osdisk_diskencryptionset_id=None,
                node_count=3,
                nodepool_name="nodepool1",
+               nodepool_tags=None,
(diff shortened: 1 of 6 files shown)


yonzhan:
add to S166.

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -2004,6 +2010,8 @@ def aks_agentpool_update(cmd,   # pylint: disable=unused-argument
         instance.min_count = None
         instance.max_count = None
 
+    instance.tags = tags
```
haroldrandom: tags is None default, what if user doesn't pass `--tags` option, it will clear the tags by mistake right? But user doesn't want to clear tags right?
 I am not sure about the context, correct me if I misunderstand.

zqingqing1: No, it won't clear tags if no --tags passed. Only when customer input --tags "", the tags will be cleaned

zqingqing1:
@haroldrandom  please help to merge if no other concerns and questions :)

PR #1267, opened 2020-02-12
[Network]hide create for management-ip-config create

### LINKED ISSUE #1270: Firewall management IP configuration should only be created while creating a new firewall
- If the issue is to do with Azure CLI 2.0 in-particular, create an issue here at [Azure/azure-cli](https://github.com/Azure/azure-cli/issues)

Network (firewall)

### Description of issue (in as much detail as possible)
Currently the only way to attach a Management IP Configuration appears to be with this workflow:
az network firewall create 
az network firewall ip-config create
az network firewall management-ip-config create

Each of these commands executes a PUT against firewall. 
However, this cannot work, as creating a firewall with forced tunneling requires both the “data” and “management” IP configs to be attached to the firewall at the same time.

Attaching the data ip-config first, and then management ip-config later becomes a migration scenario ion the back end, which is not supported today.

Further, we need to prevent customers from attempting to modify already existing firewalls and attaching management IP configs to them, for the same reason.
(There is work planned to block this scenario in NRP, but it’s not there yet.)

Basically, attaching a management IP configuration should only be possible one time, at firewall creation time, and it requires both types of ip-config in one PUT call.

Customers should be able to add more “data” IP configs. (but only a single “management”  IP config)
Additionally, customers should be able to change the public IP referenced in the “management” IP config. 

-----

MyronFanQiu:
Due to service limitation. hide management-ip-config create.

1. should we hide this command or not?  We can just left here and told user that it cannot work now.

Fix #1270 

---

### General Guidelines

For new extensions:


yonzhan:
add to S165.

src/azure-firewall/azext_firewall/tests/latest/test_azure_firewall_scenario.py
```diff
@@ -95,22 +95,27 @@ def test_azure_firewall_management_ip_config(self, resource_group):
...
 
+        '''
+        # Disable it due to service limitation.
         self.cmd(
             'network firewall management-ip-config create -g {rg} -n {management_ipconfig} -f {af} --public-ip-address {management_pubip} --vnet-name {management_vnet}',
             checks=[
                 self.check('name', '{management_ipconfig}'),
                 self.check('subnet.id', subnet_id_management_ip_config)
             ])
+        '''
```
haroldrandom: I think it's not a convention to comment via DocString style. how about replace it with `#`?

MyronFanQiu: fix

src/azure-firewall/azext_firewall/_params.py
```diff
@@ -113,6 +113,18 @@ def load_arguments(self, _):
...
+        c.argument('management_item_name', options_list=['--m-name'], arg_group="Management Ip Config",
+                   help='Name of the management IP configuration.', is_preview=True)
+        c.argument('management_subnet', validator=get_management_subnet_validator(), help=argparse.SUPPRESS, is_preview=True)
+        c.argument('management_virtual_network_name', virtual_network_name_type, options_list=['--m-vnet-name'],
+                   arg_group="Management Ip Config",
+                   help='The virtual network (VNet) name for management ip configuation. '
+                        'It should contain one subnet called "AzureFirewallManagementSubnet".', is_preview=True)
+        c.argument('management_public_ip_address', help='Name or ID of the public IP to use for management ip configuation.',
+                   arg_group="Management Ip Config", options_list=['--m-public-ip-address'],
+                   validator=get_management_public_ip_validator(), is_preview=True)
```
haroldrandom: need `--m` prefix? Can we removed it? I don't know whether it will cause coflict?

MyronFanQiu: no. We already have --public-ip-address in ip-config create

MyronFanQiu: we have four almost same arguments. The only difference is the `m-` prefix.

haroldrandom: How about group them in a argument group **Management** to give user an instinct hint?

MyronFanQiu: Have done

PR #1281, opened 2020-02-17
Shiqiu/update index json0217

ShichaoQiu:
Update index.json for asc cli extension 0217 release


src/spring-cloud/HISTORY.rst
```diff
@@ -5,7 +5,7 @@ Release History
 
 0.2.1
 ++++++
-* Add command "az spring-cloud app log" to replace "az spring-cloud app log tail" for log streaming.
+* Add command "az spring-cloud app logs" to replace "az spring-cloud app log tail" for log streaming.
```
qwordy: This history entry is not consistent with the code changes.

qwordy: I am confused with log and logs. Why sometimes it's single form and sometimes it's plural form?

src/spring-cloud/azext_spring_cloud/commands.py
```diff
@@ -56,8 +56,11 @@ def load_command_table(self, _):
         g.custom_command('start', 'app_start', supports_no_wait=True)
         g.custom_command('stop', 'app_stop', supports_no_wait=True)
         g.custom_command('restart', 'app_restart', supports_no_wait=True)
-        g.custom_command('log', 'app_tail_log')
-        g.custom_command('log tail', 'app_tail_log', deprecate_info=g.deprecate(redirect='az spring-cloud app log', hide=True))
+        g.custom_command('logs', 'app_tail_log')
+
+    with self.command_group('spring-cloud app log', client_factory=cf_spring_cloud,
+                            deprecate_info=g.deprecate(redirect='az spring-cloud app logs', hide=True)) as g:
```
qwordy: Maybe you can add when it's not available any more.

ShichaoQiu: Thanks a lot for your advice. I'm afraid that we are not sure when it will be not available any more. Once we have a specific time, we will update it.

src/spring-cloud/HISTORY.rst
ShichaoQiu: Hava synced about this offline. Thanks.

PR #1315, opened 2020-02-25
IoT extension update 0.9.0 + re-branding to azure-iot

digimaun:
This is a re-branding/re-aliasing of the Azure CLI IoT extension. Instead of the old `azure-cli-iot-ext` the new alias is `azure-iot`. The old references are not removed yet because we don't want to break our users, and we need time to get various docs and script references updated. We'll get a deprecation timeline in play and remove the old alias after.

---

### General Guidelines

For new extensions:


blackchoey:
@digimaun @qwordy Can you please hold on merging this change? We have an Azure Pipelines task depending on this extension and will be impacted by this change. I'm working to update our task and expected to finish by tomorrow. Thank you very much!

blackchoey:
Thanks @yungezz @digimaun @qwordy Please go ahead and merge this change.

yungezz:
suggest to add descriptive information in existing extension to route user to the new extension.

digimaun:
> suggest to add descriptive information in existing extension to route user to the new extension.

We were thinking of doing that after this merge in the following elements:
- IoT extension github readme with clear visible indicators
- Work with the docs team to replace `azure-cli-iot-ext` with `azure-iot` in all references
- Update of the IoT command module 'try out the extension' event handler (for the base CLI) to point to the new name
- Recording a Channel 9 video
- In various communication with users/stakeholders

We could update the existing/old extension information but keep in mind its the same code base and project. That would mean a new release of "v0.8.10" with the old alias and the primary change being to the package description (to state its deprecated).

With this PR, the old extension is not going away until the deprecation timeline ends. Because of this, I think we have more time/flexibility to remove the existing alias in a graceful manner. We just want to make sure the new alias is available on the index (first).

yungezz:
thanks for the detail plan sharing.

digimaun:
Bump. @qwordy , please let me know if you need anything else to merge.

PR #1318, opened 2020-02-25
Add support for pg v2

temandr:
---

### General Guidelines

For new extensions:


yonzhan:
add to S166.

temandr:
@Juliehzl Please let me know what changes are needed. Otherwise, please merge this change.

temandr:
@Juliehzl Is there any update on this? It looks like the travis also started failing without any changes from end.

src/index.json
```diff
@@ -1372,8 +1372,8 @@
         ],
         "dms-preview": [
```
Juliehzl: As you know, dms-preview value is an array, I will prefer append your new download link with previous one. In this way. we can keep the capability to install with previous version.

temandr: We don't really want to retain the previous version since it would enable support for a deprecated scenario. This change is meant to replace the default request for the PG scenario to a different call.

temandr:
@Juliehzl I created and updated the history file. 
Regarding the SDK change, the change is meant to replace the existing functionality of the command thus no need for command changes.

PR #1325, opened 2020-02-27
[AKS] az aks: Validate network plugin to be either "azure" or "kubenet".

JunSun17:
---

### General Guidelines

For new extensions:


src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -798,6 +798,11 @@ def aks_create(cmd,     # pylint: disable=too-many-locals,too-many-statements,to
 
     outbound_type = _set_outbound_type(outbound_type, network_plugin, load_balancer_sku, load_balancer_profile)
 
+    # Validate network plugin to be either "azure" or "kubenet"
+    if network_plugin:
+        if network_plugin not in ('azure', 'kubenet'):
```
yungezz: lower()?

JunSun17: We only take lower case "azure" and "kubenet", it is case sensitive if you mean network_plugin should be convert to lower case.

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -855,6 +855,11 @@ def aks_create(cmd,     # pylint: disable=too-many-locals,too-many-statements,to
...
+            raise CLIError('Network plugin type needs to be either \"azure\" or \"kubenet\"')
```
fengzhou-msft: can we use enum_type for network_plugin?

JunSun17: network_plugin type is defined from data model and we can not change it here unless we redefine and publish a new data model.

fengzhou-msft: You can add `arg_type=get_enum_type(['azure', 'kubenet'])` [here](https://github.com/Azure/azure-cli-extensions/blob/997ab297f1b3c3c1eb4d9b6a0eedc1d9169b5994/src/aks-preview/azext_aks_preview/_params.py#L77). The underlying data type does not need to change. This will list the allowed values in help message and raise error when the value is not in the enum list.

JunSun17: sure, added arg_type, is there any other places that need to be changed due to introduction of this arg_type?

fengzhou-msft: no other changes needed.

Issue #1329, opened 2020-02-29
az hack fails to create an app if any free app service plan in a different location already exists


michimune:
- If the issue is to do with Azure CLI 2.0 in-particular, create an issue here at [Azure/azure-cli](https://github.com/Azure/azure-cli/issues)

### Extension name (the extension in question)
`hack`

### Description of issue (in as much detail as possible)
When executing the following command:
```
az hack create --name signdemo --runtime python --ai --location westus2 --output yaml
```
I got the following exception:
```
Creating website...
The command failed with an unexpected error. Here is the traceback:

 
update_site_config() takes 2 positional arguments but 3 were given
Traceback (most recent call last):
  File "/opt/az/lib/python3.6/site-packages/knack/cli.py", line 206, in invoke
    cmd_result = self.invocation.execute(args)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 608, in execute
    raise ex
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 666, in _run_jobs_serially
    results.append(self._run_job(expanded_arg, cmd_copy))
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 659, in _run_job
    six.reraise(*sys.exc_info())
  File "/opt/az/lib/python3.6/site-packages/six.py", line 693, in reraise
    raise value
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 636, in _run_job
    result = cmd_copy(params)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/commands/__init__.py", line 306, in __call__
    return self.handler(*args, **kwargs)
  File "/opt/az/lib/python3.6/site-packages/azure/cli/core/__init__.py", line 493, in default_command_handler
    return op(**command_args)
  File "/home/michimune/.azure/cliextensions/hack/azext_hack/custom.py", line 40, in create_hack
    website.create()
  File "/home/michimune/.azure/cliextensions/hack/azext_hack/_website_utils.py", line 67, in create
    webapp = self.__create_webapp(app_service_plan)
  File "/home/michimune/.azure/cliextensions/hack/azext_hack/_website_utils.py", line 144, in __create_webapp
    plan=app_service_plan.name, runtime=runtime_setting['name'], deployment_local_git=True)
[...]

PR #1330, opened 2020-02-29
Fix selection logic of app service plan, make logs a bit better

michimune:
This PR fixes the issue: https://github.com/Azure/azure-cli-extensions/issues/1329

It also adds some logging so you can see the command's progress a little better.


yonzhan:
add to S166.

PR #1363, opened 2020-03-06
SSH extension

rlrossiter:
Add an extension used for AAD-based SSH for Azure VMs.

---

### General Guidelines

For new extensions:


yonzhan:
SSH for Azure VM

yungezz:
hi @rlrossiter we'll look at the issue and keep offline syncing with you.

src/ssh/azext_ssh/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+    short-summary: SSH into Azure VMs
+"""
+
+helps['ssh vm'] = """
+    type: command
+    short-summary: SSH into Azure VMs
+    examples:
+        - name: Give a resource group and VM to SSH to
+          text: |
+            az ssh vm --resource-group [group] --name [vmName] : az ssh vm --resource-group mygroup --name myvm
```
arrownj, line 20: Normally we do not need the part before `:`. Same with other examples. I think it's better to remove it to align with other modules/extensions.

src/ssh/azext_ssh/azext_metadata.json
```diff
@@ -0,0 +1,3 @@
+{
+    "azext.isPreview": true
```
arrownj, line 2: Better do define a `minCLICoreVersion` here. You can reference [here](https://github.com/Azure/azure-cli-extensions/blob/e99c1374cacb559bcf8883147eaae5a6df0118c6/src/stream-analytics/azext_stream_analytics/azext_metadata.json#L3).

src/ssh/azext_ssh/ssh_utils.py
```diff
@@ -0,0 +1,74 @@
...
+    lines.append("\tIdentityFile " + private_key_file)
+
+    with open(config_path, 'w') as f:
+        f.write('\n'.join(lines))
+
+
+def _get_ssh_path():
+    ssh_path = "ssh"
+
+    if platform.system() == 'Windows':
```
arrownj, line 50: Is it possible there is no `ssh` in other platform ?

src/ssh/azext_ssh/tests/latest/test_custom.py
```diff
@@ -0,0 +1,196 @@
...
+        custom.ssh_vm(cmd, "rg", "vm", "ip", "public", "private")
+
+        mock_do_op.assert_called_once_with(
+            cmd, "rg", "vm", "ip", "public", "private", mock_ssh_utils.start_ssh_connection)
+
+    @unittest.skip('have problems with patching functools.partial, will enable it after getting the root cause.')
+    @mock.patch('azext_ssh.custom._do_ssh_op')
+    @mock.patch('azext_ssh.custom.ssh_utils')
+    @mock.patch('functools.partial')
+    def test_ssh_config(self, mock_partial, mock_ssh_utils, mock_do_op):
```
arrownj, line 28: Hi @rlrossiter , which python version are you using to run this test ? It always failed from my side (python 3.8).

src/ssh/setup.py
```diff
@@ -0,0 +1,42 @@
...
+    'Programming Language :: Python',
+    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.6',
+    'Programming Language :: Python :: 3.7',
+    'Programming Language :: Python :: 3.8',
+    'License :: OSI Approved :: MIT License',
+]
+
+DEPENDENCIES = [
+    'paramiko==2.6.0',
```
arrownj, line 25: Hi @rlrossiter, seems we do not use these dependencies in our code any more ? Can we remove them ?

src/ssh/azext_ssh/custom.py
```diff
@@ -0,0 +1,115 @@
...
+    data = _prepare_jwk_data(public_key_file)
+    from azure.cli.core._profile import Profile
+    profile = Profile(cli_ctx=cmd.cli_ctx)
+    username, certificate = profile.get_msal_token(scopes, data)
+
+    cert_file = _write_cert_file(public_key_file, username, certificate)
+    op_call(ssh_ip, username, cert_file, private_key_file)
+
+
+def _prepare_jwk_data(public_key_file):
```
arrownj, line 47: These codes were previously in azure-cli-core. I move it here because we think it is ssh certificate specific. In azure-cli-core, we expose a function `get_msal_token`  in _profile.Profile to make it simple and clean. @rlrossiter @danybeam, could you please help review it ?

src/ssh/azext_ssh/custom.py
```diff
@@ -0,0 +1,115 @@
...
+    if not ssh_ip:
+        raise util.CLIError(f"VM '{vm_name}' does not have a public IP address to SSH to")
+
+    scopes = ["https://pas.windows.net/CheckMyAccess/Linux/user_impersonation"]
+    data = _prepare_jwk_data(public_key_file)
+    from azure.cli.core._profile import Profile
+    profile = Profile(cli_ctx=cmd.cli_ctx)
+    username, certificate = profile.get_msal_token(scopes, data)
+
+    cert_file = _write_cert_file(public_key_file, username, certificate)
```
arrownj, line 43: Hi @rlrossiter , previously there is a fixed `ssh-rsa-cert-v01@openssh.com ` in certificate head, which is weird to me. I replace it with `username`, could you please help check whether it is right ?

**Previous code in azure-cli-core**
```
class SSHCredentials(object):
    def __init__(self, username, cert):
        self.username = username
        self.certificate = "ssh-rsa-cert-v01@openssh.com " + cert
```

rlrossiter: The original was correct as that was the type of certificate the file was (it's required by SSH). The username is encoded as part of the base64 bytes written to the file. The username was passed back alongside the credentials but not part of the certificate in order to know who to SSH as

src/ssh/azext_ssh/custom.py
rlrossiter: I don't like the use of the `_profile` module from core outside of core because modules that start with _ usually denote that it's a private module that shouldn't be used outside of the repo it lives in. That's why I put this in azure-cli-core to start

src/ssh/azext_ssh/tests/latest/test_custom.py
rlrossiter: The version isn't the problem, the code changed and broke the unit tests. A lot of them are now failing for me too because of the code changes (Python 3.6).

arrownj: Hi @rlrossiter, what other tests also failed for you ? As you can see from the CI log, all other tests are passed. Did you use the latest `azure-cli` dev branch and rerun `azdev setup` ? It should be only this test failed. And the reason has something to do with `functions.partial`. I'm not sure the root cause currently, will investigate deeper.

rlrossiter: I guess I didn't have all of the commits you added to this branch. Will re-pull and take a look if I get a chance.

src/ssh/azext_ssh/custom.py
arrownj: Yes, I agree with you about the private module part. For this `_profile` case, as there are lots of examples in extensions to use `_profile` directly, so I think we can just follow it. From cli-core perspective, we have an internal discussion, and we think we'd better not include SSH specific logic in it.

src/ssh/azext_ssh/custom.py
arrownj: Thanks for your explain. Yes, I decode the certificate and find `ssh-ras-cert-v01@openssh.com` is in it. I have change it back. You can have a look.

Could you please share more about this? Is there any document which explain it? It's still weird to me that why it is a fixed name in the certificate. Thanks in advance.

PR #1371, opened 2020-03-10
Add .NET core support in ASC

metacpp:
---

### General Guidelines

For new extensions:

---
This change is [Reviewable](https://reviewable.io/reviews/azure/azure-cli-extensions/1371)


yungezz:
@metacpp seems this is pure vendored SDK change. you'll add more commits later right?

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -19,7 +19,7 @@
 from ast import literal_eval
 from azure.cli.core.commands import cached_put
 from ._utils import _get_rg_location
-from six.moves.urllib import parse
+from six.moves import parse
```
fengzhou-msft: Is this change caused by upgrade of six?

src/spring-cloud/azext_spring_cloud/_params.py
```diff
@@ -84,6 +84,8 @@ def load_arguments(self, _):
                        help='Runtime version of used language')
             c.argument('jvm_options', type=str,
                        help="A string containing jvm options, use '=' instead of ' ' for this argument to avoid bash parse error, eg: --jvm-options='-Xms1024m -Xmx2048m'")
+            c.argument('net_core_main_entry_path', type=str,
```
yuwzho: @yucwan Could you please work with @metacpp about the command parameter interface for steeltoe?

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -99,6 +99,7 @@ def app_create(cmd, client, resource_group, service, name,
                instance_count=None,
                runtime_version=None,
                jvm_options=None,
+               net_core_main_entry_path=None,
```
JunyiYi: We don't need to set NetCoreMainEntryPath during app creation, this should always be `null`.

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -127,6 +128,7 @@ def app_create(cmd, client, resource_group, service, name,
         instance_count=instance_count,
         environment_variables=env,
         jvm_options=jvm_options,
+        net_core_main_entry_path=net_core_main_entry_path,
         runtime_version=runtime_version,)
     user_source_info = models.UserSourceInfo(
         relative_path='<default>', type='Jar')
```
JunyiYi: It's completely OK to pass `<default>` `JAR` to .net core app (deployment), but might be weird for users using REST API calls.

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -315,6 +320,9 @@ def app_deploy(cmd, client, resource_group, service, name,
 
     file_type, file_path = _get_upload_local_file(jar_path)
 
+    if runtime_version == "NetCore_31":
```
JunyiYi: Just a quick question, can we reuse any enumerations defined in the SDK (e.g. `_app_platform_management_client_enums.py`) instead of raw string here?

src/spring-cloud/azext_spring_cloud/_params.py
```diff
@@ -84,6 +84,8 @@ def load_arguments(self, _):
...
+                       help="A string containing the path to the .NET executable relative to zip root")
             c.argument('env', env_type)
 
     for scope in ['spring-cloud app create', 'spring-cloud app deployment create']:
```
JunyiYi: Do we need to rename (or maybe add new argument for backward compatibility) the `jar_path` argument (https://github.com/Azure/azure-cli-extensions/pull/1371/files#diff-31fe45cdcec6a27226e759744f994e10R112) here? For example, `artifact_path` / `binary_path`, since we will cover both `JAR` and `ZIP` here.

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -328,6 +336,7 @@ def app_deploy(cmd, client, resource_group, service, name,
                        memory,
                        instance_count,
                        env,
+                       net_core_main_entry_path,
                        target_module,
                        no_wait,
                        file_type,
```
JunyiYi: We should exclude `NetCoreZip` from displaying build logs as well: https://github.com/Azure/azure-cli-extensions/pull/1371/files#diff-6f0c4fd16bbeedad43a6ad85cfb907e3R376.

src/spring-cloud/azext_spring_cloud/_params.py
yucwan: the name seems too long to me. Suggest "main_entry_path" or "main_entry". It's enough to have .net in the discription.

src/spring-cloud/azext_spring_cloud/_params.py
yucwan: From UX perspective i'm ok to deprecate --jar-path and use --artifact-path instead.
@ShichaoQiu any input from eng point of view? will supporting both .NET and java artifact with one parameter cause any problems?

yucwan:
@metacpp Hi Su, please add @ShichaoQiu as reviewer as well. He is currently the dev owner for the extension.

src/spring-cloud/azext_spring_cloud/_utils.py
```diff
@@ -14,15 +14,16 @@
...
 
 
-def _get_upload_local_file(jar_path=None):
-    file_path = jar_path
-    file_type = "Jar"
+def _get_upload_local_file(runtime_version, artifact_path=None):
+    file_path = artifact_path
+    file_type = "NetCoreZip" if runtime_version == AppPlatformEnums.RuntimeVersion.net_core_31 else "Jar"
 
     if file_path is None:
```
yuwzho, line 28: Do you need to check file_type != NetCoreZip here?

src/spring-cloud/azext_spring_cloud/azure_storage_file/_deserialization.py
```diff
@@ -203,16 +203,16 @@ def _convert_xml_to_directories_and_files(response):
 
     for file_element in entries_element.findall('File'):
         # Name element
-        file = File()
-        file.name = file_element.findtext('Name')
+        file_ = File()
```
yuwzho, line 206: `_file` is the private field naming style

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -209,9 +210,13 @@ def app_create(cmd, client, resource_group, service, name,
         instance_count=instance_count,
         environment_variables=env,
         jvm_options=jvm_options,
-        runtime_version=runtime_version,)
+        net_core_main_entry_path=None,
+        runtime_version=runtime_version)
+
+    file_type = "NetCoreZip" if runtime_version == AppPlatformEnums.RuntimeVersion.net_core_31 else "Jar"
```
yuwzho, line 216: CaseIgnore comparasion?

NIT: All the validation, default value set you can put to [validator](https://github.com/Azure/azure-cli-extensions/blob/master/src/spring-cloud/azext_spring_cloud/_validators.py) and set `namespace.runtime_version=XXX`, thus can skip the file_type setting here, `_get_upload_local_file()` method as well.

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -460,8 +469,8 @@ def app_get_build_log(cmd, client, resource_group, service, name, deployment=Non
         raise CLIError(NO_PRODUCTION_DEPLOYMENT_ERROR)
     deployment_properties = client.deployments.get(
         resource_group, service, name, deployment).properties
-    if deployment_properties.source.type == "Jar":
-        raise CLIError("Jar deployment has no build logs.")
+    if deployment_properties.source.type == "Jar" or deployment_properties.source.type == "NetCoreZip":
```
yuwzho, line 472: NIT: Use enum comparasion may avoid some string change in the future.

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -611,10 +620,11 @@ def app_set_deployment(cmd, client, resource_group, service, name, deployment):
 def deployment_create(cmd, client, resource_group, service, app, name,
                       skip_clone_settings=False,
                       version=None,
-                      jar_path=None,
+                      artifact_path=None,
```
yuwzho, line 623: This will ignore `jar_path` parameter input in `az spring-cloud app deployment create` command. 
If this is on purpose, _param need to be updated as well, and the version bump up should be a major version, and add PM for awareness.

src/spring-cloud/azext_spring_cloud/_params.py
```diff
@@ -83,6 +83,8 @@ def load_arguments(self, _):
         with self.argument_context(scope) as c:
             c.argument('deployment', options_list=[
                 '--deployment', '-d'], help='Name of an existing deployment of the app. Default to the production deployment if not specified.', validator=validate_deployment_name)
+            c.argument('main_entry', options_list=[
+                '--main-entry', '-m'], help="A string containing the path to the .NET executable relative to zip root.")
```
fengzhou-msft, line 87: `The path to the .NET executable relative to zip root.`  would be more succinct.

yungezz:
seems no test for this steeltoe. Can you pls add some test?

src/spring-cloud/azext_spring_cloud/_utils.py
metacpp: Don't quite understand the question here. 3 different scenarios:

1. source deployment: file_type = jar and file_path is None
2. jar deployment: file_type =jar and file_path is not none
3. netcorezip deployment: file_type=netcorezip and file_path is not none

src/spring-cloud/azext_spring_cloud/custom.py
metacpp: That's the legacy string, we can have another PR to do refactor in a later PR.

src/spring-cloud/azext_spring_cloud/custom.py
metacpp: At internal function level, we unified all jar-path to artifact-path, but in parameter, we still keep the old one and add new one.

src/spring-cloud/azext_spring_cloud/custom.py
metacpp: don't quite understand your comment, let's sync it offline.

src/spring-cloud/azext_spring_cloud/_utils.py
yuwzho: I see, I didn't realize the file_path is not required only when source deployment. The code miss led me to consider the un-existing scenario file_path w/ file_type = netcorezip. Would you mind commenting your scenraios in the code?

fengzhou-msft (approved):

PR #1380, opened 2020-03-12
[Notification Hub] fix sku and help

fengzhou-msft:
There's a weird bug that in _params.py. I forgot to rename the `sku-name` option to `sku`, but it turns out `--sku` can still be used and the value will be assigned to `sku_name` correctly. That's why previous tests and linter all passed with `--sku`. 
 
Fix https://github.com/Azure/azure-cli/issues/12590.

---

### General Guidelines

For new extensions:


src/notification-hub/azext_notification_hub/_params.py
```diff
@@ -81,14 +81,14 @@ def load_arguments(self, _):
         c.argument('notification_hub_name', options_list=['--name', '-n'], help='The notification hub name.')
         c.argument('location', arg_type=get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
         c.argument('tags', tags_type)
-        c.argument('sku_name', arg_type=get_enum_type(['Free', 'Basic', 'Standard']), help='Name of the notification hub sku')
+        c.ignore('sku_name')
```
MyronFanQiu, line 84: This would be a breaking change and user would meet errors. Just want to confirm. If the usage is quite low, it's fine to do it.

src/notification-hub/azext_notification_hub/custom.py
```diff
@@ -132,13 +149,10 @@ def update_notificationhubs_hub(cmd, client,
                                 resource_group_name,
                                 namespace_name,
                                 notification_hub_name,
-                                tags=None,
-                                sku_name=None):
```
MyronFanQiu, line 136: If you delete it here, I think we don't need the `c.ignore`

MyronFanQiu: ok. got it. create command still have sku

src/notification-hub/azext_notification_hub/azext_metadata.json
```diff
@@ -1,4 +1,4 @@
 {
-    "azext.isPreview": true,
-    "azext.minCliCoreVersion": "2.0.67"
+    "azext.isExperimental": true,
+    "azext.minCliCoreVersion": "2.3.0"
```
fengzhou-msft, line 3:
```suggestion
    "azext.minCliCoreVersion": "2.3.1"
```

PR #1383, opened 2020-03-12
[AKS] aks update: Override outbound IPs with input values when update.

JunSun17:
Twin: https://github.com/Azure/azure-cli/pull/12572

---

### General Guidelines

For new extensions:

### About Extension Publish


qwordy:
Usually, in `update`, if a parameter is not set, it remains the original value.
1. Why make this change?
2. How about noticing users the change in help message?

src/aks-preview/azext_aks_preview/_loadbalancer.py
```diff
@@ -57,18 +57,22 @@ def configure_load_balancer_profile(managed_outbound_ip_count, outbound_ips, out
...
-            public_ips=outbound_ip_resources
-        )
-    if outbound_ip_prefix_resources:
-        profile.outbound_ip_prefixes = ManagedClusterLoadBalancerProfileOutboundIPPrefixes(
-            public_ip_prefixes=outbound_ip_prefix_resources
-        )
+    if managed_outbound_ip_count or outbound_ip_resources or outbound_ip_prefix_resources:
+        profile.managed_outbound_ips = None
+        profile.outbound_ips = None
+        profile.outbound_ip_prefixes = None
```
fengzhou-msft: So the logic for update is changed to that if one of the 3 parameters is provided, clean up the other 2 if not provided. Will any default values be set on service side for None-value parameters? Are there any constraints among these 3 parameters? If yes, can we add validators to check the constraints?

yungezz: there's similiar comments on this behavior in  PR in core, https://github.com/Azure/azure-cli/pull/12572. hi @JunSun17 could you pls have a look and ping us if you want to talk offline. thanks.

gtracer: @fengzhou-msft the behavior that Jun that implemented is our desired behavior. To reiterate:
1. Yes we want to wipe the exiting values if any new value is provided
2. Yes set a default value in the RP but we don't want to fill that default value in a client like the CLI today (leaving that up to a layer in the RP)
3. There are very few constraints. Only notable ones are a) cannot set more than one value at once b) cannot have a managed_outbound_ips count of > 100 currently. But these constraints are currently held in the RP front-end because there is a possibility that they could change quickly and are rarely triggered

JunSun17: Yes, it is also explained in another PR, this change is intentionally made.

PR #1386, opened 2020-03-13
[Breadth Coverage] Stream Analytics

arrownj:
Onboard Stream Analytics

---

### General Guidelines

For new extensions:

### About Extension Publish


src/stream-analytics/azext_stream_analytics/_help.py
```diff
@@ -0,0 +1,370 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+# pylint: disable=too-many-lines
+# pylint: disable=line-too-long
+from knack.help_files import helps  # pylint: disable=unused-import
+
+
+helps['stream-analytics job'] = """
```
yungezz: `-` in command name is not recommend practice right?

src/stream-analytics/azext_stream_analytics/_help.py
```diff
@@ -0,0 +1,370 @@
...
+    short-summary: Commands to manage stream-analytics streaming job.
+"""
+
+helps['stream-analytics job create'] = """
+    type: command
+    short-summary: Creates a streaming job or replaces an already existing streaming job.
+    examples:
+      - name: Create a streaming job
+        text: |-
+               az stream-analytics job create --resource-group "sjrg6936" --name "sj59" \\
```
yungezz: for resource names in sample, let's consistent with `myResource`, such as myResourceGrouop etc.

src/stream-analytics/azext_stream_analytics/_help.py
```diff
@@ -0,0 +1,370 @@
...
+    examples:
+      - name: Get a streaming job
+        text: |-
+               az stream-analytics streaming-job show --resource-group "sjrg6936" --name "sj59"
+      - name: Get a streaming job and expand its inputs, outputs, transformation, and functions
+        text: |-
+               az stream-analytics streaming-job show --resource-group "sjrg3276" --name "sj7804" --expand
+"""
+
+helps['stream-analytics job list'] = """
```
yungezz: is there list by subs?

src/stream-analytics/azext_stream_analytics/_help.py
```diff
@@ -0,0 +1,370 @@
...
+    examples:
+      - name: List all streaming jobs
+        text: |-
+               az stream-analytics job list --resource-group "sjrg3276"
+      - name: List all streaming jobs and expand their inputs, outputs, transformation, and functions
+        text: |-
+               az stream-analytics job list --resource-group "sjrg6936" --expand
+"""
+
+helps['stream-analytics job start'] = """
```
yungezz: is the start a long running operation?

src/stream-analytics/azext_stream_analytics/_help.py
```diff
@@ -0,0 +1,370 @@
...
+    short-summary: Commands to manage stream-analytics input.
+"""
+
+helps['stream-analytics input create'] = """
+    type: command
+    short-summary: Creates an input or replaces an already existing input under an existing streaming job.
+    examples:
+      - name: Create an input
+        text: |-
+               az stream-analytics input create --resource-group "sjrg8440" --job-name "sj9597" --name \\
```
yungezz: is the serialization necessary? can we define default deliminitor?

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,180 @@
...
+    with self.argument_context('stream-analytics job create') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('name', options_list=['--name', '-n'], help='The name of the streaming job.')
+        c.argument('location', arg_type=get_location_type(self.cli_ctx))
+        c.argument('tags', tags_type)
+        c.argument('events_out_of_order_policy', arg_type=get_enum_type(EventsOutOfOrderPolicy),
+                   help='Indicates the policy to apply to events that arrive out of order in the input event stream.')
+        c.argument('output_error_policy', arg_type=get_enum_type(OutputErrorPolicy),
+                   help='Indicates the policy to apply to events that arrive at the output.')
+        c.argument('events_out_of_order_max_delay_in_seconds', type=int,
```
yungezz: there's general guideline not to specify unit of number in parameter name for future extensibility, but in help message

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,180 @@
...
+        c.argument('events_late_arrival_max_delay_in_seconds', type=int,
+                   help='The maximum tolerable delay in seconds where events arriving late could be included.  Supported range is -1 to 1814399 (20.23:59:59 days) and -1 is used to specify wait indefinitely.')
+        c.argument('data_locale',
+                   help='The data locale of the stream analytics job. Defaults to \'en-US\' if none specified.')
+
+    with self.argument_context('stream-analytics job update') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('name', options_list=['--name', '-n'], help='The name of the streaming job.')
+        c.argument('tags', tags_type)
+        c.argument('events_out_of_order_policy', arg_type=get_enum_type(EventsOutOfOrderPolicy),
```
yungezz: too long.. event-outoforder-policy?

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,180 @@
...
+                   help='The data locale of the stream analytics job.')
+
+    with self.argument_context('stream-analytics job delete') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('name', options_list=['--name', '-n'], help='The name of the streaming job.')
+
+    with self.argument_context('stream-analytics job show') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('name', options_list=['--name', '-n'], help='The name of the streaming job.')
+        c.argument('expand', action='store_true',
```
yungezz: what's general naming of this kind of argument in cli ? expand or verbose?

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,180 @@
...
+
+    with self.argument_context('stream-analytics job stop') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('name', options_list=['--name', '-n'], help='The name of the streaming job.')
+
+    with self.argument_context('stream-analytics input create') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('job_name', help='The name of the streaming job.')
+        c.argument('name', options_list=['--name', '-n'], help='The name of the input.')
+        c.argument('input_type', get_enum_type(['Stream', 'Reference']), help='The type of the input.')
```
yungezz: type is enough. command is input create, `input` is redundant

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,180 @@
...
+        c.argument('output_start_time',
+                   help='Output start time, must have a value if --output-start-mode is set to CustomTime.')
+
+    with self.argument_context('stream-analytics job stop') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('name', options_list=['--name', '-n'], help='The name of the streaming job.')
+
+    with self.argument_context('stream-analytics input create') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('job_name', help='The name of the streaming job.')
```
yungezz: should be job, and supports both name and resource id right?

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,180 @@
...
+        c.argument('name', options_list=['--name', '-n'], help='The name of the output.')
+
+    with self.argument_context('stream-analytics transformation create') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('job_name', help='The name of the streaming job.')
+        c.argument('name', options_list=['--name', '-n'], help='The name of the transformation.')
+        c.argument('streaming_units', help='The number of streaming units that the streaming job uses.')
+        c.argument('transformation_query', help='The query that will be run in the streaming job. You can learn more about the Stream Analytics Query Language (SAQL) here: https://msdn.microsoft.com/library/azure/dn834998 .')
```
yungezz: query

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,180 @@
...
+        c.argument('job_name', help='The name of the streaming job.')
+        c.argument('name', options_list=['--name', '-n'], help='The name of the transformation.')
+
+    with self.argument_context('stream-analytics function create') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('job_name', help='The name of the streaming job.')
+        c.argument('name', options_list=['--name', '-n'], help='The name of the function.')
+        c.argument('function_inputs', type=validate_file_or_dict, help='The inputs of the function.')
+        c.argument('function_output', type=validate_file_or_dict, help='The output of the function.')
+        c.argument('function_binding', type=validate_file_or_dict, help='The binding of the function.')
```
yungezz: remove function_ prefix from above 3 since redundant.

src/stream-analytics/azext_stream_analytics/custom.py
```diff
@@ -0,0 +1,199 @@
...
+def create_stream_analytics_job(cmd, client, resource_group, name, location=None, tags=None,
+                                events_out_of_order_policy=None, output_error_policy=None,
+                                events_out_of_order_max_delay_in_seconds=None,
+                                events_late_arrival_max_delay_in_seconds=None, data_locale=None):
+
+    body = {
+        'location': location or _get_resource_group_location(cmd.cli_ctx, resource_group),
+        'tags': tags,
+        'properties': {
+            'sku': {'name': 'Standard'},
```
yungezz: hardcode sku name?

src/stream-analytics/azext_stream_analytics/custom.py
```diff
@@ -0,0 +1,199 @@
...
+                                    transformation_name=name)
+
+
+def get_stream_analytics_transformation(cmd, client, resource_group, job_name, name):
+    return client.get(resource_group_name=resource_group, job_name=job_name, transformation_name=name)
+
+
+def create_stream_analytics_function(cmd, client, resource_group, job_name, name, function_inputs, function_output,
+                                     function_binding):
+    properties = {
```
yungezz: file exists check and any schema validation on input/output/binding?

src/stream-analytics/azext_stream_analytics/custom.py
```diff
@@ -0,0 +1,199 @@
...
+        'type': 'Scalar',
```
yungezz: hardcode type?

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: It  will conflict with the global `query` parameter, so add this prefix. There is a similar case for `output` below.

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: same as `query`, `output` will have conflict problem.

src/stream-analytics/azext_stream_analytics/custom.py
arrownj: yes, currently only `Standard` is available, that and this property is marked as required for REST api. So we need to hard code it here.

src/stream-analytics/azext_stream_analytics/custom.py
arrownj: in api 2016-03-01, only Scalar is support.

qwordy:
Can you share the speclet here?

src/stream-analytics/azext_stream_analytics/_help.py
arrownj: There are many commands have `-`, especially in extensions. Can we use a abbreviation such as `sa` for it ? I prefer `stream-analytics`, because I think it is clearer. Anyone else can give your opinions ?

src/stream-analytics/azext_stream_analytics/_help.py
arrownj: there are several different types of serialization(json, csv, avro, others), each one may have different properties.

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: `verbose` will conflict with global `verbose` parameter. Maybe we can use `expand-all` instead.

arrownj:
> Can you share the speclet here?

Here is the speclet: https://microsoft.sharepoint.com/:w:/r/teams/AzureManagementExperience/_layouts/15/guestaccess.aspx?e=4dwWBD&share=EdkCjo-EyDxFm8dpGrzOnCwBowMITT2zU98URXfkRN2SPg

src/stream-analytics/azext_stream_analytics/_params.py
yungezz: got it. how about the other 2

src/stream-analytics/azext_stream_analytics/custom.py
yungezz: what's type of sku? enum? if yes, we should expose sku as argument, with enum contains one value: Standard, maybe with default value

src/stream-analytics/azext_stream_analytics/custom.py
yungezz: same as sku

src/stream-analytics/azext_stream_analytics/_help.py
yungezz: ok

fengzhou-msft:
Add a `__init__.py` file in `vendored_sdks` so it will be packaged in wheel like [this](https://github.com/Azure/azure-cli-extensions/pull/1299/commits/25ad40151ad5d444968e4a7daf74fe10e5b05a2f).

src/stream-analytics/azext_stream_analytics/custom.py
arrownj: yes, enum value defined in spec. I consider this before, however give it up as last. Because I think actually user can do nothing with this parameter. When more values are supported, manual works are still needed even if we add it as a parameter currently.

And another similar property is `compatibility_level`, in the rest spec, it says the only value is 1.0(set to 1.1 can succeed, 1.2 will failed. currently portal support 1.0, 1.1, 1.2). So I just emit these properties and leave them to the new api version support in the future.

arrownj: for `compatibility_level`, as it is not marked as required, service will give it a default value 1.0 if it is not in the request body.

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,180 @@
...
+        c.argument('tags', tags_type)
+        c.argument('events_out_of_order_policy', arg_type=get_enum_type(EventsOutOfOrderPolicy),
+                   help='Indicates the policy to apply to events that arrive out of order in the input event stream.')
+        c.argument('output_error_policy', arg_type=get_enum_type(OutputErrorPolicy),
+                   help='Indicates the policy to apply to events that arrive at the output.')
+        c.argument('events_out_of_order_max_delay_in_seconds', type=int,
+                   help='The maximum tolerable delay in seconds where out-of-order events can be adjusted to be back in order.')
+        c.argument('events_late_arrival_max_delay_in_seconds', type=int,
+                   help='The maximum tolerable delay in seconds where events arriving late could be included.  Supported range is -1 to 1814399 (20.23:59:59 days) and -1 is used to specify wait indefinitely.')
+        c.argument('data_locale',
```
MyronFanQiu: should it be an enum?

src/stream-analytics/azext_stream_analytics/custom.py
yungezz: got it. it make sense when considering from implementation perspective. while from user perspective, user need to understand that which default sku is used. sku is key elemenet of pricing.

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,180 @@
...
+    with self.argument_context('stream-analytics input test') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('job_name', help='The name of the streaming job.')
+        c.argument('name', options_list=['--name', '-n'], help='The name of the input.')
+
+    with self.argument_context('stream-analytics output create') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('job_name', help='The name of the streaming job.')
+        c.argument('name', options_list=['--name', '-n'], help='The name of the output.')
+        c.argument('datasource', type=validate_file_or_dict, help='The datasource of the output.')
```
MyronFanQiu: is there any link for this argument?

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,180 @@
...
+        c.argument('serialization', type=validate_file_or_dict, help='The serialization of the output.')
```
MyronFanQiu: same here.

src/stream-analytics/azext_stream_analytics/_help.py
MyronFanQiu: can we provide examples for json, avro, others too?

src/stream-analytics/azext_stream_analytics/vendored_sdks/streamanalytics/models/_models.py
```diff
@@ -0,0 +1,2455 @@
...
+
+    _validation = {
+        'type': {'required': True},
+    }
+
+    _attribute_map = {
+        'type': {'key': 'type', 'type': 'str'},
+    }
+
+    _subtype_map = {
```
MyronFanQiu: well. It's quite complicate.

MyronFanQiu: I agree, for such a complicate definition, we can support json input. Is there any way to let user know how to use it?

src/stream-analytics/setup.py
```diff
@@ -0,0 +1,61 @@
...
+
+# The full list of classifiers is available at
+# https://pypi.python.org/pypi?%3Aaction=list_classifiers
+CLASSIFIERS = [
+    'Development Status :: 4 - Beta',
+    'Intended Audience :: Developers',
+    'Intended Audience :: System Administrators',
+    'Programming Language :: Python',
+    'Programming Language :: Python :: 2',
+    'Programming Language :: Python :: 2.7',
```
MyronFanQiu: should be removed.

src/stream-analytics/azext_stream_analytics/vendored_sdks/streamanalytics/models/_models.py
arrownj: yes, it's complicate for these parameters. Actually there are some examples in the rest spec examples, can we link it directly, or write them down in our README file or CLI examples (around 10 examples for each)?

MyronFanQiu: On the one hand, I totally agree that it's quite complicated and it would take too much effort to support without json argument. On the other hand, for such a complicated argument, it's quite hard to use a json argument for users. examples is good. Is it able to add some commands like `az vm diagnostics get-default-config -h` to give examples to users. Meanwhile add some simple checks and hints for users?

src/stream-analytics/azext_stream_analytics/_help.py
arrownj: thanks, updated.

src/stream-analytics/azext_stream_analytics/_help.py
arrownj: nice finding, list by subscription added.

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: parameter name updated.

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: update for both events_outoforder_policy and events_outoforder_max_delay.

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: user --expand-all now.

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: updated.

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: yes, id_part added, support both id and name now.

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: parameter name will keep `--transformation-query`

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: function_inputs -> inputs, function_binding -> binding.

src/stream-analytics/azext_stream_analytics/custom.py
arrownj: `--sku-name` added.

src/stream-analytics/azext_stream_analytics/custom.py
arrownj: this is handle by the common validator `validate_file_or_dict`, which do not raise error if file not exists. Will treat it as a string when file not exists. The json schema is validated in service side. There are several different types of these parameter, I think the service side will just look for the properties it needs, if not found or wrong, raise error.

src/stream-analytics/azext_stream_analytics/custom.py
arrownj: `type` added.

src/stream-analytics/setup.py
arrownj: nice finding, thanks. removed.

src/stream-analytics/azext_stream_analytics/custom.py
yungezz: ok. cli layer should do at least basic file schema check.

src/stream-analytics/azext_stream_analytics/_help.py
arrownj: yes. `job create`, `job delete`, `job start`, `job stop`, `input test`, `output test`, `function test`, all these are long running operations. no_wait added for all of these. Thanks for your findings.

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: From rest api spec perspective, it is defined as a string, though I also think it should be an enum. From CLI perspective, as sdk/spec didn't provide the value list, we also do not know all the available values. The original description of this property is: `The data locale of the stream analytics job. Value should be the name of a supported .NET Culture from the set https://msdn.microsoft.com/en-us/library/system.globalization.culturetypes(v=vs.110).aspx. Defaults to 'en-US' if none specified.` It is actually an enum of the .NET framework which have a long value list(a little bit different with normal locale value).
I will keep this description in the help message to give user advice about it.

src/stream-analytics/azext_stream_analytics/vendored_sdks/streamanalytics/models/_models.py
arrownj: @MyronFanQiu I add more json format examples for `input create`, `output create` and `function create`. Not cover all the cases, but most of them are covered.

src/stream-analytics/README.md
```diff
@@ -0,0 +1,181 @@
...
+```
+az stream-analytics function create \
+    --resource-group groupName \
+    --job-name jobName \
+    --name functionName \
+    --inputs @function_inputs.json \
+    --function-output @function_output.json \
+    --binding @function_binding.json
+```
+function_inputs.json contains the following content
```
yungezz: update to 'inputs' to align with argument name

src/stream-analytics/README.md
```diff
@@ -0,0 +1,181 @@
...
+    }
+]
+```
+function_output.json contains the following content
+```json
+{
+    "dataType": "Any"
+}
+```
+function_binding.json contains the following content
```
yungezz: update to 'binding' to align with argument name

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,186 @@
...
+
+    with self.argument_context('stream-analytics job show') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('name', options_list=['--name', '-n'], help='The name of the streaming job.')
+        c.argument('expand_all', action='store_true',
+                   help='Expand inputs, transformation, outputs and functions of the streaming job')
+
+    with self.argument_context('stream-analytics job list') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('expand_all', action='store_true',
```
yungezz: actually expand is enough..

src/stream-analytics/README.md
arrownj: updated

src/stream-analytics/README.md
arrownj: updated

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: change back^^

src/stream-analytics/azext_stream_analytics/_help.py
```diff
@@ -0,0 +1,547 @@
...
+
+
+helps['stream-analytics job'] = """
+    type: group
+    short-summary: Commands to manage stream-analytics streaming job.
+"""
+
+helps['stream-analytics job create'] = """
+    type: command
+    short-summary: Creates a streaming job or replaces an already existing streaming job.
```
fengzhou-msft: All short summaries in this file use third-person-singular verb form. Shall we use basic form?

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def load_arguments(self, _):
+
+    with self.argument_context('stream-analytics job create') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('name', options_list=['--name', '-n'], help='The name of the streaming job.')
+        c.argument('location', arg_type=get_location_type(self.cli_ctx))
+        c.argument('tags', tags_type)
+        c.argument('sku_name', arg_type=get_enum_type(SkuName), help='The name of the SKU.')
```
fengzhou-msft: use `--sku`?

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,186 @@
...
+    SkuName, OutputStartMode, EventsOutOfOrderPolicy, OutputErrorPolicy, UdfType, CompatibilityLevel
+)
+
+
+def load_arguments(self, _):
+
+    with self.argument_context('stream-analytics job create') as c:
+        c.argument('resource_group', resource_group_name_type)
+        c.argument('name', options_list=['--name', '-n'], help='The name of the streaming job.')
+        c.argument('location', arg_type=get_location_type(self.cli_ctx))
```
fengzhou-msft: do you want to get default location from resource group?

src/stream-analytics/azext_stream_analytics/_params.py
```diff
@@ -0,0 +1,186 @@
...
+from azure.cli.core.commands.validators import validate_file_or_dict
+from .vendored_sdks.streamanalytics.models import (
+    SkuName, OutputStartMode, EventsOutOfOrderPolicy, OutputErrorPolicy, UdfType, CompatibilityLevel
+)
+
+
+def load_arguments(self, _):
+
+    with self.argument_context('stream-analytics job create') as c:
+        c.argument('resource_group', resource_group_name_type)
```
fengzhou-msft: This parameter should be named `resource_group_name` in custom.py and then you can omit it in `_params.py` if you like.

src/stream-analytics/azext_stream_analytics/commands.py
```diff
@@ -0,0 +1,77 @@
...
+        operations_tmpl='azext_stream_analytics.vendored_sdks.streamanalytics.operations._streaming_jobs_operations#StreamingJobsOperations.{}',
+        client_factory=cf_jobs)
+    with self.command_group('stream-analytics job', stream_analytics_jobs, client_factory=cf_jobs) as g:
+        g.custom_command('create', 'create_stream_analytics_job', supports_no_wait=True)
+        g.custom_command('update', 'update_stream_analytics_job')
+        g.custom_command('delete', 'delete_stream_analytics_job', supports_no_wait=True)
+        g.custom_show_command('show', 'get_stream_analytics_job')
+        g.custom_command('list', 'list_stream_analytics_job')
+        g.custom_command('start', 'start_stream_analytics_job', validator=validate_streaming_job_start, supports_no_wait=True)
+        g.custom_command('stop', 'stop_stream_analytics_job', supports_no_wait=True)
```
fengzhou-msft: Add a wait command for command groups with `supports_no_wait=True` command.

src/stream-analytics/azext_stream_analytics/custom.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def get_stream_analytics_job(cmd, client, resource_group, name, expand=False):
+    expand = EXPANDED_ALL_STRING if expand else None
+    return client.get(resource_group_name=resource_group, job_name=name, expand=expand)
+
+
+def list_stream_analytics_job(cmd, client, resource_group=None, expand=False):
+    expand = EXPANDED_ALL_STRING if expand else None
+    if resource_group is None:
```
fengzhou-msft: Use `if not resource_group_name:` in order to use the workaround `-g=` or `-g ""` to temparorily disable appending the default resource group.

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: updated

src/stream-analytics/azext_stream_analytics/_help.py
arrownj: updated

src/stream-analytics/azext_stream_analytics/_params.py
arrownj: updated

src/stream-analytics/azext_stream_analytics/custom.py
arrownj: updated

src/stream-analytics/azext_stream_analytics/commands.py
arrownj: there are 4 subgroups which `has support_no_wait=True` command(`job`, `input`, `output`, `function`), however each of them reference the provisioningState of `job`. So I add `wait` command for `job` only. Could you please advice whether it is right ?

fengzhou-msft: If they use the same get api for job, then I think it is the right way.

PR #1388, opened 2020-03-13
[CI] Update trigger condition

mitchdenny:
This is just a suggestion for how I would setup my PR and CI triggers on your YAML file.


yonzhan:
add to S167

PR #1405, opened 2020-03-17
enable kubenet for udr

gtracer:
---

### General Guidelines

For new extensions:

### About Extension Publish


gtracer:
Not updating the index as I am not in a rush for this code to release. It can go out with the next index update

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -118,7 +118,7 @@
         - name: --outbound-type
           type: string
           short-summary: How outbound traffic will be configured for a cluster.
-          long-summary: Select between loadBalancer and userDefinedRouting. If not set, defaults to type loadBalancer. Requires --network-plugin to be azure and a standard load balancer SKU.
+          long-summary: Select between loadBalancer and userDefinedRouting. If not set, defaults to type loadBalancer. Requires --vnet-subnet-id to be provided with a preconfigured route table and --load-balancer-sku to be Standard.
```
yungezz: seems `--vnet-subnet-id` is an existing param. it's not a good name, normally we jsut call it `--subnet-id` in cli. anyway, not related to this PR direclty

yungezz: and this seems a breaking change, change from `network-plugin` to 'vnet-subnet-id`, users who replying on `network-plugin` is not able to set correct outbound-type? any compatibile code?

gtracer: Regardless, I appreciate the advice. If I have a free cycle I will try and send a PR to clean it up

gtracer: I see why it looks like a breaking change but it actually isn't. 

Today, the RP blocks you if you don't provide the vnet-subnet-id. The feature happens to be one where user's can configure their own networking so they must bring a subnet with an associated route table. So I am just bringing that functionality into the CLI. Overall experience for customer's should be identical

The second change which removes the network-plugin is actually relaxing the requirements. Instead of needing type azure you can use any type. We don't want to call it out in the release notes yet though as we are only sharing it with customer's that have communicated a desire for it who will put some miles on it with us

yungezz: thanks for explanation, make sense

gtracer:
Good to merge from my end @yungezz

PR #1412, opened 2020-03-19
[db-up] Fix bug for server name and enable ssl enforecement

Juliehzl:
---
Fix https://github.com/Azure/azure-cli/issues/11633 , https://github.com/Azure/azure-cli/issues/11635, https://github.com/Azure/azure-cli/issues/12340

The PR will change the behavior for some commands:
1. ssl_enforcement will be enabled by default
2. When resource group name in namespace is different from what in cache, reset it and create new server name
3. Bump psycopg2 in MacOS, refering https://github.com/Azure/azure-cli-extensions/pull/1675/files

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/db-up/azext_db_up/_validators.py
+++ b/src/db-up/azext_db_up/_validators.py
@@ -67,8 +67,8 @@ def _process_db_up_namespace(cmd, namespace, db_type=None):
         try:
             resource_client.resource_groups.get(namespace.resource_group_name)
             create_resource_group = False
-        except CloudError as ex:  # throw exception when resource group name is invalid
-            raise ex
+        except CloudError:  # throw exception when resource group name is invalid
+            pass
 
     if create_resource_group:
         # create new resource group
--- a/src/db-up/azext_db_up/_validators.py
+++ b/src/db-up/azext_db_up/_validators.py
@@ -30,6 +30,7 @@ def db_down_namespace_processor(db_type):
     return lambda cmd, namespace: _process_db_down_namespace(namespace, db_type=db_type)
 
 
+# pylint: disable=bare-except
 def _process_db_up_namespace(cmd, namespace, db_type=None):
     # populate from cache if existing when no resource group name provided
     resource_client = resource_client_factory(cmd.cli_ctx)
@@ -37,9 +38,8 @@ def _process_db_up_namespace(cmd, namespace, db_type=None):
         _set_value(db_type, namespace, 'resource_group_name', 'group', cache=False)
         try:
             resource_client.resource_groups.get(namespace.resource_group_name)
-        except Exception:  # Clear resource group name information when it is invalid
+        except:  # Clear resource group name information when it is invalid
             namespace.resource_group_name = None
-            pass
 
     # populate from cache if existing when no location provided
     if namespace.location is None:
--- a/src/db-up/azext_db_up/_validators.py
+++ b/src/db-up/azext_db_up/_validators.py
@@ -31,42 +31,58 @@ def db_down_namespace_processor(db_type):
 
 
 def _process_db_up_namespace(cmd, namespace, db_type=None):
-    # populate from cache if existing
-    _set_value(db_type, namespace, 'location', 'location', cache=False)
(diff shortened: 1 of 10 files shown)


yonzhan:
add to S167

src/db-up/azext_db_up/_validators.py
```diff
@@ -30,43 +30,59 @@ def db_down_namespace_processor(db_type):
...
 def _process_db_up_namespace(cmd, namespace, db_type=None):
-    # populate from cache if existing
-    _set_value(db_type, namespace, 'location', 'location', cache=False)
-    _set_value(db_type, namespace, 'resource_group_name', 'group', cache=False)
+    # populate from cache if existing when no resource group name provided
+    resource_client = resource_client_factory(cmd.cli_ctx)
+    if namespace.resource_group_name is None:
+        _set_value(db_type, namespace, 'resource_group_name', 'group', cache=False)
+        try:
+            resource_client.resource_groups.get(namespace.resource_group_name)
```
zhoxing-ms: Why is there no operation to assign the returned result to a variable?

src/db-up/azext_db_up/_validators.py
```diff
@@ -30,43 +30,59 @@ def db_down_namespace_processor(db_type):
...
+        _set_value(db_type, namespace, 'location', 'location', cache=False)
+    # generate smart defaults when namespace.location is None
     if _get_value(db_type, namespace, 'location', 'location') is None:
         try:
             get_default_location_from_resource_group(cmd, namespace)
         except (CLIError, ValidationError):
-            namespace.location = 'westus2'
-    _set_value(db_type, namespace, 'location', 'location', namespace.location)
+            namespace.location = 'eastus'
+    _set_value(db_type, namespace, 'location', 'location', default=namespace.location)
```
zhoxing-ms: Why is it changed from `westus2` to `eastus`

src/db-up/azext_db_up/util.py
```diff
@@ -30,8 +30,12 @@ def create_random_resource_name(prefix='azure', length=15):
...
     config_section = CONFIG_MAP[db_type]
     if fallback == '_fallback_none':
-        return DB_CONFIG.get(config_section, option)
-    return DB_CONFIG.get(config_section, option, fallback=fallback)
+        result = DB_CONFIG.get(config_section, option)
+    else:
+        result = DB_CONFIG.get(config_section, option, fallback=fallback)
+    if result:
+        return result
+    return None
```
zhoxing-ms: If the `result` is `0` or `False`, the result of function returns None, is that as expected?

src/db-up/azext_db_up/_validators.py
```diff
@@ -30,43 +30,59 @@ def db_down_namespace_processor(db_type):
...
+    if namespace.resource_group_name is None:  # No resource group provided and in cache
         namespace.resource_group_name = create_random_resource_name('group')
     else:
         try:
             resource_client.resource_groups.get(namespace.resource_group_name)
             create_resource_group = False
-        except CloudError:
-            pass
+        except CloudError as ex:  # throw exception when resource group name is invalid
+            raise ex
```
qianwens: just remove the try except

zhoxing-ms: Below is the logic of create resource group. The original logic was that if the resource group was not found in the Service, the resource group was created on the Service and the rest of the process was followed.
Is it in line with the expectation to directly exit the following process?

Juliehzl:
It is hard to record such scenario with playback test, so make them live_only. And all tests passed locally:
image

src/db-up/azext_db_up/util.py
Juliehzl: yes

src/db-up/azext_db_up/_validators.py
Juliehzl: westus2 will be not available in some cases.

src/db-up/azext_db_up/_validators.py
Juliehzl: yes. Here is to check if the resource group name in cache is still available.

PR #1418, opened 2020-03-19
Sync extensions for JEDI

fengzhou-msft:
This script will be used in an internal [pipeline](https://dev.azure.com/azure-sdk/internal/_build?definitionId=1608&_a=summary) to sync up new extensions, triggered by new commits in master branch.

It also supports cleaning up and syncing all extensions from the master branch index.json file.

---

### General Guidelines

For new extensions:

### About Extension Publish


scripts/ci/sync_extensions.py
```diff
@@ -0,0 +1,141 @@
...
+from subprocess import check_output
+
+DEFAULT_TARGET_INDEX_URL = os.getenv('AZURE_EXTENSION_TARGET_INDEX_URL')
+STORAGE_ACCOUNT_KEY = os.getenv('AZURE_EXTENSION_TARGET_STORAGE_ACCOUNT_KEY')
+STORAGE_ACCOUNT = os.getenv('AZURE_EXTENSION_TARGET_STORAGE_ACCOUNT')
+STORAGE_CONTAINER = os.getenv('AZURE_EXTENSION_TARGET_STORAGE_CONTAINER')
+
+
+def _get_updated_extension_names():
+    cmd = 'git --no-pager diff --diff-filter=ACMRT HEAD~1 -- src/index.json'
```
yungezz: I'm still little concern on replying on change on index.json. what if the pipeline run failed? new run need to catch up extension blob changes in last several index.json change.

scripts/ci/sync_extensions.py
```diff
@@ -0,0 +1,141 @@
...
+    updated_content = check_output(cmd.split()).decode('utf-8')
+    updated_urls = [line.replace('+', '') for line in updated_content.splitlines() if line.startswith('+') and not line.startswith('+++') and 'downloadUrl' in line]
```
yungezz: why not load the json content and get from key directly?

scripts/ci/sync_extensions.py
```diff
@@ -0,0 +1,141 @@
...
+    return updated_exts
+
+
+def _download_file(url):
+    import requests
+    count = 3
+    the_ex = None
+    while count > 0:
+        try:
+            response = requests.get(url, allow_redirects=True)
```
yungezz: since whl package size may be very big, is this lib supporting resume from break-point? if there any more reliable/dedicated package downloading library?

haroldrandom:
Can't we move this inside ADO

fengzhou-msft:
> Can't we move this inside ADO

code too long, ADO inline script will truncate it.

haroldrandom:
> > Can't we move this inside ADO
> 
> code too long, ADO inline script will truncate it.

I was wondering whether there is a better place to place this script instead of exposing it to the public because user don't need it.

scripts/ci/sync_extensions.py
fengzhou-msft: Here we only extract updated downloadUrl from git diff for incremental sync.

scripts/ci/sync_extensions.py
fengzhou-msft: Added environment variable to control commit numbers to diff. So in case of failure, we can manually trigger the pipeline and pick up all the changes for the last n commits.

scripts/ci/sync_extensions.py
fengzhou-msft: Changed to stream downloading so we can handle big files well.

scripts/ci/sync_extensions.py
```diff
@@ -0,0 +1,146 @@
...
+                curr_index['extensions'][extension_name].append(entry)
+
+    with open(os.path.join(target_index_path), 'w') as outfile:
+        outfile.write(json.dumps(curr_index, indent=4, sort_keys=True))
+
+
+def main():
+    import shutil
+    import tempfile
+    from azure.storage.blob import BlockBlobService
```
haroldrandom: install azure.storage.blob in Bash task?

fengzhou-msft: yes, as a prerequisite.

scripts/ci/sync_extensions.py
```diff
@@ -0,0 +1,146 @@
...
+def _get_updated_extension_names():
+    cmd = 'git --no-pager diff --diff-filter=ACMRT HEAD~{} -- src/index.json'.format(COMMIT_NUM)
+    updated_content = check_output(cmd.split()).decode('utf-8')
+    updated_urls = [line.replace('+', '') for line in updated_content.splitlines() if line.startswith('+') and not line.startswith('+++') and 'downloadUrl' in line]
+    updated_exts = []
+    NAME_REGEX = r'.*/([^/]*)-\d+.\d+.\d+'
+    if not updated_urls:
+        return updated_exts
+    for line in updated_urls:
+        search_result = re.search(r'"downloadUrl":\s+"(.*?)"', line)
```
haroldrandom: What if user manually update src/index.json and input `'downloadUrl'`?

fengzhou-msft: This also detects such case and is the same process as update by bot.

haroldrandom: I guess the re expression `r'"downloadUrl":\s+"(.*?)"` won't be ablt to search `'download'`

scripts/ci/sync_extensions.py
```diff
@@ -0,0 +1,168 @@
...
+STORAGE_ACCOUNT = os.getenv('AZURE_EXTENSION_TARGET_STORAGE_ACCOUNT')
+STORAGE_CONTAINER = os.getenv('AZURE_EXTENSION_TARGET_STORAGE_CONTAINER')
+COMMIT_NUM = os.getenv('AZURE_EXTENSION_COMMIT_NUM') or 1
+
+
+def _get_updated_extension_filenames():
+    cmd = 'git --no-pager diff --diff-filter=ACMRT HEAD~{} -- src/index.json'.format(COMMIT_NUM)
+    updated_content = check_output(cmd.split()).decode('utf-8')
+    FILENAME_REGEX = r'"filename":\s+"(.*?)"'
+    added_ext_filenames = [re.findall(FILENAME_REGEX, line)[0] for line in updated_content.splitlines() if line.startswith('+') and not line.startswith('+++') and 'filename' in line]
```
yungezz: I remember --name could show changed file name. suggest to check git diff command

yungezz: found it :) my sample used before: git diff $branch remotes/origin/master --name-only  -- * --no-pager

scripts/ci/sync_extensions.py
fengzhou-msft: no, it won't.

PR #1428, opened 2020-03-21
[Breadth Coverage] Portal

YalinLi0312:
Speclet: https://aka.ms/AA7uas2

Files have customizations should focus on:
-  src/portal/azext_portal/generated/commands.py
-  src/portal/azext_portal/generated/custom.py
-  src/portal/azext_portal/generated/_help.py
-  src/portal/azext_portal/generated/_params.py
-  src/portal/azext_portal/tests/latest/dashboard.json
-  src/portal/azext_portal/tests/latest/properties-update.json
-  src/portal/azext_portal/tests/latest/properties.json
-  src/portal/azext_portal/tests/latest/test_portal_scenario.py
-  src/portal/azext_portal/vendored_sdks/portal/operations/_dashboard_operations.py
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
Portal

YalinLi0312:
> Portal

Nice catch! :)

src/portal/azext_portal/generated/_params.py
```diff
@@ -0,0 +1,46 @@
...
+    with self.argument_context('portal dashboard show') as c:
+        c.argument('resource_group_name', resource_group_name_type, help='The name of the resource group.')
+        c.argument('name', help='The name of the dashboard.')
+
+    with self.argument_context('portal dashboard create') as c:
+        c.argument('resource_group_name', resource_group_name_type, help='The name of the resource group.')
+        c.argument('name', help='The name of the dashboard.')
+        c.argument('location', arg_type=get_location_type(self.cli_ctx), help='Resource location')
+        c.argument('tags', tags_type, help='Resource tags')
+        c.argument('input_path', type=file_type, help='The path to the dashboard properties JSON file.', completer=FilesCompleter())
```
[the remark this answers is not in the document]
markcowl: from inspection of the api call, looks ike the comm and line will be preferred, which, for this api call would be thr right thing

src/portal/azext_portal/generated/custom.py
```diff
@@ -0,0 +1,92 @@
...
+        raise CLIError('File not Found: {}'.format(str(ex)))
+
+
+def parse_dashboard_json(input_path):
+    try:
+        with open(input_path) as json_file:
+            try:
+                dashboard = json.load(json_file)
+            except json.decoder.JSONDecodeError as ex:
+                raise CLIError('JSON decode error for {}: {}'.format(json_file, str(ex)))
```
markcowl:
```suggestion
                raise CLIError('There was an error decoding the JSON file {}: {}'.format(json_file, str(ex)))
```

src/portal/azext_portal/generated/custom.py
```diff
@@ -0,0 +1,92 @@
...
+            if 'location' not in dashboard:
+                raise CLIError(str(json_file) + " does not contain the property 'location'")
```
markcowl: This decoding works for the import method, but should we have a different validation for the create and update methods, since these do not use the location?

src/portal/azext_portal/tests/latest/test_portal_scenario.py
```diff
@@ -0,0 +1,62 @@
...
+            'testDashboard': self.create_random_name(prefix='cli_test_dashboards'[:9], length=24)
+        })
+
+        self.cmd('az portal dashboard create '
+                 '--location "eastus" '
+                 '--input- "src/portal/azext_portal/tests/latest/properties.json" '
+                 '--tags aKey=aValue anotherKey=anotherValue '
+                 '--name "{testDashboard}" '
+                 '--resource-group "{rg}"',
+                 checks=[JMESPathCheck('name', self.kwargs.get('testDashboard', ''))])
```
markcowl: I am not familiar with how these checks work, are these checking that the name of the returned value matches 'testDashboard', or is this a check on the request?

markcowl: If this checks the output, it would be nice to have a check on each of the properties, rather than just one - for create, for update, and for show.  This essentially verifies that the api is modeled correctly

src/portal/azext_portal/tests/latest/test_portal_scenario.py
```diff
@@ -0,0 +1,62 @@
...
+        self.cmd('az portal dashboard delete '
+                 '--name "{testDashboard}" '
+                 '--resource-group "{rg}"',
+                 checks=[])
+
+        self.cmd('az portal dashboard import '
+                 '--input-path "src/portal/azext_portal/tests/latest/dashboard.json" '
+                 '--name "{testDashboard}" '
+                 '--resource-group "{rg}"',
+                 checks=[JMESPathCheck('name', 'testdashboard')])
```
markcowl: We should have an az portal dashboard show check after this, to verify that the dashboard was correcty imported, and has the expected properties.

Otherwise, I really like this test

src/portal/report.md
```diff
@@ -0,0 +1,48 @@
...
+
+delete a portal dashboard.
+
+|Option|Type|Description|Path (SDK)|Path (swagger)|
+|------|----|-----------|----------|--------------|
+|**--resource_group_name**|string|The name of the resource group.|resource_group_name|resource_group_name|
+|**--name**|string|The name of the dashboard.|dashboard_name|name|
+### portal dashboard list
+
+list a portal dashboard.
```
markcowl: SHould we have import in here?

src/portal/azext_portal/generated/_params.py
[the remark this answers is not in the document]
YalinLi0312: > Since tags and location can be entered in the json file, which values will be used if the values provided in parameters and the values provided in the json file conflict?

The JSON file used in create only contains dashboard properties(lenses and metadata), so would not have value conflicts. 
I think you may ask the conflict in "import". When do import, dashboard name can exist in dashboard JSON file and parameter. If the name entered in JSON file would use this value, otherwise would use the one in parameter.

src/portal/azext_portal/generated/_params.py
```diff
@@ -0,0 +1,46 @@
...
+
+    with self.argument_context('portal dashboard update') as c:
+        c.argument('resource_group_name', resource_group_name_type, help='The name of the resource group.')
+        c.argument('name', help='The name of the dashboard.')
+        c.argument('input_path', type=file_type, help='The path to the dashboard properties JSON file.', completer=FilesCompleter())
```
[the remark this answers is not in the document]
YalinLi0312: > What happens if the user provides a change to the location or tags in the input json file? Is this data lost?

The json file used in "update" is dashboard properties(lenses and metadata) as well. If users are allowed to change the location, that would be creating a new dashboard. And "tag" is also not allowed to change here, I think I can add it as an optional parameter. What do you think?

src/portal/azext_portal/generated/_params.py
```diff
@@ -0,0 +1,46 @@
...
+
+    with self.argument_context('portal dashboard delete') as c:
+        c.argument('resource_group_name', resource_group_name_type, help='The name of the resource group.')
+        c.argument('name', help='The name of the dashboard.')
+
+    with self.argument_context('portal dashboard import') as c:
```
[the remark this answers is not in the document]
YalinLi0312: > Given the create command above, not sure what additional functionality this command provides. Seems like we should remove it,

No, we should keep it. As our offline discussion, we add "import  -g --name --dashboard" to support creating a dashboard from the json file which download from website.

src/portal/azext_portal/generated/custom.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def portal_dashboard_create(cmd, client,
+                            resource_group_name,
+                            name,
+                            location,
+                            input_path,
+                            tags=None):
+    properties_lenses, properties_metadata = parse_properties_json(input_path)
+    return client.create_or_update(resource_group_name=resource_group_name, dashboard_name=name, location=location, tags=tags, lenses=properties_lenses, metadata=properties_metadata)
```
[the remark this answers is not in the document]
YalinLi0312: > So we ignore the location and tags in the input file

Yes. location and tags are in parameters, the "input-file" here is the dashboard properties. Same as in dashboard update

src/portal/azext_portal/generated/custom.py
```diff
@@ -0,0 +1,92 @@
...
+                            name):
+    return client.delete(resource_group_name=resource_group_name, dashboard_name=name)
+
+
+def portal_dashboard_import(cmd, client,
+                            resource_group_name,
+                            input_path,
+                            name=None):
+    dashboard = parse_dashboard_json(input_path)
+    return client.dashboard_import(resource_group_name=resource_group_name, dashboard_name=dashboard.get('name', name), dashboard=dashboard)
```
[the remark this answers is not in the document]
YalinLi0312: > where is this method implemented?

_dashboard_operations.py: line 427

src/portal/azext_portal/tests/latest/test_portal_scenario.py
YalinLi0312: > I am not familiar with how these checks work, are these checking that the name of the returned value matches 'testDashboard', or is this a check on the request?

Checks the name in returned value

YalinLi0312: > If this checks the output, it would be nice to have a check on each of the properties, rather than just one - for create, for update, and for show. This essentially verifies that the api is modeled correctly

Good point. I will change to check every property in dashboard

src/portal/azext_portal/generated/custom.py
YalinLi0312: > This decoding works for the import method, but should we have a different validation for the create and update methods, since these do not use the location?

I added a validation for create/update in line 57 in this file.

src/portal/azext_portal/tests/latest/test_portal_scenario.py
YalinLi0312: > We should have an az portal dashboard show check after this, to verify that the dashboard was correcty imported, and has the expected properties.
> 
> Otherwise, I really like this test

I don't think we need "az dashboard show" after to check the result. Because "import" have returned the deserialized dashboard response(line484/487 in _dashboard_operations.py), which is same as the result in "show".

src/portal/report.md
YalinLi0312: > SHould we have import in here?

idk where this file in use, but it looks we should add "import" here as well.

yungezz:
hi @fengzhou-msft could you pls help Yalin reviewing this? thanks.

src/portal/azext_portal/generated/_params.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def load_arguments(self, _):
+
+    with self.argument_context('portal dashboard list') as c:
+        c.argument('resource_group_name', resource_group_name_type, help='The name of the resource group.')
+
+    with self.argument_context('portal dashboard show') as c:
+        c.argument('resource_group_name', resource_group_name_type, help='The name of the resource group.')
+        c.argument('name', help='The name of the dashboard.')
```
fengzhou-msft: Please change the argument name to the full `dashboard_name`, but add `options_list=['--name', '-n']`.

src/portal/azext_portal/generated/_params.py
```diff
@@ -0,0 +1,46 @@
...
+
+    with self.argument_context('portal dashboard create') as c:
+        c.argument('resource_group_name', resource_group_name_type, help='The name of the resource group.')
+        c.argument('name', help='The name of the dashboard.')
+        c.argument('location', arg_type=get_location_type(self.cli_ctx), help='Resource location')
+        c.argument('tags', tags_type, help='Resource tags')
```
fengzhou-msft: Remove `help='Resource tags'` since `tags_type` carries a more detailed help message.

src/portal/azext_portal/generated/_params.py
```diff
@@ -0,0 +1,46 @@
...
+        c.argument('resource_group_name', resource_group_name_type, help='The name of the resource group.')
+
+    with self.argument_context('portal dashboard show') as c:
+        c.argument('resource_group_name', resource_group_name_type, help='The name of the resource group.')
+        c.argument('name', help='The name of the dashboard.')
+
+    with self.argument_context('portal dashboard create') as c:
+        c.argument('resource_group_name', resource_group_name_type, help='The name of the resource group.')
+        c.argument('name', help='The name of the dashboard.')
+        c.argument('location', arg_type=get_location_type(self.cli_ctx), help='Resource location')
```
fengzhou-msft: Recommend to add `validator=get_default_location_from_resource_group`, when --location is not provided (and the default location is also not set), it will use the location of the resource group. Also the help message can be removed since `get_location_type` has a more detailed help message.

src/portal/azext_portal/generated/commands.py
```diff
@@ -0,0 +1,21 @@
...
+    from azext_portal.generated._client_factory import cf_dashboard
+    portal_dashboard = CliCommandType(
+        operations_tmpl='azext_portal.vendored_sdks.portal.operations._dashboard_operations#DashboardOperations.{}',
+        client_factory=cf_dashboard)
+    with self.command_group('portal dashboard', portal_dashboard, client_factory=cf_dashboard) as g:
+        g.custom_command('list', 'portal_dashboard_list')
+        g.custom_show_command('show', 'portal_dashboard_show')
+        g.custom_command('create', 'portal_dashboard_create')
+        g.custom_command('update', 'portal_dashboard_update')
+        g.custom_command('delete', 'portal_dashboard_delete')
```
fengzhou-msft: Recommend to add `confirmation=True` for delete.

src/portal/azext_portal/generated/custom.py
```diff
@@ -0,0 +1,92 @@
...
+# pylint: disable=line-too-long
+# pylint: disable=too-many-lines
+
+import json
+from knack.util import CLIError
+
+
+def portal_dashboard_list(cmd, client,
+                          resource_group_name=None):
+    if resource_group_name is not None:
```
fengzhou-msft: Please use `if resource_group_name` so we can use the `-g ""` workaround to disable the passing of default resource group.

src/portal/setup.py
```diff
@@ -0,0 +1,60 @@
...
+# The full list of classifiers is available at
+# https://pypi.python.org/pypi?%3Aaction=list_classifiers
+CLASSIFIERS = [
+    'Development Status :: 4 - Beta',
+    'Intended Audience :: Developers',
+    'Intended Audience :: System Administrators',
+    'Programming Language :: Python',
+    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.4',
+    'Programming Language :: Python :: 3.5',
```
fengzhou-msft: Python 3.4 and 3.5 could be removed as we dropped support for them in Azure CLI.

src/portal/azext_portal/generated/_params.py
YalinLi0312: Why you prefer `dashboard_name`?
I think 1)"name" is cleaner and clearer since every cmd followed `az portal dashboard XXX`, the name param is definitely for the portal dashboard. 2)For consistency consideration, I notice other extensions are using `--name` like storage, VM, container registry, etc.

src/portal/azext_portal/generated/custom.py
YalinLi0312: I'm not clear what you mean. Here `-g` is an optional parameter, if it is not None will "list by this resource group", otherwise will "list by the subscription" which configured in your az account. I don't see there is a place that should `disable the passing of default resource group`.
Pls correct me if I misunderstood.

YalinLi0312:
@fengzhou-msft Thanks for your review, and addressed your comments. PTAL

src/portal/azext_portal/generated/custom.py
fengzhou-msft: When users run `az portal dashboard list`, they expect to list by subscription. But if they have run  `az configure -d group=myDefaultResourceGroup`, then the previous command will append `-g myDefaultResourceGroup` by CLI. If users still want to list by subscription in this case, they can use the workaround `az portal dashboard list -g ""` or `az portal dashboard list -g=`. So here we allow users to pass an empty resource group name to avoid appending the default resource group. 

We may want to have an explicit switch to temporarily turn off the default values, but before that is implemented, let's keep the workaround working for list command.

src/portal/azext_portal/generated/_params.py
fengzhou-msft: Using `options_list=['--name', '-n']` also allows you to use `--name` as the option plus the more convenient short option `-n`, but for argument name, there are cases that you could have more than three resource names in the same command. To make the names clear during development and maintenance, the convention is to use the full name for argument.

src/portal/azext_portal/generated/custom.py
fengzhou-msft: @RodgeFu can we make this change in codegen for the list function that could either list by resource group or list by subscription? Use `if resource_group_name` instead of `if resource_group_name is not None`.

src/portal/azext_portal/generated/_params.py
YalinLi0312: I renamed the parameter though the configuration in readme.az.md file in azure-rest-api-specs repo. That's why the argument name also changed to `name`. But it looks cmd params can be renamed in `option-list` as well. I'm confused about which one is better to use.
@fengzhou-msft @RodgeFu Any suggestion?

src/portal/azext_portal/generated/custom.py
YalinLi0312: > When users run `az portal dashboard list`, they expect to list by subscription. But if they have run `az configure -d group=myDefaultResourceGroup`, then the previous command will append `-g myDefaultResourceGroup` by CLI. If users still want to list by subscription in this case, they can use the workaround `az portal dashboard list -g ""` or `az portal dashboard list -g=`. So here we allow users to pass an empty resource group name to avoid appending the default resource group. The original `if resource_group_name is not None` will still list by resource group with the empty resource group, so we'd better use `if resource_group_name`.
> 
> We may want to have an explicit switch to temporarily turn off the default values, but before that is implemented, let's keep the workaround working for list command.

Got it, I didn't know the default rg setting.

src/portal/azext_portal/tests/latest/test_portal_scenario.py
markcowl: Using the show verifies that the import worked correctly, including that it created the dashboard with the correct resource id.  It is generally a good practice to use a get after a set to verify that the service correctly created and can retrieve the object

src/portal/report.md
zikalino: well, this file may be entirely removed. it's an extra.

qwordy:
In `az portal dashboard create`, there is only one parameter `--input-path`. I suggest expose the parameters to users instead of asking for a json file. I don't know how my json file should be like and I can't get help from examples unless I read the REST reference. Only advanced users have the ability and intention to learn REST reference. This design violates the purpose of Azure CLI.

Example in swagger,
```
{
  "parameters": {
    "dashboardName": "testDashboard",
    "resourceGroupName": "testRG",
    "api-version": "2019-01-01-preview",
    "subscriptionId": "00000000-0000-0000-0000-000000000000",
    "dashboard": {
      "properties": {
        "lenses": {
          "aLens": {
            "order": 1,
            "parts": {
              "aPart": {
                "position": {
                  "x": 1,
                  "y": 2,
                  "rowSpan": 4,
                  "colSpan": 3
                }
              },
              "bPart": {
                "position": {
                  "x": 5,
                  "y": 5,
                  "rowSpan": 6,
                  "colSpan": 6
                }
              }
            }
          },
          "bLens": {
            "order": 2,
            "parts": {}
          }
        },
        "metadata": {
          "metadata": {
            "X": 4,
            "Y": 3,
            "RowSpan": 1,
            "ColSpan": 2
          }
        }
      },
      "location": "eastus",
      "tags": {
        "aKey": "aValue",
        "anotherKey": "anotherValue"
      }
    }
  }
```

bquantump:
> In `az portal dashboard create`, there is only one parameter `--input-path`. I suggest expose the parameters to users instead of asking for a json file. I don't know how my json file should be like and I can't get help from examples unless I read the REST reference. Only advanced users have the ability and intention to learn REST reference. This design violates the purpose of Azure CLI.
> 
> Example in swagger,
> 
> ```
> {
>   "parameters": {
>     "dashboardName": "testDashboard",
>     "resourceGroupName": "testRG",
>     "api-version": "2019-01-01-preview",
>     "subscriptionId": "00000000-0000-0000-0000-000000000000",
>     "dashboard": {
>       "properties": {
>         "lenses": {
>           "aLens": {
>             "order": 1,
>             "parts": {
>               "aPart": {
>                 "position": {
>                   "x": 1,
>                   "y": 2,
>                   "rowSpan": 4,
>                   "colSpan": 3
>                 }
>               },
>               "bPart": {
>                 "position": {
>                   "x": 5,
>                   "y": 5,
>                   "rowSpan": 6,
>                   "colSpan": 6
>                 }
>               }
>             }
>           },
>           "bLens": {
>             "order": 2,
>             "parts": {}
>           }
>         },
>         "metadata": {
>           "metadata": {
>             "X": 4,
>             "Y": 3,
>             "RowSpan": 1,
>             "ColSpan": 2
>           }
>         }
>       },
>       "location": "eastus",
>       "tags": {
>         "aKey": "aValue",
>         "anotherKey": "anotherValue"
>       }
>     }
>   }
> ```

A user should not be expected to configure this file themselves. This files is generated from the portal designer and the user just grabs it from there. They don't need any advanced knowledge or anything of that sorts as you are suggesting. Not sure if the service team was directly consulted on this, though. Also, comments "this design violates the purpose of Azure CLI" aren't very helpful without further explanation.

YalinLi0312:
> In `az portal dashboard create`, there is only one parameter `--input-path`. I suggest expose the parameters to users instead of asking for a json file. I don't know how my json file should be like and I can't get help from examples unless I read the REST reference. Only advanced users have the ability and intention to learn REST reference. This design violates the purpose of Azure CLI.
> 
> Example in swagger,
> 
> ```
> {
>   "parameters": {
>     "dashboardName": "testDashboard",
>     "resourceGroupName": "testRG",
>     "api-version": "2019-01-01-preview",
>     "subscriptionId": "00000000-0000-0000-0000-000000000000",
>     "dashboard": {
>       "properties": {
>         "lenses": {
>           "aLens": {
>             "order": 1,
>             "parts": {
>               "aPart": {
>                 "position": {
>                   "x": 1,
>                   "y": 2,
>                   "rowSpan": 4,
>                   "colSpan": 3
>                 }
>               },
>               "bPart": {
>                 "position": {
>                   "x": 5,
>                   "y": 5,
>                   "rowSpan": 6,
>                   "colSpan": 6
>                 }
>               }
>             }
>           },
>           "bLens": {
>             "order": 2,
>             "parts": {}
>           }
>         },
>         "metadata": {
>           "metadata": {
>             "X": 4,
>             "Y": 3,
>             "RowSpan": 1,
>             "ColSpan": 2
>           }
>         }
>       },
>       "location": "eastus",
>       "tags": {
>         "aKey": "aValue",
>         "anotherKey": "anotherValue"
>       }
>     }
>   }
> ```


The dashboard JSON template can export from the Azure website, this is an easy way user know how the JSON file should look like. For creating a dashboard the cmd is `az portal dashboard create --name  --location -rg --input-path <the path to dashboard properties JSON file> --tag`. I was planned to expose every parameter to users, but `lenses` and `metadata` can be very complex, that's the main reason I put them to one parameter in a JSON file.

qwordy:
If the file can be generated, it is great. My personal understanding of one of goals of Azure CLI: provide a more user-friendly way to manage resource compared to writing and sending HTTP request.
Personally, I like JSON file. It saves lots of time developing CLI commands. But usually we need to expose parameters and add help for them to users. This is how most CLI commands are designed.

src/portal/README.rst
```diff
@@ -0,0 +1,5 @@
+Microsoft Azure CLI 'portal' Extension
+==========================================
+
+This package is for the 'portal' extension.
+i.e. 'az portal'
```
fengzhou-msft, line 5: Please put more content in this readme file introducing this extension. An example is [here](https://github.com/Azure/azure-cli-extensions/tree/master/src/storage-preview).

src/portal/azext_portal/generated/_params.py
```diff
@@ -0,0 +1,62 @@
...
+
+    with self.argument_context('portal dashboard show') as c:
+        c.argument('resource_group_name', resource_group_name_type,
+                   help='The name of the resource group.')
+        c.argument('name', options_list=[
+                   '--name', '-n'], help='The name of the dashboard.')
+
+    with self.argument_context('portal dashboard create') as c:
+        c.argument('resource_group_name', resource_group_name_type,
+                   help='The name of the resource group.')
```
fengzhou-msft, line 33: Plese remove the help text. `resource_group_name_type` has provided a more detailed help message.

src/portal/azext_portal/azext_metadata.json
```diff
@@ -0,0 +1,4 @@
+{
+    "azext.isPreview": true,
+    "azext.minCliCoreVersion": "2.0.67"
```
fengzhou-msft, line 3: Requested by PM, we release all Breadth Coverage extensions initially with an Experimental tag and use 2.3.1 as the minimal version which added the support for this tag. 
```suggestion
    "azext.isExperimental": true,
    "azext.minCliCoreVersion": "2.3.1"
```

src/portal/azext_portal/generated/commands.py
```diff
@@ -0,0 +1,22 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from azure.cli.core.commands import CliCommandType
+
+
+def load_command_table(self, _):
```
fengzhou-msft, line 9: Add experimental tag for the command group `logic`, you can reference [here](https://github.com/arrownj/azure-cli-extensions/blob/a9c5b01192ba687f6465a7f9a96bc25005f9c339/src/stream-analytics/azext_stream_analytics/commands.py#L16-L17).

src/portal/setup.cfg
```diff
@@ -0,0 +1,2 @@
+[bdist_wheel]
+universal=1
```
fengzhou-msft, line 2: These 2 lines can be removed as we only support python 3 now so the pakcage is no longer universal.

fengzhou-msft:
The [CI](https://github.com/Azure/azure-cli-extensions/pull/1428/checks?check_run_id=607266939) actually passed, but the status is not updated in the PR. It won't block PR merge.

src/portal/azext_portal/generated/_validators.py
```diff
@@ -0,0 +1,18 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+
+def example_name_or_id_validator(cmd, namespace):
```
qwordy, line 7: Maybe you should delete it.

qwordy:
I find it does not support --ids.
https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-the-ids-parameter

src/portal/azext_portal/tests/latest/dashboard.json
```diff
@@ -0,0 +1,61 @@
...
+      }
+    }
+  },
+  "name": "7c0464ec-b1cd-4a98-a4a5-1ebe2d980260",
+  "type": "Microsoft.Portal/dashboards",
+  "location": "eastus",
+  "tags": {
+    "hidden-title": "test dashboard"
+  },
+  "apiVersion": "2019-01-01-preview"
```
qwordy, line 60: I don't think user can choose api version. It's a constant value in your SDK.

YalinLi0312: This is an example used in test

src/portal/azext_portal/generated/_validators.py
YalinLi0312: This is auto-generated, can you explain more about why it should be deleted?

qwordy:
> 
> 
> I find it does not support --ids.
> https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-the-ids-parameter

This comment has not been resolved.

src/portal/azext_portal/tests/latest/dashboard.json
qwordy: Can users choose api version in dashboard.json? If not, please delete this line.

src/portal/azext_portal/generated/commands.py
```diff
@@ -0,0 +1,25 @@
...
+
+    with self.command_group('stream-analytics', is_experimental=True):
```
fengzhou-msft, line 11: command group name is not right.

YalinLi0312: Since this pr already merged, I will address it in another pr. Thanks for your catch!

YalinLi0312:
> > I find it does not support --ids.
> > https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-the-ids-parameter
> 
> This comment has not been resolved.

since this pr is for portal breadth coverage, it may not be included. 
@RodgeFu do you think this support can be involved in CodeGen to avoid customization?

RodgeFu:
> > > I find it does not support --ids.
> > > https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#supporting-the-ids-parameter
> > 
> > 
> > This comment has not been resolved.
> 
> since this pr is for portal breadth coverage, it may not be included.
> @RodgeFu Rodge Fu FTE do you think this support can be involved in CodeGen to avoid customization?

sure, issue https://github.com/Azure/autorest.az/issues/339 created for this. thanks.

PR #1441, opened 2020-03-24
[aks-preview] add universal back

arrownj:
Currently `az extension add` will skip whl which do not contain .py2. So the previous change will cause the extension can't be added.

Changes for `az extension add` are already submit by @fengzhou-msft in https://github.com/Azure/azure-cli/pull/12727. 

Here we add universal back and leave it when the new `az extension add` behavior go online.

---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
add to S167

haroldrandom:
Hi, I remove the Python 2 version limit on wheel name in CI. https://github.com/Azure/azure-cli-extensions/pull/1439. This won't be a problem I think.

karataliu:
Do we need to increase package version for those changes?

The following blob seems already uploaded with the old config.
https://azurecliprod.blob.core.windows.net/cli-extensions/aks_preview-0.4.38-py3-none-any.whl

arrownj:
> Do we need to increase package version for those changes?
> The following blob seems already uploaded with the old config.
> https://azurecliprod.blob.core.windows.net/cli-extensions/aks_preview-0.4.38-py3-none-any.whl

no need, the change in the index.json is not merged.

PR #1451, opened 2020-03-26
Presentation aks fix: The name displayed for the storage is inreality everything before -hcp-, this fix helps clearing that confusion.

Tatsinnit:
Detail this PR trying to achieve: 

3 changes:.

* Fix `Pylint` failures.

*  Right representation of the Storage URL matching the blobname which gets created. More detail could be found here: https://github.com/Azure/aks-periscope/issues/12 and currently when the storage blob gets created its following the convention under following ARM convention : https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/error-storage-account-name Hence when using persicope user navigate to `storage explorer` they will see blobname you will see logs uploaded under the blob container: **example only** `aksaspen2-dns-22701a84` and not `aksaspen2-dns-22701a84-hcp-eastus2-azmk8s-io`

* The `return` is not necessary to be explicit in python, AFAIK. essentially `return` or `return None` or jsut not mentioning is same. (Unless there is something specific reason, hence detailing here)

---

### General Guidelines

For new extensions:

### About Extension Publish


Tatsinnit:
Tag: `AKS` 

Ping reviewer /assign : @xizhamsft + @JunSun17 , Thanks heaps!

yonzhan:
S167

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1500,7 +1500,21 @@ def aks_kollect(cmd,    # pylint: disable=too-many-statements,too-many-locals
     else:
         display_diagnostics_report(temp_kubeconfig_path)
 
-    return
+
+def _trim_fqdn_name_containing_hcp(normalized_fqdn: str) -> str:
```
gtracer: Can we please move to this to a helper file

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1500,7 +1500,21 @@ def aks_kollect(cmd,    # pylint: disable=too-many-statements,too-many-locals
...
+    """
+    Trims the storage blob name and takes everything prior to "-hcp-".
+    Currently it is displayed wrong: i.e. at time of creation cli has
+    following limitation:
+    https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/
```
gtracer: To be specific, you are saying that the problems are:
1. the `-` in the name are causing it to fail 
2. having hcp in every name is not ideal as it leads to less characters that are truly unique in the name 

Is that right?

JunSun17:
LGTM, just make sure you have tested it.

src/aks-preview/azext_aks_preview/custom.py
xizhamsft: @gtracer , the issue here is that the aks-periscope pods will create a blob for MC with name 'xxx-xxx-hcp-xxxx' like this "xxx-xxx". Basically removing everything after "-hcp" (including -hcp). But our CLI code still uses the full MC name in the URL. Hence the mismatch.

yungezz:
hi @Tatsinnit could you pls add tests to it? thanks.

src/aks-preview/azext_aks_preview/custom.py
Tatsinnit: @gtracer Cool, sounds good, i have moved this to helper:

* Moving fwd plan is to refactor this file. Essentially, I have a plan to do another PR with some refactor for `custom.py` move rest of the helper `functs` out.

Tatsinnit:
:bulb: Test et. al. added, like mentioned I would do another drop next for the generic things I see can be corrected for `aks-preview`. Thanks + ping: cc: @yungezz + @gtracer 

* Also `pylint` was broken which is fixed 

Below was the lint failure.

FYI: 

image

PR #1458, opened 2020-03-30
spring-cloud: add location to Apps operation

hellokangning:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
add to S168

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -172,9 +175,12 @@ def app_update(cmd, client, resource_group, service, name,
     if enable_persistent_storage is False:
         properties.persistent_disk = models.PersistentDisk(size_in_gb=0)
 
+    resource = client.services.get(resource_group, service)
+    location = resource.location
+
     logger.warning("[1/2] updating app '{}'".format(name))
     poller = client.apps.update(
-        resource_group, service, name, properties)
+        resource_group, service, name, properties, location)
```
yungezz: why location required in each rest call. suppose it should not in update parameter in swagger

hellokangning: We are migrating Microsoft/Spring/Apps from proxy resource to tracked resource. And `location` is a must for tracked resource. I've also updated the manifest change in https://github.com/Azure/azure-rest-api-specs/pull/8851

fengzhou-msft: If this location is always the same as the location of the resource_group, why not fetch it from the resource group on the service side? Or location could be different in the future?

hellokangning: Not the same with the location of resource_group, but with the one of `Microsoft.AppPlatform/Spring`. We cannot do it on service side, since it's ARM consuming this property, before our RP.

fengzhou-msft: I see. Thanks for the explanation.

PR #1603, opened 2020-04-26
Support Databricks CMK/BYOK

fengzhou-msft:
---

### General Guidelines

For new extensions:

### About Extension Publish


fengzhou-msft:
Wait for service ready to add tests.

yonzhan:
Databricks

src/databricks/azext_databricks/_help.py
```diff
@@ -34,6 +34,16 @@
       - name: Clean the workspace's tags.
         text: |-
                az databricks workspace update --resource-group MyResourceGroup --name MyWorkspace --tags ""
+      - name: Assign identity for storage account.
+        text: |-
+               az databricks workspace update --resource-group MyResourceGroup --name MyWorkspace --assign-identity
```
yungezz, line 39: does it support configuring identity type?

yungezz: just confirm, spec doesn't mention identity be configured at creation time

src/databricks/azext_databricks/_params.py
```diff
@@ -30,6 +31,11 @@ def load_arguments(self, _):
     with self.argument_context('databricks workspace update') as c:
         c.argument('workspace_name', options_list=['--name', '-n'], id_part='name', help='The name of the workspace.')
         c.argument('tags', tags_type)
+        c.extra('assign_identity', action='store_true', help='Enable the Managed Identity for managed storage account.')
+        c.argument('encryption_key_source', options_list=['--key-source'], arg_group='Encryption', arg_type=get_enum_type(['Default', 'Microsoft.Keyvault']), validator=validate_encryption_values, help='The encryption key source (provider).')
```
yungezz, line 35: worth to sync with PM to see is it necessary to define consistence on parms name now

src/databricks/azext_databricks/_validators.py
```diff
@@ -37,3 +38,14 @@ def validate_workspace_values(cmd, namespace):
...
+
+
+def validate_encryption_values(namespace):
+    from knack.util import CLIError
+    if namespace.encryption_key_source:
+        if namespace.encryption_key_source == 'Default' and any(v is not None for v in [namespace.encryption_key_name, namespace.encryption_key_version, namespace.encryption_key_vault]):
+            raise CLIError('--key-name, --key-version, --key-vault should not be provided when --key-source is Default')
+        if namespace.encryption_key_source == 'Microsoft.Keyvault' and any(v is None for v in [namespace.encryption_key_name, namespace.encryption_key_version, namespace.encryption_key_vault]):
+            raise CLIError('--key-name, --key-version, --key-vault are required when --key-source is Microsoft.Keyvault')
+    elif any(v is not None for v in [namespace.encryption_key_name, namespace.encryption_key_version, namespace.encryption_key_vault]):
```
yungezz, line 50: if key_source is None, why still need check key_name, version etc?

src/databricks/azext_databricks/azext_metadata.json
```diff
@@ -1,4 +1,4 @@
 {
-    "azext.isExprimental": true,
+    "azext.isExperimental": true,
```
yungezz, line 2: command test will not fail?

src/databricks/azext_databricks/_help.py
fengzhou-msft: identity type cannot be configured, it's managed by databricks service. It can only be enabled in update command in the spec.

src/databricks/azext_databricks/azext_metadata.json
fengzhou-msft: This typo only affects the message shown when installing the extensions. Current tests will not catch it, but there is a plan to add metadata validation to avoid such issue.

src/databricks/azext_databricks/_validators.py
fengzhou-msft: To consider the case when key_source is forgotten but other key options are provided.

bim-msft: Hi @fengzhou-msft , same concern with Catherine here: what if users don't specify any of these 4 parameters? It seems this validator would be passed.

fengzhou-msft: It is a valid case, the key properties are not updated. They are not required in this update command.

src/databricks/azext_databricks/_params.py
```diff
@@ -26,10 +27,16 @@ def load_arguments(self, _):
...
 
     with self.argument_context('databricks workspace update') as c:
         c.argument('workspace_name', options_list=['--name', '-n'], id_part='name', help='The name of the workspace.')
         c.argument('tags', tags_type)
+        c.argument('prepare_encryption', action='store_true', help='Flag to enable the Managed Identity for managed storage account to prepare for CMK encryption.')
+        c.argument('encryption_key_source', options_list=['--key-source'], arg_group='Encryption', arg_type=get_enum_type(['Default', 'Microsoft.Keyvault']), validator=validate_encryption_values, help='The encryption key source (provider).')
+        c.argument('encryption_key_name', options_list=['--key-name'], arg_group='Encryption', help='The name of KeyVault key.')
+        c.argument('encryption_key_version', options_list=['--key-version'], arg_group='Encryption', help='The version of KeyVault key.')
+        c.argument('encryption_key_vault', options_list=['--key-vault'], arg_group='Encryption', help='The Uri of KeyVault.')
```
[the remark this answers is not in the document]
fengzhou-msft: You mean add an option, not support uri or name with one option?

The design basically followed the interface of [storage](https://github.com/Azure/azure-cli/blob/06f6ec5e07bf3db0881475cb6204e6bb82c04696/src/azure-cli/azure/cli/command_modules/storage/_params.py#L271-L278) keys.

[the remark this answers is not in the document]
bim-msft: @fengzhou-msft Thanks for the clarification, I think it's better if users are able to just specify the vault name instead of the whole URI. Anyway, aligning with the previous design is not bad.

PR #1606, opened 2020-04-26
[AKS] Unblock agentpool min count range check

marwanad:
---

### General Guidelines

For new extensions:

### About Extension Publish

Unblock updating minCount in both `aks update` and `aks nodepool update` when minCount < actual node count < max Count.

With this, `minCount` can be greater than the actual node count. Later scale operations equal to or greater than minCount, will respect that minCount in operations.

### DIFF
--- a/src/aks-preview/azext_aks_preview/custom.py
+++ b/src/aks-preview/azext_aks_preview/custom.py
@@ -1140,14 +1140,12 @@ def aks_update(cmd,     # pylint: disable=too-many-statements,too-many-branches,
 
     if min_count is None or max_count is None:
         if enable_cluster_autoscaler or update_cluster_autoscaler:
-            raise CLIError('Please specifying both min-count and max-count when --enable-cluster-autoscaler or '
+            raise CLIError('Please specify both min-count and max-count when --enable-cluster-autoscaler or '
                            '--update-cluster-autoscaler set.')
 
     if min_count is not None and max_count is not None:
         if int(min_count) > int(max_count):
             raise CLIError('value of min-count should be less than or equal to value of max-count.')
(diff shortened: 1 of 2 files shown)


yonzhan:
AKS

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -2183,17 +2179,14 @@ def aks_agentpool_update(cmd,   # pylint: disable=unused-argument
...
 
     if min_count is None or max_count is None:
         if enable_cluster_autoscaler or update_cluster_autoscaler:
-            raise CLIError('Please specifying both min-count and max-count when --enable-cluster-autoscaler or '
+            raise CLIError('Please specify both min-count and max-count when --enable-cluster-autoscaler or '
                            '--update-cluster-autoscaler set.')
     if min_count is not None and max_count is not None:
         if int(min_count) > int(max_count):
             raise CLIError('value of min-count should be less than or equal to value of max-count.')
-        if int(node_count) < int(min_count) or int(node_count) > int(max_count):
```
yungezz, line 2195: curious why remove this validation?

marwanad: I'll given an example:

You start with 1 nodes in a cluster, min count: 1 and max count: 50. You later decide that the desired min count is 2 nodes in anticipation of some increase in workload where you'd like to keep 1 extra node available. We want to unblock this operation as changing the min/max count should be independent of the actual node count.

Same idea for max count: let's say at a point in time, you have 60 nodes in the cluster (minCount: 1, maxCount: 70) but you later decide that the desirable maximum count is 50 nodes (because you're anticipating a large scale-down for example that would leave many nodes eligible for scaling down that would otherwise not be scaled-down because of that maxCount restriction).

PR #1634, opened 2020-05-02
update help context of "--uptime-sla"

zqingqing1:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/azext_aks_preview/_help.py
+++ b/src/aks-preview/azext_aks_preview/_help.py
@@ -199,7 +199,7 @@ helps['aks create'] = """
           short-summary: The node resource group is the resource group where all customer's resources will be created in, such as virtual machines.
         - name: --uptime-sla
           type: bool
-          short-summary: Enable paid managed cluster service with high availability.
+          short-summary: Enable a paid managed cluster service with a financially backed SLA.
         - name: --attach-acr
           type: string
           short-summary: Grant the 'acrpull' role assignment to the ACR specified by name or resource ID.


yonzhan:
add to S169

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -199,7 +199,7 @@
           short-summary: The node resource group is the resource group where all customer's resources will be created in, such as virtual machines.
         - name: --uptime-sla
           type: bool
-          short-summary: Enable paid managed cluster service with high availability.
+          short-summary: Enable a paid managed cluster service with a financially backed SLA.
```
gtracer, line 202: I wonder if it is worth adding a link here. Because there are a lot of details about what the SLA includes (HA components etc.)

gtracer: I would consider adding a word like "premium offering" or "high availability" etc. But will leave it to your discretion

gtracer: actually I think the best way to add more information is to also add a long-summary with more details and a link 

long-summary: xyz

zqingqing1: Per PM, He suggests to have summary like this. Since the service is not yet announced, with current summary should be fine.

gtracer: @jluk thoughts on adding a long summary?

jluk: I'm not sure a long summary is necessary, you could add an aka link (aka.ms/aks/paid-tier) or something like that which explains the details.

The desire to avoid the term "HA" implies you cannot get HA without turning this feature on. We need to describe the merits of the paid SLA without detracting from the free tier.

jluk: IMO we should defer and keep the changes Qingqing has added, if we get feedback the cli help text was their only route to get more info and it's insufficient we can revisit.

yungezz: I suggest add a link. help text in cli will be rendered as help doc at docs.ms https://docs.microsoft.com/en-us/cli/azure/ext/aks-preview/aks?view=azure-cli-latest#ext-aks-preview-az-aks-create-required-parameters

zqingqing1:
@arrownj @qwordy  Can anyone of you merge this PR? Thanks :)

Issue #1677, opened 2020-05-09
az monitor log-analytics solution create  - remove Remove --plan-publisher and --plan-product


dougbrad:
- If the issue is to do with Azure CLI 2.0 in-particular, create an issue here at [Azure/azure-cli](https://github.com/Azure/azure-cli/issues)

### Extension name (the extension in question)

az monitor log-analytics solution create 

### Description of issue (in as much detail as possible)

Remove --plan-publisher and --plan-product
Remove any mention to Microsoft or 3rd party solutions - this is for Microsoft solutions only.

-----

yonzhan:
add to S170

zhoxing-ms:
@dougbrad Thanks for your feedback, I will annotate `--plan-publisher`, `--name` and `--plan-product` with `deprecate_info`, add a new parameter `--solution-type` and remove any mention to Microsoft or 3rd party solutions.
Since the solution does not modify the logic of any old parameters, they are only declared to be deleted temporarily and will be deleted when the customers no longer uses them, so there will be no breaking change.

PR #1680, opened 2020-05-11
[Log Analytics Solution] deprecate `--plan-publisher`, `--name`, `--plan-product` and add `--solution-type` for create

zhoxing-ms:
Issue: #1677

Because log analytics solution does not support third party solutions,  so annotate `--plan-publisher`, `--name` and` --plan-product` with `deprecate_info`, add a new parameter `--solution-type` and remove any mention to Microsoft or 3rd party solutions.
Since the solution does not modify the logic of any old parameters, they are only declared to be deleted temporarily and will be deleted when the customers no longer uses them, so there will be no breaking change.

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/log-analytics-solution/HISTORY.rst
+++ b/src/log-analytics-solution/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+0.1.1
++++++++++++++++
+* deprecate --plan-publisher, --name, --plan-product and add --solution-type for create
+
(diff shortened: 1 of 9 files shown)


yonzhan:
add to S169

zhoxing-ms:
> You may need to update the version in setup.py to trigger the release pipeline.

Done

PR #1798, opened 2020-06-01
vm-repair: Add support for encrypted VMs

swbae31:
Current PR will enable the repair VM creation of encrypted VMs:
- Linux and Windows managed disk VMs which are encrypted using the single pass method.

Main changes are to the command

**az vm repair create**
with the parameter **--unlock-encrypted-vm** added to skip the confirmation prompt when the VM is encrypted.

Credits to @dibaskar for AZ disk encryption/decryption expertise and main implementation of this PR.


yonzhan:
vm-repair

src/vm-repair/azext_vm_repair/encryption_types.py
```diff
@@ -0,0 +1,13 @@
...
+# --------------------------------------------------------------------------------------------
+
+import enum
+
+
+class Encryption(enum.Enum):
+    NONE = 1                  # Not an encrypted VM.
+    SINGLE_WITH_KEK = 2       # Its an encryped VM using single pass method with kek option.
+    SINGLE_WITHOUT_KEK = 3    # Its an encrypted VM using single pass method without kek option.
+    DUAL = 4                  # Its an encrypted VM using dual pass method.
```
qwordy, line 13: I am curious about what is dual pass.

dibaskar: Dual pass encryption happens using keyvault and AAD crdentials.
Single pass encryption happens using keyvault but without AAD credetntails.

Single pass encryption is the latest one, which have additional features like encryption option from portal and disk swap option etc..... 

where as in Dual pass we need to perform encryption only using powershell or cli commands and we need to delete and recreate VM while troubleshooting dual pass encrypted VMs , as swap disk option is not available for them.

PR #1821, opened 2020-06-04
[Breadth Coverage] Attestation (mgmt operations)

YalinLi0312:
[Speclet doc](https://microsoft.sharepoint.com/:w:/r/teams/AzureManagementExperience/_layouts/15/doc2.aspx?sourcedoc=%7B00203144-2E2A-4EA0-B13B-644ABE3FF9E5%7D&file=ATTESTATION%20RP%20Breadth%20Speclet.docx&action=default&mobileredirect=true&cid=f3f0dcfe-5ca8-40f6-8f4d-66957b7af657).

Attestation has cmdlets from both resource manager and data plane. To simplify the review work, this PR only contains mgmt operations in attestation breadth coverage. I will create another PR for data plane operations.

Here are files I did customizations, you may need focus on when doing review:
* src/attestation/azext_attestation/generated/_help.py
* src/attestation/azext_attestation/generated/_params.py
* src/attestation/azext_attestation/generated/commands.py
* src/attestation/azext_attestation/generated/custom.py
* src/attestation/azext_attestation/tests/test_attestation_scenario.py

---

### General Guidelines

For new extensions:

### About Extension Publish


src/attestation/azext_attestation/generated/_help.py
```diff
@@ -0,0 +1,98 @@
...
+        text: |-
+               az attestation show --name "myattestationprovider" --resource-group "MyResourceGroup"
+"""
+
+helps['attestation create'] = """
+    type: command
+    short-summary: Creates or updates an attestation.
+    parameters:
+      - name: --policy-signing-certificates-keys
+        short-summary: The value of the "keys" parameter is an array of JWK values.  Bydefault, the order of the JWK va\
```
bquantump, line 46: Does this show up on the command line with attestation create -h?
If so, this very verbose for command line display. Maybe this should move to the help file.

src/attestation/azext_attestation/tests/latest/test_attestation_scenario.py
```diff
@@ -0,0 +1,116 @@
...
+             '--resource-group "{rg}" '
+             '--yes',
+             checks=[])
+    test.cmd('az attestation list '
+             '--resource-group "{rg}"',
+             checks=[test.check('length(value)', 0)])
+
+
+@try_manual
+def cleanup(test, rg):
```
bquantump, line 90: This does nothing. Is this here because some of the decorates need to call it later?

src/attestation/azext_attestation/generated/_params.py
```diff
@@ -0,0 +1,48 @@
...
+        c.argument('resource_group_name', resource_group_name_type)
+        c.argument('provider_name', options_list=['--name', '-n'], help='Name of the attestation service instance',
+                   id_part='name')
+
+    with self.argument_context('attestation create') as c:
+        c.argument('resource_group_name', resource_group_name_type)
+        c.argument('provider_name', options_list=['--name', '-n'], help='Name of the attestation service')
+        c.argument('location', arg_type=get_location_type(self.cli_ctx),
+                   validator=get_default_location_from_resource_group,
+                   help='List of available regions for the resource type is \'eastus2,centralus,uksouth\'')
```
bquantump, line 37: If the service teams decides to add new locations, this will have to be updated every time they do so. Has the service team indicated that these are fixed and they won't be updated?

src/attestation/azext_attestation/generated/_params.py
```diff
@@ -0,0 +1,48 @@
...
+        c.argument('tags', tags_type)
+        c.argument('attestation_policy', help='Name of attestation policy.')
```
bquantump, line 39: I don't see a description of this policy in the README.md and this is one vague. Can you add something more descriptive here on in the README.md or put a link in the README.md to a description?

src/attestation/azext_attestation/generated/_params.py
```diff
@@ -0,0 +1,48 @@
...
+        c.argument('policy_signing_certificates_keys', action=AddPolicySigningCertificatesKeys, nargs='+', help='The va'
```
bquantump, line 40: Do we want this to be a JSON or file to be read in.? Given the help.py, there is a lot of options here and customers my find it easier to have JSON files for the different crypto algorithms and settings for them they want to use.

yonzhan:
Attestation

src/attestation/azext_attestation/tests/latest/test_attestation_scenario.py
haroldrandom: No need. We recommend write test like the old way like CLI.

You can comment it or delete it.

YalinLi0312: > No need. We recommend write test like the old way like CLI.
> 
> You could comment it or delete it.

What's the old way you mean? Can you give an example?
If no need to have this function, I will delete it.

src/attestation/azext_attestation/generated/_params.py
YalinLi0312: It's not fixed, they will add more in the future. 
To avoid keep tracking and updating here, I will remove the help section, cli response will show which regions are currently supported when customer choose an invalid region.

src/attestation/azext_attestation/tests/latest/test_attestation_scenario.py
haroldrandom: Like what you did in https://github.com/Azure/azure-cli-extensions/blob/master/src/portal/azext_portal/tests/latest/test_portal_scenario.py

src/attestation/azext_attestation/generated/_params.py
YalinLi0312: This arg is the name of an attestation policy. There is another cmdlet `az attestation policy`, which is in data plane, I think the description for policy should be here.

src/attestation/azext_attestation/tests/latest/test_attestation_scenario.py
YalinLi0312: @RodgeFu Do you know if there has any specific reason that changing the tests from the old way?

src/attestation/azext_attestation/generated/_params.py
YalinLi0312: Good suggestion, JSON would be easier for a complex input. I will update in next commit.

src/attestation/azext_attestation/tests/latest/test_attestation_scenario.py
RodgeFu: The setup/cleanup method are there to support easier test customization. By default, they does nothing, but if needed, you can easy override them to provide extra setup/cleanup logic for test.

YalinLi0312:
FYI. This PR cannot be merged, az attestation create will fail with --policy-signing-certificates-keys

YalinLi0312:
Updates:
1. To align with PSH cmdlet, the param --policy-signing-certificates-keys changed to --certs-input-file. 
2. Confirmed with the service team, update the required property in JSONWebKey.

PTAL

PR #1865, opened 2020-06-09
[AKS] add uptime-sla for updating

zqingqing1:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/azext_aks_preview/_help.py
+++ b/src/aks-preview/azext_aks_preview/_help.py
@@ -311,6 +311,9 @@ helps['aks update'] = """
         - name: --max-count
           type: int
           short-summary: Maximum nodes count used for autoscaler, when "--enable-cluster-autoscaler" specified. Please specify the value in the range of [1, 100]
+         - name: --uptime-sla
+          type: bool
+          short-summary: Enable a paid managed cluster service with a financially backed SLA.
         - name: --cluster-autoscaler-profile
           type: list
           short-summary: Space-separated list of key=value pairs for configuring cluster autoscaler. Pass an empty string to clear the profile.
--- a/src/aks-preview/azext_aks_preview/_help.py
+++ b/src/aks-preview/azext_aks_preview/_help.py
@@ -311,7 +311,7 @@ helps['aks update'] = """
         - name: --max-count
           type: int
           short-summary: Maximum nodes count used for autoscaler, when "--enable-cluster-autoscaler" specified. Please specify the value in the range of [1, 100]
-         - name: --uptime-sla
+        - name: --uptime-sla
           type: bool
           short-summary: Enable a paid managed cluster service with a financially backed SLA.
         - name: --cluster-autoscaler-profile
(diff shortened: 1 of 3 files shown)


yonzhan:
AKS

zqingqing1:
@arrownj  we need to. From my previous experience. extension overrides cli change, if extension doesn't have the "--uptime-sla" in "az aks update", the overrode one won't have even cli has.

palma21:
Any blockers to get this merged? We have customer reports of the feature not working

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -311,6 +311,9 @@
         - name: --max-count
           type: int
           short-summary: Maximum nodes count used for autoscaler, when "--enable-cluster-autoscaler" specified. Please specify the value in the range of [1, 100]
+        - name: --uptime-sla
```
gtracer, line 314: I wonder what naming other services use that have a paid and free version. Wonder if we can be consistent in azure

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1240,6 +1243,12 @@ def aks_update(cmd,     # pylint: disable=too-many-statements,too-many-branches,
     if attach_acr and detach_acr:
         raise CLIError('Cannot specify "--attach-acr" and "--detach-acr" at the same time.')
 
+    if uptime_sla:
```
gtracer, line 1246: Not blocking but my preference is you create a new helper func (in a different file) that handles all SLA specific code. It is not blocking because this is only a few lines of code but I am guessing this code will grow and having it separated is preferable

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -109,6 +109,7 @@ def load_arguments(self, _):
         c.argument('cluster_autoscaler_profile', nargs='+', validator=validate_cluster_autoscaler_profile)
         c.argument('min_count', type=int, validator=validate_nodes_count)
         c.argument('max_count', type=int, validator=validate_nodes_count)
+        c.argument('uptime_sla', action='store_true')
```
gtracer, line 112: Question (non-blocking): What is this? Never seen an action before

src/aks-preview/azext_aks_preview/custom.py
gtracer: Actually, I see it is in both create and update - would really prefer if it it was in a short function that sets and returns the MC

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1131,6 +1131,7 @@ def aks_update(cmd,     # pylint: disable=too-many-statements,too-many-branches,
                disable_pod_security_policy=False,
                attach_acr=None,
                detach_acr=None,
+               uptime_sla=False,
```
gtracer, line 1134: Thanks for settings as False

xiazhan:
> Any blockers to get this merged? We have customer reports of the feature not working

+1 We need this extension to unblock customers soon. @arrownj  Can we get it merged?

src/aks-preview/azext_aks_preview/_help.py
zqingqing1: The naming is requested by PM, and agreed with several people who already has discussion about the naming. And the naming has been used for a while. It is not a good idea to change it.

src/aks-preview/azext_aks_preview/_params.py
zqingqing1: if the value used, then set it to True

src/aks-preview/azext_aks_preview/custom.py
zqingqing1: Just 3 lines of code, directly setting is not bad as well.

Issue #1584, opened 2020-04-23
az image copy dangerous cleanup with temporary_resource_group_name


suppix:
- If the issue is to do with Azure CLI 2.0 in-particular, create an issue here at [Azure/azure-cli](https://github.com/Azure/azure-cli/issues)

### Extension name (the extension in question)
az image copy

### Description of issue (in as much detail as possible)

With the option cleanup you simply delete the whole temporary resource group, but with the option 
--temporary_resource_group_name it can be dangerous.  I think it's better to delete only what you really create. With this option cleanup and temporary_resource_group_name you can delete my existing resource group which can contain important data.

yonzhan:
@qwordy  please take a look

qwordy:
If you omit `--temporary-resource-group-name`, a temporary resource group image-copy-rg will be used. I suggest omit it or use a different temporary resource group from source resource group.

suppix:
az image copy --help

PR #1880, opened 2020-06-12
[Imagecopy] Validation on resource group; Improve logging

qwordy:
1.
Resolve https://github.com/Azure/azure-cli-extensions/issues/1584
`--cleanup` : Include this switch to delete temporary resources upon completion.
`--temporary-resource-group-name`    : Resource Group name where temporary storage account will be created.  Default: image-copy-rg.
If `--cleanup` is set and `--temporary-resource-group-name` is set an existing resource group, this resource group will be deleted, even the resources created by the user before.

Now it is forbidden to use existing resource group name is `--cleanup` is set.

2.
Add more logs. Print running commands.

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/image-copy/azext_imagecopy/__init__.py
+++ b/src/image-copy/azext_imagecopy/__init__.py
@@ -47,7 +47,7 @@ class ImageCopyCommandsLoader(AzCommandsLoader):
             c.argument('timeout', options_list=['--timeout'], type=int, default=3600,
                        help='Time in seconds for the copy operation to finish. Increase this time if '
                        'you are going to copy large images (disks) like 512GB or more.')
-            c.argument('temporary_resource_group_name', options_list=['--temporary_resource_group_name'],
+            c.argument('temporary_resource_group_name', options_list=['--temporary-resource-group-name'],
                        default='image-copy-rg',
                        help='Resource Group name where temporary storage account will be created.')
             c.argument('export_as_snapshot', options_list=['--export-as-snapshot'], action='store_true', default=False,
--- a/src/image-copy/azext_imagecopy/cli_utils.py
+++ b/src/image-copy/azext_imagecopy/cli_utils.py
@@ -91,6 +91,6 @@ def get_storage_account_id_from_blob_path(cmd, blob_path, resource_group, subscr
 
 
 def format_cmd(cmd: List[str]):
-    if len(cmd)
-    terms = cmd[3:]
-    return 'Running: az ' + ' '.join(cmd[3:])
+    if len(cmd) > 3:
+        return 'Running: az ' + ' '.join(cmd[3:])
+    return 'Running: ' + ' '.join(cmd)
--- a/src/image-copy/azext_imagecopy/cli_utils.py
+++ b/src/image-copy/azext_imagecopy/cli_utils.py
@@ -7,6 +7,8 @@ import sys
 import json
 
 from subprocess import check_output, STDOUT, CalledProcessError
+from typing import List
+
 from knack.util import CLIError
 
 from knack.log import get_logger
@@ -86,3 +88,9 @@ def get_storage_account_id_from_blob_path(cmd, blob_path, resource_group, subscr
         namespace='Microsoft.Storage', type='storageAccounts', name=storage_account_name)
 
     return storage_account_id
+
+
+def format_cmd(cmd: List[str]):
+    if len(cmd)
+    terms = cmd[3:]
+    return 'Running: az ' + ' '.join(cmd[3:])
--- a/src/image-copy/azext_imagecopy/create_target.py
+++ b/src/image-copy/azext_imagecopy/create_target.py
@@ -9,7 +9,9 @@ import time
 from knack.util import CLIError
 from knack.log import get_logger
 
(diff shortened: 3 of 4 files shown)


yonzhan:
Image copy

src/image-copy/azext_imagecopy/__init__.py
```diff
@@ -47,7 +47,7 @@ def load_arguments(self, _):
             c.argument('timeout', options_list=['--timeout'], type=int, default=3600,
                        help='Time in seconds for the copy operation to finish. Increase this time if '
                        'you are going to copy large images (disks) like 512GB or more.')
-            c.argument('temporary_resource_group_name', options_list=['--temporary_resource_group_name'],
+            c.argument('temporary_resource_group_name', options_list=['--temporary-resource-group-name'],
```
arrownj, line 50: break change ?

qwordy: The old name is wrong. It is a breaking change, but it should be updated.

src/image-copy/azext_imagecopy/cli_utils.py
```diff
@@ -86,3 +88,9 @@ def get_storage_account_id_from_blob_path(cmd, blob_path, resource_group, subscr
         namespace='Microsoft.Storage', type='storageAccounts', name=storage_account_name)
 
     return storage_account_id
+
+
+def format_cmd(cmd: List[str]):
+    if len(cmd) > 3:
+        return 'Running: az ' + ' '.join(cmd[3:])
```
jiasli, line 95: Is this safe on all platforms with different entry scripts? If so, add some comments, otherwise using the magic number `3` in `cmd[3:]` may cause trouble.

src/image-copy/azext_imagecopy/cli_utils.py
```diff
@@ -86,3 +88,9 @@ def get_storage_account_id_from_blob_path(cmd, blob_path, resource_group, subscr
...
+    return 'Running: ' + ' '.join(cmd)
```
jiasli, line 96: This may be confusing when `cmd`'s items contain spaces or quotes. Say `hello there` will become `az hello there`, instead of `az "hello there"`. `"hello there"` will become `az "hello there"`, instead of `az "\"hello there\""`.

src/image-copy/azext_imagecopy/create_target.py
```diff
@@ -37,6 +39,7 @@ def create_target_image(cmd, location, transient_resource_group_name, source_typ
                                    '--sku', 'Standard_LRS'],
                                   subscription=target_subscription)
 
+    logger.warning(format_cmd(cli_cmd))
```
tamirkamara, line 42: Looks like you're printing every command. What about doing it just 1 time inside of the run_cli_command function?
BTW, that function is already printing it out, but as a "debug" message... why do we need it twice?

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -19,14 +21,24 @@
 # pylint: disable=too-many-branches
 def imagecopy(cmd, source_resource_group_name, source_object_name, target_location,
               target_resource_group_name, temporary_resource_group_name='image-copy-rg',
-              source_type='image', cleanup='false', parallel_degree=-1, tags=None, target_name=None,
+              source_type='image', cleanup=None, parallel_degree=-1, tags=None, target_name=None,
               target_subscription=None, export_as_snapshot='false', timeout=3600):
+    # If --cleanup is set, forbid using an existing temporary resource group name.
+    # It is dangerous to clean up an existing resource group.
+    if cleanup:
```
tamirkamara, line 28: There's a person that also had an issue when using the same source & target resource group name with cleanup=true. That essentially deleted the source group which included a bunch of other VMs.
Another check I had in mind is not allowing same source/target (on the same subscription) when cleaning up. But that might be for another PR...

src/image-copy/azext_imagecopy/cli_utils.py
qwordy: You are right. I haven't tested in other platforms except Powershell.

src/image-copy/azext_imagecopy/cli_utils.py
qwordy: Yes. The returned string may not recover the intent of cmd variable.

qwordy:
I will split this PR into two PRs.

src/image-copy/azext_imagecopy/create_target.py
qwordy: Almost. But I skip `wait_for_blob_copy_operation` because `storage blob show` is called repeatedly.
The reason I make the change is that I want to print CLI command that can be executed directly.

src/image-copy/azext_imagecopy/custom.py
qwordy: Do you mean temporary resource group and target resource group?

tamirkamara: sorry, I confused it a bit. 
I meant the source and the temporary one. There was an incident where a user gave the same same name for the temp one as the source and the process deleted many production resources.

qwordy: Source resource group is an existing resource group, so this problem can be covered by this piece of code.

PR #1885, opened 2020-06-15
[Imagecopy] Fix a bug of tags

qwordy:
This bug is reported from Boeing on IcM. When a user tries to specify tags, it will fail.
It is a dict instead of a list.
This PR fixed this bug.

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/image-copy/azext_imagecopy/cli_utils.py
+++ b/src/image-copy/azext_imagecopy/cli_utils.py
@@ -66,7 +66,10 @@ def prepare_cli_command(cmd, output_as_json=True, tags=None, subscription=None):
 
         if tags is not None:
             for k, v in tags.items():
-                full_cmd.append(k + '=' + v)
+                if v == '':
+                    full_cmd.append(k)
+                else:
+                    full_cmd.append(k + '=' + v)
 
     return full_cmd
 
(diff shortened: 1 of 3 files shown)


src/image-copy/azext_imagecopy/cli_utils.py
```diff
@@ -65,7 +65,8 @@ def prepare_cli_command(cmd, output_as_json=True, tags=None, subscription=None):
         full_cmd += ['--tags', EXTENSION_TAG_STRING]
 
         if tags is not None:
-            full_cmd += tags.split()
+            for k, v in tags.items():
```
yungezz, line 68: tags could be key only

yungezz: don't we have common tags utilities in azure-cli core repo. how about leverage it?

yonzhan:
Imagecopy

src/image-copy/azext_imagecopy/cli_utils.py
qwordy: The purpose of this piece of code is to transform parsed tags to raw parameter values.

PR #1894, opened 2020-06-17
Refine the tracing and tags experiences

ShichaoQiu:
* Enable distributed tracing by default when creating the service
* Enable to update tags and distributed tracing settings by using "az spring-cloud update"


yonzhan:
spring cloud

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -39,30 +41,75 @@
...
+
+    app_insights_target_status = False
+    if app_insights is not None or app_insights_key is not None or disable_distributed_tracing is False:
+        app_insights_target_status = True
+        if resource_properties.trace.enabled is False:
+            update_app_insights = True
+    elif disable_distributed_tracing is True:
+        app_insights_target_status = False
+        if resource_properties.trace.enabled is True:
+            update_app_insights = True
```
fengzhou-msft, line 88: Based on the logic, if app_insights is not None, disable_distributed_tracing is True, it will be treated the same as disable_distributed_tracing is False.

I suggest to use `action='store_true'` for disable_distributed_tracing. When you want to enable distributed tracing, you just provide app_insights or app_insights_key. If you provide disable_distributed_tracing, it means true and can have higher priority or throw error when both disable_distributed_tracing and (app_insights or app_insights_key) are provided.

src/spring-cloud/azext_spring_cloud/_params.py
```diff
@@ -37,6 +37,17 @@ def load_arguments(self, _):
...
 
+    for scope in ['spring-cloud create', 'spring-cloud update']:
+        with self.argument_context(scope) as c:
+            c.argument('app_insights_key',
+                       help="Instrumentation key of the existing Application Insights to be added for the distributed tracing")
+            c.argument('app_insights',
+                       help="Name of the existing Application Insights in the same Resource Group. Or Resource ID of the existing Application Insights in a different Resource Group.")
+            c.argument('disable_distributed_tracing',
+                       arg_type=get_three_state_flag(),
+                       help="Disable distributed tracing, if not disabled and no existing Application Insights specified with --app-insights-key or --app-insights, will create a new Application Insights instance in the same resource group.")
```
fengzhou-msft, line 48: Another reason to use `action='store_true'` is that this is not a property of resource, but a switch to control the usage of application insights. See the last example of [Registering Flags](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#registering-flags).

src/spring-cloud/azext_spring_cloud/custom.py
ShichaoQiu: Have fixed, thanks a lot

src/spring-cloud/azext_spring_cloud/_params.py
ShichaoQiu: Have fixed, thanks a lot

PR #1897, opened 2020-06-17
 [Imagecopy] Validation on temporary resource group 

qwordy:
Resolve https://github.com/Azure/azure-cli-extensions/issues/1584
`--cleanup` : Include this switch to delete temporary resources upon completion.
`--temporary-resource-group-name`    : Resource Group name where temporary storage account will be created.  Default: image-copy-rg.
If `--cleanup` is set and `--temporary-resource-group-name` is set an existing resource group, this resource group will be deleted, even the resources created by the user before.

Now it is forbidden to use existing resource group name is `--cleanup` is set.

---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
Imagecopy

src/image-copy/azext_imagecopy/__init__.py
```diff
@@ -47,7 +47,7 @@ def load_arguments(self, _):
             c.argument('timeout', options_list=['--timeout'], type=int, default=3600,
                        help='Time in seconds for the copy operation to finish. Increase this time if '
                        'you are going to copy large images (disks) like 512GB or more.')
-            c.argument('temporary_resource_group_name', options_list=['--temporary_resource_group_name'],
+            c.argument('temporary_resource_group_name', options_list=['--temporary-resource-group-name'],
```
arrownj, line 50: This should be a break change, right ?

src/image-copy/azext_imagecopy/custom.py
```diff
@@ -21,9 +21,17 @@ def imagecopy(cmd, source_resource_group_name, source_object_name, target_locati
               target_resource_group_name, temporary_resource_group_name='image-copy-rg',
               source_type='image', cleanup='false', parallel_degree=-1, tags=None, target_name=None,
               target_subscription=None, export_as_snapshot='false', timeout=3600):
+    # If --cleanup is set, forbid using an existing temporary resource group name.
+    # It is dangerous to clean up an existing resource group.
+    cli_cmd = prepare_cli_command(['group', 'exists', '-n', temporary_resource_group_name],
```
arrownj, line 26: Is there any existing function to do this check ? I'm not sure whether it is a good idea to run another cli command in current one.

src/image-copy/azext_imagecopy/__init__.py
qwordy: Yes. The old name is wrong. We should use "-" instead of "_".

src/image-copy/azext_imagecopy/custom.py
qwordy: We can also use SDK. This extension runs cli commands instead of using SDK. I follow this style.

arrownj: okay... Personally, I do not suggest to follow this bad pattern.🤔

tamirkamara: @arrownj, 
This pattern was the **recommendation of the Azure CLI team** when this extension was developed. A refactor after a couple of years is always an option...

I know it might be difficult to have full context of the situation which is why I personally try to learn the history before voicing my opinion.

src/image-copy/azext_imagecopy/__init__.py
fengzhou-msft: Is it possible to add `--temporary-resource-group-name` while deprecating `--temporary_resource_group_name`? Then you can remove `--temporary_resource_group_name` after a while.

src/image-copy/azext_imagecopy/custom.py
fengzhou-msft, line 26: Where is the code for checking `If --cleanup is set`?

qwordy: Fixed

src/image-copy/azext_imagecopy/__init__.py
qwordy: I support both now.

PR #1919, opened 2020-06-22
Support 2020-06-01 API version and add ppg parameter

tilnl:
---

### General Guidelines

For new extensions:

### About Extension Publish


src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -172,6 +172,9 @@
         - name: --vnet-subnet-id
           type: string
           short-summary: The ID of a subnet in an existing VNet into which to deploy the cluster.
+        - name: --ppg
```
zqingqing1, line 175: --ppg-id?

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -83,6 +83,7 @@ def load_arguments(self, _):
         c.argument('pod_cidr')
         c.argument('service_cidr')
         c.argument('vnet_subnet_id', type=str, validator=validate_vnet_subnet_id)
+        c.argument('ppg')
```
zqingqing1, line 86: Do you need to add some validation fo the ppg string?

tilnl: There is validation in rp sync code.

src/aks-preview/azext_aks_preview/_help.py
tilnl: According to cli command guidelines https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#argument-naming-conventions 
ID Arguments:
Arguments that expect ARM IDs should omit -id but call out that an ARM ID is allowed in the help text. These are often used as part of the "name or ID" convention (i.e: --storage-account can often accept a storage account name OR ARM ID).

JunSun17:
LGTM.

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -484,6 +487,9 @@
...
+          type: string
+          short-summary: The ID of a PPG.
```
gtxistxgao, line 492: use full name of PPG may be better.

tilnl: Following vmss, use -ppg.

PR #1975, opened 2020-07-06
[ai-examples] Fix missing subscription or disabled telemetry bug

mirdaki:
The Aladdin service expects to receive a GUID for both the correlationId and subscriptionId if they are sent. If users were not logged in or had disabled their telemetry, this service would return a 400, since the values passed in were considered strings, not GUIDs.

I believe https://github.com/Azure/azure-cli/issues/14094 was caused by this issue in `az find`, though the user was unable to reproduce it, so there could be other factors at play. The PR to fix `az find` is https://github.com/Azure/azure-cli/pull/14243.

---

### General Guidelines

### About Extension Publish

### DIFF
--- a/src/ai-examples/azext_ai_examples/custom.py
+++ b/src/ai-examples/azext_ai_examples/custom.py
@@ -13,8 +13,6 @@ from azure.cli.core import telemetry as telemetry_core
 from azure.cli.core import __version__ as core_version
 from azure.cli.core._help import HelpExample
 
-DEFAULT_GUID = "00000000-0000-0000-0000-000000000000"
-
 
 # Commands
 def check_connection_aladdin():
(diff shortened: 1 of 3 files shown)


yungezz:
hi @fengzhou-msft @bim-msft  could you pls help to review this one?

src/ai-examples/azext_ai_examples/custom.py
```diff
@@ -78,20 +80,20 @@ def ping_aladdin_service():
 
 
 def call_aladdin_service(query):
-    correlation_id = telemetry_core._session.correlation_id  # pylint: disable=protected-access
-    subscription_id = telemetry_core._get_azure_subscription_id()  # pylint: disable=protected-access
     version = str(parse_version(core_version))
 
+    # Only pull in the contextual values if we have consent
+    telemetry_consent = telemetry_core.is_telemetry_enabled()
+    correlation_id = telemetry_core._session.correlation_id if telemetry_consent else DEFAULT_GUID  # pylint: disable=protected-access
```
yungezz, line 87: when user chose opt out of telemetry collecting, we should not collect telemetry. then what's the default value used for?

mirdaki: Our service expects to receive a GUID value for either `correlationId` or `subscriptionId` if one is supplied. The `DEFAULT_GUID` just acts as a dummy value (being all zeros) to make the service happy. The other option would be to not include `correlationId` or `subscriptionId` in the request at all, but that made the flow a little more complex, so I thought this would be a simpler solution. I'm willing to switch to the other way if you think that would be preferred.

yungezz: thanks @mirdaki . Just to confirm, when user opting out, is there another other data sent to the service besides correlationId and subsId? We follows Microsoft general practice to allow user disabling telemetry. I'm not sure is it right to still send data when user opting out, even some dummy data.

mirdaki: That's a good point. Thank you for bringing it up. I will get an update out later today!

mirdaki: @yungezz Pushed the update. Let me know what you think!

And to answer your question about what data we send (sorry, forgot to reply before), in addition to the optional `correlationId` and `subscriptionId`, we need the command the user is asking about and the version of the CLI so we can provide the appropriate examples. Those last two items are required for us to give useful examples, so we don't consider them telemetry.

The reason we collect the `correlationId` and `subscriptionId` (for the users who do allow it) is so we can better understand the context users are running commands in to improve our ML model.

PR #1987, opened 2020-07-08
Update Frontdoor extension to use latest SDK version - RP API v2020-05-01

MoustafaEid:
---

### General Guidelines

For new extensions:

### About Extension Publish

Changes:
* Added the following fields to the `Backend` model:
   - `PrivateLinkAlias`
    - `PrivateLinkResourceId`
    - `PrivateLinkLocation`
    - `PrivateEndpointStatus`
    - `PrivateLinkApprovalMessage`


MoustafaEid:
Frontdoor tests are passing, but the storage tests are failing. Why is that? Can we bypass this check?
image

yonzhan:
add to S173

yonzhan:
@Juliehzl  please take a look at this PR.

MoustafaEid:
@Juliehzl any updates on this?

src/front-door/azext_front_door/_params.py
```diff
@@ -122,6 +122,10 @@ def load_arguments(self, _):
             c.argument('http_port', type=int, help='HTTP TCP port number.')
             c.argument('https_port', type=int, help='HTTPS TCP port number.')
             c.argument('weight', type=int, help='Weight of this endpoint for load balancing purposes.')
+            c.argument('private_link_alias', help='The Alias of the Private Link resource. Populating this optional field indicates that this backend is \'Private\'.')
+            c.argument('private_link_resource_id', help='The Resource Id of the Private Link. Populating this optional field indicates that this backend is \'Private\'.')
```
MyronFanQiu, line 126: What's private link resource id? How can user obtain this value?

src/front-door/azext_front_door/_params.py
```diff
@@ -122,6 +122,10 @@ def load_arguments(self, _):
...
+            c.argument('private_link_location', help='The location of the Private Link resource. Required only if \'privateLinkResourceId\' is populated.')
```
MyronFanQiu, line 127: what's the meaning of location? Can user choose one location?

src/front-door/azext_front_door/_params.py
MoustafaEid: The private link ID will be shared by the owners of the private link, they're usually an owner of a service that can be privately consumed without exposing that service publicly.

src/front-door/azext_front_door/_params.py
MoustafaEid: The location designates the location of the private link to which a frontdoor can connect. While the location can be any location that is supported by Frontdoor private link feature, the user will almost always choose the location of or closest to the private link provided to minimize latency of the request consuming said service behind the private link

MyronFanQiu: I thought user need to provide private link resource id already, why do they need support location one more time?

src/front-door/azext_front_door/_params.py
MyronFanQiu: Does it mean the private link service id?

src/front-door/azext_front_door/_params.py
MoustafaEid: Once a user has submitted a frontdoor with a private backend (by providing alias or resource id/location), Frontdoor create a private endpoint in a Frontdoor-owned virtual network. Requests coming to this private backend will be forwarded through said private endpoint, having it closer to the private link lowers the latency for the request.

Also, not all private links have locations, some can be global. It's up to the user to pick the location that makes sense to their use case, but it doesn't have to be a particular value

src/front-door/azext_front_door/_params.py
MoustafaEid: I'm not sure if Service ID is synonymous to Resource ID -- It's an ARM resource ID (e.g. `/subscriptions/da61bba1-cbd5-438c-a738-c717a6b2d59f/resourceGroups/moeidrg/providers/Microsoft.Network/privateLinkServices/pls-east`)

The naming convention matches up to what the owner of the private link expects to see when they attempt to copy the ID from portal
image

MyronFanQiu: Private Link is a quite general word. It represent private link service, private endpoint and  private endpoint connection together. Based on the resource id `Microsoft.Network/privateLinkServices`. It should be a ARM resource ID of private link service. Better to add more detail help message.

MoustafaEid: I get your point -- that's good feedback, we'll probably wait to gather more feedback from users re: help text/tips and make the appropriate adjustments in the next API release

PR #1994, opened 2020-07-09
{CI} Temporarily disable command index in CI

MyronFanQiu:
CI in extension use `pip` to install `whl` package and command index cannot detect such changes. Disable command index first. We need to refine CI later.
image


yonzhan:
add to S172

PR #2018, opened 2020-07-14
wrap <xxx> in backticks for better documentation Part 2

daxianji007:
## Issue
Our documentation treats the summary as markdown. If it has <xxx> it could be rendered as an html tag.

similar to https://github.com/Azure/azure-cli-extensions/pull/2016

## Common solution
Common solution is wrap <xxx> in backticks, as the PR shows.


yonzhan:
add to S173

PR #2031, opened 2020-07-16
[Monitor] az monitor app-insights api-key: Add warning to remove permission and example

### LINKED ISSUE #1945: az cli application-insights api-key creates with wrong permissions
### Description of issue

The default behavior of creating an api-key with the az cli application-insights extension is that it will add full permissions if  None is supplied.
If i would like to create a read only or write only api-key, then i cant seem to do it with the az cli. 

1) `az monitor app-insights api-key create --read-properties ReadTelemetry ..` 
would always result in WriteAnnotations as well, because nothing is provided for --write-properties and the default is WriteAnnotations

2) `az monitor app-insights api-key create --write-properties WriteAnnotations  .. `
would always result in ReadTelemetry and  AuthenticateSDKControlChannel as well, because nothing is provided for --read-properties and the default is 'ReadTelemetry', 'AuthenticateSDKControlChannel'.

My assumption and as it seems to behave everywhere else is that i need to say the permission set i want or if i supply no read and write permissions, then default is all. IF i supply my permission set, then only that will be used. 

I believe the bug is here:
https://github.com/Azure/azure-cli-extensions/blob/master/src/application-insights/azext_applicationinsights/custom.py

Line 142-145:
```
    if read_properties is None:
        read_properties = ['ReadTelemetry', 'AuthenticateSDKControlChannel']
    if write_properties is None:
        write_properties = ['WriteAnnotations']
```

MyronFanQiu:
---

Fix #1945 

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/application-insights/HISTORY.rst
+++ b/src/application-insights/HISTORY.rst
@@ -2,6 +2,7 @@
 ++++++++++++++++++
 
 * Argument `--cap` in `az monitor app-insights component billing update` supports float value.
+* [Incoming Breaking Change] Default write permission would be removed for `az monitor app-insights api-key` in the future.
 
 0.1.8
 ++++++++++++++++++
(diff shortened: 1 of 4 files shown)


yonzhan:
add to S174

PR #2053, opened 2020-07-22
[Account] Support sovereign clouds

jiasli:
As confirmed with @navysingla from service team. This extension now supports sovereign clouds.

Set `base_url_bound=True` so that the base URL for sovereign clouds' ARM endpoint is used.


yonzhan:
add to S173

jiasli:
Merged into https://github.com/Azure/azure-cli-extensions/pull/2308.

PR #2067, opened 2020-07-24
vm-repair: Fix vm name bug when 8th character in a VM name is a hyphen

swbae31:
- Minor fixes to address bug reports. 
- Minor fixes to help text

The default name value for newly created repair vms will now have a '_' at the end.

### DIFF
--- a/src/vm-repair/azext_vm_repair/_validators.py
+++ b/src/vm-repair/azext_vm_repair/_validators.py
@@ -41,7 +41,7 @@ def validate_create(cmd, namespace):
     if namespace.repair_vm_name:
         _validate_vm_name(namespace.repair_vm_name, is_linux)
     else:
-        namespace.repair_vm_name = ('repair-' + namespace.vm_name)[:15]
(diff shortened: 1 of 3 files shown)


yonzhan:
vm-repair

src/vm-repair/azext_vm_repair/_validators.py
```diff
@@ -41,7 +41,7 @@ def validate_create(cmd, namespace):
     if namespace.repair_vm_name:
         _validate_vm_name(namespace.repair_vm_name, is_linux)
     else:
-        namespace.repair_vm_name = ('repair-' + namespace.vm_name)[:15]
+        namespace.repair_vm_name = ('repair-' + namespace.vm_name)[:14] + '_'
```
qwordy, line 44: May I know why "-" is allowed after "repair" but not at the end?

swbae31: Its how Azure resource naming is set up. It cannot end with a hyphen.

PR #2074, opened 2020-07-28
[ai-examples] Add fallback to use built-in values if there is an issue with the service

mirdaki:
Now if the service provides no values or the client is unable to connect to the service, it falls back to using the built in examples.

---

### General Guidelines


yonzhan:
ai-examples

PR #2104, opened 2020-07-31
Shiqiu/release0729

ShichaoQiu:
* Remove 'cpu', 'memory' and 'instance-count' from 'az spring-cloud app deploy' command
* Fix log streaming feature proxy issues


yonzhan:
spring cloud

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -116,6 +114,7 @@ def spring_cloud_update(cmd, client, resource_group, name, app_insights_key=None
 
 
 def spring_cloud_delete(cmd, client, resource_group, name, no_wait=False):
+    logger.warning("Stop using Azure Spring Cloud? We appreciate your feedback: aka.ms/springclouddeletesurvey")
```
fengzhou-msft, line 117: Use the full `https://aka.ms/springclouddeletesurvey` so the link can be recognized and clicked from shell.

ShichaoQiu: @fengzhou-msft Good idea, have updated it, thanks

PR #2124, opened 2020-08-04
webapp: add support for using new OneDeploy API for deployments  in CLI

dannysongg:
- Allows users to deploy various types of artifacts such as app.war, app.jar, static files, and zip files
- Supports asynchronous deployments, and deployments to a custom path.
- Deployment type is inferred from the extension of the file being deployed. This can be overridden using the --type parameter.

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/webapp/azext_webapp/__init__.py
+++ b/src/webapp/azext_webapp/__init__.py
@@ -25,6 +25,7 @@ class WebappExtCommandLoader(AzCommandsLoader):
         with self.command_group('webapp') as g:
             g.custom_command('container up', 'create_deploy_container_app', exception_handler=ex_handler_factory())
             g.custom_command('remote-connection create', 'create_tunnel')
+            g.custom_command('deploy', 'enable_one_deploy')
 
         with self.command_group('webapp scan') as g:
             g.custom_command('start', 'start_scan')
--- a/src/webapp/azext_webapp/__init__.py
+++ b/src/webapp/azext_webapp/__init__.py
@@ -52,14 +52,17 @@ class WebappExtCommandLoader(AzCommandsLoader):
             c.argument('dryrun', help="show summary of the create and deploy operation instead of executing it", default=False, action='store_true')
             c.argument('registry_rg', help="the resource group of the Azure Container Registry")
             c.argument('registry_name', help="the name of the Azure Container Registry")
+            c.argument('slot', help="the deployment slot")
         with self.argument_context('webapp remote-connection create') as c:
             c.argument('port', options_list=['--port', '-p'],
                        help='Port for the remote connection. Default: Random available port', type=int)
             c.argument('name', options_list=['--name', '-n'], help='Name of the webapp to connect to')
+            c.argument('slot', help="Name of the deployment slot to use")
         with self.argument_context('webapp scan') as c:
             c.argument('name', options_list=['--name', '-n'], help='Name of the webapp to connect to')
             c.argument('scan_id', options_list=['--scan-id'], help='Unique scan id')
             c.argument('timeout', options_list=['--timeout'], help='Timeout for operation in milliseconds')
+            c.argument('slot', help="Name of the deployment slot to use")
 
         with self.argument_context('webapp deploy') as c:
             c.argument('name', options_list=['--name'], help='Name of the webapp to connect to')
@@ -68,6 +71,7 @@ class WebappExtCommandLoader(AzCommandsLoader):
             c.argument('is_async', options_list=['--async'], help='Asynchronous deployment', type=bool)
             c.argument('target_path', options_list=['--target-path'], help='Target path to which the file will be deployed to.')
             c.argument('timeout', options_list=['--timeout'], help='Timeout for operation in milliseconds')
+            c.argument('slot', help="Name of the deployment slot to use")
 
 
 COMMAND_LOADER_CLS = WebappExtCommandLoader
--- a/src/webapp/azext_webapp/__init__.py
+++ b/src/webapp/azext_webapp/__init__.py
@@ -52,7 +52,7 @@ class WebappExtCommandLoader(AzCommandsLoader):
             c.argument('dryrun', help="show summary of the create and deploy operation instead of executing it", default=False, action='store_true')
             c.argument('registry_rg', help="the resource group of the Azure Container Registry")
             c.argument('registry_name', help="the name of the Azure Container Registry")
-            c.argument('slot', help="the deployment slot")
+            c.argument('slot', help="Name of the deployment slot to use")
         with self.argument_context('webapp remote-connection create') as c:
             c.argument('port', options_list=['--port', '-p'],
                        help='Port for the remote connection. Default: Random available port', type=int)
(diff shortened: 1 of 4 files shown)


yonzhan:
webapp

src/webapp/azext_webapp/_help.py
```diff
@@ -59,3 +59,8 @@
         - name: Upload files from the current directory to an Azure Container Registry, then build a container image and deploy it to a web app. The Azure Container Registry must already exist.
           text: az webapp container up -n AppName --registry-rg ContainerRegistryResourceGroup --registry-name ContainerRegistryName
 """
+
+helps['webapp deploy'] = """
+    type: command
+    short-summary: Deploys a given artifact to Azure Web Apps through OneDeploy.
```
panchagnula, line 65: isn't this going to be scoped only for certain extensions currently? Like .war, .zip so would be good to be clear. Please check these help content with your PM as well

src/webapp/azext_webapp/custom.py
```diff
@@ -308,3 +310,75 @@ def _should_create_new_asp(cmd, rg_name, asp_name, location):
...
+    # check if there's an ongoing process
+    if res.status_code == 409:
+        raise CLIError("There may be an ongoing deployment or your app setting has WEBSITE_RUN_FROM_PACKAGE. "
+                       "Please track your deployment in {} and ensure the WEBSITE_RUN_FROM_PACKAGE app setting "
+                       "is removed.".format(deployment_status_url))
+
+    # check the status of async deployment
+    if res.status_code == 202:
+        logger.warning("Asynchronous deployment request has been recieved")
+        response = _check_zip_deployment_status(cmd, resource_group_name, name, deployment_status_url, authorization, timeout)
```
panchagnula, line 381: Have verified that this check deployment status works here as well?

src/webapp/azext_webapp/__init__.py
```diff
@@ -25,6 +25,7 @@ def load_command_table(self, _):
         with self.command_group('webapp') as g:
             g.custom_command('container up', 'create_deploy_container_app', exception_handler=ex_handler_factory())
             g.custom_command('remote-connection create', 'create_tunnel')
+            g.custom_command('deploy', 'enable_one_deploy')
```
panchagnula, line 28: We need the arguments defined for this command inside load_arguments below.

src/webapp/azext_webapp/_help.py
dannysongg: The OneDeploy API can deploy any extension to a webapp as a static file. However it does take specific actions for certain extensions such as .war, .jar, and .zip.

src/webapp/azext_webapp/custom.py
dannysongg: Check deployment status has the same functionality as it does when calling 'az webapp deployment source config-zip'

src/webapp/azext_webapp/custom.py
```diff
@@ -308,3 +310,75 @@ def _should_create_new_asp(cmd, rg_name, asp_name, location):
...
+    headers['Cache-Control'] = 'no-cache'
+    headers['User-Agent'] = get_az_user_agent()
+
+    import requests
+    import os
+
+    # Read file content
+    with open(os.path.realpath(os.path.expanduser(src)), 'rb') as fs:
+        artifact_content = fs.read()
+        logger.warning("Starting deployment...")
```
qwordy, line 365: It is more like an info not a warning

src/webapp/azext_webapp/custom.py
```diff
@@ -308,3 +310,75 @@ def _should_create_new_asp(cmd, rg_name, asp_name, location):
...
+    if res.status_code == 409:
+        raise CLIError("There may be an ongoing deployment or your app setting has WEBSITE_RUN_FROM_PACKAGE. "
+                       "Please track your deployment in {} and ensure the WEBSITE_RUN_FROM_PACKAGE app setting "
+                       "is removed.".format(deployment_status_url))
+
+    # check the status of async deployment
+    if res.status_code == 202:
+        logger.warning("Asynchronous deployment request has been recieved")
+        response = _check_zip_deployment_status(cmd, resource_group_name, name, deployment_status_url, authorization, timeout)
+        return response
```
qwordy, line 382: Is there any other status code that should be handled? And how to handle exception like connection error?

src/webapp/azext_webapp/custom.py
dannysongg: Other commands use logger.warning to print information to the terminal. I used here to remain consistent with the other code.

src/webapp/azext_webapp/custom.py
dannysongg: The code currently handles all possible status codes the can be returned during a OneDeploy deployment. I believe that other errors such as connection errors will be handled externally. Please correct me if I am incorrect about this.

yonzhan:
webapp

src/webapp/azext_webapp/_help.py
```diff
@@ -59,3 +59,8 @@
         - name: Upload files from the current directory to an Azure Container Registry, then build a container image and deploy it to a web app. The Azure Container Registry must already exist.
           text: az webapp container up -n AppName --registry-rg ContainerRegistryResourceGroup --registry-name ContainerRegistryName
 """
+
+helps['webapp deploy'] = """
+    type: command
```
panchagnula, line 64: Can you add some examples here for different types and if the path can be relative vs absolute path?

src/webapp/azext_webapp/custom.py
```diff
@@ -308,3 +310,75 @@ def _should_create_new_asp(cmd, rg_name, asp_name, location):
                 item.location == location):
             return False
     return True
+
+
+def enable_one_deploy(cmd, resource_group_name, name, src, deploy_type=None, is_async=None, target_path=None, timeout=None, slot=None):
+    import ntpath
+    logger.warning("Preparing for deployment")
```
panchagnula, line 317: can you make this logger.info() so the logs are minimal

src/webapp/azext_webapp/custom.py
```diff
@@ -308,3 +310,75 @@ def _should_create_new_asp(cmd, rg_name, asp_name, location):
...
+    user_name, password = _get_site_credential(cmd.cli_ctx, resource_group_name, name, slot)
+
+    try:
+        scm_url = _get_scm_url(cmd, resource_group_name, name, slot)
+    except ValueError:
+        raise CLIError('Failed to fetch scm url for function app')
```
panchagnula, line 323: why does this say function app here? can you update as failed to fetch sam url for app {} -> show app name as well.

src/webapp/azext_webapp/__init__.py
```diff
@@ -51,14 +52,26 @@ def load_arguments(self, _):
             c.argument('dryrun', help="show summary of the create and deploy operation instead of executing it", default=False, action='store_true')
             c.argument('registry_rg', help="the resource group of the Azure Container Registry")
             c.argument('registry_name', help="the name of the Azure Container Registry")
+            c.argument('slot', help="the deployment slot")
```
qwordy, line 55: Maybe you need to make the help messages consistent.
In the other two places, you use
```
Name of the deployment slot to use
```

PR #2282, opened 2020-09-02
[Next] add an experimental command `az next`

Edmund-Shi:
This pr add an experimental command az next which can recommend next command for you.

# Architecture Overview

There are mainly 3 classes in scheduled task:

## Kusto Client

You can use `get_data` function to get data from Kusto database. Please note that the Kusto query string hard codes the time period. If you need to change the function's schedule, you also need to change the time period in the query string.

### Returned Value

Returned value is a dictionary whose key is `uid` and value is a list of tuples which contain `command` `param` `os_type` `source` `begin_time` `end_time`. 

## Cosmos Client

You can use the Cosmos client to query data and insert items to Cosmos DB. We query all items in Cosmos DB to restore previous state of the `CommandTable`. And then use data from Kusto client to update the `CommandTable`. Finally, write back new data to Cosmos DB.

### Returned Value

Query return an iterable dict. You can find the API reference [here](https://docs.microsoft.com/en-us/python/api/azure-cosmos/azure.cosmos.containerproxy?view=azure-python#query-items-query--parameters-none--partition-key-none--enable-cross-partition-query-none--max-item-count-none--enable-scan-in-query-none--populate-query-metrics-none----kwargs-).

## Command Table

This table is used for calculating the next command for each command.

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -147,3 +147,5 @@
 /src/guestconfig/ @gehuan
 
 /src/swiftlet/ @qwordy
+
+/src/azext_next/ @Edmund-Shi
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -148,4 +148,4 @@
 
 /src/swiftlet/ @qwordy
 
-/src/azext_next/ @Edmund-Shi
+/src/azext_next/ @Edmund-Shi @zhoxing-ms 
--- a/src/next/azext_next/custom.py
+++ b/src/next/azext_next/custom.py
@@ -66,7 +66,9 @@ def _give_recommends(recommends):
             reason = " people use this command in next step."
         idx += 1
         print("{}. az {}".format(idx, rec['command']))
-        print("Recommended reason: {:.1f}% {}".format(rec['ratio'] * 100, reason))
+        if rec['ratio']:
+               reason = "{:.1f}% {}".format(rec['ratio'] * 100, reason)
+        print("Recommended reason: {}".format(reason))
 
 
 def handle_next(cmd):
--- a/src/next/azext_next/custom.py
+++ b/src/next/azext_next/custom.py
@@ -12,10 +12,10 @@ def _get_api_url():
     return "https://cli-recommendation.azurewebsites.net/api/RecommendationService"
 
 
-def _get_last_cmd():
+def _get_last_cmd(cmd):
     '''Get last executed command from local log files'''
     import os
-    his_file_name = os.path.join(os.environ['HOME'], '.azure', 'recommendation', 'cmd_history.log')
+    his_file_name = os.path.join(cmd.cli_ctx.config.config_dir, 'recommendation', 'cmd_history.log')
     with open(his_file_name, "r") as f:
         lines = f.read().splitlines()
         lines = [x for x in lines if x != 'next']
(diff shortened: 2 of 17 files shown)


yonzhan:
az next

.github/CODEOWNERS
```diff
@@ -147,3 +147,5 @@
 /src/guestconfig/ @gehuan
 
 /src/swiftlet/ @qwordy
+
+/src/azext_next/ @Edmund-Shi
```
zhoxing-ms, line 151:
```suggestion
/src/azext_next/ @Edmund-Shi @zhoxing-ms 
```

src/next/azext_next/custom.py
```diff
@@ -0,0 +1,119 @@
...
+3. command: Only the commands with high correlation with previously executed commands are recommend
+4. resource: Only the resources related to previously created resources are recommended
+5. senario: Only the E2E scenarios related to current usage scenarios are recommended
+'''
+    print(msg)
+    option = _read_int("What kind of recommendation do you want? (RETURN is to set all): ", 1)
+    last_cmd = _get_last_cmd()
+    recommends = _get_recommend_from_api(last_cmd, option)
+    if not recommends:
+        raise CLIError("Failed to get recommend for '{}'.".format(last_cmd))
```
zhoxing-ms, line 86: If the recommendation data is not obtained, it may not be because the query failed, but also be that there is no recommended data to meet the target after calculation.
Here we need to determine whether there is no data or query error according to the status code or error information

src/next/azext_next/custom.py
```diff
@@ -0,0 +1,119 @@
...
+    print("")
+    idx = 0
+    for rec in recommends:
+        if 'reason' in rec:
+            reason = rec['reason']
+        else:
+            reason = " people use this command in next step."
+        idx += 1
+        print("{}. az {}".format(idx, rec['command']))
+        print("Recommended reason: {:.1f}% {}".format(rec['ratio'] * 100, reason))
```
zhoxing-ms, line 69: We may also need to consider the case where `rec['ratio']` does not exist
```suggestion
        if rec['ratio']:
               reason = "{:.1f}% {}".format(rec['ratio'] * 100, reason)
        print("Recommended reason: {}".format(reason))
```

src/next/azext_next/custom.py
```diff
@@ -0,0 +1,119 @@
...
+        option = _read_int("Which one is helpful to you? (If none, please input 0) :")
+    else:
+        option = input("Does it helpful to you? (y/n): ")
+        if option in ["y", "yes", "Y", "Yes", "YES"]:
+            option = 1
+        else:
+            option = 0
+    if option == 0:
+        # we can send feedback here
+        return "recommend abort"
```
zhoxing-ms, line 99: Feedback should be sent for both adoption and non adoption.

src/next/azext_next/custom.py
Edmund-Shi: This may be also caused by bad network connection.

zhoxing-ms: Yes~

zhoxing-ms (approved):

zhoxing-ms:
Replaced by PR #2449

PR #2526, opened 2020-10-15
{storage-blob-preview} Add support for blob upload-batch, download-batch, delete-batch and copy start-batch

Juliehzl:
---

Test passes locally:
image

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
storage-blob-preview

src/storage-blob-preview/azext_storage_blob_preview/_params.py
```diff
@@ -179,6 +182,14 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
 
     with self.argument_context('storage blob') as c:
         c.argument('blob_name', options_list=('--name', '-n'), arg_type=blob_name_type)
+        c.argument('destination_path', help='The destination path that will be appended to the blob name.')
+        c.argument('socket_timeout', deprecate_info=c.deprecate(),
+                   help='The socket timeout(secs), used by the service to regulate data flow.')
```
zhoxing-ms, line 187: If parameter `socket_timeout` is no longer needed and `deprecate_info` is added just to avoid breaking users, could we hide it?

Juliehzl: deprecate_info will not break customers from my understanding. I don't want to hide it because I want to let customer know it will be deprecated soon and remove the usage in scripts.

src/storage-blob-preview/azext_storage_blob_preview/operations/blob.py
```diff
@@ -195,64 +191,86 @@ def _download_blob(blob_service, container, destination_folder, normalized_blob_
...
+                progress_callback.message = '{}/{}: "{}"'.format(
+                    index + 1, len(blobs_to_download), blobs_to_download[blob_normed])
+            blob_client = client.get_blob_client(container=container_name,
+                                                 blob=blobs_to_download[blob_normed])
+            destination_path = os.path.join(destination, os.path.normpath(blob_normed))
+            destination_folder = os.path.dirname(destination_path)
+            # Failed when there is same name for file and folder
+            if os.path.isfile(destination_folder) and os.path.exists(destination_folder):
+                raise CLIError("There is a file having the same name with directory to create. Please rename the file "
+                               "and retry the command.")
```
zhoxing-ms, line 228: Do we need to print out the duplicate file name here?
By the way, I have a little suggestion: maybe we can consider adding options to support users to choose to overwrite or skip the path with the same name. In this way, the whole download process will not be interrupted by one of the duplicate name of folder or file.

src/storage-blob-preview/azext_storage_blob_preview/_params.py
zhoxing-ms: I know there's no breaking change, I'm just not perfer to show a parameter which is deprecated for Track2. Because this field hide is more friendly to new users and reduces their learning costs.
However, displaying this field is better for users who write scripts to notice this change. But I personally think it's better to have the log output the deprecated information when executing the script, rather than let the users interactively execute `-h` command to find the deprecated information.

By the way, that's a small question and I can accept whether this parameter is displayed or not~

src/storage-blob-preview/azext_storage_blob_preview/_client_factory.py
```diff
@@ -138,3 +134,26 @@ def cf_adls_directory(cli_ctx, kwargs):
...
+    account_name = kwargs.pop('account_name', None)
+    if connection_string:
+        return t_share_service.from_connection_string(conn_str=connection_string)
+
+    account_url = get_account_url(cli_ctx, account_name=account_name, service='file')
+    credential = account_key or sas_token or token_credential
+
+    if account_url and credential:
+        return t_share_service(account_url=account_url, credential=credential)
+    return None
```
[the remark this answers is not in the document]
Juliehzl: why do we need to raise an exception here? If there is an exception, SDK side will raise it directly.

src/storage-blob-preview/azext_storage_blob_preview/_params.py
Juliehzl: Yes, I will hide it then.

src/storage-blob-preview/azext_storage_blob_preview/operations/blob.py
Juliehzl: print file name here. But I don't want to skip them because it may change import files existed in the folder for customer.

src/storage-blob-preview/azext_storage_blob_preview/_params.py
```diff
@@ -118,6 +119,20 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
...
+        help='Required if the blob has associated snapshots. "only": Deletes only the blobs snapshots. '
+             '"include": Deletes the blob along with all snapshots.')
+    overwrite_type = CLIArgumentType(
+        arg_type=get_three_state_flag(),
+        help='Whether the blob to be uploaded should overwrite the current data. If True, upload_blob will '
+        'overwrite the existing data. If set to False, the operation will fail with ResourceExistsError. '
+        'The exception to the above is with Append blob types: if set to False and the data already exists, '
+        'an error will not be raised and the data will be appended to the existing blob. If set '
+        'overwrite=True, then the existing append blob will be deleted, and a new one created. '
+        'Defaults to False.')
```
zhoxing-ms, line 135: Should we change the `upload_blob` to some name that users can easily understand
Such as: `upload_blob` --> the operation of uploading blob

src/storage-blob-preview/azext_storage_blob_preview/_params.py
```diff
@@ -263,6 +299,17 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
...
+    with self.argument_context('storage blob download-batch') as c:
+        # c.register_precondition_options()
+        c.ignore('container_name')
+        c.argument('destination', options_list=('--destination', '-d'))
+        c.argument('source', options_list=('--source', '-s'))
+        c.extra('max_concurrency', options_list='--max-connections', type=int, default=2,
+                help='The number of parallel connections with which to download.')
+        c.extra('no_progress', progress_type)
+        c.argument('max_connections', type=int,
+                   help='Maximum number of parallel connections to use when the blob size exceeds 64MB.')
```
zhoxing-ms, line 311: Why do you define a similar parameter `max_connections` when we have parameter `max_concurrencys`? What is the difference between them

src/storage-blob-preview/azext_storage_blob_preview/_validators.py
```diff
@@ -1059,3 +1075,199 @@ def validate_upload_blob(namespace):
...
+            full_path = os.path.join(root, f)
+            if not pattern or _match_path(full_path, pattern):
+                yield (full_path, full_path[len_folder_path:])
+
+
+def process_blob_download_batch_parameters(cmd, namespace):
+    """Process the parameters for storage blob download command"""
+    # 1. quick check
+    if not os.path.exists(namespace.destination) or not os.path.isdir(namespace.destination):
+        raise ValueError('incorrect usage: destination must be an existing directory')
```
zhoxing-ms, line 1102: Could we use the more specific error type `InvalidArgumentValueError` instead of `ValueError`?

src/storage-blob-preview/azext_storage_blob_preview/_validators.py
```diff
@@ -1059,3 +1075,199 @@ def validate_upload_blob(namespace):
...
+
+    # 4. determine blob type
+    if namespace.blob_type is None:
+        vhd_files = [f for f in namespace.source_files if f[0].endswith('.vhd')]
+        if any(vhd_files) and len(vhd_files) == len(namespace.source_files):
+            # when all the listed files are vhd files use page
+            namespace.blob_type = 'page'
+        elif any(vhd_files):
+            # source files contain vhd files but not all of them
+            raise CLIError("""Fail to guess the required blob type. Type of the files to be
```
zhoxing-ms, line 1132: Could we use the more specific error type `ArgumentUsageError` instead of `CLIError`?

src/storage-blob-preview/azext_storage_blob_preview/_params.py
Juliehzl: remove the last one

src/storage-blob-preview/azext_storage_blob_preview/_params.py
Juliehzl: updated

src/storage-blob-preview/azext_storage_blob_preview/_validators.py
Juliehzl: updated

src/storage-blob-preview/azext_storage_blob_preview/_validators.py
Juliehzl: updated

zhoxing-ms (approved):

PR #2680, opened 2020-11-17
[AKS] Fix addon's permission issue after updating to MSI cluster

norshtein:
---
For clusters updating from SPN cluster to MSI cluster, if the cluster is using addons before updating, when updating finishes, each addon will get a dedicated user assigned identity. Some permission is needed on the user assigned identity to make the addon correctly running. However, currently the permission is not grant, so this PR adds extracts a function `_put_managed_cluster_ensuring_permission` from creating path and execute it in updating path to ensure permission.

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
AKS

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -3002,3 +2983,69 @@ def get_aks_custom_headers(aks_custom_headers=None):
...
+                              resource_name=name,
+                              parameters=managed_cluster,
+                              custom_headers=headers)
+
+    return cluster
+
+
+def _is_msi_cluster(managed_cluster):
+    return (managed_cluster and managed_cluster.identity and
+            (managed_cluster.identity.type.casefold() == "systemassigned" or managed_cluster.identity.type.casefold() == "userassigned"))
```
arrownj, line 3051: Can `managed_cluster.identity.type` be None ?

norshtein: Yes, it can be 'None', when it is set to 'None', the cluster is not  an MSI cluster so this function should return false.

arrownj: The code will raise exception if `managed_cluster.identity.type` is None, right ?

arrownj (approved):

PR #2720, opened 2020-11-25
[VirtualWan] az network vpn-gateway connection: Change the underlying operations from VpnGatewayOperations to VpnConnectionOperations

haroldrandom:
Close https://github.com/Azure/azure-cli/issues/15036

- Changing the operations from `VpnGatewayOperations` to `VpnConnectionOperations`, which brings breaking change to user when user creates a `vpn-gateway connection` as the response content is the data strucutre of `VpnConnection` instead of the `VpnGateway`.
- `network vpn-gateway show` raise `SystemExit` instead of `CLIError`.

Also, fix the bug that virtual-wan cannot detect tests.

---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
VirtualWan

src/virtual-wan/azext_vwan/_params.py
```diff
@@ -121,7 +121,9 @@ def load_arguments(self, _):
         c.argument('shared_key', help='Shared key.')
 
     with self.argument_context('network vpn-gateway connection list') as c:
+        # List commands cannot use --ids flag
         c.argument('resource_name', vpn_gateway_name_type, id_part=None)
+        c.argument('gateway_name', id_part=None)
```
kairu-ms, line 126: Is it a breaking change, as the 'gateway_name' parameter required?

haroldrandom: As the comments said, if I don't set `id_part=None`. The argument lists would have `--ids` for list command.

After using the origin SDK method as the implementation method and setting id_part of `gateway_name` to None and , it behaves like the same as before

image

image

kairu-ms (approved):

jsntcy (approved):

PR #2736, opened 2020-12-02
[ASC] [Azure Spring Cloud] Switch api-version from 2019-05-01-preview to 2020-07-01

ShichaoQiu:
Switch api-version from 2019-05-01-preview to 2020-07-01


yonzhan:
ASC

src/spring-cloud/azext_spring_cloud/_validators.py
```diff
@@ -43,6 +43,12 @@ def validate_sku(namespace):
             raise CLIError("The pricing tier only accepts value [Basic, Standard]")
 
 
+def validate_instance_count(namespace):
+    if namespace.instance_count is not None:
+        if namespace.instance_count < 1:
+            raise CLIError("--instance-count must be more than 0")
```
fengzhou-msft, line 49: Suggest to use `greater than`

src/spring-cloud/azext_spring_cloud/custom.py
```diff
@@ -60,16 +59,23 @@ def spring_cloud_create(cmd, client, resource_group, name, location=None, app_in
...
-                          app_insights_key, app_insights, disable_distributed_tracing)
-
     resource = models.ServiceResource(location=location, sku=full_sku, properties=properties, tags=tags)
-    return sdk_no_wait(no_wait, client.create_or_update, resource_group_name=resource_group, service_name=name, resource=resource)
+
+    poller = client.services.create_or_update(
+        resource_group, name, resource)
+    logger.warning(" - Creating Service ..")
+    while poller.done() is False:
+        sleep(5)
```
fengzhou-msft, line 70: Is it possible that in case of some failure, the returned response is not a poller?

src/spring-cloud/azext_spring_cloud/_validators.py
ShichaoQiu: have updated it, thanks

fengzhou-msft (approved):

PR #2815, opened 2020-12-21
{storage-blob-preview} Support blob url for blob related commands

Juliehzl:
---

Support blob url for blob related commands:
- you could use blob url with sas token
- you also could use blob url without sas token, but provide --account-key, --sas-token.. credential

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
storage-blob-preview

src/storage-blob-preview/azext_storage_blob_preview/_client_factory.py
```diff
@@ -76,6 +78,18 @@ def cf_blob_service(cli_ctx, kwargs):
...
+        credential = get_credential(kwargs)
+        # del unused kwargs
+        kwargs.pop('connection_string')
+        kwargs.pop('account_name')
+        kwargs.pop('container_name')
+        kwargs.pop('blob_name')
+        return t_blob_client.from_blob_url(blob_url=kwargs.pop('blob_url'),
+                                           credential=credential,
+                                           snapshot=kwargs.pop('snapshot', None))
+    kwargs.pop('blob_url')
```
evelyn-ys, line 92: It seems `blob_url` can be either `None` or `''` here. Why should we pop it out?

src/storage-blob-preview/azext_storage_blob_preview/_params.py
```diff
@@ -219,6 +219,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals, too-many-statem
         c.extra('lease', lease_type)
         c.extra('snapshot', snapshot_type)
         c.extra('version_id', version_id_type)
+        c.argument('delete_snapshots', arg_type=get_enum_type(['only', 'include']),
```
evelyn-ys, line 222: Can we use the enum definition in sdk?

evelyn-ys (changes requested):
Don't forget to upgrade setup.py

evelyn-ys (approved):

src/storage-blob-preview/azext_storage_blob_preview/_client_factory.py
Juliehzl: Actually if blob url is none or '', it would be invalid. added check for blob url and blob name.

src/storage-blob-preview/azext_storage_blob_preview/_params.py
Juliehzl: yes

Juliehzl:
> Don't forget to upgrade setup.py

I would update in next PR to make two PRs in one release.

PR #2867, opened 2021-01-06
[storage-blob-preview] `az storage blob upload`: support jumbo block blob

evelyn-ys:
The maximum size supported for a block blob increases from **4.75 TiB** to **190.7 TiB** (50,000 blocks, each block from **100MB** to **4000MB**).

---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
storage-blob-preview

src/storage-blob-preview/azext_storage_blob_preview/operations/blob.py
```diff
@@ -586,6 +586,22 @@ def snapshot_blob(client, metadata=None, **kwargs):
...
+    if not blob_type or blob_type != 'block':
+        return
+    # increase the block size to 100MB when the block list will contain more than
+    # 50,000 blocks(each block 4MB)
+    if length > 50000 * 4 * 1024 * 1024:
+        client._config.max_block_size = 100 * 1024 * 1024
+        client._config.max_single_put_size = 256 * 1024 * 1024
+    # increase the block size to 4000MB when the block list will contain more than
+    # 50,000 blocks(each block 100MB)
+    if length > 50000 * 100 * 1024 * 1024:
```
Juliehzl, line 600: with two `if`, if the length > 50000 * 100* 1024 * 1024, `client_config` will be change twice. Could we use `if ... elif ...`?

evelyn-ys: You are right, the value will be overwrote twice. I just want the code flow more close to human mind. When 5000 blocks of 4MB is not enough, we should increase the block size to 100MB. When 100MB is not enough again, we should increase to 4000MB.

src/storage-blob-preview/azext_storage_blob_preview/_help.py
```diff
@@ -6,6 +6,29 @@
 
 from knack.help_files import helps  # pylint: disable=unused-import
 
+helps['storage account blob-service-properties update'] = """
```
evelyn-ys, line 9: This change has nothing to do with jumbo blob, it's aimed to fix CI.

As in cli main repo, a new argument(`--default-service-version`) is added to this command and the help example added the usage of this new argument. But this command is overwritten in extension to apply track2 sdk and no new argument added, so if we keep using help from main repo, CI will fail because in extension CI it can't recognize the new argument.

Juliehzl: As extension is preview for storage module, we should make it newer than main repo.
For the feature added in main repo, we should support it in this extension, too.
Could you add `--default-service-version` in extension command, just copy my code in main repo?

evelyn-ys: Sure

src/storage-blob-preview/azext_storage_blob_preview/operations/blob.py
```diff
@@ -574,6 +574,22 @@ def snapshot_blob(client, metadata=None, **kwargs):
     return client.get_blob_properties()
 
 
+# pylint: disable=protected-access
+def _adjust_block_blob_size(client, blob_type, length):
+    if not blob_type or blob_type != 'block':
+        return
+    # increase the block size to 100MB when the block list will contain more than
+    # 50,000 blocks(each block 4MB)
+    if length > 50000 * 4 * 1024 * 1024:
```
Juliehzl, line 583: should also make sure length <= 50000 * 100 * 1024 * 1024

src/storage-blob-preview/azext_storage_blob_preview/operations/blob.py
```diff
@@ -618,11 +634,13 @@ def upload_blob(cmd, client, file_path=None, container_name=None, blob_name=None
     try:
         if file_path:
             length = os.path.getsize(file_path)
+            _adjust_block_blob_size(client, blob_type, length)
```
Juliehzl, line 637: this should only applied to block blob, not page or append blob. please add check.

evelyn-ys: It's checked in `_adjust_block_blob_size`:
```Python
def _adjust_block_blob_size(client, blob_type, length):
    if not blob_type or blob_type != 'block':
        return
```

src/storage-blob-preview/azext_storage_blob_preview/operations/blob.py
evelyn-ys: For those larger than 50000 * 100 * 1024 * 1024, the best thing we can do is making the config biggest value. So I didn't add length <= 50000 * 100 * 1024 * 1024 here.

src/storage-blob-preview/azext_storage_blob_preview/operations/blob.py
Juliehzl: Because we already have the capability to PUT with block size `4000 * 1024 * 1024`, could we just remove the first `if`?

src/storage-blob-preview/azext_storage_blob_preview/tests/latest/test_storage_blob_live_scenarios.py
```diff
@@ -45,14 +45,20 @@ def test_storage_blob_upload_10G_file(self, resource_group, storage_account):
         self.verify_blob_upload_and_download(resource_group, storage_account, 10 * 1024 * 1024,
                                              'block', skip_download=True)
 
+    @ResourceGroupPreparer()
+    @StorageAccountPreparer()
+    def test_storage_blob_upload_5G_file_with_fixed_block_size(self, resource_group, storage_account):
```
Juliehzl, line 50: with previous api version, we already have the capability to upload 10GB blob. How can we test the new api version support jumbo blob then?

src/storage-blob-preview/azext_storage_blob_preview/tests/latest/test_storage_blob_live_scenarios.py
```diff
@@ -68,8 +74,19 @@ def verify_blob_upload_and_download(self, group, account, file_size_kb, blob_typ
...
+        def fix_block_blob_size(client, blob_type, length):
+            client._config.max_block_size = 4000 * 1024 * 1024
+            client._config.max_single_put_size = 5000 * 1024 * 1024
+
+        if fix_block_size:
+            import mock
+            with mock.patch('azext_storage_blob_preview.operations.blob._adjust_block_blob_size',
+                            side_effect=fix_block_blob_size):
+                self.cmd('storage blob upload -c {} -f "{}" -n {} --type {} --timeout 1200'
+                         .format(container, local_file, blob_name, blob_type))
```
evelyn-ys, line 86: Mock function `_adjust_block_blob_size` here so that no matter how large the file is, `client._config` is always taking the largest value. In this way, jumbo blob can be tested.

src/storage-blob-preview/azext_storage_blob_preview/tests/latest/test_storage_blob_live_scenarios.py
evelyn-ys: See https://github.com/Azure/azure-cli-extensions/pull/2867#discussion_r584463312

src/storage-blob-preview/azext_storage_blob_preview/operations/blob.py
evelyn-ys: Do you mean for all `length > 50000 * 4 * 1024 * 1024`, we set the `_config.max_block_size` to 4000MB, just skip 100MB? I think it's workable.

evelyn-ys: Made the change.

Juliehzl (approved):

PR #2893, opened 2021-01-13
Fix UT for azure spring cloud

gavinfish:
---

For Python, we need assert raised exceptions out of `self.assertRaises()`, referring to [here](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) for details. Currently lots of UT in Azure Spring Cloud misuses it, and this will lead regression someday since all these UT do not work as excepted today.

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/spring-cloud/azext_spring_cloud/tests/latest/test_asc_validator.py
+++ b/src/spring-cloud/azext_spring_cloud/tests/latest/test_asc_validator.py
@@ -121,14 +121,14 @@ class TestValidateIPRanges(unittest.TestCase):
                        reserved_cidr_range='10.0.0.0/14', sku=None, location='eastus')
         with self.assertRaises(CLIError) as context:
             validate_vnet_required_parameters(ns)
-            self.assertEqual('--app-subnet, --service-runtime-subnet must be set when deploying to VNet',
-                             str(context.exception))
+        self.assertEqual('--app-subnet, --service-runtime-subnet must be set when deploying to VNet',
+                         str(context.exception))
 
     def test_single_cidr(self):
         ns = Namespace(reserved_cidr_range='10.0.0.0/8')
         with self.assertRaises(CLIError) as context:
             _validate_cidr_range(ns)
-            self.assertEqual('--reserved-cidr-range should be 3 unused /16 IP ranges', str(context.exception))
+        self.assertEqual('--reserved-cidr-range should be 3 unused /16 IP ranges', str(context.exception))
 
     def test_multi_cidr(self):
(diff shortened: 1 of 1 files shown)


src/spring-cloud/azext_spring_cloud/tests/latest/test_asc_validator.py
```diff
@@ -230,13 +230,9 @@ def test_set_default_cidr_range(self):
             _set_default_cidr_range(
                 ['128.0.0.0/1', '0.0.0.0/2', '64.0.0.0/3', '96.0.0.0/4', '112.0.0.0/5', '120.0.0.0/6', '124.0.0.0/7',
                  '126.0.0.0/8'])
-            self.assertEqual(
-                'Cannot set "reserved-cidr-range" automatically.Please specify "--reserved-cidr-range" with 3 unused CIDR ranges in your network environment.',
-                str(context.exception))
-            _set_default_cidr_range(['0.0.0.0/1'])
```
gavinfish, line 236: @yuwzho please help to verify whether this case is safe to remove. Currently it will not throw exception as its original expectation.

yonzhan:
spring cloud

src/spring-cloud/azext_spring_cloud/tests/latest/test_asc_validator.py
yuwzho: LGTM

yuwzho (approved):

src/spring-cloud/azext_spring_cloud/tests/latest/test_asc_validator.py
```diff
@@ -121,14 +121,14 @@ def test_lack_of_parameters(self):
                        reserved_cidr_range='10.0.0.0/14', sku=None, location='eastus')
         with self.assertRaises(CLIError) as context:
             validate_vnet_required_parameters(ns)
-            self.assertEqual('--app-subnet, --service-runtime-subnet must be set when deploying to VNet',
-                             str(context.exception))
+        self.assertEqual('--app-subnet, --service-runtime-subnet must be set when deploying to VNet',
+                         str(context.exception))
```
fengzhou-msft, line 125: What happens if assertEqual fails in the original code?

gavinfish: in the original code, the assertEqual will never be executed since `validate_vnet_required_parameters` will throw exception and be caught by `self.assertRaises(CLIError)`

fengzhou-msft (approved):

PR #2921, opened 2021-01-21
[SSH] Add support for getting the username from the certificate

N6UDP:
* Change to use the first in the list of validprincipals as the default username
* Remove old paramiko dependency
This is an ugly hack with a real fix to come as part of a client and server side combined fix.  This will fix guest users who are using their mail (e.g. first.last@contso.com) but logged into az cli using a different UPN (e.g. flast@contso.com)

---

### General Guidelines

### About Extension Publish

### DIFF
--- a/src/ssh/HISTORY.md
+++ b/src/ssh/HISTORY.md
@@ -1,6 +1,11 @@
 Release History
 ===============
 
+0.1.4
+-----
+* Change to use the first in the list of validprincipals as the default username
+* Remove old paramiko dependency
+
 0.1.3
 -----
 * Add support for using private IPs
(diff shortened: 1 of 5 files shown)


N6UDP:
@fengzhou-msft this should fix your guest issue for now; we had a long discussion earlier about the long term solution to this but we agreed we should just read it from the cert for now.  I may eventually do so in python instead of via ssh-keygen but there didn't appear to be a read-to-go library.

yungezz:
hi @fengzhou-msft could you pls help to review? thanks

fengzhou-msft (approved):

PR #2964, opened 2021-02-02
[AKS]Assign required permission for virtual-node MSI

gossion:
the required permission is correctly assigned when `create`ing a new AKS cluster, but still missed in `eanble-addons` subcommand. This PR is to fill that gap.


yonzhan:
AKS

feiskyer (approved):

fengzhou-msft (approved):

PR #2973, opened 2021-02-04
[Databricks] Add --enable-no-public-ip

fengzhou-msft:
`az databricks workspace create`: Add `--enable-no-public-ip` to allow creating workspace with no public ip

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/databricks/azext_databricks/custom.py
+++ b/src/databricks/azext_databricks/custom.py
@@ -24,6 +24,7 @@ def create_databricks_workspace(cmd, client,
                                 tags=None,
                                 prepare_encryption=None,
                                 require_infrastructure_encryption=None,
+                                enable_no_public_ip=None,
                                 no_wait=False):
     body = {}
     body['tags'] = tags  # dictionary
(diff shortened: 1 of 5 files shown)


yonzhan:
Databricks

qwordy:
Does it support `--enable-no-public-ip false`?

src/databricks/azext_databricks/custom.py
```diff
@@ -37,6 +38,7 @@ def create_databricks_workspace(cmd, client,
     _set_parameter_value(parameters, 'custom_private_subnet_name', custom_private_subnet_name)  # str
     _set_parameter_value(parameters, 'prepare_encryption', prepare_encryption)
     _set_parameter_value(parameters, 'require_infrastructure_encryption', require_infrastructure_encryption)
+    _set_parameter_value(parameters, 'enable_no_public_ip', enable_no_public_ip)
```
qwordy, line 41: Does this function support nested path like 'foo.bar'?

yungezz (approved):

fengzhou-msft:
> Does it support `--enable-no-public-ip false`?

Current command interface does not support this. If this option is not provided, the property is none, but the behavior should be the same as setting to false. Once you provide `--enable-no-public-ip`, you cannot update it to false.

qwordy (approved):

src/databricks/azext_databricks/custom.py
fengzhou-msft: Currently there is only one layer properties under `parameters` and the `_set_parameter_value` method only supports a single-value path as well. 

The swagger for this service is a bit different from others. Each property is defined as an object which has `type` and `value` properties (`type` is readonly and we only need to set `value`) instead of a simple string or boolean, that's why I defined this method to set value.

PR #2987, opened 2021-02-08
{Script} Add retry in get_whl_from_url and only verify latest version of an extension

fengzhou-msft:
To mitigate the increasing failures in `Verify Extensions Index` CI task recently.

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/scripts/ci/test_index.py
+++ b/scripts/ci/test_index.py
@@ -16,7 +16,7 @@ import unittest
 import hashlib
 import shutil
 
-from distutils.version import LooseVersion
+from packaging import version
 from wheel.install import WHEEL_INFO_RE
 
 from util import get_ext_metadata, get_whl_from_url, get_index_data, verify_dependency
@@ -107,7 +107,7 @@ class TestIndex(unittest.TestCase):
     def test_checksums(self):
         for exts in self.index['extensions'].values():
             # only test the latest version
-            item = max(exts, key=lambda ext: LooseVersion(ext['metadata']['version']))
+            item = max(exts, key=lambda ext: version.parse(ext['metadata']['version']))
             ext_file = get_whl_from_url(item['downloadUrl'], item['filename'],
                                         self.whl_cache_dir, self.whl_cache)
             print(ext_file)
@@ -136,7 +136,7 @@ class TestIndex(unittest.TestCase):
         extensions_dir = tempfile.mkdtemp()
         for ext_name, exts in self.index['extensions'].items():
             # only test the latest version
-            item = max(exts, key=lambda ext: LooseVersion(ext['metadata']['version']))
+            item = max(exts, key=lambda ext: version.parse(ext['metadata']['version']))
             ext_dir = tempfile.mkdtemp(dir=extensions_dir)
             ext_file = get_whl_from_url(item['downloadUrl'], item['filename'],
                                         self.whl_cache_dir, self.whl_cache)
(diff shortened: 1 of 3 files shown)


yonzhan:
Script

scripts/ci/test_index.py
```diff
@@ -21,6 +21,10 @@
 
 from util import get_ext_metadata, get_whl_from_url, get_index_data, verify_dependency
 
+# Some extensions are published through manually submitted PRs and the latest version may not be in the end of
+# the version list in index.json. Let's just test all versions for these extensions.
+EXTS_TEST_ALL_VERSIONS = {'azure-cli-ml', 'alias', 'azure_iot'}
```
jiasli, line 26: Maybe we can sort all versions and only test the latest one?

scripts/ci/test_index.py
```diff
@@ -105,8 +109,12 @@ def test_filename_duplicates(self):
 
     @unittest.skipUnless(os.getenv('CI'), 'Skipped as not running on CI')
     def test_checksums(self):
-        for exts in self.index['extensions'].values():
-            for item in exts:
+        for ext_name, exts in self.index['extensions'].items():
+            end = len(exts) - 1
+            for index, item in enumerate(exts):
```
jiasli, line 114: Redefine the name `index` can be confusing to readers. 😉

scripts/ci/test_index.py
jiasli: Python's built-in function [`max`](https://docs.python.org/3/library/functions.html#max) supports a custom key extraction function and we can use [`packaging.version`](https://packaging.pypa.io/en/latest/version.html) for version comparison.

fengzhou-msft: updated with picking a `max` version so the order of versions does not matter now.

scripts/ci/test_index.py
```diff
@@ -106,14 +106,16 @@ def test_filename_duplicates(self):
...
-            for item in exts:
-                ext_file = get_whl_from_url(item['downloadUrl'], item['filename'],
-                                            self.whl_cache_dir, self.whl_cache)
-                computed_hash = get_sha256sum(ext_file)
-                self.assertEqual(computed_hash, item['sha256Digest'],
-                                 "Computed {} but found {} in index for {}".format(computed_hash,
-                                                                                   item['sha256Digest'],
-                                                                                   item['filename']))
+            # only test the latest version
+            item = max(exts, key=lambda ext: LooseVersion(ext['metadata']['version']))
```
jiasli, line 110: `distutils.version.LooseVersion` is legacy and conformant only to the superseded [PEP 386](https://www.python.org/dev/peps/pep-0386/) (by [PEP 440](https://www.python.org/dev/peps/pep-0440/)). Perhaps we can migrate to the latest [`packaging.version`](https://packaging.pypa.io/en/latest/version.html).

See https://stackoverflow.com/a/11887885/2199657

jiasli:
```py
>>> from distutils.version import LooseVersion, StrictVersion
>>> StrictVersion('1.0.0a1') < StrictVersion('1.0.0')
True
>>> LooseVersion('1.0.0a1') < LooseVersion('1.0.0')
False
```

scripts/ci/test_index.py
```diff
@@ -16,7 +16,7 @@
 import hashlib
 import shutil
 
-from distutils.version import LooseVersion
+from packaging import version
```
fengzhou-msft, line 19: `distutils.version.LooseVersion` does not support PEP440 which we follow as discussed in https://stackoverflow.com/a/11887885/2199657

jiasli (approved):

PR #3165, opened 2021-03-18
Requiring the list of providers as argument on workspace creation

ricardo-espinoza:


qwordy (approved):

qwordy:
@ricardo-espinoza Can you share why remove the default provider? I like short command. It is harder to use when more parameters become required.

ricardo-espinoza:
@qwordy , we removed the default provider because of a change on the service side. We can't assume that any of the providers will be available in all subscriptions, and as such, some might have Microsoft/Basic and not others. We're discussing if we will have that in the future, but for now, requiring the user to be explicit on the provider ensures that they will only specify those available to them.

PR #3213, opened 2021-03-31
Changes to support new resource-graph api version

fenwickt:
- Updated sdk with the new track2 python package
- Added new param management_groups for querying the management groups
- Added new param allow_partial_scopes for not failing the query and returning partial results when query overreaches max scopes allowed by server
- Added new param skip_token so that users can paginate through the query
- Modified max value of top to be 1000 instead of 5000
- Changed cmdl output to dictionary format with the fileds 'data' and 'skip_token'
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
resource-graph

fenwickt:
@kairu-ms would really appreciate the PR review!

src/resource-graph/azext_resourcegraph/_params.py
```diff
@@ -30,19 +28,16 @@ def load_arguments(self, _):
     with self.argument_context('graph query') as c:
         c.argument('graph_query', options_list=['--graph-query', '--q', '-q'], required=True,
                    completer=get_generic_completion_list(_QUERY_EXAMPLES), help='Resource Graph query to execute.')
-        c.argument('first', options_list=['--first'], required=False, type=int, default=100,
-                   help='The maximum number of objects to return. Accepted range: 1-5000.')
-        c.argument('skip', options_list=['--skip'], required=False, type=int, default=0,
-                   help='Ignores the first N objects and then gets the remaining objects.')
+        c.argument('first', options_list=['--first'], required=False, type=int, default=None,
+                   help='The maximum number of objects to return. Accepted range: 1-1000. Default value is 100.')
```
kairu-ms, line 32: Why the default is None but it's 100 in help message?

src/resource-graph/azext_resourcegraph/_params.py
```diff
@@ -30,19 +28,16 @@ def load_arguments(self, _):
...
+        c.argument('skip', options_list=['--skip'], required=False, type=int, default=None,
+                   help='Ignores the first N objects and then gets the remaining objects. Default value is 0.')
```
kairu-ms, line 34: Same question as above

src/resource-graph/azext_resourcegraph/azext_metadata.json
```diff
@@ -1,4 +1,4 @@
 {
     "azext.minCliCoreVersion": "2.3.1",
```
kairu-ms, line 2: Recommend to update min version to 2.22.0

src/resource-graph/azext_resourcegraph/_validators.py
```diff
@@ -7,8 +7,11 @@
...
-        raise CLIError("Value of --first has to be between 1 and 5000.")
+    if namespace.first and not 1 <= namespace.first <= 1000:
+        raise CLIError("Value of --first has to be between 1 and 1000.")
 
-    if not namespace.skip >= 0:
+    if namespace.skip and not namespace.skip >= 0:
         raise CLIError("Value of --skip cannot be negative.")
+
+    if namespace.subscriptions is not None and namespace.management_groups is not None:
+        raise CLIError("Both --subscriptions and --management-groups cannot be passed together at the same time.")
```
kairu-ms, line 17: Please follow the [guideline](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md) to deprecate CLIError.

src/resource-graph/azext_resourcegraph/custom.py
```diff
@@ -14,90 +14,95 @@
...
                              "To rewrite the query and enable paging, "
                              "see the docs for an example: https://aka.ms/arg-results-truncated")
 
-    except ErrorResponseException as ex:
-        raise CLIError(json.dumps(_to_dict(ex.error), indent=4))
-
-    return results
-
+    except HttpResponseError as ex:
+        raise CLIError(json.dumps(_to_dict(ex.model.error), indent=4))
```
kairu-ms, line 100: Please follow the [guideline](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md) to deprecate CLIError.

src/resource-graph/azext_resourcegraph/custom.py
```diff
@@ -14,90 +14,95 @@
...
-
-            request = QueryRequest(query=full_query, subscriptions=subs_list, options=request_options)
-            response = client.resources(request)  # type: QueryResponse
-            if response.result_truncated == ResultTruncated.true:
-                result_truncated = True
 
+        if first is not None:
+            first = min(first, __ROWS_PER_PAGE)
+        elif skip_token is None:
+            first = __FIRST
```
kairu-ms, line 66: Why assign first when skip_token is None

src/resource-graph/azext_resourcegraph/custom.py
```diff
@@ -11,93 +11,97 @@
...
-                                            query=graph_query,
-                                            tags=tags,
-                                            location=location)
-    return client.graph_query.create_or_update(resource_group_name=resource_group_name,
-                                               resource_name=resource_name,
-                                               properties=graph_shared_query)
+    result_dict = dict(); 
+    result_dict['data']=results
+    result_dict['skip_token']=skip_token
+    return result_dict
```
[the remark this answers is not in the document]
kairu-ms: That's ok.

src/resource-graph/azext_resourcegraph/custom.py
```diff
@@ -14,90 +14,95 @@
...
-            results.extend(response.data)
-
-            if len(results) >= first or skip_token is None:
-                break
 
         if result_truncated and len(results) < first:
             __logger.warning("Unable to paginate the results of the query. "
                              "Some resources may be missing from the results. "
                              "To rewrite the query and enable paging, "
                              "see the docs for an example: https://aka.ms/arg-results-truncated")
```
kairu-ms, line 97: According to about question, first will be None and it will cause error

src/resource-graph/azext_resourcegraph/custom.py
```diff
@@ -14,90 +14,95 @@
...
-            if response.result_truncated == ResultTruncated.true:
-                result_truncated = True
 
+        if first is not None:
+            first = min(first, __ROWS_PER_PAGE)
+        elif skip_token is None:
+            first = __FIRST
+
+        if skip is None and skip_token is None:
+            skip = __SKIP
```
kairu-ms, line 69: This part of code should be in `validate_query_args`

src/resource-graph/azext_resourcegraph/custom.py
```diff
@@ -14,90 +14,95 @@
...
-                                            query=graph_query,
-                                            tags=tags,
-                                            location=location)
-    return client.graph_query.create_or_update(resource_group_name=resource_group_name,
-                                               resource_name=resource_name,
-                                               properties=graph_shared_query)
+    result_dict = dict()
+    result_dict['data'] = results
+    result_dict['skip_token'] = skip_token
+    return result_dict
```
kairu-ms, line 105: Why not directly return response?

src/resource-graph/HISTORY.rst
```diff
@@ -1,3 +1,8 @@
+2.0.0
```
kairu-ms, line 1: 3.0.0?

fenwickt: Changed setup.py to 2.0

src/resource-graph/azext_resourcegraph/custom.py
fenwickt: Actually that would work. Changed to return response.

src/resource-graph/azext_resourcegraph/custom.py
fenwickt: Moved

src/resource-graph/azext_resourcegraph/custom.py
fenwickt: Good catch, thank you! Added a check

src/resource-graph/azext_resourcegraph/_params.py
fenwickt: 100 is set be default in the custom.py if skip token is None. This is to know whether 100 was set explicitly by user or user left it empty. Please let me know if there is a better way to do it!

We have this logic where skip and first override skip token if present. Basically if user is passing us skipToken we're decoding the skip token. But if user passes us skip token AND skip or first, we will override the values from the skip token to those passed by user. 
That's why it is important to know when user explicitly wanted to override the top and skip and when the default values are passed which means no override should be done.

src/resource-graph/azext_resourcegraph/_params.py
fenwickt: Same as above - we need to know whether user overwrote this value or left it blank. The default value is set in customs.py

src/resource-graph/azext_resourcegraph/custom.py
fenwickt: Yes, it's page size. It's actually renamed to 'top' on a server side but I didn't want to introduce this breraking change.

This is the case where is user didn't pass neither skip token nor first explicitly, we're setting first to a default value from help message which is 100.

src/resource-graph/azext_resourcegraph/custom.py
fenwickt: We don't want to expose extra info to the user like facets, which is available in API but there is no support for it in CLI. This way is kind of trimming the response to the values that are within CLI scope.

fenwickt:
@kairu-ms sorry had a delay to fix your comments because was on-call. Updated the changes, please let me know if you have any questions.

kairu-ms:
@fenwickt Why your PR contains 2548 files changed?

kairu-ms:
You should use `git rebase master` instead of `git pull master`

kairu-ms:
BTW, you should create a feature branch instead of master branch on your repo.

kairu-ms:
This PR is invalid and I will close it. Please create a new PR. Thanks.

Issue #3151, opened 2021-03-11
How to specify "Latest" in --secret-version argument?


wasker:
## Describe the bug

When viewed in Azure Portal, custom domain functionality of AFD recommends using latest version of certificate to avoid service interruption:

![image](https://user-images.githubusercontent.com/1999365/110867031-f4d2aa80-827a-11eb-8b64-fb35bba784e2.png)

Apparently, it's impossible to specify "Latest" version from the CLI.

**Command Name**
```
az network front-door frontend-endpoint enable-https

Extension Name: front-door. Version: 1.0.11.
```

**Errors:**

## To Reproduce:

Steps to reproduce the behavior. Note that argument values have been redacted, as they may contain sensitive information.

Either

`az network front-door frontend-endpoint enable-https --resource-group myrg --front-door-name myfd --name domainname --certificate-source AzureKeyVault --vault-id /subscriptions/my-sub-id/resourceGroups/myrg/providers/Microsoft.KeyVault/vaults/mykv --secret-name cert--ssl-frontdoor`

or 

`az network front-door frontend-endpoint enable-https --resource-group myrg --front-door-name myfd --name domainname --certificate-source AzureKeyVault --vault-id /subscriptions/my-sub-id/resourceGroups/myrg/providers/Microsoft.KeyVault/vaults/mykv --secret-name cert--ssl-frontdoor --secret-version Latest`

or 

`az network front-door frontend-endpoint enable-https --resource-group myrg --front-door-name myfd --name domainname --certificate-source AzureKeyVault --vault-id /subscriptions/my-sub-id/resourceGroups/myrg/providers/Microsoft.KeyVault/vaults/mykv --secret-name cert--ssl-frontdoor --secret-version ""`

result in

`ERROR: usage error: --certificate-source AzureKeyVault --vault-id ID --secret-name NAME --secret-version VERSION`

## Expected Behavior

Use latest version of the certificate.

## Environment Summary
```
Windows-10-10.0.19041-SP0
Python 3.6.8
Installer: MSI

azure-cli 2.20.0

Extensions:
aks-preview 0.5.5
front-door 1.0.11
log-analytics 0.2.1
resource-graph 1.1.0

```

yonzhan:
front-door

t-bzhan:
Just update the front-door extension to 1.0.13, the secret version parameter is now optional and your certificate will be auto-rotated when a new version of secret generated if secret version is not specified.

mossywell:
[...]

PR #3218, opened 2021-04-01
[FrontDoor] az network front-door frontend-endpoint enable-https: allow secret-version to be optional to use Latest version

t-bzhan:
This PR will fix https://github.com/Azure/azure-cli-extensions/issues/3151
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
FrontDoor

src/front-door/azext_front_door/_help.py
```diff
@@ -148,6 +148,14 @@
     type: command
     short-summary: Disable HTTPS protocol for a custom domain.
 """
+
+helps['network front-door frontend-endpoint wait'] = """
```
kairu-ms, line 152: Please add test for this command

t-bzhan: Thanks, added.

src/front-door/azext_front_door/_params.py
```diff
@@ -65,7 +65,8 @@ def load_arguments(self, _):
     with self.argument_context('network front-door', arg_group='HTTPS') as c:
         c.argument('certificate_source', arg_type=get_enum_type(FrontDoorCertificateSource), help='Certificate source to enable HTTPS.')
         c.argument('secret_name', help='The name of the Key Vault secret representing the full certificate PFX')
-        c.argument('secret_version', help='The version of the Key Vault secret representing the full certificate PFX')
+        c.argument('secret_version', help='The version of the Key Vault secret representing the full certificate PFX, '
```
jsntcy, line 68: Are there any other versions supported besides `Latest`?

t-bzhan: Yes, I've added two test: One for Latest version and one for specific version.

For the time being, those tests will fail as our existing enable-https implementation has ignored the LROPoller which will cause failures in TearDown, I've marked them as skipped.

Another way to work around this is we could wait for the operations to finish using az network frontend-endpoint wait, but that only works in live_only model, it seems that VCR.py does not work well with the ignored underlying LRO poll, and each test will take more than 1 hour to finish.

jsntcy (approved):
:shipit:

PR #3259, opened 2021-04-13
[Blueprint] Close #2723: add export feature for blueprint and artifacts

### LINKED ISSUE #2723: Export Azure Blueprint with Artifacts
Hi @yungezz,

I am creating a separate issue for easier tracking. I see a new issue has been created and awaiting a merge. I would like to inquire about being able to export the blueprint and artifacts from the Azure CLI like you can with Export-AzBlueprintWithArtifact. We are trying to keep consistency and use the Azure CLI as much as possible, and would prefer not to rely on having to install PowerShell modules for other users if we can do everything through the Azure CLI. Thanks!

V/r

08nholloway:
---

### General Guidelines


fengzhou-msft:
Nice work! Please add a release note in `HISTORY.rst` and bump the minor version in `setup.py`.

src/blueprint/azext_blueprint/_params.py
```diff
@@ -85,6 +86,11 @@ def load_arguments(self, _):
     with self.argument_context('blueprint list') as c:
         pass
 
+    with self.argument_context('blueprint export') as c:
+        c.argument('blueprint_name', options_list=['--name', '-n'], help='Name of the blueprint definition to export.')
+        c.argument('output_path', type=file_type, help='The directory path for json definitions of the blueprint and artifacts. The blueprint definition file will be named blueprint.json. Artifacts json files will be in a subdirectory named artifacts.', completer=FilesCompleter())
+        c.argument('force', arg_type=get_three_state_flag(invert=True, return_label=True), options_list=['--force', '-f'], help='When set, if directory does not exist, it will be created. If the directory exists and has contents, they will be overwritten. If not set, user will be prompted for permission to proceed')
```
fengzhou-msft, line 92: `-Force` is the common name used for Powershell cmdlet option to skip user confirmation, here I think you can use the Azure CLI equivalent option name `--yes`. 
`get_three_state_flag` is usually used for resource properties that will be persisted in service side while `action='store_true'` is used to control command behavior (see more in [Registering Flags](https://github.com/Azure/azure-cli/blob/dev/doc/authoring_command_modules/authoring_commands.md#registering-flags)): 
```suggestion
        c.argument('force', action='store_true', options_list=['--yes', '-y'], help='When set, if directory does not exist, it will be created. If the directory exists and has contents, they will be overwritten. If not set, user will be prompted for permission to proceed')

```

src/blueprint/HISTORY.rst
```diff
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+0.3.1
```
fengzhou-msft, line 6: Please use `0.3.0`.

08nholloway: resolved: 97180a3b9f795a15281e4366432a0ad3cca1ceb4

src/blueprint/setup.py
```diff
@@ -16,7 +16,7 @@
 
 # TODO: Confirm this is the right version number you want and it matches your
 # HISTORY.rst entry.
-VERSION = '0.2.1'
+VERSION = '0.3.1'
```
fengzhou-msft, line 19:
```suggestion
VERSION = '0.3.0'

```

fengzhou-msft (approved):

PR #3267, opened 2021-04-14
[FrontDoor] onboard Track2 SDK

t-bzhan:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
FrontDoor

src/front-door/azext_front_door/custom.py
```diff
@@ -288,8 +319,9 @@ def create_fd_frontend_endpoints(cmd, resource_group_name, front_door_name, item
 
 
 def configure_fd_frontend_endpoint_disable_https(cmd, resource_group_name, front_door_name, item_name):
```
jsntcy, line 321: Better to pass `no_wait` as a parameter for method `configure_fd_frontend_endpoint_disable_https` and add `supports_no_wait=True` to g.custom_command(), so that CLI users can decide how to use the command (wait or no_wait).

You can refer to https://github.com/Azure/azure-cli-extensions/pull/3267/files#diff-b8eaf23fb42849ff925daab3d11895eb5d1aca345af11d422c41d2d6712554d1R142

t-bzhan: I've considered this. But that would cause breaking changes. The previous version's behavior is not waiting. And add `supports_no_wait=True` will make the default behavior to waiting, you have to add the --no-wait explicitly to achieve the same old effect. I guess we need to support some similar flag like "--wait" for those commands whose default behavior is not waiting.

jsntcy: ok, let's still keep the previous behavior.

jsntcy (approved):

jsntcy (approved):

PR #3346, opened 2021-05-07
Add blob prefix when synchronizing extensions

fengzhou-msft:
As teleport only supports transferring blobs in a specified container of a storage account, we need to use folders to organize different blobs like extensions, MSI packages and other files. This PR updates the tasks to sync extensions with the support of a blob prefix.

This PR also deletes some old versions of extensions in `index.json` as they are rarely used and would increase the time for the initial extension sync up. 

---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
@Juliehzl for awareness

scripts/ci/sync_extensions.py
```diff
@@ -15,14 +15,15 @@
 STORAGE_ACCOUNT = os.getenv('AZURE_EXTENSION_TARGET_STORAGE_ACCOUNT')
 STORAGE_CONTAINER = os.getenv('AZURE_EXTENSION_TARGET_STORAGE_CONTAINER')
 COMMIT_NUM = os.getenv('AZURE_EXTENSION_COMMIT_NUM') or 1
+BLOB_PREFIX = os.getenv('AZURE_EXTENSION_BLOB_PREFIX')
```
qwordy, line 18: What will happen if it is not set?

scripts/ci/sync_extensions.py
```diff
@@ -119,25 +126,33 @@ def main():
     target_index = DEFAULT_TARGET_INDEX_URL
     os.mkdir(os.path.join(temp_dir, 'target'))
     target_index_path = os.path.join(temp_dir, 'target', 'index.json')
-    download_file(target_index, target_index_path)
-
+    try:
+        download_file(target_index, target_index_path)
+    except Exception as ex:
```
qwordy, line 131: Too broad exception.

src/index.json
```diff
@@ -720,8 +720,8 @@
         ],
         "aks-preview": [
             {
-                "downloadUrl": "https://azurecliaks.blob.core.windows.net/azure-cli-extension/aks_preview-0.4.35-py2.py3-none-any.whl",
-                "filename": "aks_preview-0.4.35-py2.py3-none-any.whl",
+                "downloadUrl": "https://azcliprod.blob.core.windows.net/cli-extensions/aks_preview-0.5.9-py2.py3-none-any.whl",
```
qwordy, line 723: What if the users insist to use old version?

qwordy (approved):

scripts/ci/sync_extensions.py
fengzhou-msft: It will be `None` and the blob will be uploaded without a prefix in the container.

src/index.json
fengzhou-msft: Ask them to install with `--source` as the file is still there.

scripts/ci/sync_extensions.py
fengzhou-msft: As it's not customer facing, we can just fail and retry later.

PR #3392, opened 2021-05-18
[Connectedk8s] Adding troubleshoot command for ConnectedClusters

alphaWizard:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
Connectedk8s

src/connectedk8s/HISTORY.rst
```diff
@@ -7,6 +7,7 @@ Release History
 ++++++
 * Add compatible logic for the track 2 migration of resource dependence
```
alphaWizard, line 9: Have to bump version

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1421,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+                current_time = datetime.now(timezone.utc)
+                if cert_expirn_time != datetime.min and cert_expirn_time < current_time:
+                    tr_logger.error("MSI certificate on the cluster has expired.")
+                    logger.error("MSI certificate on the cluster has expired.")
+            else:
+                tr_logger.info("'managedIdentityCertificateExpirationTime' is still not populated in CC object")
+        except Exception as ex:
+            tr_logger.error("Error occured while checking if the MSI certificate has expired: {}".format(str(ex)), exc_info=True)
+
+        utils.try_upload_log_file(storage_account, sas_token, tr_logger)
```
alphaWizard, line 1496: To add this part - in progress.

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -389,3 +485,116 @@ def names(self, names):
...
+        else:
+            storage_account_id = storage_account
+
+        if is_valid_resource_id(storage_account_id):
+            try:
+                parsed_storage_account = parse_resource_id(storage_account_id)
+            except CloudError as ex:
+                raise ValidationError(ex.message)
+        else:
+            raise ValidationError("Invalid storage account id %s" % storage_account_id)
```
alphaWizard, line 578: TO-DO: Change to warning

src/connectedk8s/setup.py
```diff
@@ -37,6 +37,7 @@
 DEPENDENCIES = [
     'kubernetes==11.0.0',
     'pycryptodome==3.9.8',
+    'azure-storage-blob'
```
alphaWizard, line 40: TO-DO: pin it to v12.8.1

src/connectedk8s/azext_connectedk8s/_params.py
```diff
@@ -67,6 +67,15 @@ def load_arguments(self, _):
...
 
+    with self.argument_context('connectedk8s troubleshoot') as c:
+        c.argument('location', arg_type=get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
+        c.argument('cluster_name', options_list=['--name', '-n'], id_part='name', help='The name of the connected cluster.')
+        c.argument('kube_config', options_list=['--kube-config'], help='Path to the kube config file.')
+        c.argument('kube_context', options_list=['--kube-context'], help='Kubconfig context from current machine.')
+        c.argument('storage_account', options_list=['--storage-account'], help='Name or ID of the storage account to save the diagnostic information')
+        c.argument('sas_token', options_list=['--sas-token'], help='The SAS token with writable permission for the storage account.')
+        c.argument('output_file', options_list=['--output-file'], type=file_type, default=os.path.join(os.path.expanduser('~'), '.azure', 'az_connectedk8s_troubleshoot_output.tar.gz'), help="Output zipped file path for the logs collected during troubleshoot.")
```
alphaWizard, line 77: This should a tar file collecting all logs. Non-periscope + periscope.

yonzhan:
Connectedk8s

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -389,3 +490,359 @@ def names(self, names):
...
+
+    print(f'{colorama.Fore.GREEN}Your logs are being uploaded to storage account {format_bright(storage_account_name)}...')
+
+    print()
+    print(f'You can download Azure Storage Explorer here '
+          f'{format_hyperlink("https://azure.microsoft.com/en-us/features/storage-explorer/")}'
+          f' to check the logs by accessing the storage account {storage_account_name}.')
+    # f' to check the logs by adding the storage account using the following URL:')
+    # print(f'{format_hyperlink(log_storage_account_url)}')
+
```
alphaWizard, line 812: Also have to add the local archived log file - using kubectl cp

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+                    logger.error("MSI certificate on the cluster has expired.")
+            else:
+                tr_logger.info("'managedIdentityCertificateExpirationTime' is still not populated in CC object")
+        except Exception as ex:
+            tr_logger.error("Error occured while checking if the MSI certificate has expired: {}".format(str(ex)))
+
+        storage_account_name, sas_token, readonly_sas_token = utils.setup_validate_storage_account(cmd.cli_ctx, storage_account, sas_token, resource_group_name)
+        if storage_account_name:  # When validated the storage account
+            utils.try_upload_log_file(cluster_name, storage_account_name, sas_token, troubleshoot_log_path)
+            utils.try_archive_log_file(troubleshoot_log_path, output_file)
```
alphaWizard, line 1503: Common line in IF and ELSE, will move outside.

src/connectedk8s/azext_connectedk8s/_help.py
```diff
@@ -117,3 +117,24 @@
   - name: Disable multiple features.
     text: az connectedk8s disable-features -n clusterName -g resourceGroupName --features custom-locations azure-rbac
 """
+
+helps['connectedk8s troubleshoot'] = """
+  type: command
+  short-summary: Collects diagnose infomation and  gets logs on the connected cluster.
+  parameters:
```
akashkeshari, line 124: what about --resource-group and --name

src/connectedk8s/azext_connectedk8s/_params.py
```diff
@@ -69,6 +69,15 @@ def load_arguments(self, _):
         c.argument('features', features_types, options_list=['--features'], help='Space-separated list of features you want to disable.')
         c.argument('yes', options_list=['--yes', '-y'], help='Do not prompt for confirmation.', action='store_true')
 
+    with self.argument_context('connectedk8s troubleshoot') as c:
+        c.argument('location', arg_type=get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
```
akashkeshari, line 73: Why is 'location' argument reqd? Should not be necessary.

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -379,6 +399,87 @@ def is_guid(guid):
...
+    extensions = get_extensions()
+    if extensions:
+        for ext in extensions:
+            if ext.name == extension_name:
+                return ext.version or 'Unknown'
+
+    return 'NotFound'
+
+
+def check_connectivity(url='https://azure.microsoft.com', max_retries=5, timeout=1):
```
akashkeshari, line 438: Why are we checking connectivity to azure.microsoft.com.
Shouldn't this be azure resource manager endpoint(management.azure.com) ?
Also this value changes based on cloud the customer has set. We first need to check the current cloud and accordingly check connectivity to cloud specific endpoint

akashkeshari: Another point is that also reach out to the graph endpont(also cloud specific) for fetching the custom locations oid. Probably that connection can also be verified

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -379,6 +399,87 @@ def is_guid(guid):
...
+    import timeit
+    start = timeit.default_timer()
+    success = None
+    try:
+        with requests.Session() as s:
+            s.mount(url, requests.adapters.HTTPAdapter(max_retries=max_retries))
+            s.head(url, timeout=timeout)
+            success = True
+    except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as ex:
```
akashkeshari, line 447: If any other exception occurs, troubleshoot cmd will crash. Can we handle that too and give appropriate message

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+                 sas_token=None, output_file=os.path.join(os.path.expanduser('~'), '.azure', 'az_connectedk8s_troubleshoot_output.tar.gz')):
+    colorama.init()
+    print(f"{colorama.Fore.YELLOW}Troubleshooting the ConnectedCluster for possible issues...")
+    utils.check_connectivity()  # Checks internet connectivity
+    troubleshoot_log_path = os.path.join(os.path.expanduser('~'), '.azure', 'connected8s_troubleshoot.log')
+    utils.setup_logger('connectedk8s_troubleshoot', troubleshoot_log_path)
+    tr_logger = logging.getLogger('connectedk8s_troubleshoot')  # logger for troubleshooting, onto a log file
+
+    # Send cloud information to telemetry
+    send_cloud_telemetry(cmd)
```
akashkeshari, line 1433: Sending cloud telemetry in troubleshoot dosen't make sense. Instead we should collect cloud info in the troubleshooting logs

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+
+    # setting kubeconfig
+    kube_config = set_kube_config(kube_config)
+
+    # Loading the kubeconfig file in kubernetes client configuration
+    load_kube_config(kube_config, kube_context)
```
akashkeshari, line 1439: We should do only kubernetes related troubleshooting after this. Collection of connectedk8s extension version, checking provider registration, connectivity to dp endpoint etc. can be done before this.

yonzhan:
Connectedk8s

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+    configuration = kube_client.Configuration()
+    validate_release_namespace(client, cluster_name, resource_group_name, configuration, kube_config, kube_context)
```
akashkeshari, line 1441: Why are we doing this in connectedk8s troubleshoot?  We already execute this function in 'connectedk8s connect' or 'enable/disable features'. User will end up getting the same error message. I don't see any use case here, is there?

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -379,6 +399,87 @@ def is_guid(guid):
         return False
 
 
+def get_latest_extension_version(extension_name='connectedk8s'):
+    try:
+        import re
+        git_url = "https://raw.githubusercontent.com/Azure/azure-cli-extensions/master/src/{}/setup.py".format(extension_name)
```
akashkeshari, line 405: Version present in seup.py might not always be the latest released version. Latest released version is always found in index.json: https://raw.githubusercontent.com/Azure/azure-cli-extensions/main/src/index.json
Alternatively, you could also do what this cmd does: 'az extension list-versions -n connectedk8s'

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -379,6 +399,87 @@ def is_guid(guid):
...
+                    match = re.search(r'VERSION = \"(.*)\"$', txt)
+                    if match:
+                        return match.group(1)
+        return None
+    except Exception as ex:  # pylint: disable=broad-except
+        logger.info("Failed to get the latest version from '%s'. %s", git_url, str(ex))
+        return None
+
+
+def get_existing_extension_version(extension_name='connectedk8s'):
```
akashkeshari, line 427: Use try catch

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+    try:
+        latest_connectedk8s_version = utils.get_latest_extension_version()
+        local_connectedk8s_version = utils.get_existing_extension_version()
+        tr_logger.info("Latest available connectedk8s version: {}".format(latest_connectedk8s_version))
+        tr_logger.info("Local connectedk8s version: {}".format(local_connectedk8s_version))
+        from azure.cli.core import __version__ as azure_cli_core_version
+        tr_logger.info("azure-cli-core version installed locally: {}".format(azure_cli_core_version))
+        if latest_connectedk8s_version and local_connectedk8s_version != 'Unknown' and local_connectedk8s_version != 'NotFound':
+            if version.parse(local_connectedk8s_version) < version.parse(latest_connectedk8s_version):
+                print("You have an update pending. You can update the connectedk8s extension to latest v{} using 'az extension update -n connectedk8s'".format(latest_connectedk8s_version))
```
akashkeshari, line 1451: Would be kind of good to also add the min az core version reqd for updating to latest connectedk8s version

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
     return str.format(consts.Successfully_Disabled_Features, features, connected_cluster.name)
 
 
+def troubleshoot(cmd, client, resource_group_name, cluster_name, kube_config=None, kube_context=None, location=None, storage_account=None,
```
akashkeshari, line 1423: Is it okay to mandate the cluster_name and rg_name. A customer can run into onboarding issue even before the PUT cc call. For them, these paramters won't make any sense

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+            tr_logger.error("Couldn't check the existence of Connected cluster resource. Error: {}".format(str(ex)))
+
+        kapi_instance = kube_client.CoreV1Api(kube_client.ApiClient(configuration))
+        try:
+            pod_list = kapi_instance.list_namespaced_pod(consts.Arc_Namespace)
+            pods_count = 0
+            for pod in pod_list.items:
+                pods_count += 1
+                if pod.status.phase != 'Running':
+                    tr_logger.warning("Pod {} is in {} state. Reason: {}. Container statuses: {}".format(pod.metadata.name, pod.status.phase, pod.status.reason, pod.status.container_statuses))
```
akashkeshari, line 1473: Can we have one log file for each container in the pod and package all log files in the final zip? That would ease the debugging process for us. If we keep logging all troubleshooting info i a single log file, it will be very difficult to debug

akashkeshari: We should also capture container logs and not just container statuses. Error logs in the containers are more useful

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+
+            if pods_count == 0:
+                tr_logger.warning("No pods found in azure-arc namespace.")
```
akashkeshari, line 1476: Why don't we check the presence of 'azure-arc' ns/helm release first? Again, there can be scenarios where connect cmd fails even before onboarding (eg: connection to dp). In such cases, we don't need to collect k8s specific diagnostic logs and just mention that helm release 'azure-arc' is not present on the cluster

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -389,3 +490,359 @@ def names(self, names):
...
+
+def get_kubernetes_secret(api_instance, namespace, secret_name, custom_logger=None):
+    try:
+        return api_instance.read_namespaced_secret(secret_name, namespace)
+    except Exception as e:
+        handle_logging_error(custom_logger, "Error occurred when retrieving secret '{}': ".format(secret_name) + str(e))
+
+
+def handle_logging_error(custom_logger, error_string):
+    if custom_logger:
```
akashkeshari, line 503: Why if else here? Is there a scenario where custom_logger is not passed?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+
+        except Exception as ex:
+            tr_logger.error("Error occured while fetching pod's statues : {}".format(str(ex)))
+
+        cert_secret = utils.get_kubernetes_secret(kapi_instance, consts.Arc_Namespace, consts.AZURE_IDENTITY_CERTIFICATE_SECRET, custom_logger=tr_logger)
+        if (not cert_secret) or (not hasattr(cert_secret, 'data')) or (consts.AZURE_IDENTITY_CERTIFICATE_SECRET not in cert_secret.data):
```
akashkeshari, line 1482: Probably do the same check for AAD TOKEN SECRET too.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+                pass
+            tr_logger.error("Couldn't check the existence of Connected cluster resource. Error: {}".format(str(ex)))
+
+        kapi_instance = kube_client.CoreV1Api(kube_client.ApiClient(configuration))
+        try:
+            pod_list = kapi_instance.list_namespaced_pod(consts.Arc_Namespace)
+            pods_count = 0
+            for pod in pod_list.items:
+                pods_count += 1
+                if pod.status.phase != 'Running':
```
akashkeshari, line 1472: Pods status will show as running even if just the fluent-bit container is running in config-agent pod. In this case we might miss out on any issues with readiness probe

akashkeshari: I think we should always collect the arc-agent pod logs & status when the troubleshoot cmd is run and not keep it behind a condition like pod in 'running' state

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1420,97 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+
+        except Exception as ex:
+            tr_logger.error("Error occured while fetching pod's statues : {}".format(str(ex)))
+
+        cert_secret = utils.get_kubernetes_secret(kapi_instance, consts.Arc_Namespace, consts.AZURE_IDENTITY_CERTIFICATE_SECRET, custom_logger=tr_logger)
+        if (not cert_secret) or (not hasattr(cert_secret, 'data')) or (consts.AZURE_IDENTITY_CERTIFICATE_SECRET not in cert_secret.data):
+            tr_logger.error("{} secret is not present on the kubernetes cluster.".format(consts.AZURE_IDENTITY_CERTIFICATE_SECRET))
+            logger.error("{} secret is not present on the kubernetes cluster.".format(consts.AZURE_IDENTITY_CERTIFICATE_SECRET))
+
+        try:
```
akashkeshari, line 1486: Do we really need to do the exercise of determining if msi cert has expired in the code? I saw we already collect the cc get response in the logs above. We can directly confirm the cert expiry time from there

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -389,3 +490,359 @@ def names(self, names):
...
+            else:
+                diag["Status"] = f'{colorama.Fore.GREEN}{diag["Status"]}{colorama.Style.RESET_ALL}'
+
+    return diag_status
+
+
+def collect_periscope_logs(resource_group_name, name, storage_account_name=None, sas_token=None, readonly_sas_token=None, kube_context=None, kube_config=None):
+    colorama.init()
+
+    if not which('kubectl'):
```
akashkeshari, line 722: Why does this have a dependency on kubectl?

src/connectedk8s/azext_connectedk8s/custom.py
akashkeshari: Are we saying that customer's should run the troubleshoot cmd only when they see failures in 'helm install' step?

src/connectedk8s/azext_connectedk8s/_utils.py
alphaWizard: This is general internet connectivity.

src/connectedk8s/azext_connectedk8s/_utils.py
alphaWizard: Yes, this I also had in mind, Earlier Feng had told to use setup.py itself, but later I realized, the latest is there in index.json. Will update.

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: validate_release_namespace will do that validation.

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: container logs are part of periscope.

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: That validation is done internally by cli itself.

src/connectedk8s/azext_connectedk8s/_utils.py
alphaWizard: I think, this Sophie implemented this similar to what aks-periscope does. https://github.com/Azure/azure-cli-extensions/blob/a243ab0cdfc77570341063fa5aa3b7660e107e31/src/aks-preview/azext_aks_preview/custom.py#L1988

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: Makes sense, removing the extra log line from log file, but keeping it as a log error to the end user.

src/connectedk8s/azext_connectedk8s/_utils.py
alphaWizard: Yes, it can None too.

I am utilising the same function based on whether custom_logger passed is or not.  Otherwise, to use the default logger to debug.
This can thus be reused later.

Example: 
https://github.com/alphaWizard/azure-cli-extensions/blob/667907ec0a1b5f6ef6c95292df836c9c5b0e8056/src/connectedk8s/azext_connectedk8s/_utils.py#L493

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: If they face any issue onboarding, they will run this command, with the same -n and -g, In the troubleshoot, we are anyway doing the get on that, which will be null if onboarding to ARM itself didnt happen.

The troubleshoot has to be run when they face any issue.

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: Yes, these are in those commands too. 

This is just to complete the full circle of sanity checks in the troubleshoot also, so that these things can be verified, before running into periscope checks. 
Also, this answers your https://github.com/Azure/azure-cli-extensions/pull/3392#discussion_r669581316

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: Makes sense, will update.

src/connectedk8s/azext_connectedk8s/_utils.py
alphaWizard: This was same as per what azure-cli does. https://github.com/Azure/azure-cli/blob/b24817537d6f589439ac19178d0ab7672e420598/src/azure-cli-core/azure/cli/core/util.py#L786

But, yes, will add a generic exception beyond that too.

src/connectedk8s/azext_connectedk8s/_utils.py
alphaWizard: https://github.com/Azure/azure-cli/blob/b24817537d6f589439ac19178d0ab7672e420598/src/azure-cli-core/azure/cli/core/util.py#L786

src/connectedk8s/azext_connectedk8s/_params.py
alphaWizard: It's not required. We can remove, I had earlier added, thinking it might be required. Will remove.

src/connectedk8s/azext_connectedk8s/_help.py
alphaWizard: Its there  in -h of troubleshoot.

Resource Id Arguments
    --ids               : One or more resource IDs (space-delimited). It should be a complete
                          resource ID containing all information of 'Resource Id' arguments. You
                          should provide either --ids or other 'Resource Id' arguments.
    **--name -n           : The name of the connected cluster.
    --resource-group -g : Name of resource group. You can configure the default group using `az
                          configure --defaults group=<name>`.**
    --subscription      : Name or ID of subscription. You can configure the default subscription
                          using `az account set -s NAME_OR_ID`.

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: @sophsoph321 For visibility.

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: While one are you referring too:

PS C:\Users\arpgu\repos\cli_extensions\azure-cli-extensions\src\connectedk8s> kubectl get secrets -n azure-arc
_**NAME                                                                                      TYPE                                  DATA   AGE
azure-arc-connect-privatekey                                                              Opaque                                1      85m
azure-arc-kube-aad-proxy-sa-token-kc277                                                   kubernetes.io/service-account-token   3      85m
azure-arc-kube-aad-proxy-tls                                                              kubernetes.io/tls                     2      85m
azure-arc-operatorsa-token-2m8cf                                                          kubernetes.io/service-account-token   3      85m
azure-identity-certificate                                                                Opaque                                3      85m
connectproxy-agent-identity-request-token                                                 Opaque                                1      85m
default-token-mdl8q                                                                       kubernetes.io/service-account-token   3      85m
identity-request-2a051a512c1afcd426dd4090206c017a675c0f002bf329cc3165a7ba3abdcc97-token   Opaque                                1      85m**_

src/connectedk8s/azext_connectedk8s/_help.py
akashkeshari: i get that. just wondering why we defining two parameters specifically here. In params.py, we already write the help for all parameters which then come up in the -h message. Wanted to know the use case here

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -379,6 +399,85 @@ def is_guid(guid):
...
+    except Exception as ex:
+        logger.warning("Not able to reach azure management endpoints. Exception: " + str(ex))
+
+
+def check_provider_registrations(cli_ctx):
+    try:
+        rp_client = _resource_providers_client(cli_ctx)
+        cc_registration_state = rp_client.get(consts.Connected_Cluster_Provider_Namespace).registration_state
+        if cc_registration_state != "Registered":
+            logger.error("{} provider is not registered".format(consts.Connected_Cluster_Provider_Namespace))
```
akashkeshari, line 465: Will this get displayed to the user?

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -379,6 +399,85 @@ def is_guid(guid):
...
+        kc_registration_state = rp_client.get(consts.Kubernetes_Configuration_Provider_Namespace).registration_state
+        if kc_registration_state != "Registered":
+            logger.warning("{} provider is not registered".format(consts.Kubernetes_Configuration_Provider_Namespace))
```
akashkeshari, line 468: It would be better to fail here if the state in not 'Registered'. Instead of waiting for cc creation step and encountering failure there. Collect exception telemetry also and set user fault

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -379,6 +399,85 @@ def is_guid(guid):
...
+        get_subscription_client().get(subscription_id)
+    except Exception as ex:
+        logger.warning("Not able to reach azure management endpoints. Exception: " + str(ex))
+
+
+def check_provider_registrations(cli_ctx):
+    try:
+        rp_client = _resource_providers_client(cli_ctx)
+        cc_registration_state = rp_client.get(consts.Connected_Cluster_Provider_Namespace).registration_state
+        if cc_registration_state != "Registered":
```
akashkeshari, line 464: It would be better to fail here if the state in not 'Registered'. Instead of waiting for cc creation step and encountering failure there. Collect exception telemetry also and set user fault

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -118,11 +124,15 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, https_pr
                                 summary="Couldn't find any node on the kubernetes cluster with the architecture type 'amd64' and OS 'linux'")
         logger.warning("Please ensure that this Kubernetes cluster have any nodes with OS 'linux' and architecture 'amd64', for scheduling the Arc-Agents onto and connecting to Azure. Learn more at {}".format("https://aka.ms/ArcK8sSupportedOSArchitecture"))
 
+    crb_permission = utils.can_create_clusterrolebindings(configuration)
+    if not crb_permission:
+        logger.error("Kubectl credentials doesn't have permission to create clusterrolebindings on this kubernetes cluster.")
```
akashkeshari, line 129: Better to terminate with failure here rather than waiting for the failure at helm install step. Also collect exception telemetry and set user fault

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -666,6 +680,7 @@ def delete_connectedk8s(cmd, client, resource_group_name, cluster_name,
 
     # Deleting the azure-arc agents
     utils.delete_arc_agents(release_namespace, kube_config, kube_context, configuration)
+    utils.check_delete_job(configuration, consts.Arc_Namespace)  # Checks for the completion/absence of delete job
```
akashkeshari, line 683: We already wait for 'azure-arc' namespace termination in utils.delete_arc_agents. This check is irrelevant here

src/connectedk8s/azext_connectedk8s/_utils.py
akashkeshari: okay, but what value is it adding to connectedk8s troubleshoot command then? The customer will already get this error in connectedk8s connect

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1424,102 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+                 sas_token=None, output_file=os.path.join(os.path.expanduser('~'), '.azure', 'az_connectedk8s_troubleshoot_output.tar.gz')):
+    colorama.init()
+    print(f"{colorama.Fore.YELLOW}Troubleshooting the ConnectedCluster for possible issues...")
+    utils.check_connectivity()  # Checks internet connectivity
+    troubleshoot_log_path = os.path.join(os.path.expanduser('~'), '.azure', 'connected8s_troubleshoot.log')
+    utils.setup_logger('connectedk8s_troubleshoot', troubleshoot_log_path)
+    tr_logger = logging.getLogger('connectedk8s_troubleshoot')  # logger for troubleshooting, onto a log file
+
+    cloud_name = cmd.cli_ctx.cloud.name.upper()
+    # Setting cloud name to format that is understood by golang SDK.
```
akashkeshari, line 1437: Nit: Here we need not update cloudname as per golang convention

src/connectedk8s/azext_connectedk8s/custom.py
akashkeshari: okay

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1424,102 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+                    tr_logger.error("Connected cluster resource doesn't exist. " + str(ex))
+                    logger.error("The specified ConnectedCluster resource doesn't exist.")
+            except AttributeError:
+                pass
+            tr_logger.error("Couldn't check the existence of Connected cluster resource. Error: {}".format(str(ex)))
+
+        # setting kubeconfig
+        kube_config = set_kube_config(kube_config)
+        # Loading the kubeconfig file in kubernetes client configuration
+        load_kube_config(kube_config, kube_context)
```
akashkeshari, line 1471: Better to have these in try catch too. If failure occurs here due to an issue with python SDK, we might not be able to deploy periscope container

src/connectedk8s/azext_connectedk8s/custom.py
akashkeshari: I think one should be enough: identity-request-2a051a512c1afcd426dd4090206c017a675c0f002bf329cc3165a7ba3abdcc97-token

src/connectedk8s/azext_connectedk8s/custom.py
akashkeshari: should be fine i guess

src/connectedk8s/azext_connectedk8s/custom.py
akashkeshari: Just a point: If container logs are collected by periscope, then successfully executing periscope becomes very critical. (Most of the times, we ask for arc-agent pod logs from the customer for troubleshooting).

src/connectedk8s/azext_connectedk8s/_utils.py
akashkeshari: should be fine

src/connectedk8s/azext_connectedk8s/custom.py
akashkeshari: okay

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: This is different. If you remember, we get issues sometimes, where namespace is deleted, but delete job is stuck. 
Can you have a look at this function **_check_delete_job()_**: https://github.com/alphaWizard/azure-cli-extensions/blob/0b06cf230e5032b2b6eb9c34a90d0f3ae6f6ec00/src/connectedk8s/azext_connectedk8s/_utils.py#L524

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -389,3 +488,359 @@ def names(self, names):
...
+        if diag["Status"]:
+            if "Error:" in diag["Status"]:
+                diag["Status"] = f'{colorama.Fore.RED}{diag["Status"]}{colorama.Style.RESET_ALL}'
+            else:
+                diag["Status"] = f'{colorama.Fore.GREEN}{diag["Status"]}{colorama.Style.RESET_ALL}'
+
+    return diag_status
+
+
+def collect_periscope_logs(resource_group_name, name, storage_account_name=None, sas_token=None, readonly_sas_token=None, kube_context=None, kube_config=None):
```
akashkeshari, line 717: Will periscope also collect CPU/memory information from the nodes? This would be useful for cases where pods are stuck in 'Pending' due to not enough resources on k8s cluster

alphaWizard: @sophsoph321 Can you please answer this?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1409,6 +1424,102 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+                    logger.error("MSI certificate on the cluster has expired.")
+        except Exception as ex:
+            pass
+
+        storage_account_name, sas_token, readonly_sas_token = utils.setup_validate_storage_account(cmd.cli_ctx, storage_account, sas_token, resource_group_name)
+        if storage_account_name:  # When validated the storage account
+            utils.try_upload_log_file(cluster_name, storage_account_name, sas_token, troubleshoot_log_path)
+            utils.try_archive_log_file(troubleshoot_log_path, output_file)
+            # token_in_storage_account_url = readonly_sas_token if readonly_sas_token is not None else sas_token.
+            utils.collect_periscope_logs(resource_group_name, cluster_name, storage_account_name, sas_token, readonly_sas_token, kube_context, kube_config)
```
akashkeshari, line 1514: One imp concern I have regarding periscope:
For cases when the cluster is behind proxy, periscope will not be able to push pod logs to the storage which defeats the whole purpose of troubleshoot. Somehow, we should find a way to get pod logs on the local machine. Since we have seen many issues due to proxy in the past.

alphaWizard: That must be handled by the aks-periscope internally I guess, because aks-periscope when collecting logs for aks cluster behind the proxy, should have implemented that. 
@sophsoph321 Can you confirm back on this once?

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: This is some ID generated by identity operator pod IG - 2a051a512c1afcd426dd4090206c017a675c0f002bf329cc3165a7ba3abdcc9

Do you think we should add this?

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: That is true. Periscope checks are for the deeper level investigation into the issue. **As per design, Periscope is the integral part of this troubleshoot command.**

src/connectedk8s/azext_connectedk8s/_help.py
alphaWizard: Got the point, will remove.

src/connectedk8s/azext_connectedk8s/_utils.py
alphaWizard: This is in troubleshoot command, we can't do some of the things we will be doing in troubleshoot command. If general internet connectivity itself is broken, we are capturing that in logs.

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: We can discuss on call later, if required.

src/connectedk8s/azext_connectedk8s/_utils.py
sophsoph321: Periscope currently does not collect CPU/memory information from the nodes.

alphaWizard: I think, in pod status we are capturing, will reflect any pod scheduling issue.
 Additionally, we can add the node CPU/memory from cli itself, by extracting that from **_describe node_** information. Will add that.

src/connectedk8s/azext_connectedk8s/custom.py
sophsoph321: Yes, part of the implementation for getting the pod logs on the local machine must be handled by aks-periscope. Currently, this work will not be in scope for the first release of this troubleshoot cli command, but will be done in the next release.

src/connectedk8s/azext_connectedk8s/_utils.py
alphaWizard: Added linux/amd64 nodes's capacity: CPU and memory

yonzhan:
Connectedk8s

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -817,6 +816,26 @@ def collect_periscope_logs(resource_group_name, name, storage_account_name=None,
...
+
+        subprocess_cmd = kubectl_prior + ["delete", "ClusterRole", "aks-periscope-role", "--ignore-not-found"]
+        subprocess.call(subprocess_cmd, stderr=subprocess.STDOUT)
+
+        subprocess_cmd = kubectl_prior + ["delete", "--all", "apd", "-n", "aks-periscope", "--ignore-not-found"]
+        subprocess.call(subprocess_cmd, stderr=subprocess.STDOUT)
+
+        subprocess_cmd = kubectl_prior + ["delete", "CustomResourceDefinition", "diagnostics.aks-periscope.azure.github.com", "--ignore-not-found"]
+        subprocess.call(subprocess_cmd, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as err:
```
alphaWizard, line 836: lets's add a broader generic Exception below, just for catching any unexpected exception.

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -389,3 +490,378 @@ def names(self, names):
...
+    # print(f'{format_hyperlink(log_storage_account_url)}')
+
+    print()
+    if not prompt_y_n('Do you want to see analysis results now?', default="n"):
+        print(f"You can rerun 'az connectedk8s troubleshoot -g {resource_group_name} -n {name}' "
+              f"anytime to check the analysis results.")
+    else:
+        display_diagnostics_report(kubectl_prior)
+
+    try:
```
alphaWizard, line 819: If this is reused, let's create a function here in utils itself.

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -389,3 +490,378 @@ def names(self, names):
...
+        subprocess_cmd = kubectl_prior + ["delete", "ClusterRole", "aks-periscope-role", "--ignore-not-found"]
+        subprocess.call(subprocess_cmd, stderr=subprocess.STDOUT)
+
+        subprocess_cmd = kubectl_prior + ["delete", "--all", "apd", "-n", "aks-periscope", "--ignore-not-found"]
+        subprocess.call(subprocess_cmd, stderr=subprocess.STDOUT)
+
+        subprocess_cmd = kubectl_prior + ["delete", "CustomResourceDefinition", "diagnostics.aks-periscope.azure.github.com", "--ignore-not-found"]
+        subprocess.call(subprocess_cmd, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as err:
+        raise CLIInternalError(err.output)
```
alphaWizard, line 837: Don't raise **CLIInternalError** here(CLI will exit from here itself in this case), rather just use Fplain Exception, so that this can be caught by the outer try-catch.

and Maybe add some descriptive log instead on just err.output

something like - **_raise Exception("Error occurred while deleting the aks-periscope logs. Error: {}".format(err.output))_**

src/connectedk8s/azext_connectedk8s/custom.py
alphaWizard: Link to job yaml definition: https://github.com/Azure/ClusterConfigurationAgent/blob/4aad9e9cf04897ffb4f42b0dd5b620b0ac07f83d/charts/azure-arc-k8sagents/templates/hooks.yaml#L14

akashkeshari: The delete job pod runs in the azure-arc namespace itself. So, the 'azure-arc' ns should not go away until the pod has deleted the CRDs. Did you see otherwise on some setup?

src/connectedk8s/azext_connectedk8s/_help.py
```diff
@@ -123,9 +123,9 @@
   short-summary: Collects diagnose infomation and  gets logs on the connected cluster.
   examples:
       - name: using storage account name and a shared access signature token with write permission
-        text: az connectedk8s troubleshoot -g MyResourceGroup -n ConnectedCluster --storage-account MyStorageAccount --sas-token "MySasToken"
+        text: az connectedk8s troubleshoot -g MyResourceGroup -n ConnectedCluster --storage-account MyStorageAccount --sas-token "MySasToken" --container-name "mycontainername"
```
alphaWizard, line 130: Probably, keep storage-container-name

otherwise, it creates a confusion with container on k8s cluster.

src/connectedk8s/azext_connectedk8s/_params.py
```diff
@@ -75,6 +75,7 @@ def load_arguments(self, _):
         c.argument('kube_context', options_list=['--kube-context'], arg_group='Target Kubernetes Cluster', help='Kubconfig context from current machine.')
         c.argument('storage_account', options_list=['--storage-account'], arg_group='Storage Account', help='Name or ID of the storage account to save the diagnostic information')
         c.argument('sas_token', options_list=['--sas-token'], arg_group='Storage Account', help='The SAS token with writable permission for the storage account.')
+        c.argument('container_name', options_list=['--container-name'], help = 'The name of the storage container.')
```
alphaWizard: make it a part of **_arg_group='Storage Account'_**

As in line 77.

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -538,13 +538,12 @@ def check_delete_job(configuration, namespace):
         logger.debug("Error occurred while retrieving status of the delete job: {}".format(str(e)))
 
 
-def try_upload_log_file(cluster_name, storage_account_name, storage_token, log_file_path):
+def try_upload_log_file(storage_account_name, storage_token, container_name, log_file_path):
```
alphaWizard, line 550: Lets not make container name a required parameter.

Probably keep some default name as well, like "connectedk8s_troubleshoot_logs" or something

src/connectedk8s/azext_connectedk8s/_params.py
alphaWizard: Lets not make container name a required parameter.

Probably keep some default name as well, like "connectedk8s_troubleshoot_logs" or something

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -431,3 +530,339 @@ def can_create_clusterrolebindings(configuration):
...
+                                              (base64.b64encode(bytes("?" + sas_token, 'ascii'))).decode('ascii'))
+    deployment_yaml = deployment_yaml.replace("aksrepos.azurecr.io/staging/aks-periscope:v0.4", "aksrepos.azurecr.io/staging/aks-periscope:0.5")
+    container_logs = "azure-arc"
+    kube_objects = "azure-arc/pod azure-arc/service"
+    yaml_lines = deployment_yaml.splitlines()
+    for index, line in enumerate(yaml_lines):
+        if "DIAGNOSTIC_CONTAINERLOGS_LIST" in line:
+            yaml_lines[index] = line + ' ' + container_logs
+        if "AZURE_BLOB_SAS_KEY" in line:
+            yaml_lines[index] = line + '\n' + '  AZURE_BLOB_CONTAINER_NAME: ' + base64.b64encode(bytes(container_name, 'ascii')).decode('ascii')
```
alphaWizard, line 770: If custom container name is not provided by user, then, will it face issue in decoding?

yonzhan:
Connectedk8s

yonzhan:
Connectedk8s

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1442,6 +1446,129 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
     return str.format(consts.Successfully_Disabled_Features, features, connected_cluster.name)
 
 
+def troubleshoot(cmd, client, resource_group_name, cluster_name, storage_account, sas_token=None, container_name="connectedk8stroubleshoot", kube_config=None, kube_context=None,
+                 output_file=os.path.join(os.path.expanduser('~'), '.azure', 'az_connectedk8s_troubleshoot_output.tar.gz')):
+    colorama.init()
+    print(f"{colorama.Fore.YELLOW}Troubleshooting the ConnectedCluster for possible issues...")
```
mallika-trivedi, line 1452: I'm good with Shashank's suggestion here to make it:
**"Troubleshooting issues with your Azure Arc-enabled Kubernetes cluster..."**

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1442,6 +1446,129 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+    try:
+        latest_connectedk8s_version = utils.get_latest_extension_version(tr_logger)
+        local_connectedk8s_version = utils.get_existing_extension_version()
+        tr_logger.info("Latest available connectedk8s version: {}".format(latest_connectedk8s_version))
+        tr_logger.info("Local connectedk8s version: {}".format(local_connectedk8s_version))
+        from azure.cli.core import __version__ as azure_cli_core_version
+        tr_logger.info("azure-cli-core version installed locally: {}".format(azure_cli_core_version))
+        if latest_connectedk8s_version and local_connectedk8s_version != 'Unknown' and local_connectedk8s_version != 'NotFound':
+            if version.parse(local_connectedk8s_version) < version.parse(latest_connectedk8s_version):
+                print("You have an update pending. You can update the connectedk8s extension to latest v{} using 'az extension update -n connectedk8s'".format(latest_connectedk8s_version))
```
mallika-trivedi, line 1470: Let's change this to: **"Your connectedk8s CLI extension is not on the latest version. We recommend that you update this to the latest version v{} using the command 'az extension update -n connectedk8s'".**

Color: Yellow

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1442,6 +1446,129 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+            required_node_counts = 0
+            for item in api_response.items:
+                node_arch = item.metadata.labels.get("kubernetes.io/arch")
+                node_os = item.metadata.labels.get("kubernetes.io/os")
+                if node_arch == "amd64" and node_os == "linux":
+                    tr_logger.info("Node {} has capacity: CPU: {}, and Memory: {} ".format(required_node_counts + 1, item.status.capacity.get("cpu"), item.status.capacity.get("memory")))
+                    required_node_counts += 1
+            if required_node_counts == 0:
+                tr_logger.error("There are no linux/amd64 nodes available on the kubernetes cluster.")
+                logger.error("This kubernetes cluster doesn't have any nodes with OS 'linux' and architecture 'amd64', for scheduling the Arc-Agents onto and connecting to Azure. Learn more at {}".format("https://aka.ms/ArcK8sSupportedOSArchitecture"))
```
mallika-trivedi, line 1520: Can you clarify if this logger is just to the storage account or also to the CLI output? 

CLI should read:
****"The cluster does not have any nodes with****...rest is the same as existing string".****

Color: Red

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1442,6 +1446,129 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+            pod_list = kapi_instance.list_namespaced_pod(consts.Arc_Namespace)
+            pods_count = 0
+            pods_pending = 0
+            for pod in pod_list.items:
+                pods_count += 1
+                tr_logger.debug("Pod {} is in {} state. Reason: {}. Container statuses: {}".format(pod.metadata.name, pod.status.phase, pod.status.reason, pod.status.container_statuses))
+                if pod.status.phase == "Pending":
+                    pods_pending += 1
+            if pods_pending > 0:
+                logger.warning("You have pods stuck in pending state. A possible cause is insuffient resources on the Kubernetes cluster you are trying to onboard to azure-arc. Please run kubectl get pods -n azure-arc to ensure all pods are in running state.")
```
mallika-trivedi, line 1534: CLI output should read:

**"Some pods in the azure-arc namespace are stuck in a pending state. Insufficient resources may sometimes cause this. Please run kubectl get pods -n azure-arc to ensure your pods are in a running state."**

Color: Red

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1442,6 +1446,129 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+            if pods_count == 0:
+                tr_logger.warning("No pods found in azure-arc namespace.")
+
+        except Exception as ex:
+            tr_logger.error("Error occured while fetching pod's statues : {}".format(str(ex)))
+
+        cert_secret = utils.get_kubernetes_secret(kapi_instance, consts.Arc_Namespace, consts.AZURE_IDENTITY_CERTIFICATE_SECRET, custom_logger=tr_logger)
+        if (not cert_secret) or (not hasattr(cert_secret, 'data')) or (consts.AZURE_IDENTITY_CERTIFICATE_SECRET not in cert_secret.data):
+            tr_logger.error("{} secret is not present on the kubernetes cluster.".format(consts.AZURE_IDENTITY_CERTIFICATE_SECRET))
+            logger.error("{} secret is not present on the kubernetes cluster.".format(consts.AZURE_IDENTITY_CERTIFICATE_SECRET))
```
mallika-trivedi, line 1544: **CLI output should read same as on line 1544. except just say "cluster" instead of kubernetes cluster to keep it consistent with other strings** 

@shashankbarsin can you help suggest what the customer can do if this secret is not present on the cluster. Is there a particular step they may have done incorrectly? It should be created by connect agent right?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1442,6 +1446,129 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+
+        try:
+            current_time = datetime.now(timezone.utc)
+            cert_expirn_time = connected_cluster.managed_identity_certificate_expiration_time
+            if cert_expirn_time != datetime.min and cert_expirn_time < current_time:
+                logger.error("MSI certificate on the cluster has expired.")
```
mallika-trivedi, line 1550: CLI output:

"The *whatever name is used in the azure_identity_certificate_secret*on the cluster has expired. This should renew automatically when the cluster has internet connectivity. Please check your internet connectivity." 

Instead of MSI please keep the naming consistent between this and the previous secret not present output. The customer doesn't necessarily need to know about the secret vs. certificate. 

Color: Red

@shashankbarsin can you verify the above.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1442,6 +1446,129 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+    tr_logger.info(str.format("Cloud Name: {}", cloud_name))
+
+    try:
+        latest_connectedk8s_version = utils.get_latest_extension_version(tr_logger)
+        local_connectedk8s_version = utils.get_existing_extension_version()
+        tr_logger.info("Latest available connectedk8s version: {}".format(latest_connectedk8s_version))
+        tr_logger.info("Local connectedk8s version: {}".format(local_connectedk8s_version))
+        from azure.cli.core import __version__ as azure_cli_core_version
+        tr_logger.info("azure-cli-core version installed locally: {}".format(azure_cli_core_version))
+        if latest_connectedk8s_version and local_connectedk8s_version != 'Unknown' and local_connectedk8s_version != 'NotFound':
```
mallika-trivedi, line 1468: I don't see the check for the agent version. Can you point me to where we're ensuring it's n-2.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1442,6 +1446,129 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
...
+                logger.error("MSI certificate on the cluster has expired.")
+        except ValueError as ex:
+            tr_logger.error(str(ex))
+        except Exception as ex:
+            tr_logger.error("Error while checking MSI cert expiration time: {}".format(str(ex)))
+
+        storage_account_name, sas_token, readonly_sas_token = utils.setup_validate_storage_account(cmd.cli_ctx, storage_account, sas_token, resource_group_name)
+        container_name = container_name + "-" + time.strftime("%Y%m%d-%H%M%S")
+        if storage_account_name:  # When validated the storage account
+            print(f"Logs will be uploaded to the following storage account: {storage_account_name} in container {container_name}")
```
mallika-trivedi, line 1559: CLI output:
**"The troubleshooting logs will be uploaded to the following storage account..."**

mallika-trivedi (commented):
@sirireddy12 I left some comments on phrases we should change. Three overall comments:
1. I didn't see the agent version check. We can sync on this one and what the wording should say here.
2. Not sure if custom_logger, trlogger are going to be printed to CLI output the same as the print statements. The commends I've left are all for CLI output but we definitely want them logged as well so that CSS can take a look at those files but I'm less concerned about the wording for that at least for MVP. Let's just make sure there is some parity between what is shown in CLI and logged in storage account.
3. I want to say something like "All checks were successful" if we didn't have anything to report but let's see how the Periscope checks go and maybe we want to print one overall successful line. I'll get back to you on this one.
3.

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -433,6 +532,342 @@ def can_create_clusterrolebindings(configuration):
...
+        logger.debug(node_line)
+        if columns[1] != "Ready":
+            logger.warning("Node %s is not Ready. Current state is: %s.", columns[0], columns[1])
+        else:
+            ready_nodes[columns[0]] = False
+
+    logger.debug('There are %s ready nodes in the cluster', str(len(ready_nodes)))
+
+    if not ready_nodes:
+        logger.warning('No nodes are ready in the current cluster. Diagnostics info might not be available.')
```
sirireddy12, line 639: @sophsoph321 , Hey Sophie, these nodes which we're talking about here are the persicope nodes right? Correct me if I'm wrong.

If yes, then should we really be telling the user in CLI output that "no nodes are ready"? wouldn't the user confuse it with the nodes of his own k8s cluster?

@mallika-trivedi / @shashankbarsin for suggestions

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -433,6 +532,342 @@ def can_create_clusterrolebindings(configuration):
...
+        subprocess_cmd,
+        universal_newlines=True)
+    logger.debug(nodes)
+    node_lines = nodes.splitlines()
+    ready_nodes = {}
+    for node_line in node_lines:
+        columns = node_line.split()
+        logger.debug(node_line)
+        if columns[1] != "Ready":
+            logger.warning("Node %s is not Ready. Current state is: %s.", columns[0], columns[1])
```
sirireddy12, line 632: @sophsoph321 Hey sophie, what exactly are these nodes that we're checking for? Are these the nodes which are already present as part of user's k8s cluster or is it something that periscope is adding?

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -433,6 +532,342 @@ def can_create_clusterrolebindings(configuration):
...
+            for node_name in ready_nodes:
+                if ready_nodes[node_name]:
+                    continue
+                apdName = "aks-periscope-diagnostic-" + node_name
+                try:
+                    subprocess_cmd = kubectl_prior + ["get", "apd", apdName, "-n", "aks-periscope", "-o=jsonpath={.spec.networkconfig}"]
+                    network_config = subprocess.check_output(
+                        subprocess_cmd,
+                        universal_newlines=True)
+                    logger.debug('Dns status for node %s is %s', node_name, network_config)
```
sirireddy12, line 677: @sophsoph321 should we also add error in red when we find out that DNS is not working on user's cluster?

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -433,6 +532,342 @@ def can_create_clusterrolebindings(configuration):
...
+                    raise CLIInternalError(err.output)
+
+    print()
+    if network_config_array:
+        print("Below are the network configuration for each node: ")
+        print()
+        print(tabulate(network_config_array, headers="keys", tablefmt='simple'))
+        print()
+    else:
+        logger.warning("Could not get network config. "
```
sirireddy12, line 704: Let's enhance this part to make it clear to the user that there is an issue with network outbound connecitivity on his/her cluster

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -433,6 +532,342 @@ def can_create_clusterrolebindings(configuration):
         return "Unknown"
 
 
+def check_delete_job(configuration, namespace):
```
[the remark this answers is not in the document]
sirireddy12: removed

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -386,6 +405,72 @@ def is_guid(guid):
...
+        success = False
+    except Exception as ex:  # pylint: disable=broad-except
+        logger.error("Failed to check Internet connectivity. Error: %s", str(ex))
+        success = False
+    stop = timeit.default_timer()
+    logger.debug('Connectivity check: %s sec', stop - start)
+    return success
+
+
+def validate_azure_management_reachability(subscription_id):
```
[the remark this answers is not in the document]
sirireddy12: removed it

src/connectedk8s/azext_connectedk8s/_utils.py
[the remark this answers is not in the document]
sirireddy12: removed it

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: I realized the agent version check was not implemented . Let me reconfirm offline as to why it wasn't

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: done

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: done

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: Yes it's the output to CLI as well. Whatever is coming with "logger" is the output to cli and whatever is coming with "tr_logger" is the output to the log file

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: done

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: done

src/connectedk8s/azext_connectedk8s/_utils.py
sophsoph321: @sirireddy12 These are nodes which are already present as part of user's k8s cluster.

src/connectedk8s/azext_connectedk8s/_utils.py
sophsoph321: Sure, @mallika-trivedi to suggest here.

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -433,13 +525,360 @@ def can_create_clusterrolebindings(configuration):
...
+
+            apd_lines = apd.splitlines()
+            if apd_lines and 'No resources found' in apd_lines[0]:
+                apd_lines.pop(0)
+
+            print("Got {} diagnostic results for {} ready nodes{}\r".format(len(apd_lines),
+                                                                            len(ready_nodes),
+                                                                            '.' * retry), end='')
+            if len(apd_lines) < len(ready_nodes):
+                logger.warning("Warning: There are %s apd resources, but there are %s nodes running on the cluster."
```
sirireddy12, line 658: User might not understand what an "apd" resource is right? @sophsoph321 Should we mention diagnostic resource instead ?

@shashankbarsin / @mallika-trivedi to suggest too

yonzhan:
Connectedk8s

PR #3398, opened 2021-05-19
Add support for user agent string in calls to AQ via CLI

ricardo-espinoza:
As part of this change we are:
- Defining a user-agent-like string to represent the client used to submit jobs via the CLI.
- Append this to the user agent of the Control Plane and Data Plane clients.
- Set this as an environment variable for child processes to read it, as in the case of invoked standalone executables.

### DIFF
--- a/src/quantum/azext_quantum/__init__.py
+++ b/src/quantum/azext_quantum/__init__.py
@@ -8,6 +8,12 @@ from azure.cli.core import AzCommandsLoader
 import azext_quantum._help  # pylint: disable=unused-import
 
 
+# This is the version reported by the CLI to the service when submitting requests.
+# This should be in sync with the extension version in 'setup.py', unless we need to
+# submit using a different version.
+CLI_REPORTED_VERSION = "0.4.0"
+
+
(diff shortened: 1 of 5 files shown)


yonzhan:
quantum

src/quantum/azext_quantum/__init__.py
```diff
@@ -8,6 +8,12 @@
 import azext_quantum._help  # pylint: disable=unused-import
 
 
+# This is the version reported by the CLI to the service when submitting requests.
+# This should be in sync with the extension version in 'setup.py', unless we need to
+# submit using a different version.
+CLI_REPORTED_VERSION = "0.5.0"
```
qwordy, line 14: Can we use one variable? `VERSION` is also defined in `setup.py`.

qwordy (approved):

src/quantum/azext_quantum/__init__.py
ricardo-espinoza: I agree we should consolidate, for now, this is a way to have the version in development in `master` reporting a certain value without triggering the release to happen.

PR #3486, opened 2021-06-10
Add support for 0.5 core / 512 Mi resource requests for spring-cloud deployments

allxiao:
---

### General Guidelines

    (for the newly added code)

For new extensions:

### About Extension Publish


yonzhan:
spring cloud

src/spring-cloud/azext_spring_cloud/_params.py
```diff
@@ -79,10 +79,10 @@ def load_arguments(self, _):
...
         c.argument('assign_identity', arg_type=get_three_state_flag(),
                    help='If true, assign managed service identity.')
-        c.argument('cpu', type=int, default=1,
-                   help='Number of virtual cpu cores per instance.')
-        c.argument('memory', type=int, default=1,
-                   help='Number of GB of memory per instance.')
+        c.argument('cpu', type=str, default="1",
+                   help='CPU resource quantity. Should be 500m or number of CPU cores.')
+        c.argument('memory', type=str, default="1Gi",
+                   help='Memory resource quantity. Should be 512Mi or #Gi, e.g., 1Gi, 3Gi.')
```
zhoxing-ms, line 85: May I ask this is a breaking change for users, right?

src/spring-cloud/azext_spring_cloud/_params.py
```diff
@@ -140,8 +140,8 @@ def prepare_logs_argument(c):
             c.argument('env', env_type)
 
     with self.argument_context('spring-cloud app scale') as c:
-        c.argument('cpu', type=int, help='Number of virtual cpu cores per instance.')
-        c.argument('memory', type=int, help='Number of GB of memory per instance.')
+        c.argument('cpu', type=str, help='CPU resource quantity. Should be 500m or number of CPU cores.')
+        c.argument('memory', type=str, help='Memory resource quantity. Should be 512Mi or #Gi, e.g., 1Gi, 3Gi.')
```
zhoxing-ms, line 144: Same as above

src/spring-cloud/azext_spring_cloud/_params.py
```diff
@@ -160,8 +160,8 @@ def prepare_logs_argument(c):
     with self.argument_context('spring-cloud app deployment create') as c:
         c.argument('skip_clone_settings', help='Create staging deployment will automatically copy settings from production deployment.',
                    action='store_true')
-        c.argument('cpu', type=int, help='Number of virtual cpu cores per instance.')
-        c.argument('memory', type=int, help='Number of GB of memory per instance.')
+        c.argument('cpu', type=str, help='CPU resource quantity. Should be 500m or number of CPU cores.')
+        c.argument('memory', type=str, help='Memory resource quantity. Should be 512Mi or #Gi, e.g., 1Gi, 3Gi.')
```
zhoxing-ms, line 164: Same as above

src/spring-cloud/azext_spring_cloud/_resource_quantity.py
```diff
@@ -0,0 +1,34 @@
...
+from knack.log import get_logger
+
+logger = get_logger(__name__)
+
+def validate_cpu(cpu):
+    if cpu is None:
+        return None
+
+    if not re.match(r"^\d+m?$", cpu):
+        raise CLIError("CPU quantity should be millis (500m) or integer (1, 2, ...)")
```
zhoxing-ms, line 12: Could you add some explanation for the function of the regular expression?

src/spring-cloud/azext_spring_cloud/_resource_quantity.py
```diff
@@ -0,0 +1,34 @@
...
+    try:
+        # For backward compatibility, to convert integer value to value with Gi unit
+        int(memory)
+        logger.warning("Memory quantity [--memory] should be specified with unit, such as 512Mi, 1Gi. "
+                       "Support for integer quantity will be dropped in future release.")
+        unified = memory + "Gi"
+    except ValueError:
+        pass
+
+    if not re.match(r"^\d+[MG]i$", unified):
```
zhoxing-ms, line 31: Could you add some explanation for the function of the regular expression?

src/spring-cloud/azext_spring_cloud/_resource_quantity.py
```diff
@@ -0,0 +1,34 @@
...
+        raise CLIError("Memory quantity should be integer followed by unit (Mi/Gi)")
```
zhoxing-ms, line 32: Please use the specific error type `InvalidArgumentValueError` instead of `CLIError`

src/spring-cloud/azext_spring_cloud/_params.py
allxiao: The shell environment does not distinguish int or str arguments. The following commands works both in the previous version and the version being reviewed:

```
az spring-cloud app update -n test-app --cpu 1 --memory 1
```

The update should be a super set of the existing functionality. Is this still considered as breaking change?

src/spring-cloud/azext_spring_cloud/_resource_quantity.py
allxiao: The current released version only supports integer quantity ([1, 8]). The change being reviewed introduced the notation `###m` for millis, and the only supported fractional amount is `500m`. However, the user is also allowed to specify `1000m` for `1`.

The regex here is to validate the input from the client side, with some digits followed by optional `m`. I added some comments in the updated code.

src/spring-cloud/azext_spring_cloud/_resource_quantity.py
allxiao: We allow the user to specify the memory quantity with either `###Mi` or `###Gi`, for example, `1024Mi`, `3Gi`. The code above this line is to add backward compatibility for legacy input, which uses integer for `Gi` values.

The regular expression here add some client side validation for the user input, to restrict it to be some digits followed by `Mi` or `Gi`. (Note that integer input will be converted to `###Gi` before this validation).

src/spring-cloud/azext_spring_cloud/_resource_quantity.py
allxiao: updated.

src/spring-cloud/azext_spring_cloud/_params.py
zhoxing-ms: OK, I think it should be OK~

zhoxing-ms (approved):

PR #3500, opened 2021-06-11
David/aks monitoring aad auth

daweim0:
---

### General Guidelines

For new extensions:

This PR adds support for enabling monitoring on AKS clusters with msi auth (instead of using a log analytics workspace's shared key). The feature is entering private preview soon. 

I added unit tests, but commented most them out. They require feature flags to be set on a subscription-by-subscription basis, so they will fail in a CI pipeline. They are still useful during development though, so I did not delete them before pushing.

Just general info, the long term plan is to move most of the functionality added here to the AKS resource provider. That can not happen yet though, so for now the functionality must live in the cli.


yonzhan:
aks

ganga1980 (commented):
initial feedback

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -91,6 +91,7 @@ def load_arguments(self, _):
         c.argument('pod_subnet_id', type=str, validator=validate_pod_subnet_id)
         c.argument('ppg')
         c.argument('workspace_resource_id')
+        c.argument('enable_msi_auth_for_monitoring', arg_type=get_three_state_flag(), is_preview=True)
```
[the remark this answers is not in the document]
daweim0: The three states are --enable-msi-auth-for-monitoring false, --enable-msi-auth-for-monitoring true, or --enable-msi-auth-for-monitoring

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -2687,104 +2701,179 @@ def _ensure_container_insights_for_monitoring(cmd, addon):
...
+            json_response = json.loads(r.text)
+            for resource in json_response["resourceTypes"]:
+                region_ids = map(lambda x: region_names_to_id[x], resource["locations"])  # map is lazy, so doing this for every region isn't slow
+                if resource["resourceType"] == "dataCollectionRules" and location not in region_ids:
+                    raise CLIError(f'Data Collection Rules are not enabled for LA workspace region {location}')
+                elif resource["resourceType"] == "dataCollectionRuleAssociations" and cluster_region not in region_ids:
+                    raise CLIError(f'Data Collection Rule Associations are not enabled for cluster region {location}')
+
+            # create the DCR
+            dcr_creation_body = json.dumps({"location": location,
```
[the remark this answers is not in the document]
daweim0: If the DCR already exists then re-creating it is a no-op.

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1460,14 +1463,21 @@ def aks_create(cmd,     # pylint: disable=too-many-locals,too-many-statements,to
...
             if 'not found in Active Directory tenant' in ex.message:
                 time.sleep(3)
             else:
                 raise ex
-    raise retry_exception
+    else:
+        raise retry_exception
+
+    # Create the DCR Association here, it must be done after the cluster is created.
+    if monitoring and enable_msi_auth_for_monitoring:
```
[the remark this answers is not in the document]
daweim0: There is now a wait for the cluster to be created

daweim0:
One thing I'm wondering about: The arguments --workspace-resource-id is registered for `az aks create` ([here](https://github.com/Azure/azure-cli-extensions/blob/ff20effe1c4aca1dc71bf57251685e43a3f0a646/src/aks-preview/azext_aks_preview/_params.py#L93)) but not `az aks enable-addons` ([here](https://github.com/Azure/azure-cli-extensions/blob/ff20effe1c4aca1dc71bf57251685e43a3f0a646/src/aks-preview/azext_aks_preview/_params.py#L252)), even though it's used for both commands. 

Is this intentional or should I also add the argument to `az aks enable-addons`?

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -220,8 +220,11 @@
           type: bool
           short-summary: Use FIPS-enabled OS on agent nodes.
         - name: --workspace-resource-id
-          type: string
+          type: bool
```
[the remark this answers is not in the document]
daweim0: oops, thanks for catching that

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1460,14 +1469,30 @@ def aks_create(cmd,     # pylint: disable=too-many-locals,too-many-statements,to
...
+        try:
```
daweim0, line 1484: Not sure if there's a more concise way to wait for cluster creation to finish (since the object created_cluster can be a waitable type or not depending on if --no-wait is passed). Open to suggestions here.

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -3274,17 +3414,29 @@ def aks_disable_addons(cmd, client, resource_group_name, name, addons, no_wait=F
...
                               appgw_name=appgw_name, appgw_subnet_prefix=appgw_subnet_prefix, appgw_subnet_cidr=appgw_subnet_cidr, appgw_id=appgw_id, appgw_subnet_id=appgw_subnet_id, appgw_watch_namespace=appgw_watch_namespace,
                               enable_sgxquotehelper=enable_sgxquotehelper, enable_secret_rotation=enable_secret_rotation, no_wait=no_wait)
 
     if CONST_MONITORING_ADDON_NAME in instance.addon_profiles and instance.addon_profiles[CONST_MONITORING_ADDON_NAME].enabled:
-        _ensure_container_insights_for_monitoring(
-            cmd, instance.addon_profiles[CONST_MONITORING_ADDON_NAME])
+        if instance.addon_profiles[CONST_MONITORING_ADDON_NAME].config[CONST_MONITORING_USING_AAD_MSI_AUTH]:
+            # Ensure the cluster does not use service principal auth
+            try:
+                if instance.servicePrincipalProfile.clientId != "":  # just checking if this field exists
```
daweim0, line 3430: Not sure if this is the best way to check if a cluster uses service principal auth. Please let me know if there is a better way.

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -222,6 +222,9 @@
         - name: --workspace-resource-id
           type: string
           short-summary: The resource ID of an existing Log Analytics Workspace to use for storing monitoring data. If not specified, uses the default Log Analytics Workspace if it exists, otherwise creates one.
+        - name: --enable-msi-auth-for-monitoring
+          type: bool
+          short-summary: Sends monitoring data to Log Analytics using the cluster's assigned identity (instead of the Log Analytics Workspace's shared key).
```
zhoxing-ms, line 227:
```suggestion
          short-summary: Send monitoring data to Log Analytics using the cluster's assigned identity (instead of the Log Analytics Workspace's shared key).
```

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -1042,6 +1045,9 @@
   - name: --workspace-resource-id
     type: string
     short-summary: The resource ID of an existing Log Analytics Workspace to use for storing monitoring data.
+  - name: --enable-msi-auth-for-monitoring
+    type: bool
+    short-summary: Sends monitoring data to Log Analytics using the cluster's assigned identity (instead of the Log Analytics Workspace's shared key).
```
zhoxing-ms, line 1050:
```suggestion
    short-summary: Send monitoring data to Log Analytics using the cluster's assigned identity (instead of the Log Analytics Workspace's shared key).
```

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1250,28 +1252,34 @@ def aks_create(cmd,     # pylint: disable=too-many-locals,too-many-statements,to
...
+        vnet_subnet_id = vnet_subnet_id,
+        enable_secret_rotation = enable_secret_rotation,        
     )
     monitoring = False
     if CONST_MONITORING_ADDON_NAME in addon_profiles:
         monitoring = True
-        _ensure_container_insights_for_monitoring(
-            cmd, addon_profiles[CONST_MONITORING_ADDON_NAME])
+        if enable_msi_auth_for_monitoring and not enable_managed_identity:
+            raise CLIError("--enable-msi-auth-for-monitoring can not be used on clusters with service principal auth.")
```
zhoxing-ms, line 1277: Could you please use the specific error type `ArgumentUsageError` instead of `CLIError`

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1460,14 +1468,30 @@ def aks_create(cmd,     # pylint: disable=too-many-locals,too-many-statements,to
                 attach_acr,
                 headers,
                 no_wait)
-            return created_cluster
+            break
```
zhoxing-ms, line 1471: May I ask why you change `return created_cluster` to `break`?

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1460,14 +1468,30 @@ def aks_create(cmd,     # pylint: disable=too-many-locals,too-many-statements,to
...
-    raise retry_exception
+    else:
+        raise retry_exception
+
+    # Create the DCR Association here, it must be done after the cluster is created.
+    if monitoring and enable_msi_auth_for_monitoring:
+        try:
+            LongRunningOperation(cmd.cli_ctx)(created_cluster)
+        except AttributeError:
+            # this means the cluster was already created
```
zhoxing-ms, line 1486: Does `AttributeError` only correspond to `this means the cluster was already created` here? If not, could we capture more specific errors, such as parsing error code?

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -2687,104 +2730,212 @@ def _ensure_container_insights_for_monitoring(cmd, addon):
...
+                    r = send_raw_request(cmd.cli_ctx, "GET", feature_check_url)
+                    break
+                except CLIError as e:
+                    pass
+            else:
+                raise e
+            json_response = json.loads(r.text)
+            for resource in json_response["resourceTypes"]:
+                region_ids = map(lambda x: region_names_to_id[x], resource["locations"])  # map is lazy, so doing this for every region isn't slow
+                if resource["resourceType"] == "dataCollectionRules" and location not in region_ids:
```
zhoxing-ms, line 2788: Do we need to consider case sensitivity when matching `resourceType`?

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -2687,104 +2730,212 @@ def _ensure_container_insights_for_monitoring(cmd, addon):
...
+                    raise CLIError(f'Data Collection Rules are not supported for LA workspace region {location}')
+                elif resource["resourceType"] == "dataCollectionRuleAssociations" and cluster_region not in region_ids:
+                    raise CLIError(f'Data Collection Rule Associations are not supported for cluster region {location}')
```
zhoxing-ms, line 2791: Please use a specific error type from `azure.cli.core.azclierror` instead of `CLIError`

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -3274,17 +3445,29 @@ def aks_disable_addons(cmd, client, resource_group_name, name, addons, no_wait=F
...
 
     if CONST_MONITORING_ADDON_NAME in instance.addon_profiles and instance.addon_profiles[CONST_MONITORING_ADDON_NAME].enabled:
-        _ensure_container_insights_for_monitoring(
-            cmd, instance.addon_profiles[CONST_MONITORING_ADDON_NAME])
+        if instance.addon_profiles[CONST_MONITORING_ADDON_NAME].config[CONST_MONITORING_USING_AAD_MSI_AUTH]:
+            # Ensure the cluster does not use service principal auth
+            try:
+                if instance.servicePrincipalProfile.clientId != "":  # just checking if this field exists
+                    pass
+                raise CLIError("--enable-msi-auth-for-monitoring can not be used on clusters with service principal auth.")
```
zhoxing-ms, line 3463: Please use the specific error type `ArgumentUsageError` instead of `CLIError`

zhoxing-ms:
> One thing I'm wondering about: The arguments --workspace-resource-id is registered for `az aks create` ([here](https://github.com/Azure/azure-cli-extensions/blob/ff20effe1c4aca1dc71bf57251685e43a3f0a646/src/aks-preview/azext_aks_preview/_params.py#L93)) but not `az aks enable-addons` ([here](https://github.com/Azure/azure-cli-extensions/blob/ff20effe1c4aca1dc71bf57251685e43a3f0a646/src/aks-preview/azext_aks_preview/_params.py#L252)), even though it's used for both commands.
> 
> Is this intentional or should I also add the argument to `az aks enable-addons`?

The parameters registered under `az aks create` are only for `az aks create` and `az aks create`'s subcommands (if it has). You need to register this parameter under `az aks enable-addons` again~
But for the common properties of this parameter, you can configure it as `CLIArgumentType` and reuse it in those registered places.

src/aks-preview/azext_aks_preview/custom.py
zhoxing-ms: When the condition `monitoring and enable_msi_auth_for_monitoring` is met and then call method `_put_managed_cluster_ensuring_permission()`, how about fixing parameter `no_wait` to False in this situation?
Because in this case,  `no_wait=True` is meaningless and ineffective

src/aks-preview/azext_aks_preview/custom.py
zhoxing-ms: I don't know much about service principal auth. @jiasli Could you help to answer this question?

src/aks-preview/azext_aks_preview/custom.py
daweim0: Makes sense. An error is thrown if --attach-acr, --enable-managed-identity, and --no-wait are all three specified, should I follow the same pattern or silently disable --no-wait?

src/aks-preview/azext_aks_preview/custom.py
daweim0: Because I need to create a DCR-A after the cluster is created. The logic to return created_cluster was moved out of that loop. Now that you ask, I think it would be cleaner to move the DCR-A creation call into the for loop.

src/aks-preview/azext_aks_preview/custom.py
daweim0: Removed the need for this error

src/aks-preview/azext_aks_preview/custom.py
daweim0: Not sure, but it's safer to make it case insensitive

src/aks-preview/azext_aks_preview/custom.py
jiasli: This doesn't seem to be related to CLI's authentication. I have zero knowledge about AKS instance. 😥

ganga1980: Hi, @daweim0 , if the cluster using managed identity, clientId will be "msi" - https://docs.microsoft.com/en-us/azure/aks/use-managed-identity#obtain-and-use-the-system-assigned-managed-identity-for-your-aks-cluster. I think, we should get validated with Tongyao for both SAI & UAI.

sshadmin@gangams-ci-2:~$ az aks show -g gangams-aks-cpu-throttle-test -n gangams-aks-cpu-throttle-test --query "servicePrincipalProfile"
{
  "clientId": "msi"
}

daweim0: That works perfectly, thanks

ganga1980: Hi, @daweim0 , as discussed, this needs to be fixed. The fix should be **if instance.servicePrincipalProfile.clientId != "msi"**

daweim0:
I believe all comments are addressed. Please let me know if anything else needs to change before this PR is merged.

ganga1980 (approved):

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -1149,6 +1149,161 @@ def test_aks_update_to_msi_cluster_with_addons(self, resource_group, resource_gr
         self.cmd(
             'aks delete -g {resource_group} -n {name} --yes --no-wait', checks=[self.is_empty()])
 
+    # # These tests require subscription-level feature flags to be set by AKS and Azure Monitor. The features will be 
+    # # broadly rolling out soon, then these tests can be uncommented. (written on 6/29/2021)
```
FumingZhang, line 1153: Could you please uncomment the test and filter out this test case by adding its name to the filter defined [here](https://github.com/daweim0/azure-cli-extensions/blob/david/aks-monitoring-aad-auth/src/aks-preview/azcli_aks_live_test/ext_matrix_default.json#L29-L30) in your repo or [here](https://github.com/Azure/azure-cli-extensions/blob/main/src/aks-preview/azcli_aks_live_test/configs/ext_matrix_default.json#L29-L30) in the latest main branch?

FumingZhang (changes requested):

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -2687,104 +2727,219 @@ def _ensure_container_insights_for_monitoring(cmd, addon):
...
+                # This will run if the above for loop was not broken out of. This means all three requests failed
+                raise e
+            json_response = json.loads(r.text)
+            for region_data in json_response["value"]:
+                region_names_to_id[region_data["displayName"]] = region_data["name"]
+
+            # check if region supports DCRs and DCR-A
+            for _ in range(3):
+                try:
+                    feature_check_url = f"https://management.azure.com/subscriptions/{subscription_id}/providers/Microsoft.Insights?api-version=2020-10-01"
```
zhoxing-ms, line 2779: At present, are there corresponding methods in SDK that can call those REST API? If so, why not consider using the Python SDK methods instead of directly requesting REST?

daweim0: I'm not sure if there are python SDK methods to check for feature availability. I directly used rest because I couldn't figure out how to call "https://management.azure.com/subscriptions/{subscription_id}/providers/Microsoft.Insights?api-version=2020-10-01" and "https://management.azure.com/subscriptions/{subscription_id}/locations?api-version=2019-11-01" with the Python SKD.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
daweim0: done, thanks for explaining how to fix the issue

zhoxing-ms (approved):

src/aks-preview/azext_aks_preview/custom.py
zhoxing-ms: At present, there is no document for Python SDK. It depends on whether there is a swagger defined for those REST
It seems that there is no support from SDK methods, so this PR can be merge first

FumingZhang (approved):

PR #3575, opened 2021-07-02
Updating the quantum extension to add simulator support

bettinaheim:
This PR edits the output function to allow for jobs that run on the simulator to have output logged during execution that is not in json format. For simulator jobs, we expect the job result to be printed as the last item on its own line, and string output to be enclosed with `"`.

---

### General Guidelines

For new extensions:

- [N/A] My extension description/summary conforms to the [Extension Summary Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/extensions/extension_summary_guidelines.md).

### About Extension Publish


yonzhan:
quantum

kairu-ms:
@bettinaheim Please add description for your PR and linter related issues if exist. And you should also add tests to cover your code.

src/quantum/azext_quantum/commands.py
```diff
@@ -64,7 +64,7 @@ def one(key, value):
         return OrderedDict([
             ('Result', key),
             ('Frequency', f"{value:10.8f}"),
-            ('', f"\u2590{barra:<22} |"),
+            ('', f"\u007C{barra:^20}\u007C")  # f"\u23B9{barra:^22}\u23B8")
```
ricardo-espinoza, line 67: @bettinaheim , is there a reason why we are drawing the histogram this way?
Also, is it intentional to leave the comment with U+23B9 and U+23B8 or is it there because of experimentation?

bettinaheim: The bars for the simulator looked confusing to me because you often get a bar of 100%, but as it was, it didn't look like the bar was entirely filled up (it has no gap on the left side, and then a gap before the final dash on the right side, if I remember correctly). Ideally, I would have made it so that the bar just looks full, but for reasons I can't remember now I couldn't get that with the characters that looked like they were commonly supported in the terminal while also keeping a bar on the right to indicate where the 100% would sit. I am happy to adjust if you have other suggestions. The comment I believe was the better option in character choice but didn't look like it was commonly supported. I removed it.

bettinaheim:
@kairu-ms I edited the description for the PR. Regarding the tests, we are still in the process of integrating support for simulators to our service. The made edits are a prerequisite to allow us to proceed with that. We also have integration tests on our end.

kairu-ms (approved):

kairu-ms:
@bettinaheim If you want to release, please update extension version and history. Thanks.

PR #3617, opened 2021-07-09
Fix issue during submission of jobs where status is reported incorrectly.

ricardo-espinoza:
This PR is a minor bug fixed related to job ids being retrieved incorrectly from standard output of a caller process.

The fix strips additional information from the output so only the job-id is used.

### DIFF
--- a/src/quantum/azext_quantum/operations/job.py
+++ b/src/quantum/azext_quantum/operations/job.py
@@ -150,7 +150,10 @@ def submit(cmd, program_args, resource_group_name=None, workspace_name=None, loc
     result = subprocess.run(args, stdout=subprocess.PIPE, check=False)
 
     if result.returncode == 0:
-        job_id = result.stdout.decode('ascii').strip()
+        output = result.stdout.decode('ascii').strip()
+        # Retrieve the job-id as the last line from standard output.
+        job_id = output.split()[-1]
+        # Query for the job and return status to caller.
         return get(cmd, job_id, resource_group_name, workspace_name, location)
 
     raise CLIError("Failed to submit job.")


src/quantum/azext_quantum/operations/job.py
```diff
@@ -150,7 +150,10 @@ def submit(cmd, program_args, resource_group_name=None, workspace_name=None, loc
     result = subprocess.run(args, stdout=subprocess.PIPE, check=False)
 
     if result.returncode == 0:
-        job_id = result.stdout.decode('ascii').strip()
+        output = result.stdout.decode('ascii').strip()
+        # Retrieve the job-id as the last line from standard output.
+        job_id = output.split()[-1]
```
george-moussa, line 155: is there any other way of getting the job_id? relying on stdout is error-prone. 
Also if we end up relying on stdout, can we make sure there is a comment in the standalone executable that explains how the output is used in CLI to avoid breaking it again?

anpaz (approved):
lgtm.
we need another PR for the qsharp-runtime such that the output of the Message does not get printed to the console when submitting to azure quantum.

src/quantum/azext_quantum/operations/job.py
anpaz: no, there is currently no other way to get the corresponding job id.
there is already a comment, in fact, there is already a parameter for the standalone executable to return the make the job id the output of the program. the fact that the Messages from the Q# program are getting directed to the console in this case is a bit unrelated.

yonzhan:
quantum

kairu-ms (approved):

PR #3658, opened 2021-07-15
[AKS][aks-preview] `az aks nodepool upgrade`: Add new parameter `--aks-custom-headers`

mainred:
---

### General Guidelines

For new extensions:

### About Extension Publish

Test:

`az aks nodepool upgrade --resource-group $rg --cluster-name k8s --name dr2 --kubernetes-version 1.21.1 --aks-custom-headers WindowsContainerRuntime=containerd --debug`

returns successfully and the header value can ben seen below:
```
cli.azure.cli.core.sdk.policies: Request headers:
cli.azure.cli.core.sdk.policies:     'x-ms-client-request-id': 'b7eaadd7-e556-11eb-9394-000d3ac57d3d'
cli.azure.cli.core.sdk.policies:     'CommandName': 'aks nodepool upgrade'
cli.azure.cli.core.sdk.policies:     'ParameterSetName': '--resource-group --cluster-name --name --kubernetes-version --aks-custom-headers --debug'
cli.azure.cli.core.sdk.policies:     'WindowsContainerRuntime': 'containerd'
```

### DIFF
--- a/src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
+++ b/src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
@@ -9,6 +9,7 @@ import tempfile
 
 from azure.cli.testsdk import (
     ResourceGroupPreparer, RoleBasedServicePrincipalPreparer, ScenarioTest, live_only)
+from azure.cli.command_modules.acs._format import version_to_tuple
 from azure_devtools.scenario_tests import AllowLargeResponse
 
 from .recording_processors import KeyReplacer
@@ -25,6 +26,18 @@ class AzureKubernetesServiceScenarioTest(ScenarioTest):
         super(AzureKubernetesServiceScenarioTest, self).__init__(
             method_name, recording_processors=[KeyReplacer()]
         )
+
+    def _get_versions(self, location):
+        """Return the previous and current Kubernetes minor release versions, such as ("1.11.6", "1.12.4")."""
+        versions = self.cmd(
+            "az aks get-versions -l westus2 --query 'orchestrators[].orchestratorVersion'").get_output_in_json()
+        # sort by semantic version, from newest to oldest
+        versions = sorted(versions, key=version_to_tuple, reverse=True)
+        upgrade_version = versions[0]
+        # find the first version that doesn't start with the latest major.minor.
+        prefix = upgrade_version[:upgrade_version.rfind('.')]
+        create_version = next(x for x in versions if not x.startswith(prefix))
+        return create_version, upgrade_version
 
     @live_only()  # without live only fails with need az login
     @AllowLargeResponse()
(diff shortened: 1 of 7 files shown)


yonzhan:
aks

mainred:
This is to support upgrading Windows nodepools from docker to containerd during the Windows containerd public preview.

cc  @keikhara  @AbelHu @palma21

feiskyer (approved):

mainred:
@zhoxing-ms , could you please take a look? Thanks.

AbelHu (approved):

zhoxing-ms:
@mainred Could you add some test cases for this new parameter?

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -885,6 +885,7 @@ def test_aks_upgrade_node_image_only_nodepool(self, resource_group, resource_gro
                                                '--cluster-name {name} ' \
                                                '-n {node_pool_name} ' \
                                                '--node-image-only ' \
+                                               '--aks-custom-headers fake-key1=value1,fake-key2=value2' \
```
mainred, line 888: `--aks-custom-headers` is basically used for preview features thus a specific header is temporarily supported. Instead of using a specific header, we use a fake header and value, which is also supported but brings no actual effect on the command.
We can discuss here.

mainred: cc @zhoxing-ms

zhoxing-ms: May I ask why don't you use the specific header here?

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -896,6 +909,63 @@ def test_aks_upgrade_node_image_only_nodepool(self, resource_group, resource_gro
...
+            'nodepool2_name': 'npwin',
+            'k8s_version': create_version,
+            'upgrade_k8s_version': upgrade_version,
+        })
+
+        # create AKS cluster
+        create_cmd = 'aks create --resource-group={resource_group} --name={name} --location={location} ' \
+                     '--dns-name-prefix={dns_name_prefix} --node-count=1 --generate-ssh-keys ' \
+                     '--windows-admin-username={windows_admin_username} --windows-admin-password={windows_admin_password} ' \
+                     '--load-balancer-sku=standard --vm-set-type=virtualmachinescalesets --network-plugin=azure' \
```
FumingZhang, line 939:
```suggestion
                     '--load-balancer-sku=standard --vm-set-type=virtualmachinescalesets --network-plugin=azure ' \
```

FumingZhang (changes requested):
need to add a space to separate options

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
mainred: Done, thanks.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
mainred: @zhoxing-ms , I added a new test to do aks nodepool upgrade with the customer header, thanks for your advice.

FumingZhang (approved):

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -896,6 +909,62 @@ def test_aks_upgrade_node_image_only_nodepool(self, resource_group, resource_gro
             self.check('provisioningState', 'UpgradingNodeImageVersion')
         ])
 
+    @live_only() # live only is required for test environment setup like `az login`
```
zhoxing-ms, line 912: May I ask why `az login` is required for these tests in test environment setup? What's the problem with using the current subscription?

FumingZhang: Let me help explain. I think the comment here is somewhat misunderstanding. This test case needs to be executed in live mode. Without the @live_only tag, when check-in CI is running in recording mode, if the recording file corresponding to this case is not found, this case will be executed in live mode. However, az login is not configured in CI. In turn, the test will fail, prompting that az login is required.

FumingZhang: Maybe just keep the @live_only tag and delete the confusing comment.

mainred: Per @FumingZhang 's explanation, I don't think it's confusing or misunderstanding. tests like this require live_only to setup the environment, and `az login` is one of the steps. 
I guess somebody else hit the same issue requiring `az login`, and add a comment like this but introduced grammatical mistake, _without live only fails with need az login_

zhoxing-ms: Actually, in check-in CI, the pipeline uses the subscription of our team to run the test, so no additional login is required. Could you show me the information of the error report?

mainred: This is the pipeline build that asks for `az login`.
https://msazure.visualstudio.com/09706533-03bf-4b43-9a9b-b49c75429646/_apis/build/builds/44691106/logs/12

mainred: After reading through the doc, I found the comment is a just a solution for a fix, but not really suitable, it doesn't have to be live_only test, but we need your help to explain why it requires `az login` without live_only decorator

mainred: Removed all the misleading comments for live_only decorator.

mainred: @zhoxing-ms please merge this PR when you think you have nowhere to change. Thanks.

mainred: Ha, got what you mean. Let me upload the YAML file then we can remove live only. Thanks for pointing this out.

mainred: Added generated recording file.

zhoxing-ms (approved):

PR #3676, opened 2021-07-20
[k8s-configuration] Update sourceControlConfiguration SDK Models to Track2

jonathan-innis:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
k8s-configuration

src/k8s-configuration/azext_k8s_configuration/_validators.py
```diff
@@ -44,3 +58,86 @@ def __validate_k8s_name(param_value, param_name, max_len):
...
+            except ValueError:
+                try:
+                    key_obj = io.StringIO(_from_base64(ssh_private_key_data).decode('utf-8'))
+                    Ed25519Key(file_obj=key_obj)
+                    return
+                except SSHException:
+                    raise InvalidArgumentValueError(
+                        'Error! --ssh-private-key provided in invalid format',
+                        'Verify the key provided is a valid PEM-formatted key of type RSA, ECC, DSA, or Ed25519')
+
```
zhoxing-ms, line 129: Do we have a more efficient way to get the type of encryption algorithm instead of trying one by one?

src/k8s-configuration/azext_k8s_configuration/_validators.py
```diff
@@ -44,3 +58,86 @@ def __validate_k8s_name(param_value, param_name, max_len):
...
+        rp_client = _resource_providers_client(cmd.cli_ctx)
+        registration_state = rp_client.get(consts.PROVIDER_NAMESPACE).registration_state
+
+        if registration_state.lower() != consts.REGISTERED.lower():
+            logger.warning("'Source Control Configuration' cannot be used because '%s' provider has not been "
+                           "registered. More details for registering this provider can be found here - "
+                           "https://aka.ms/RegisterKubernetesConfigurationProvider", consts.PROVIDER_NAMESPACE)
+    except Exception:
+        logger.warning("Unable to fetch registration state of '%s' provider. "
+                       "Failed to enable 'source control configuration' feature...", consts.PROVIDER_NAMESPACE)
```
zhoxing-ms, line 143: There is no need to throw an exception in these cases, right?

zhoxing-ms:
Could you run some live tests for these changes?

jonathan-innis:
> Could you run some live tests for these changes?
> 
> 
> 
> 

What kind of live tests are you looking for? We have had issue with live tests previously because we require an AKS cluster and connected cluster to be pre-provisioned in order to run tests against it

src/k8s-configuration/azext_k8s_configuration/_validators.py
jonathan-innis: We found the error that was returned by default from the CLI wasn't specific enough and didn't help customers figure out how to fix the registration issue

src/k8s-configuration/azext_k8s_configuration/_validators.py
jonathan-innis: Unfortunately, not really. This is the way that we have been validating the private keys up to this point.

src/k8s-configuration/azext_k8s_configuration/_validators.py
zhoxing-ms: I mean, even if there's a issue here, we just need to log it here, it doesn't affect the following process, right?

src/k8s-configuration/azext_k8s_configuration/_validators.py
zhoxing-ms: OK~

src/k8s-configuration/azext_k8s_configuration/_validators.py
jonathan-innis: That's what we're doing right?

zhoxing-ms: Yes. That's what I want to confirm~ If the exception here will affect the following process, it needs to be thrown early. If not, it is fine to only print the log

jonathan-innis:
Record mode is difficult to run as well because some of these resources take identity which is not repeatable in record mode. We have an E2E test suite that we run on our end that covers all cases

src/k8s-configuration/azext_k8s_configuration/_validators.py
jonathan-innis: It will not affect the following process, just a log to give more information on what the user needs to do without the provider enabled. Similar logic is found for [connectedk8s](https://github.com/Azure/azure-cli-extensions/blob/main/src/connectedk8s/azext_connectedk8s/custom.py#L1969)

zhoxing-ms: Got it, thanks~

zhoxing-ms:
OK, it is suggested to fix those issues related tests in live mode/record mode in the future, so that our pipeline can better maintain and test the code quality

zhoxing-ms (approved):

PR #3698, opened 2021-07-26
[SSH] Various fixes for Github Issues

N6UDP:
---
* Add support for direct MSAL usage in newer azure cli (beta currently)
* Add support for port option and ssh additional arguments option
* Remove directory creation as part of ssh_config creation
* Try .pub for public key if only private key is specified
* Add --hostname to --ip argument
* Add fallback to private IP with warning
* Add USGov and China cloud support

### General Guidelines

### About Extension Publish


yonzhan:
SSH

src/ssh/azext_ssh/custom.py
```diff
@@ -105,7 +115,10 @@ def _check_or_create_public_private_files(public_key_file, private_key_file):
         ssh_utils.create_ssh_keyfile(private_key_file)
 
     if not public_key_file:
-        raise util.CLIError(f"Public key file not specified")
+        if private_key_file:
+            public_key_file = private_key_file + ".pub"
+        else:
+            raise util.CLIError("Public key file not specified")
```
zhoxing-ms, line 121: Could you please use a specific error type instead of `CLIError`

src/ssh/HISTORY.md
```diff
@@ -1,5 +1,14 @@
 Release History
 ===============
+0.1.6
+-----
+* Add support for direct MSAL usage in newer azure cli (beta currently)
```
zhoxing-ms, line 5:
```suggestion
* Add support for direct MSAL usage in newer Azure CLI (beta currently)
```

src/ssh/azext_ssh/_help.py
```diff
@@ -20,6 +20,7 @@
         - name: Give the public IP (or hostname) of a VM to SSH to
           text: |
             az ssh vm --ip 1.2.3.4
+            az ssh vm --hostname example.com
```
zhoxing-ms, line 23: We need to add a new example name for this new text

zhoxing-ms:
@N6UDP Any update for those comments?

jiasli (dismissed):

src/ssh/azext_ssh/_help.py
N6UDP: what do you mean?  example.com is a valid IETF name

src/ssh/azext_ssh/custom.py
N6UDP: Are there docs or examples? The existing code through the ssh extension seems to raise only two error types.

src/ssh/azext_ssh/_params.py
```diff
@@ -8,11 +8,14 @@ def load_arguments(self, _):
...
-        c.argument('ssh_ip', options_list=['--ip'], help='The public IP address (or hostname) of the VM')
+        c.argument('ssh_ip', options_list=['--ip', '--hostname'],
+                   help='The public (or reachable private) IP address (or hostname) of the VM')
         c.argument('public_key_file', options_list=['--public-key-file', '-p'], help='The RSA public key file path')
         c.argument('private_key_file', options_list=['--private-key-file', '-i'], help='The RSA private key file path')
         c.argument('use_private_ip', options_list=['--prefer-private-ip'],
-                   help='Will use a private IP if available. By default only public IPs are used.')
+                   help='Will prefer private IP.  Requires connectivity to the private IP.')
+        c.argument('port', options_list=['--port'], help='SSH port')
+        c.argument('ssh_args', options_list=['--ssh-args'], help='SSH Additional arguments')
```
jiasli, line 18: I don't think `--ssh-args` will work as expected, as it only takes exactly one value. The user may usually want to provide more than one value. (https://github.com/Azure/azure-cli/issues/19101)

Also, if not handled correctly, leading dashes in parameters like `--ssh-args -o SomeSshOption=3` will corrupt the parsing. See details at https://github.com/Azure/azure-cli/issues/18869.

src/ssh/azext_ssh/ssh_utils.py
```diff
@@ -9,14 +9,15 @@
...
-from . import file_utils
-
 logger = log.get_logger(__name__)
 
 
-def start_ssh_connection(ip, username, cert_file, private_key_file):
+def start_ssh_connection(port, ssh_args, ip, username, cert_file, private_key_file):
+    ssh_arg_list = []
+    if ssh_args:
+        ssh_arg_list = [ssh_args]
```
jiasli, line 18: Even if `ssh_args` was taken as `--ssh-args="-o SomeSshOption=3 -L 8080:localhost:8080"`, you will need to use [`shlex.split`](https://docs.python.org/3/library/shlex.html#shlex.split) to split the values into 

```py
>>> shlex.split('-o SomeSshOption=3 -L 8080:localhost:8080')
['-o', 'SomeSshOption=3', '-L', '8080:localhost:8080']
```

instead of simply wrapping `ssh_args` in a list.

jiasli:
If possible, consider splitting this PR to multiple PRs and link to corresponding issues. This simplifies the review process and our discussion.

src/ssh/azext_ssh/_help.py
jiasli: I believe @zhoxing-ms means you may give it a new name, like L20.

src/ssh/azext_ssh/custom.py
zhoxing-ms: Please refer to this document to set specific error types [error_handling_guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md)

src/ssh/azext_ssh/custom.py
```diff
@@ -50,12 +51,30 @@ def _do_ssh_op(cmd, resource_group, vm_name, ssh_ip, public_key_file, private_ke
 
 
 def _get_and_write_certificate(cmd, public_key_file, cert_file):
-    scopes = ["https://pas.windows.net/CheckMyAccess/Linux/.default"]
+    cloudtoscope = {
+        "azurecloud": "https://pas.windows.net/CheckMyAccess/Linux/.default",
+        "azurechinacloud": "https://pas.chinacloudapi.cn/CheckMyAccess/Linux",
```
Klaas-, line 56: I think you forgot the /.default at the end

Klaas-: I tested this change, withouth the /.default at the end it does not work, with it I seem to get further along in the auth process :)

src/ssh/azext_ssh/_help.py
N6UDP: they are an alias of the same argument so I don't see why we would do so.  e.g. you can give IP to --hostname and a hostname to --ip .. I just added the `--hostname` alias.

jiasli (approved):

jiasli:
I currently don't have enough time to fully check the PR. Let's publish it for now (with the preview talisman).

PR #3723, opened 2021-07-29
[CosmosDB] Managed Cassandra Hotfix

AverageDesigner:
cli extension module version 0.8.0 introduced regression because of python sdk update. It is being fixed in this PR.

### DIFF
--- a/src/cosmosdb-preview/azext_cosmosdb_preview/custom.py
+++ b/src/cosmosdb-preview/azext_cosmosdb_preview/custom.py
@@ -20,6 +20,7 @@ from azext_cosmosdb_preview.vendored_sdks.azure_mgmt_cosmosdb.models import (
     ContinuousModeBackupPolicy,
     ClusterResource,
     ClusterResourceProperties,
+    DataCenterResource,
     DataCenterResourceProperties
 )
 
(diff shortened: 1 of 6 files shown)


src/cosmosdb-preview/azext_cosmosdb_preview/custom.py
```diff
@@ -566,7 +567,11 @@ def cli_cosmosdb_managed_cassandra_datacenter_create(client,
         base64_encoded_cassandra_yaml_fragment=base64_encoded_cassandra_yaml_fragment
     )
 
-    return client.begin_create_update(resource_group_name, cluster_name, data_center_name, data_center_properties)
+    data_center_resource = DataCenterResource(
+        data_center_properties=data_center_properties
```
MehaKaushik, line 571: Is this a bug in sdk generation or was something missed in Swagger ?? Did you make changes to the sdk manually ?? Ideally we don't want to do that, we should always fix in swagger and generate sdk from swagger.

frrossms (approved):

src/cosmosdb-preview/azext_cosmosdb_preview/custom.py
frrossms: We haven't changed the Swagger, so it has to be a bug in SDK generation.

frrossms (commented):
Are there changes for the nodetool commands that need to be added?

yonzhan:
CosmosDB

frrossms (approved):

evelyn-ys (commented):
Pls update `setup.py` and `history.rst` to upgrade version

evelyn-ys (approved):

PR #3755, opened 2021-08-05
[vm-repair] Making Public IP as optional

haagha:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
vm-repair

src/vm-repair/azext_vm_repair/_validators.py
```diff
@@ -201,6 +204,18 @@ def _prompt_repair_password(namespace):
...
+def _prompt_public_ip(namespace):
+    from knack.prompting import prompt_y_n, NoTTYException
+    try:
+        if prompt_y_n('Does repair vm requires public ip?'):
+            namespace.associate_public_ip = "yes"
+        else:
+            namespace.associate_public_ip = '""'
+
+    except NoTTYException:
+        raise CLIError('Please specify the associate-public-ip parameter in non-interactive mode.')
```
zhoxing-ms, line 216: Could you please use specific error type instead of `CLIError`. For more details, please refer to [error_handling_guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md)

zhoxing-ms:
Could you add some test cases for this new parameter?

src/vm-repair/azext_vm_repair/_validators.py
dibaskar: Specified specific error type **ValidationError** instead of **CLIError**

Below are the test cases.
**Test case 1 :**
Repair command can be executed without argument --associate-public-ip . Then command will wait for user input with below prompt.
 "Does repair vm requires public ip? (y/n)"
 a) y input should create repair vm with public ipaddress
 b) n input should create repair VM without public ipaddress
 
**Test case 2 :**
Repair command can be executed with argument--associate-public-ip . User won’t be prompted for public IP option instead repair VM will be created with Public IP associated to it.

dibaskar:
> Could you add some test cases for this new parameter?



> Could you add some test cases for this new parameter?

Below are the test cases.
**Test case 1 :**
Repair command can be executed without argument --associate-public-ip . Then command will wait for user input with below prompt.
 "Does repair vm requires public ip? (y/n)"
 a) y input should create repair vm with public ipaddress
 b) n input should create repair VM without public ipaddress
 
**Test case 2 :**
Repair command can be executed with argument--associate-public-ip . User won’t be prompted for public IP option instead repair VM will be created with Public IP associated to it.

zhoxing-ms:
@dibaskar Could you add some tests for checking this parameter?

dibaskar:
> @dibaskar Could you add some tests for checking this parameter?

@zhoxing-ms tests can be executed as below.

az vm repair create -g dual_group -n impactedvmname --repair-vm-name repairvmname --repair-username username --repair-password password --associate-public-ip --verbose --debug

az vm repair create -g dual_group -n impactedvmname --repair-vm-name repairvmname --repair-username username --repair-password password --verbose --debug

zhoxing-ms:
@dibaskar Actually, I mean adding test code to `test_repair_commands.py` file

zhoxing-ms:
@dibaskar  Why are these tests `LiveScenarioTest` and not `ScenarioTest`? Is there any reason why these tests cannot be recorded?

dibaskar:
> @dibaskar Why are these tests `LiveScenarioTest` and not `ScenarioTest`? Is there any reason why these tests cannot be recorded?

@zhoxing-ms This change will have impact to the execution flow of the command as it will wait for user input if --associate-public-ip is not specified in command. So i made use of the existing function LiveScenarioTest as this has to be tested for both linux and windows VMs ( Managed and unmanaged disks )

zhoxing-ms (approved):

PR #3765, opened 2021-08-10
Addon options

IshanArya:
---

### General Guidelines

- [✓] Have you run `azdev style <YOUR_EXT>` locally? (`pip install azdev` required)
- [✓] Have you run `python scripts/ci/test_index.py -q` locally?

### About Extension Publish


yonzhan:
aks-preview

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -370,6 +370,27 @@ def test_aks_create_with_openservicemesh_addon(self, resource_group, resource_gr
...
+    @AKSCustomResourceGroupPreparer(random_name_length=17, name_prefix='clitest', location='westus2')
+    def test_aks_addon_list(self, resource_group, resource_group_location):
+        print("what up?")
+        aks_name = self.create_random_name('cliakstest', 16)
+        self.kwargs.update({
+            'resource_group': resource_group,
+            'name': aks_name,
+        })
+
+        create_cmd = 'aks create --resource-group={resource_group} --name={name} --enable-managed-identity --generate-ssh-keys -o json'
```
FumingZhang, line 383: Please replace `--generate-ssh-keys` with `--ssh-key-value`, this could avoid race condition during concurrent testing. Refer issue [#18854](https://github.com/Azure/azure-cli/issues/18854).

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -370,6 +370,27 @@ def test_aks_create_with_openservicemesh_addon(self, resource_group, resource_gr
...
+        self.cmd(create_cmd, checks=[
+            self.check('provisioningState', 'Succeeded'),
+            self.check('addonProfiles.openServiceMesh', None),
+        ])
+
+        list_cmd = 'aks addon list --resource-group={resource_group} --name={name} -o json'
+        addon_list = self.cmd(list_cmd).get_output_in_json()
```
FumingZhang, line 390: Please use the method provided by testsdk to check the result, this could be something like ```self.check('length(@) == `2`', True)```.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -370,6 +370,27 @@ def test_aks_create_with_openservicemesh_addon(self, resource_group, resource_gr
             self.check('addonProfiles.openServiceMesh.enabled', True),
         ])
 
+    @AllowLargeResponse()
+    @AKSCustomResourceGroupPreparer(random_name_length=17, name_prefix='clitest', location='westus2')
+    def test_aks_addon_list(self, resource_group, resource_group_location):
```
FumingZhang, line 375: Please add more test cases to cover other new commands (enable, disable, update).

FumingZhang (changes requested):

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
IshanArya: Would `@` point to the list I receive from the command?

IshanArya: Also the docs recommend using this way of testing for dynamic stuff: https://github.com/Azure/azure-cli/blob/dev/doc/authoring_tests.md#sample-2-validate-the-return-value-in-json

src/aks-preview/azext_aks_preview/addonconfiguration.py
```diff
@@ -0,0 +1,900 @@
...
+    addon_args = addons.split(',')
+
+    addon_profiles = instance.addon_profiles or {}
+
+    os_type = 'Linux'
+
+    # for each addons argument
+    for addon_arg in addon_args:
+        if addon_arg not in ADDONS:
+            raise CLIError("Invalid addon name: {}.".format(addon_arg))
```
FumingZhang, line 155: Could you please replace all `CLIError` with more specific error types defined in `azure.cli.core.azclierror`? Also don’t forget to change the corresponding error type checked in test cases.

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -1003,6 +1003,164 @@
...
+    short-summary: Resource Id of an existing Subnet used to deploy the Application Gateway. Use with ingress-azure addon.
+  - name: --appgw-watch-namespace
+    type: string
+    short-summary: Specify the namespace, which AGIC should watch. This could be a single string value, or a comma-separated list of namespaces. Use with ingress-azure addon.
+  - name: --enable-sgxquotehelper
+    type: bool
+    short-summary: Enable SGX quote helper for confcom addon.
+  - name: --enable-secret-rotation
+    type: bool
+    short-summary: Enable secret rotation. Use with azure-keyvault-secrets-provider addon.
```
FumingZhang, line 1098: Could please add some test cases to cover all supported options for enable/update?

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
FumingZhang: You can check the output with JMESPath or just get it done with python. In my opinion, it is more concise to use JMESPath when the logic is not complicated. Back to this case, it's all good to check with `assert`.

src/aks-preview/azext_aks_preview/_help.py
IshanArya: I can but the base code behind each of these options is the same as the code behind `enable-addons`, which would lead to a lot of duplicated tests that are checking for the exact same thing. Do you think it is still worth it since the new code I wrote is not related to these options.

src/aks-preview/azext_aks_preview/addonconfiguration.py
IshanArya: This code was not written by me. I just moved it to organize everything. Is it ok for me to still change the errors?

src/aks-preview/azext_aks_preview/_help.py
FumingZhang: I think this should be necessary. Probably these options are passed and handled correctly for now, but future changes may introduce errors. There is no need to create a test case for each option. Options belonging to the same addon can be placed in the same test case for testing. Don't forget to test both `enable` and `update`:wink:

IshanArya: Hi Fuming, I was wondering if there is a doc that notes what each parameter does so I can test it appropriately. I realized that there actually weren't test cases for each of the options with `enable-addons`, so I'll have to make new tests.

kaiqzhan: I think adding new test cases for addon parameters is a little bit out of the scope of this pr because addon owner has more knowledge to do this. But I think we can reuse the existing unit test cases for enable/disable commands. For example, we can add extra lines to test `addon disable/enable` commands to the end of the `test_aks_byo_appgw_with_ingress_appgw_addon` test case.

kaiqzhan: There're already >20 unit test cases for various addons if you search `addonProfiles` in the test_aks_commands.py

FumingZhang (approved):
LGTM, requesting @andyliuliming for another review.

src/aks-preview/azext_aks_preview/_roleassignments.py
```diff
@@ -0,0 +1,136 @@
...
+    return list(graph_client.objects.get_objects_by_object_ids(params))
+
+
+def resolve_object_id(cli_ctx, assignee):
+    client = get_graph_rbac_management_client(cli_ctx)
+    result = None
+    if assignee.find('@') >= 0:  # looks like a user principal name
+        result = list(client.users.list(
+            filter="userPrincipalName eq '{}'".format(assignee)))
+    if not result:
```
andyliuliming, line 29: result is None

src/aks-preview/azext_aks_preview/_roleassignments.py
```diff
@@ -0,0 +1,136 @@
...
+        result = list(client.service_principals.list(
+            filter="servicePrincipalNames/any(c:c eq '{}')".format(assignee)))
+    if not result:  # assume an object id, let us verify it
```
andyliuliming, line 32: result is None

src/aks-preview/azext_aks_preview/_roleassignments.py
```diff
@@ -0,0 +1,136 @@
...
+        result = _get_object_stubs(client, [assignee])
+
+    # 2+ matches should never happen, so we only check 'no match' here
+    if not result:
```
andyliuliming, line 36: I think after L32, the result will never be None anymore.

src/aks-preview/azext_aks_preview/_roleassignments.py
```diff
@@ -0,0 +1,136 @@
...
+    return result[0].object_id
+
+
+def resolve_role_id(role, scope, definitions_client):
+    role_id = None
+    try:
+        uuid.UUID(role)
+        role_id = role
+    except ValueError:
+        pass
```
andyliuliming, line 49: I don't think it's a good idea to eat up one error silently.

src/aks-preview/azext_aks_preview/_roleassignments.py
```diff
@@ -0,0 +1,136 @@
...
+    if not role_id:  # retrieve role id
+        role_defs = list(definitions_client.list(
+            scope, "roleName eq '{}'".format(role)))
+        if not role_defs:
```
andyliuliming, line 53: more readable if we use len(role_defs) == 0

src/aks-preview/azext_aks_preview/_roleassignments.py
```diff
@@ -0,0 +1,136 @@
...
+            ids = [r.id for r in role_defs]
+            err = "More than one role matches the given name '{}'. Please pick a value from '{}'"
+            raise CLIError(err.format(role, ids))
+        role_id = role_defs[0].id
+    return role_id
+
+
+def build_role_scope(resource_group_name, scope, subscription_id):
+    subscription_scope = '/subscriptions/' + subscription_id
+    if scope:
```
andyliuliming, line 65: use scope is not None, or put the type annotation to the parameters.
scope: str, resource_group_name: str.

src/aks-preview/azext_aks_preview/_roleassignments.py
```diff
@@ -0,0 +1,136 @@
...
+
+def _create_role_assignment(cli_ctx, role, assignee,
+                            resource_group_name=None, scope=None, resolve_assignee=True):
+    from azure.cli.core.profiles import ResourceType, get_sdk
+    factory = get_auth_management_client(cli_ctx, scope)
+    assignments_client = factory.role_assignments
+    definitions_client = factory.role_definitions
+
+    scope = build_role_scope(
+        resource_group_name, scope, assignments_client.config.subscription_id)
```
andyliuliming, line 91: better to check whether assignments_client.config is None or not before use it.

src/aks-preview/azext_aks_preview/_roleassignments.py
```diff
@@ -0,0 +1,136 @@
...
+    hook = cli_ctx.get_progress_controller(True)
+    hook.add(message='Waiting for AAD role to propagate',
+             value=0, total_val=1.0)
+    logger.info('Waiting for AAD role to propagate')
+    for x in range(0, 10):
+        hook.add(message='Waiting for AAD role to propagate',
+                 value=0.1 * x, total_val=1.0)
+        try:
+            # TODO: break this out into a shared utility library
+            create_role_assignment(
```
andyliuliming, line 122: why not call _create_role_assignment?

src/aks-preview/azext_aks_preview/_roleassignments.py
```diff
@@ -0,0 +1,136 @@
...
+                cli_ctx, role, service_principal_msi_id, is_service_principal, scope=scope)
+            break
+        except CloudError as ex:
+            if ex.message == 'The role assignment already exists.':
+                break
+            logger.info(ex.message)
+        except:  # pylint: disable=bare-except
+            pass
```
andyliuliming, line 130: at least one logging.

src/aks-preview/azext_aks_preview/_roleassignments.py
```diff
@@ -0,0 +1,136 @@
...
+    for x in range(0, 10):
+        hook.add(message='Waiting for AAD role to propagate',
+                 value=0.1 * x, total_val=1.0)
+        try:
+            # TODO: break this out into a shared utility library
+            create_role_assignment(
+                cli_ctx, role, service_principal_msi_id, is_service_principal, scope=scope)
+            break
+        except CloudError as ex:
+            if ex.message == 'The role assignment already exists.':
```
andyliuliming, line 126: is there any error code we can use instead of checking the error message?

src/aks-preview/azext_aks_preview/addonconfiguration.py
```diff
@@ -0,0 +1,900 @@
+# --------------------------------------------------------------------------------------------
```
andyliuliming, line 1: I suppose this file is copied from some where?

I would suggest break out one new PR for the file renaming or function name renaming.
that will helps a lot for the code reviewers.

andyliuliming (changes requested):

src/aks-preview/azext_aks_preview/_roleassignments.py
IshanArya: Hi Andy, a lot of this code is not mine. It was simply moved around for organization purposes and to avoid circular dependencies. I ensured everything still works, but did not change any code functionality except for functions that were directly related to addons (and there were only 2 of them). I'm not sure it would be safe for me to change the functionality of these other functions without knowing there purpose.

src/aks-preview/azext_aks_preview/_roleassignments.py
IshanArya: Hi Andy, a lot of this code is not mine. It was simply moved around for organization purposes and to avoid circular dependencies. I ensured everything still works, but did not change any code functionality except for functions that were directly related to addons (and there were only 2 of them). I'm not sure it would be safe for me to change the functionality of these other functions without knowing there purpose.

src/aks-preview/azext_aks_preview/_roleassignments.py
IshanArya: Hi Andy, a lot of this code is not mine. It was simply moved around for organization purposes and to avoid circular dependencies. I ensured everything still works, but did not change any code functionality except for functions that were directly related to addons (and there were only 2 of them). I'm not sure it would be safe for me to change the functionality of these other functions without knowing there purpose.

This one seems easy enough to add in, however, if you want me to.

src/aks-preview/azext_aks_preview/_roleassignments.py
IshanArya: Hi Andy, a lot of this code is not mine. It was simply moved around for organization purposes and to avoid circular dependencies. I ensured everything still works, but did not change any code functionality except for functions that were directly related to addons (and there were only 2 of them). I'm not sure it would be safe for me to change the functionality of these other functions without knowing there purpose.

This one seems easy enough to add in, however, if you want me to, though it might break things if result could simply be a False value for whatever reason.

src/aks-preview/azext_aks_preview/_resourcegroup.py
```diff
@@ -0,0 +1,13 @@
...
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from ._client_factory import cf_resource_groups
+
+
+def get_rg_location(ctx, resource_group_name, subscription_id=None):
+    groups = cf_resource_groups(ctx, subscription_id=subscription_id)
+    # Just do the get, we don't need the result, it will error out if the group doesn't exist.
+    rg = groups.get(resource_group_name)
```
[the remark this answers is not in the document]
IshanArya: Hi Andy, a lot of this code is not mine. It was simply moved around for organization purposes and to avoid circular dependencies. I ensured everything still works, but did not change any code functionality except for functions that were directly related to addons (and there were only 2 of them). I'm not sure it would be safe for me to change the functionality of these other functions without knowing there purpose.

zhoxing-ms:
@IshanArya Could you please address those CI issues?

zhoxing-ms:
@IshanArya Any update?

PixelRobots:
Any update on this?

zhoxing-ms:
Because this PR has not been updated for a long time, so I close it first. When we need merge it again, then reopen it

tilnl-ms:
Thanks @zhoxing-ms , the owner is not working on this project now. I will create a new PR instead.

PR #3776, opened 2021-08-12
Extend handling of job output to consider empty responses

ricardo-espinoza:
It is valid for an Azure Quantum job to produce an empty file as a response. This change extends the logic to allow this condition, and also detect early mismatched quotes in the output.

---

### General Guidelines


yonzhan:
quantum

kairu-ms (approved):

PR #3794, opened 2021-08-19
Improve user output during error handling

ricardo-espinoza:
As part of this change we're making the user output more informative with the following changes:
- If the program fails during compilation, we'll show the full Q# compiler output so the user understands the error better.
- If the stand alone executable fails to submit, we'll show the corresponding error message so the user can fix issues with parameters.
- We'll also verify if 'dotnet' is available, and error out early if not.

---

### General Guidelines

### DIFF
--- a/src/quantum/azext_quantum/operations/job.py
+++ b/src/quantum/azext_quantum/operations/job.py
@@ -34,11 +34,30 @@ def get(cmd, job_id, resource_group_name=None, workspace_name=None, location=Non
     return client.get(job_id)
 
 
+def _check_dotnet_available():
+    """
+    Will fail if dotnet cannot be executed on the system.
+    """
+    args = ["dotnet", "--version"]
+
+    try:
+        import subprocess
+        result = subprocess.run(args, stdout=subprocess.PIPE, check=False)
+    except FileNotFoundError:
+        raise CLIError(f"Could not find 'dotnet' on the system.")
+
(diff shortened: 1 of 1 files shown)


yonzhan:
quantum

src/quantum/azext_quantum/operations/job.py
```diff
@@ -34,12 +34,31 @@ def get(cmd, job_id, resource_group_name=None, workspace_name=None, location=Non
...
+    args = ["dotnet", "--version"]
+
+    try:
+        import subprocess
+        result = subprocess.run(args, stdout=subprocess.PIPE, check=False)
+    except FileNotFoundError:
+        raise CLIError(f"Could not find 'dotnet' on the system.")
+
+    if result.returncode != 0:
+        raise CLIError(f"Failed to run 'dotnet'. (Error {result.returncode})")
```
kairu-ms, line 50: We are going the deprecate the usage of `CLIError` type. Could you following the [doc](https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md) to use a more specific error type.

kairu-ms (approved):

src/quantum/azext_quantum/operations/job.py
ricardo-espinoza: Thanks @kairu-ms . I've opened the following issue to track this: #3805

PR #3806, opened 2021-08-21
add support for scale down mode on the agentpool

marwanad:
Support scale down mode in agent pool create and update

---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
aks

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -3302,6 +3306,10 @@ def aks_agentpool_update(cmd,   # pylint: disable=unused-argument
         instance.max_count = None
 
     instance.tags = tags
+
+    if scale_down_mode is not None:
+        instance.scale_down_mode = scale_down_mode
```
FumingZhang, line 3311: It seems that the default `scale_down_mode` for `nodepool add` is delete, and when the `nodepool update` operation is performed, if this option is not specified, the `scale_down_mode` will be set to None. Is this expected?

marwanad: how ill it be set `None` if we only update the `instance.scale_down` if it's not `None`?

FumingZhang: Oh my fault, I missed the if statement. The code changes look good, could you please add/update some test cases to cover this option?

marwanad: I had a validator before but turns out its not needed because the enum checks are built in the SDK. Are you asking for live tests for the field?

FumingZhang: Yes, I mean the live test.

zhoxing-ms:
@marwanad Could you fix the live test issue and add some tests for those new parameters?

marwanad:
> ./scripts/transcribe_env.sh: line 13: AZCLI_ALT_SUBSCRIPTION_ID: unbound variable

> Error: No such container: azcli-aks-live-test


None of those errors are from this PR :)

zhoxing-ms:
@marwanad Could you add some tests for this change?

marwanad:
> @marwanad Could you add some tests for this change?

see the comment above on why a live test isn't possible :)

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -3251,11 +3255,11 @@ def aks_agentpool_update(cmd,   # pylint: disable=unused-argument
     update_autoscaler = enable_cluster_autoscaler + \
         disable_cluster_autoscaler + update_cluster_autoscaler
 
-    if (update_autoscaler != 1 and not tags and not mode and not max_surge):
+    if (update_autoscaler != 1 and not tags and not scale_down_mode and not mode and not max_surge):
```
andyliuliming, line 3258: this if worth some unit test.

andyliuliming (changes requested):

FumingZhang:
> @zhoxing-ms
> 
> > ./scripts/transcribe_env.sh: line 13: AZCLI_ALT_SUBSCRIPTION_ID: unbound variable
> 
> > Error: No such container: azcli-aks-live-test
> 
> None of those errors are from this PR :)

Hi, I reran the failed CI yesterday and it reported the above error message. Some configurations of the "AZCLI AKS LIVE TEST V2" pipeline was updated since the first queue time(08-21) and rerun would not include these updates. This should be resolved by triggering a new check. You can simply push an empty commit by 
```
git commit --allow-empty -m "Trigger Build"
git push
```

andyliuliming:
@FumingZhang  can we confirm that the live test is not related to this change?

@marwanad could you please file some tasks to track either the custom header supports, either the feature registration requirement is removed. thanks.

FumingZhang:
> @FumingZhang can we confirm that the live test is not related to this change?
The check-in pipeline "AZCLI AKS LIVE TEST V2" changed recently and the updates can only be applied by triggering a new check.

src/aks-preview/azext_aks_preview/custom.py
andyliuliming: since we have no unit test setup here. I think it's ready to go.
Fuming will setup the unit test structure so people can add ut later.

andyliuliming (approved):

zhoxing-ms (approved):

PR #3813, opened 2021-08-23
Connect to Arc Servers 

vthiebaut10:
- Add support for connecting to Arc Servers through the Arc Connectivity Platform
- Allow users to, not only connect to AAD accounts on Arc Servers and Azure VMs, but also to connect to local user accounts on the machines by using the "--local-user" parameter
- Support certificate-based, key-based, and password-based authentication when connecting to local users on the machine.
- Support creating certificate files for connecting to Arc Servers and Azure VMs.
- New command: 'az ssh arc' to connect to arc servers
- 'az ssh vm' and 'az ssh config' have support for both arc servers and Azure VMs.
- Users have the option of providing resource ID for identifying Azure VM/Arc Server.
- Users have the option of providing a custom port.
- Users can provide a path to the ssh.exe file they want to use for connecting using the --ssh-path parameter.

TO DO:

- Waiting on different team to provide a proxy that is production ready (right now we are using a beta version). 
- Waiting on a different team to provide the official Storage Account where we will download the proxy from.
- The API call to get the relay information is also temporary. That will be changed in the upcoming weeks. Currently, that call is only enabled for Arc Machines in canary regions.
- There is some server side setup necessary for connection to Arc Machines.  

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/index.json
+++ b/src/index.json
@@ -19296,6 +19296,49 @@
                     "version": "0.5.1"
                 },
                 "sha256Digest": "d09bf0befecb5f64b803a1873e666900d3b2d0ca9fe6eea5bd959a68254339a8"
+            },
+            {
+                "downloadUrl": "https://azcliprod.blob.core.windows.net/cli-extensions/storage_blob_preview-0.5.2-py2.py3-none-any.whl",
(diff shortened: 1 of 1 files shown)


yonzhan:
Arc Servers

PR #3817, opened 2021-08-25
Updating tests for quantum CLI extension to a new subscription

ricardo-espinoza:
This update to the extension tests includes:
- Using environment variables to define the resources used by the tests to enable them to run as part of the Azure Quantum E2E client test suite.
- Consolidating the resources used by the different tests in a single workspace.
- Re-enabling the create/destroy tests on live-only mode.
- Using a random name for workspaces that are created temporarily to avoid collisions.
---

### General Guidelines


yonzhan:
quantum

kairu-ms (approved):

PR #3867, opened 2021-09-06
Fix exception Handling for ARM requests

akashkeshari:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
k8s

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -736,7 +735,7 @@ def create_cc_resource(client, resource_group_name, cluster_name, cc, no_wait):
     try:
         return sdk_no_wait(no_wait, client.begin_create, resource_group_name=resource_group_name,
                            cluster_name=cluster_name, connected_cluster=cc)
-    except CloudError as e:
+    except Exception as e:
```
zhoxing-ms, line 738: `CloudError` has been replaced by `HttpResponseError` in the Python SDK of track 2. Why not use `HttpResponseError`? Is this because there are other exceptions that need to be caught?

akashkeshari: Yes, there could be other exceptions. We are catching it generically here and then handling the exception type in a common function 'arm_exception_handler'

akashkeshari: But just realized, i need to replace the CloudError from the 'arm_exception_handler' function and replace it with HttpResponseError

akashkeshari: Have replaced CloudError with HttpResponseError everywhere. Please review again

zhoxing-ms (approved):

zhoxing-ms:
@akashkeshari May I ask do you need to release a new extension version for this change? If so, please also modify the setup.py and HISTORY.rst

akashkeshari:
> @akashkeshari May I ask do you need to release a new extension version for this change? If so, please also modify the setup.py and HISTORY.rst

@zhoxing-ms , Both the setup and History files have already been updated to 1.1.9 in a previous merge. The current released version is 1.1.8. We wanted to take up both changes in the same release

PR #3889, opened 2021-09-15
fix(aks): check parent scopes role assignments in pod identity assignment

### LINKED ISSUE #3679: ERROR: Could not grant Managed Identity Operator permission for cluster
aks-preview

## Bug description
While running `az aks pod-identity --resource group $AKS_RG --cluster-name $AKS_NAME --namespace $NS --name $name --identity-resource-id /subscriptions/$SUBID/resourceGroups/$UAI_RG/providers/Microsoft.ManagedIdentity/userAssignedIdentities/$IDENTITY_NAME ` on a cluster with configured with User-Assigned Managed Identity there is a check done to verify that the AKS Identity has "Managed Identity Operator" role on the Identity we want to include as pod identity. The issue is that it is not enough with having the role on the Resource group of the Identity ($UAI_RG) but you must have it specifically on the Identity resource $IDENTITY_NAME. 

This should be classified as a bug since a resource inherits from its resource group, but the script does take inheritance in account but instead requires the resource to have the role assignment.

### To reproduce:
- Existing AKS Cluster (v1.21.1) running with User-Assigned Managed Identity
- Identity (that you want to add as Pod Identity) created in separate resource group
- Run az aks pod-identity add command on AKS cluster 

### Workarounds:
1.  Assign "Managed Identity Operator" role specifically on resource $IDENTITY_NAME ($UAI_RG is not enough)
2. Assign "User Access Administrator" role for the AKS UAI on $UAI_RG, thus enabling AKS UAI to grant "Managed Identity Operator" role on the resource
-----

bcho:
In previous `pod identity add`'s implementation, we check for role assignment with full scope match, which is an incorrect assumption to the [Azure RBAC assignments](https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-steps#step-3-identify-the-needed-scope). 

This pull request fixed the pod identity role assignment check as follows:

1. updated the role assignments list filter to `atScope() and assignedTo('{cluster_identity_id}')`. This could give us all role assignments targeted for the cluster identity
2. fixed scope check logic: we check with prefixed resource id

Fix #3679

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/azext_aks_preview/custom.py
+++ b/src/aks-preview/azext_aks_preview/custom.py
@@ -4184,15 +4184,25 @@ def _ensure_managed_identity_operator_permission(cli_ctx, instance, scope):
 
     factory = get_auth_management_client(cli_ctx, scope)
     assignments_client = factory.role_assignments
+    cluster_identity_object_id = cluster_identity_object_id.lower()
+    scope = scope.lower()
 
-    for i in assignments_client.list_for_scope(scope=scope, filter='atScope()'):
-        if i.scope.lower() != scope.lower():
-            continue
+    # list all assignments of the target identity (scope) that assigned to the cluster identity
+    filter_query = "atScope() and assignedTo('{}')".format(cluster_identity_object_id)
+    for i in assignments_client.list_for_scope(scope=scope, filter=filter_query):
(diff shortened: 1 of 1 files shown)


yonzhan:
aks

zhoxing-ms (approved):

PR #3900, opened 2021-09-17
Add quantum version check

warren-jones:
Once per day, the Azure CLI quantum extension will perform a version check as part of any command.  If the latest available version of the extension is greater than the one installed, the user will see a yellow warning message in the following format:
```
Version 0.7.0 of the quantum extension is installed locally, but version 0.8.0 is now available.
You can use 'az extension update -n quantum' to upgrade.
```
If the latest available version number cannot be retrieved, or it matches the installed version, no message is shown. We won’t fail the command if we can’t get the most recent version, we’ll just proceed.  We won’t check again until the date changes or the context gets cleared.

This implementation will not cover commands that our extension does not handle directly, like “--help” calls, since they are directly responded by the ‘az’ tool.

### DIFF
--- a/src/quantum/azext_quantum/__init__.py
+++ b/src/quantum/azext_quantum/__init__.py
@@ -4,6 +4,8 @@
 # --------------------------------------------------------------------------------------------
 
 from azure.cli.core import AzCommandsLoader
+from datetime import datetime
+from ._version_check_helper import check_version
 
 import azext_quantum._help  # pylint: disable=unused-import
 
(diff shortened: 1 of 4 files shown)


yonzhan:
quantum

kairu-ms (approved):

kairu-ms (changes requested):
`__init__` function of `CommandLoader` will be run when any cli command is executed. Could you move version check logic in `load_arguments` function so that it only runs when Quantum command is executed.

ricardo-espinoza:
> `__init__` function of `CommandLoader` will be run when any cli command is executed. Could you move version check logic in `load_arguments` function so that it only runs when Quantum command is executed.

Thanks @kairu-ms . Just one question... this change is only in the QuantumCommandLoader, why would this be executed if a non-quantum cli command is executed?

warren-jones:
Moved version-check logic to load_arguments, as requested by @kairu-ms

ricardo-espinoza (approved):

kairu-ms:
> > `__init__` function of `CommandLoader` will be run when any cli command is executed. Could you move version check logic in `load_arguments` function so that it only runs when Quantum command is executed.
> 
> Thanks @kairu-ms . Just one question... this change is only in the QuantumCommandLoader, why would this be executed if a non-quantum cli command is executed?

Because command loader will be create when CLI try to find a command, it can be any command.

yonzhan:
Please fix the CI first.

warren-jones:
Converted this PR to "Draft" until I finish adding the tests

warren-jones:
Moved most of the version check logic to a helper module to facilitate unit tests.  Added unit tests.

kairu-ms (approved):

Issue #3805, opened 2021-08-21
Quantum extension is using deprecated CLIError() type


ricardo-espinoza:
Per recommendation of the core Az CLI team, we'll be updating this type per the new guidelines here:

https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md

yonzhan:
route to service team

warren-jones:
This issue has been resolved by PR #3997

PR #3904, opened 2021-09-21
Add support for target specific job parameters

ricardo-espinoza:
This PR adds an optional parameter `--job-params` to the following commands.

+ `az quantum job submit`
+ `az quantum execute`
+ `az quantum run`

This parameter contains a ist of `key=value` pairs that will be passed to the Quantum execution target as part of the job submission. The support for those parameters is specific to each target.

---

### General Guidelines

### DIFF
--- a/src/quantum/azext_quantum/_help.py
+++ b/src/quantum/azext_quantum/_help.py
@@ -21,7 +21,7 @@ helps['quantum execute'] = """
       - name: Submit and wait for a Q# program from the current folder with job and program parameters.
         text: |-
             az quantum execute -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget \\
-                --job-params key1=value1 key2=value2 -- --n-qubits 3
+                --job-params key1=value1 key2=value2 -- --n-qubits=3
 """
 
 helps['quantum run'] = """
@@ -34,7 +34,7 @@ helps['quantum run'] = """
       - name: Submit and wait for a Q# program from the current folder with job and program parameters.
         text: |-
             az quantum run -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget \\
-                --job-params key1=value1 key2=value2 -- --n-qubits 3
+                --job-params key1=value1 key2=value2 -- --n-qubits=3
 """
 
 helps['quantum job'] = """
@@ -86,7 +86,7 @@ helps['quantum job submit'] = """
       - name: Submit the Q# program with program parameters (e.g. n-qubits = 2).
         text: |-
             az quantum job submit -g MyResourceGroup -w MyWorkspace -l MyLocation \\
-               --job-name MyJob -- --n-qubits 2
+               --job-name MyJob -- --n-qubits=2
 """
 
 helps['quantum job wait'] = """
--- a/src/quantum/azext_quantum/_help.py
+++ b/src/quantum/azext_quantum/_help.py
@@ -18,6 +18,10 @@ helps['quantum execute'] = """
       - name: Submit the Q# program from the current folder and wait for the result.
         text: |-
             az quantum execute -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget
+      - name: Submit and wait for a Q# program from the current folder with job and program parameters.
+        text: |-
+            az quantum execute -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget \\
+                --job-params key1=value1 key2=value2 -- --n-qubits 3
 """
 
 helps['quantum run'] = """
@@ -27,6 +31,10 @@ helps['quantum run'] = """
       - name: Submit the Q# program from the current folder and wait for the result.
         text: |-
             az quantum run -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget
+      - name: Submit and wait for a Q# program from the current folder with job and program parameters.
+        text: |-
+            az quantum execute -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget \\
+                --job-params key1=value1 key2=value2 -- --n-qubits 3
 """
 
 helps['quantum job'] = """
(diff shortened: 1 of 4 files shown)


src/quantum/azext_quantum/_params.py
```diff
@@ -4,8 +4,20 @@
...
+# pylint: disable=protected-access
+class JobParamsAction(argparse._AppendAction):
+    def __call__(self, parser, namespace, values, option_string):
+        params = {}
+        for item in values:
+            try:
+                key, value = item.split('=', 1)
+                params[key] = value
+            except ValueError:
+                raise CLIError('Usage error: {} KEY=VALUE [KEY=VALUE ...]'.format(option_string))
```
ricardo-espinoza, line 19: This usage should be deprecated, but it's added for consistency with the rest of the extension. The following item tracks the pending work to replace it across the entire extension:
https://github.com/Azure/azure-cli-extensions/issues/3805

src/quantum/azext_quantum/_help.py
```diff
@@ -18,6 +18,10 @@
       - name: Submit the Q# program from the current folder and wait for the result.
         text: |-
             az quantum execute -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget
+      - name: Submit and wait for a Q# program from the current folder with job and program parameters.
+        text: |-
+            az quantum execute -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget \\
+                --job-params key1=value1 key2=value2 -- --n-qubits 3
```
kairu-ms, line 24: I think there's redundant `--` in the example
```suggestion
                --job-params key1=value1 key2=value2 --n-qubits 3
```

src/quantum/azext_quantum/_help.py
```diff
@@ -27,6 +31,10 @@
       - name: Submit the Q# program from the current folder and wait for the result.
         text: |-
             az quantum run -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget
+      - name: Submit and wait for a Q# program from the current folder with job and program parameters.
+        text: |-
+            az quantum run -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget \\
+                --job-params key1=value1 key2=value2 -- --n-qubits 3
```
kairu-ms, line 37:
```suggestion
                --job-params key1=value1 key2=value2 --n-qubits 3
```

src/quantum/azext_quantum/_help.py
```diff
@@ -71,6 +79,14 @@
...
             az quantum job submit -g MyResourceGroup -w MyWorkspace -l MyLocation \\
                --job-name MyJob
+      - name: Submit the Q# program from the current folder with job parameters for a target.
+        text: |-
+            az quantum job submit -g MyResourceGroup -w MyWorkspace -l MyLocation \\
+               --job-name MyJob --job-params param1=value1 param2=value2
+      - name: Submit the Q# program with program parameters (e.g. n-qubits = 2).
+        text: |-
+            az quantum job submit -g MyResourceGroup -w MyWorkspace -l MyLocation \\
+               --job-name MyJob -- --n-qubits 2
```
kairu-ms, line 89:
```suggestion
               --job-name MyJob --n-qubits 2
```

src/quantum/azext_quantum/_help.py
ricardo-espinoza: @kairu-ms, ah, that's intentional. Because of legacy reasons, we have a `--` separating the CLI parameters from a positional argument that includes all the parameters that get sent to the executable itself being run in Azure Quantum. So the `--` is the existing delimiter to separate CLI extension parameters from the submitted .exe parameters.

This has been there since version 1, however, I'm driving for a redesign in future versions to do something similar to --job-params instead.

src/quantum/azext_quantum/_help.py
ricardo-espinoza: See response above.

src/quantum/azext_quantum/_help.py
ricardo-espinoza: See response above.

src/quantum/azext_quantum/_params.py
```diff
@@ -48,6 +66,7 @@ def load_arguments(self, _):
         c.argument('max_poll_wait_secs', max_poll_wait_secs_type)
 
     with self.argument_context('quantum job submit') as c:
+        c.argument('job_params', job_params_type)
```
jsntcy, line 69: Do you have tests to cover this new argument?

src/quantum/azext_quantum/_help.py
jsntcy: @ricardo-espinoza 
It's weird to add extra `--` here and it's difficult for users to understand the argument usage.
BTW, could you point us where you handle this extra `--` in code?

src/quantum/azext_quantum/_params.py
ricardo-espinoza: Following up offline on this. This argument is still not handled by the providers in our service, so there would be no effect on the tests recorded. There are tests on the next layer (our runtime) to that we exercise by calling the CLI, but we're only a pass through at this point.

Currently we can only exercise this path in our E2E suite, since we can't do job submissions without the Q# compiler/runtime and we haven't onboarded it to tests in this repo.

src/quantum/azext_quantum/_help.py
```diff
@@ -27,6 +31,10 @@
       - name: Submit the Q# program from the current folder and wait for the result.
         text: |-
             az quantum run -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget
+      - name: Submit and wait for a Q# program from the current folder with job and program parameters.
+        text: |-
+            az quantum run -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget \\
+                --job-params key1=value1 key2=value2 -- --n-qubits=3
```
jsntcy, line 37: [2:20 PM] Yu Chen
It seems the help is not consistent with the example in help it says "specified as --name=value after --", but in example, it uses as "-- --n-qubits 3" (not --name=value format)


image
image

ricardo-espinoza: This has been fixed in a more recent update, thanks Yu!

src/quantum/azext_quantum/_params.py
ricardo-espinoza: In the last commit of the PR, the generation of args using this new parameter is being tested.

src/quantum/azext_quantum/_help.py
ricardo-espinoza: This information has been provided offline. The usage of `--` is non-standard and I agree we should change it at some point, but it has been there since the first version, and is probably out of the scope of this PR.

kairu-ms (approved):

PR #3908, opened 2021-09-22
Quantum: Trim result for microsoft.simulator job only when it is a string.

ricardo-espinoza:
This is a change from @cesarzc that fixes an output processing issue for some Azure Quantum results that are not strings.

---

### General Guidelines


yonzhan:
Quantum

kairu-ms (approved):

PR #3929, opened 2021-09-29
Connect to Arc Servers

vthiebaut10:
- Add support for connecting to Arc Servers through the Arc Connectivity Platform (https://msazure.visualstudio.com/One/_wiki/wikis/One.wiki/121996/Arc-Connectivity-Platform)
- Allow users to, not only connect to AAD accounts on Arc Servers and Azure VMs, but also to connect to local user accounts on the machines by using the "--local-user" parameter
- Support certificate-based, key-based, and password-based authentication when connecting to local users on the machine.
- Support creating config files for connecting to Arc Servers and Azure VMs.
- New command: 'az ssh arc' to connect to arc servers
- 'az ssh vm' and 'az ssh config' have support for both arc servers and Azure VMs.
- Users have the option of providing resource ID for identifying Azure VM/Arc Server.
- Users have the option of providing a custom port.
- Users can provide a path to the ssh.exe file they want to use for connecting using the --ssh-client-path parameter.
- '--delete-private-key' parameter is an internal hidden parameter meant to be used only by the portal team to provide a safe one-click ssh experience using this extension.
---

### General Guidelines

    - Note: two checks are failing because code is catching non-specified exceptions, but that is by design though. What should I do about that?

For new extensions:

### About Extension Publish


yonzhan:
Arc

Issue #3766, opened 2021-08-10
AKS Periscope breaking change: Need to supply ContainerName to AKS Periscope from future release v0.5 onwards 


qpetraroia:
### Extension name (the extension in question)
AKS Periscope

### Description of issue (in as much detail as possible)

AKS Periscope used to read the container name with a hardcoded Kubeconfig. This [PR ](https://github.com/Azure/aks-periscope/pull/96) and this [issue ](https://github.com/Azure/aks-periscope/issues/99) discuss the breaking changes that will occur to each service that has a working contract with AKS Periscope. The removal of the hardcoded code in Periscope will allow for other tools to onboard onto AKS Periscope easily.

Azure-cli-extensions is one service that has a working contract with AKS Periscope. To keep the default behavior as it is pass the API Server Name as the ContainerName.

-----
CC: @Tatsinnit and @sophsoph321

yonzhan:
route to service team

Tatsinnit:
☕️ please note this work is merged as part of #3971 , please feel free to close this tracking issue or if he ever a newer version gets released. Thanks !

PR #3971, opened 2021-10-15
[AKS] Handle containerName passing and edge case for containerName.

Tatsinnit:
### What this work entails:

This PR fixes this open workitem for `cli` details under #3766 

Previously the code use to form `containerName` after the `kubectl` command was run, now it is formed earlier and then supplied through.

Thanks,

cc: @rzhang628 and @qpetraroia 

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/azext_aks_preview/custom.py
+++ b/src/aks-preview/azext_aks_preview/custom.py
@@ -1864,13 +1864,23 @@ def aks_kollect(cmd,    # pylint: disable=too-many-statements,too-many-locals
     print()
     print("Starts collecting diag info for cluster %s " % name)
 
+    # Form containerName from fqdn, as it was previosuly jsut the location of code is changed.
+    # https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names 
+    maxContainerNameLength = 63
+    fqdn = mc.fqdn if mc.fqdn is not None else mc.private_fqdn
+    normalized_container_name = fqdn.replace('.', '-')
+    len_of_container_name = normalized_container_name.index("-hcp-")
+ 	if  len_of_container_name == -1 {
+ 		len_of_container_name = maxContainerNameLength
+ 	}
+    containerName = containerName[:len_of_container_name]
+
     sas_token = sas_token.strip('?')
-    deployment_yaml = urlopen(
-        "https://raw.githubusercontent.com/Azure/aks-periscope/latest/deployment/aks-periscope.yaml").read().decode()
-    deployment_yaml = deployment_yaml.replace("# <accountName, base64 encoded>",
-                                              (base64.b64encode(bytes(storage_account_name, 'ascii'))).decode('ascii'))
+    deployment_yaml = _read_periscope_yaml()
+    deployment_yaml = deployment_yaml.replace("# <accountName, base64 encoded>", storage_account_name)
     deployment_yaml = deployment_yaml.replace("# <saskey, base64 encoded>",
                                               (base64.b64encode(bytes("?" + sas_token, 'ascii'))).decode('ascii'))
+    deployment_yaml = deployment_yaml.replace("# <containerName, string>", containerName)
 
     yaml_lines = deployment_yaml.splitlines()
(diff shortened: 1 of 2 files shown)


src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1954,6 +1962,16 @@ def aks_kollect(cmd,    # pylint: disable=too-many-statements,too-many-locals
         display_diagnostics_report(temp_kubeconfig_path)
 
 
+def _read_periscope_yaml():
+    with open("./deploymentyaml/aks-periscope.yaml", "r") as stream:
```
FumingZhang, line 1966: I am not sure if the relative path here will work and my suggestion is to form an absolute path like
```suggestion
    curr_dir = os.path.dirname(os.path.realpath(__file__))
    periscope_yaml = os.path.join(curr_dir, "deploymentyaml", "aks-periscope.yaml")
    with open(periscope_yaml, "r") as stream:
```

Tatsinnit:
opening this for PR and merge. Thank you @FumingZhang, got this working locally. ☕️

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1864,13 +1864,22 @@ def aks_kollect(cmd,    # pylint: disable=too-many-statements,too-many-locals
     print()
     print("Starts collecting diag info for cluster %s " % name)
 
+    # Form containerName from fqdn, as it was previously jsut the location of code is changed.
+    # https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata#container-names
+    maxContainerNameLength = 63
+    fqdn = mc.fqdn if mc.fqdn is not None else mc.private_fqdn
+    normalized_container_name = fqdn.replace('.', '-')
+    len_of_container_name = normalized_container_name.index("-hcp-")
```
FumingZhang, line 1872: I am not sure if the part in fqdn before "hcp" would be longer than 63, if so, this may break the limit.

Tatsinnit: https://docs.microsoft.com/en-us/azure/aks/troubleshooting#what-naming-restrictions-are-enforced-for-aks-resources-and-parameters This might help.

essentially this makes sure that in fqdn for cluster we take all character before “hpc” although the doc kind of guarantees that cluster name can’t be more then 63 character but just to be safe we take 63 in case command line or direct ARM has a loop. 

Also container name cannot be more then 63 character as well.

FumingZhang: I see, but without the background knowledge about the limit of the cluster name, what I am concerning about is that the code might get a container name longer than 63 if the fqdn is not valid (through this should not happen).

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1864,13 +1864,22 @@ def aks_kollect(cmd,    # pylint: disable=too-many-statements,too-many-locals
...
+    if len_of_container_name == -1:
+        len_of_container_name = maxContainerNameLength
```
FumingZhang, line 1874: Just an example, in this way, no matter what's the limitation about cluster name or fqdn, we will always get a container name no longer than 63.

```suggestion
    if len_of_container_name == -1:
        len_of_container_name = maxContainerNameLength
    else:
        len_of_container_name = min(len_of_container_name, maxContainerNameLength)
```

src/aks-preview/azext_aks_preview/custom.py
Tatsinnit: I think this code will only ever trigger for privatelink clusters because as the link shared above normal cluster name cannot be more then 63 but in case of privatelink cluster when they are formed the final fqdn contains “privatelink” which is nomenclature standard,

FumingZhang: Okay, since we have a strong restriction for the naming, I think the current implementation is all good.

FumingZhang (approved):

zhoxing-ms (approved):

PR #3991, opened 2021-10-19
[k8s-extension] Release for GA v1.0.0

jonathan-innis:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/k8s-extension/azext_k8s_extension/custom.py
+++ b/src/k8s-extension/azext_k8s_extension/custom.py
@@ -163,11 +163,14 @@ def update_k8s_extension(cmd, client, resource_group_name, cluster_name, name, c
     """Patch an existing Extension Instance.
     """
 
-    if configuration_settings or configuration_protected_settings or configuration_settings_file or configuration_protected_settings_file:
+    if configuration_settings or \
+       configuration_protected_settings or \
+       configuration_settings_file or \
+       configuration_protected_settings_file:
         msg = ('Updating properties in --config-settings or --config-protected-settings may lead to undesirable state'
-            ' if the cluster extension type does not support it. Please refer to the documentation of the'
-            ' cluster extension service to check if updates to these properties is supported.'
-            ' Do you wish to proceed?')
+               ' if the cluster extension type does not support it. Please refer to the documentation of the'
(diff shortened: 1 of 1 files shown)


jonathan-innis:
> Could you add some new tests for the new command and new changes?

We have some E2E tests that we run locally which don't ship with the PR. Is there a good way to perform mocking on this call rather than a scenario test? Since the request/response will not be consistent for these update calls when operations are `InProgress`

zhoxing-ms:
> Is there a good way to perform mocking on this call rather than a scenario test? Since the request/response will not be consistent for these update calls when operations are InProgress

@jonathan-innis  May I ask how long will those operations last in phase `InProgress`? If it doesn't take too long, could we wait for the completion of the operation through `az resource wait` or `time.sleep(xxx)`?

zhoxing-ms:
@NarayanThiru Thank you for your detailed clarification. I would like to ask what problems will be caused by setting up a K8s cluster in the test?

NarayanThiru:
> > Since we cannot set up a K8s cluster as part of the testing, we are not adding Patch scenario tests here.
> 
> @NarayanThiru Thank you for your detailed clarification. I would like to ask what problems will be caused by setting up a K8s cluster in the test?

Setting up a K8s cluster involves not just creating one, but also projecting it to Azure, which requires using another RP's CLI, which is an additional dependency.  In addition, this projection involves creating MSI Identity for the cluster, which can take up to 10 minutes to provision Certificate that is required for Authn.  After all of this is done, the Cluster should connect to our DataPlane and start heartbeating.  Only after this, can we create the Extension, which gets created only in the Service.  This then gets picked up by the Agent in the cluster and installed.  Then it sends back the status to the Service (DataPlane).  This can take multiple minutes depending on the connectivity and other delays.  But, most importantly these are all steps that are not under test.  This is why we have created a pipeline for all our CLI changes that exercises all these steps in our private fork.  It is only after it succeeds in that, do we push to main.   So, if you concern is that this is untested code, please rest assured, we test it thoroughly before pushing.  In addition, as I explained earlier, it continues to get tested in Production as well - as frequently as every 15 minutes.  I hope this gives details on why we don't want to do this.

PR #4077, opened 2021-11-08
aks: fix wrong behavior when applying default value for pod identity exception pod labels.

bcho:
In previous version, we accidentally allowed user to specify empty pod labels for pod identity exception, which will be converted to `None` in response. As a result, this behavior will break the exception when using 2021-09-01 model. This is because we marked the `podLabels` as required while updating it with `None` value.

To unblock customer, this pull request added support for backfilling the null `podLabels` before sending to the server side. In the meanwhile, AKS will apply change to fix the response value.

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/HISTORY.md
+++ b/src/aks-preview/HISTORY.md
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+0.5.41
+++++++
+* Fix default value behavior for pod identity esception pod labels.
+
 0.5.40
 +++++
 * Add support for new snapshot commands
--- a/src/aks-preview/HISTORY.md
+++ b/src/aks-preview/HISTORY.md
@@ -4,7 +4,7 @@ Release History
 ===============
 0.5.41
 ++++++
-* Fix default value behavior for pod identity esception pod labels.
+* Fix default value behavior for pod identity exception pod labels.
 
 0.5.40
 +++++
--- a/src/aks-preview/azext_aks_preview/custom.py
+++ b/src/aks-preview/azext_aks_preview/custom.py
@@ -3741,12 +3741,27 @@ def _ensure_pod_identity_kubenet_consent(network_profile, pod_identity_profile, 
     pod_identity_profile.allow_network_plugin_kubenet = True
 
 
+def _fill_defaults_for_pod_identity_exceptions(pod_identity_exceptions):
+    if not pod_identity_exceptions:
+        return
+
+    for exc in pod_identity_exceptions:
+        if exc.pod_labels is None:
+            # in previous version, we accidentally allowed user to specify empty pod labels,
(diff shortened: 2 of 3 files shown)


yonzhan:
aks

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -3741,13 +3741,28 @@ def _ensure_pod_identity_kubenet_consent(network_profile, pod_identity_profile,
     pod_identity_profile.allow_network_plugin_kubenet = True
 
 
+def _fill_defaults_for_pod_identity_exceptions(pod_identity_exceptions):
```
andyliuliming, line 3744: better to add some unit test.
@fumingzhang please share some samples.

FumingZhang: The PR looks good and it seems that the reason for the error could not be covered by the unit tests on the cli side.

But I still agree with liming, it might be difficult to write unit tests for entry functions like `aks_pod_identity_exception_add` or `aks_pod_identity_exception_update` since you may need to mock the command context, client and some requests. However it should be easy to test functions like `_update_addon_pod_identity` or `_fill_defaults_for_pod_identity_exceptions`, and these internal functions are used by many other functions. It is recommended to write unit test cases for pod identity related functions in a separate file.

FumingZhang (approved):

src/aks-preview/HISTORY.md
```diff
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+0.5.41
+++++++
```
wangzelin007, line 6: Should be exception.

wangzelin007 (approved):

FumingZhang (changes requested):
We already have lots of unit test cases in `test_decorator.py` which are written for `aks create`. If you want to write test cases for entry functions, may use the classes provided in `mocks.py`.

src/aks-preview/HISTORY.md
```diff
@@ -2,6 +2,10 @@
...
+* Fix default value behavior for pod identity esception pod labels.
```
wangzelin007, line 7: Should be exception.

wangzelin007 (changes requested):

src/aks-preview/azext_aks_preview/custom.py
bcho: I would suggest we address these issues in follow up PRs. Current change is not related to these proposed tests.

src/aks-preview/HISTORY.md
bcho: Updated, PTAL!

PR #4083, opened 2021-11-08
allow user to do empty az aks update to reconcile cluster

paulgmiller:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/azext_aks_preview/custom.py
+++ b/src/aks-preview/azext_aks_preview/custom.py
@@ -31,7 +31,7 @@ from six.moves.urllib.error import URLError  # pylint: disable=import-error
 import requests
 from knack.log import get_logger
 from knack.util import CLIError
-from knack.prompting import prompt_pass, NoTTYException
+from knack.prompting import prompt_pass, NoTTYException, prompt_y_n
 
 import yaml  # pylint: disable=import-error
 from dateutil.relativedelta import relativedelta  # pylint: disable=import-error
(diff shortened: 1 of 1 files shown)


src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1273,27 +1275,12 @@ def aks_update(cmd,     # pylint: disable=too-many-statements,too-many-branches,
...
-                       '"--aad-admin-group-object-ids" or '
-                       '"--enable-ahub" or '
-                       '"--disable-ahub" or'
-                       '"--enable-managed-identity"')
-
-    instance = client.get(resource_group_name, name)
+        if prompt_y_n("no arguments specicied would you like to reconcile?", default="n"):
+            #duplicating this worse than goto? 
+            headers = get_aks_custom_headers(aks_custom_headers)
+            return sdk_no_wait(no_wait, client.create_or_update, resource_group_name, name, instance, custom_headers=headers)
```
xhl873, line 1281: the instance here should be the datat model after Get, so it may be slightly different than am empty put? IMO, it should behave the same as empty put since nothing is changed, but not sure if there will be surprise or not :P

yonzhan:
aks

src/aks-preview/azext_aks_preview/custom.py
FumingZhang, line 1281: It seems that the change is made on a really really old code base, I suppose you need to update it to resolve some conflicts. Besides the function signature of `create_or_update` has changed since we adopted the track 2 SDK, thus you need to change the name of the argument `custom_headers` to `headers`.

paulgmiller:
replacing with speraet az aks reconcile command

PR #4110, opened 2021-11-15
[Quantum] Do not display update message if version numbers are not available

warren-jones:
Added logic to suppress the update message if either version number is not available, and modified the unit tests accordingly. 

This bug fix primarily affects the presentation of the update message during testing, and as such will not change the user experience.  However, in the rare event of a transient connectivity glitch during version detection, this bug fix has the added benefit of preventing the user from ever seeing a spurious "None" in place of a valid version number in the message.

---

### General Guidelines


yonzhan:
Quantum

ricardo-espinoza (approved):

kairu-ms (approved):

PR #4136, opened 2021-11-21
[Release] Update index.json for extension [ vmware ] 

azclibot:
[Release] Update index.json for extension [ vmware ] 

 Triggered by Azure CLI Extensions Release Pipeline - ADO_BUILD_ID=1206850

Last commit against main: https://github.com/Azure/azure-cli-extensions/commit/e203fc938411f4e3e53e13e392d6334b4137f3e3

PR #4167, opened 2021-11-30
publish vmware 4.0.1

cataggar:
There was a problem publishing vmware 4.0.0 #4136.  I think it was caused by removing `__init__.py` in #3992. Pretty sure I made the same mistake before. cc @jonathanhe-msft


yonzhan:
vmware

zhoxing-ms (approved):

cataggar:
Publishing worked this time. #4168

PR #4174, opened 2021-11-30
[AKS] enable `aks create` dual-stack clusters

tyler-lloyd:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
AKS

tyler-lloyd:
this is a branch of https://github.com/Azure/azure-cli-extensions/pull/4165

#4165 should be completed first, then this.

src/aks-preview/azext_aks_preview/_loadbalancer.py
```diff
@@ -26,48 +26,65 @@ def set_load_balancer_sku(sku, kubernetes_version):
...
+            else:
+                # ips -> i_ps due to track 2 naming issue
+                profile.managed_outbound_i_ps = ManagedClusterLoadBalancerProfileManagedOutboundIPs(
+                    count_ipv6=managed_outbound_ipv6_count
+                )
         if managed_outbound_ip_count:
             # ips -> i_ps due to track 2 naming issue
             profile.managed_outbound_i_ps = ManagedClusterLoadBalancerProfileManagedOutboundIPs(
                 count=managed_outbound_ip_count
             )
```
FumingZhang, line 79: The value (`count_ipv6`) might be overwritten when user set `--load-balancer-managed-outbound-ip-count` and `--load-balancer-managed-outbound-ipv6-count` at the same time. I am not sure if this is expected.

src/aks-preview/azext_aks_preview/_loadbalancer.py
```diff
@@ -26,48 +26,65 @@ def set_load_balancer_sku(sku, kubernetes_version):
...
         profile.outbound_i_ps = None
         profile.outbound_ip_prefixes = None
+        if managed_outbound_ipv6_count:
+            if profile.managed_outbound_i_ps:
+                profile.managed_outbound_i_ps.count_ipv6 = managed_outbound_ipv6_count
+            else:
+                # ips -> i_ps due to track 2 naming issue
+                profile.managed_outbound_i_ps = ManagedClusterLoadBalancerProfileManagedOutboundIPs(
+                    count_ipv6=managed_outbound_ipv6_count
+                )
```
FumingZhang, line 74: This part looks exactly the same as line 80-87. I think you can add remove one of them and add some similar checks to the `managed_outbound_ip_count` part.

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -1301,6 +1305,43 @@ def set_up_network_profile(self, mc: ManagedCluster) -> ManagedCluster:
...
+        if ip_families:
+            network_profile.ip_families = ip_families.split(',')
+
+        pod_cidrs = self.context.raw_param.get('pod_cidrs')
+        if pod_cidrs:
+            network_profile.pod_cidrs = pod_cidrs.split(',')
+
+        service_cidrs = self.context.raw_param.get('service_cidrs')
+        if service_cidrs:
+            network_profile.service_cidrs = service_cidrs.split(',')
```
FumingZhang, line 1319: I would appreciate it if you could write corresponding "getter" functions under class `AKSPreviewContext` for these new parameters. An example would be [`_get_aad_admin_group_object_ids`](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acs/decorator.py#L3381) from azure-cli. You don't need to write the internal function (function name starts with `_`) since you do not have extra validations for these parameters. Please do not modify the value if it is obtained from the `mc` object. The detailed explanation for this is as follows

> In the update mode, the parameter value is not obtained from `mc` because we just need to use the value provided by the user to overwrite the value in `mc`.
> 
> But in the create mode, `mc` is completely blank, and we want to assemble the user's input into a complete `mc`. Considering that our parameters have a certain dependency or mutual exclusion relationship, or the parameters will be normalized before putting into `mc`, there are even some parameters that do not need user to provide explicitly, but can be dynamically completed by cli.
> 
> In order to ensure the consistency of the state and make the value of the parameter more clearly determined, we created the following principle in the create mode, that is, if a parameter has been placed in the `mc` object, then when trying to get the parameter value again, (e.g. cross-validation with other parameters) the value in `mc` shall prevail.

Also in this way, we can write unit test cases for these "getter" functions in file `test_decorator.py`.

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -1301,6 +1305,43 @@ def set_up_network_profile(self, mc: ManagedCluster) -> ManagedCluster:
...
+
+        ipv6_count = self.context.raw_param.get('load_balancer_managed_outbound_ipv6_count')
+        if ipv6_count:
+            print('pre + ' + str(network_profile))
```
FumingZhang, line 1323: Please remove `print` or replace it with `logger.debug` before merging this PR.

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -1301,6 +1305,43 @@ def set_up_network_profile(self, mc: ManagedCluster) -> ManagedCluster:
...
+            else:
+                network_profile.load_balancer_profile = create_load_balancer_profile(
+                    self.context.get_load_balancer_managed_outbound_ip_count(),
+                    ipv6_count,
+                    self.context.get_load_balancer_outbound_ips(),
+                    self.context.get_load_balancer_outbound_ip_prefixes(),
+                    self.context.get_load_balancer_outbound_ports(),
+                    self.context.get_load_balancer_idle_timeout(),
+                )
+            print('post + ' + str(network_profile))
```
FumingZhang, line 1343: Also this `print`.

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -1301,6 +1305,43 @@ def set_up_network_profile(self, mc: ManagedCluster) -> ManagedCluster:
...
+
+        pod_cidrs = self.context.raw_param.get('pod_cidrs')
+        if pod_cidrs:
+            network_profile.pod_cidrs = pod_cidrs.split(',')
+
+        service_cidrs = self.context.raw_param.get('service_cidrs')
+        if service_cidrs:
+            network_profile.service_cidrs = service_cidrs.split(',')
+
+        ipv6_count = self.context.raw_param.get('load_balancer_managed_outbound_ipv6_count')
```
FumingZhang, line 1321: Please also write "getter" for this parameter.

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -1301,6 +1305,43 @@ def set_up_network_profile(self, mc: ManagedCluster) -> ManagedCluster:
...
+                network_profile.load_balancer_profile = create_load_balancer_profile(
+                    self.context.get_load_balancer_managed_outbound_ip_count(),
+                    ipv6_count,
+                    self.context.get_load_balancer_outbound_ips(),
+                    self.context.get_load_balancer_outbound_ip_prefixes(),
+                    self.context.get_load_balancer_outbound_ports(),
+                    self.context.get_load_balancer_idle_timeout(),
+                )
+            print('post + ' + str(network_profile))
+
```
FumingZhang, line 1344: We also have unit test cases for the `set_up_xxx` functions in `AKSPreviewDecorator`, please extend `test_set_up_network_profile` in file `test_decorator.py`.

src/aks-preview/azext_aks_preview/_loadbalancer.py
tyler-lloyd: yep, just left over

src/aks-preview/azext_aks_preview/decorator.py
tyler-lloyd: added getters. Although I would like to get your opinion on the behavior. I was not aware the SDK provided default values for pod_cidr and service_cidr. For new functionality we want customers to be able to just do `az aks create --pod-cidrs XYZ,ABC --service-cidrs FOO,BAR` where these do not match the defaults for pod-cidr and service-cidr. Therefore, I've made the change here to overwrite whatever is in pod_cidr and service_cidr with pod_cidrs[0] and service_cidrs[0] if it exists. This would be required as RP validates this anyway.

src/aks-preview/azext_aks_preview/decorator.py
tyler-lloyd: removed

src/aks-preview/azext_aks_preview/decorator.py
tyler-lloyd: removed

src/aks-preview/azext_aks_preview/decorator.py
tyler-lloyd: added

src/aks-preview/azext_aks_preview/decorator.py
tyler-lloyd: added unit tests in test_decorator and in test_aks_commands

src/aks-preview/azext_aks_preview/_loadbalancer.py
```diff
@@ -68,6 +69,14 @@ def configure_load_balancer_profile(managed_outbound_ip_count, outbound_ips, out
             profile.managed_outbound_i_ps = ManagedClusterLoadBalancerProfileManagedOutboundIPs(
                 count=managed_outbound_ip_count
```
FumingZhang, line 70: Should also add check for `managed_outbound_ip_count`. In an update scenario, user only specifies `managed_outbound_ip_count` this time, but we have `managed_outbound_ipv6_count` set when the cluster was created, this would overwrite `managed_outbound_ipv6_count` to `None`. Is this expected?

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -743,6 +747,45 @@ def _get_workspace_resource_id(
...
+    def get_load_balancer_managed_outbound_ipv6_count(self) -> Union[int, None]:
+        """Obtain the expected count of IPv6 managed outbound IPs.
+
+        :return: int or None
+        """
+        count_ipv6 = self.raw_param.get(
+            'load_balancer_managed_outbound_ipv6_count')
+
+        return count_ipv6 if count_ipv6 else None
```
FumingZhang, line 787: In this way, we would convert `0` to `None`. If this is expected, please add a unit test case to cover this.

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -743,6 +747,45 @@ def _get_workspace_resource_id(
...
+        """Obtain the CIDR ranges for the service subnet.
+
+        :return: List[str] or None
+        """
+        return self._get_list_attr('service_cidrs')
+
+    def _get_list_attr(self, param_key) -> Union[List[str], None]:
+        param = self.raw_param.get(param_key)
+
+        if param and isinstance(param, str):
```
FumingZhang, line 774: In this way, we would convert empty string to `None`. If this is expected, please add a unit test case to cover this.

src/aks-preview/azext_aks_preview/tests/latest/test_decorator.py
```diff
@@ -1481,6 +1521,64 @@ def test_set_up_network_profile(self):
...
+        )
+        mc_3 = self.models.ManagedCluster(location="test_location")
+        dec_mc_3 = dec_3.set_up_network_profile(mc_3)
+
+        network_profile_3 = self.models.ContainerServiceNetworkProfile(
+            network_plugin="kubenet",
+            pod_cidr="10.246.0.0/16",
+            service_cidr="10.0.0.0/16",
+            dns_service_ip=None,  # overwritten to None
+            docker_bridge_cidr=None,  # overwritten to None
```
FumingZhang, line 1557: According to the implementation (base function) in [azure-cli](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acs/decorator.py#L4731-L4755), we would overwrite `dns_service_ip/docker_bridge_cidr` to `None` when any of `network_plugin/pod_cidr/service_cidr/dns_service_ip/docker_bridge_address/network_policy` is specified but they themselves are not. Take an example, if user only specifies the new parameters like `pod_cidrs`, `service_cidrs` or `ip_families`, we would still take the default values provided by SDK for `dns_service_ip` and `docker_bridge_cidr`.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -2891,6 +2891,39 @@ def test_aks_create_with_windows_gmsa(self, resource_group, resource_group_locat
...
+            'location': resource_group_location,
+            'resource_type': 'Microsoft.ContainerService/ManagedClusters',
+            'ssh_key_value': self.generate_ssh_keys(),
+        })
+
+        # create
+        create_cmd = 'aks create --resource-group={resource_group} --name={name} --location={location} ' \
+                     '--ip-families IPv4 --pod-cidr 172.126.0.0/16 --service-cidr 172.56.0.0/16 ' \
+                     '--dns-service-ip 172.56.0.10 --pod-cidrs 172.126.0.0/16 --service-cidrs 172.56.0.0/16 ' \
+                     '--network-plugin kubenet --ssh-key-value={ssh_key_value}'
```
FumingZhang, line 2913: It seems that the new option `--load-balancer-managed-outbound-ipv6-count` is not covered by any live (E2E) test case. Please add one and test both `aks create` and `aks update`.

src/aks-preview/azext_aks_preview/decorator.py
tyler-lloyd: good catch. changed to return 0

src/aks-preview/azext_aks_preview/decorator.py
tyler-lloyd: what do we usually do in cases where the parameter is expected to be a list? convert to an empty list? `[]`?

src/aks-preview/azext_aks_preview/tests/latest/test_decorator.py
tyler-lloyd: updated to capture this functionality. if ip_families is provided then defaults should be kept (if other things are provided)

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
tyler-lloyd: currently, we do not have the toggle on for dual-stack clusters so any e2e test would fail. That should be changing shortly. My plan was to come back and add a test once it is enabled in canary (which should be by Monday)

src/aks-preview/azext_aks_preview/_loadbalancer.py
tyler-lloyd: yep, updated to preserve the ipv6 count through update commands.

src/aks-preview/azext_aks_preview/decorator.py
tyler-lloyd: Looks like for other parameters we differentiate between omitted and empty string for list properties. Will update.

src/aks-preview/azext_aks_preview/_loadbalancer.py
```diff
@@ -26,48 +26,58 @@ def set_load_balancer_sku(sku, kubernetes_version):
...
     """configure a load balancer with customer supplied values"""
     if not profile:
         return profile
 
     outbound_ip_resources = _get_load_balancer_outbound_ips(outbound_ips)
     outbound_ip_prefix_resources = _get_load_balancer_outbound_ip_prefixes(outbound_ip_prefixes)
 
-    if managed_outbound_ip_count or outbound_ip_resources or outbound_ip_prefix_resources:
-        # ips -> i_ps due to track 2 naming issue
-        profile.managed_outbound_i_ps = None
```
FumingZhang, line 62: I am not sure if we need to clear the setting of `managed_outbound_i_ps` when only `outbound_ip_resources` or `outbound_ip_prefix_resources` is specified (in an update scenario).

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -1301,6 +1351,60 @@ def set_up_network_profile(self, mc: ManagedCluster) -> ManagedCluster:
...
+            pod_cidr,
+            pod_cidrs,
+            service_cidr,
+            service_cidrs,
+        ]):
+            network_profile.dns_service_ip = dns_service_ip
+            network_profile.pod_cidr = pod_cidr
+            network_profile.pod_cidrs = pod_cidrs
+            network_profile.service_cidr = service_cidr
+            network_profile.service_cidrs = service_cidrs
```
FumingZhang, line 1382: I remember that in the previous implementation, the value of `pod_cidr` was overwritten by `pod_cidrs[0]` (similar for `service_cidr`). Is this no longer needed?

By the way, I suppose it is a bit strange to provide both options `--pod-cidr` and `--pod-cidrs` at the same time, since option `--pod-cidrs` is sufficient to cover the function of option `--pod-cidr` (except for compatibility issues); maybe we can discuss with PM whether we need to discard option `--pod-cidr` in the future?

One suggestion to the current situation is, do we need to output a warning level log message (or even raise a `MutuallyExclusiveArgumentError`) when the user uses the two options `--pod-cidr` and `--pod-cidrs` at the same time, reminding the user that the value set by `--pod-cidr` will be overwritten by `--pod-cidrs` (or this is not allowed)?

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
tyler-lloyd: these tests are now added for the ipv6-count

src/aks-preview/azext_aks_preview/decorator.py
tyler-lloyd: no need to overwrite the value as RP will take care of it if pod_cidr and service_cidr are omitted but the cidrs are supplied.

src/aks-preview/azext_aks_preview/_loadbalancer.py
tyler-lloyd: I am not really sure why this is there in the first place. It seems like this has unintended consequences. If I create a cluster then run `az aks update` with outbound_ip_resources, it seems like the managed_outbound_i_ps will be reverted back to the default values which is just `count=1`.

FumingZhang (approved):
The PR looks good to me

src/aks-preview/HISTORY.md
```diff
@@ -2,6 +2,11 @@
 
 Release History
 ===============
+0.5.47
+++++++
+* Add support for IPv4/IPv6 dual-stack networking AKS clusters
+  * `az aks create --pod-cidrs --service-cidrs --ip-families --load-balancer-managed-outbound-ipv6-count`
```
zhoxing-ms, line 8:
```suggestion
* `az aks create --pod-cidrs --service-cidrs --ip-families --load-balancer-managed-outbound-ipv6-count`
```

zhoxing-ms (approved):

PR #4194, opened 2021-12-06
Using msal token cache to fetch refresh token inside proxy command for az cli version >= 2.30

sirireddy12:
---

### General Guidelines

For new extensions:

### About Extension Publish


jiasli:
connectedk8s

zhoxing-ms (approved):

zhoxing-ms:
@sirireddy12 May I ask do you need to write the description of this change in to the history note and release a new extension for this change? If so, please modify the `HISTORY.rst` to record the this change history and modify `setup.py` to release a new version

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1635,12 +1639,16 @@ def client_side_proxy_wrapper(cmd,
...
+        msal_token_cache_user = r'.azure\msal_token_cache.bin'
+        msal_token_cache_spn = r'.azure\service_principal_entries.bin'
 
     elif(operating_system == 'Linux' or operating_system == 'Darwin'):
         install_location_string = f'.clientproxy/arcProxy{operating_system}{consts.CLIENT_PROXY_VERSION}'
         requestUri = f'{consts.CSP_Storage_Url}/{consts.RELEASE_DATE_LINUX}/arcProxy{operating_system}{consts.CLIENT_PROXY_VERSION}'
         older_version_string = f'.clientproxy/arcProxy{operating_system}*'
         creds_string = r'.azure/accessTokens.json'
+        msal_token_cache_user = r'.azure/msal_token_cache.bin'
+        msal_token_cache_spn = r'.azure/service_principal_entries.bin'
```
jiasli, line 1651: DON'T DO THIS. THIS IS NOT SUPPORTED. See https://github.com/Azure/azure-cli/pull/19853#issuecomment-984440373

jiasli (changes requested):

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1738,21 +1746,40 @@ def client_side_proxy_wrapper(cmd,
...
+        else:
+            if user_type == "user":
+                response_user_objectid = az_cli("ad signed-in-user show --query objectId -o tsv")
+                token_cache_location = os.path.expanduser(os.path.join('~', msal_token_cache_user))
+                persistence = msal_extensions.FilePersistenceWithDataProtection(token_cache_location)
+                token_cache = msal_extensions.PersistedTokenCache(persistence)
+                token_cache._reload_if_necessary()
+                home_account_id = response_user_objectid + "." + tenantId
+                owned_by_home_account = {"home_account_id": home_account_id}
+                creds_info = token_cache.find(PersistedTokenCache.CredentialType.REFRESH_TOKEN, query=owned_by_home_account)
```
jiasli, line 1773: @rayluo, I think this is another instance where people are hacking MSAL cache to get the refresh token.

rayluo: That is really unfortunate.

@sirireddy12 , why is refresh token (RT) needed? Could the usage pattern be replaced by periodically requesting an access token, possibly via Azure CLI? (@jiasli , does `az get-access-token ...` has an equivalent helper function to be called by extensions?)

FYI: Not only the token cache helpers are considered internal, MSAL would probably NOT save RT in token cache in near future.

krdhruva: @rayluo  Refresh token is needed to retrieve PoP token. Does PoP implementation support available in MSAL python which az extension can utilize?

rayluo: > @rayluo Refresh token is needed to retrieve PoP token. Does PoP implementation support available in MSAL python which az extension can utilize?

MSAL Python does not currently support PoP. However, I got an impression that the PoP implementation might be similar to [another feature named "Ssh Cert" which MSAL Python does support](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.16.0/tests/test_e2e.py#L215-L238), and then there is also an az extension "ssh" already handles the key management and feeds signed content to MSAL. You may take a look and see whether that pattern could be repurposed for PoP token.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1738,26 +1746,68 @@ def client_side_proxy_wrapper(cmd,
...
+                token_cache._reload_if_necessary()
+                home_account_id = response_user_objectid + "." + tenantId
+                owned_by_home_account = {"home_account_id": home_account_id}
+                creds_list = token_cache.find(PersistedTokenCache.CredentialType.REFRESH_TOKEN, query=owned_by_home_account)
+                creds = creds_list[0]['secret']
+            else:
+                token_cache_location = os.path.expanduser(os.path.join('~', msal_token_cache_spn))
+                try:
+                   if operating_system == 'Windows':
+                        persistence = msal_extensions.FilePersistenceWithDataProtection(token_cache_location)
```
bgavrilMS, line 1793: You really should not be doing this. MSAL's internal cache is not a public artifact and may change without notice. 

In fact, this is changing as we speak - MSAL py will use the Windows broker (WAM). Brokers don't even return RT and cache is going to be stored elsewhere (not shared anymore).

PR #4243, opened 2021-12-20
{AKS} Refactor 4th part of the update sub-command in aks-preview module

FumingZhang:
---

- Refactor updating pod identity profile.
- Move pod identity related internal functions to a separate `_podidentity.py` file.
- Sort out imports in `custom.py`.

Unit Test Coverage:

> "covered_lines": 600,
> "num_statements": 602,
> "percent_covered": 99.66777408637874,
> "missing_lines": 2,
> "excluded_lines": 21

---
Changes after code review:

- Add mutual exclusion check for `enable_pod_identity` and `disable_pod_identity`.
- Extract the verification pod identity with kubenet as an independent function.
- Use `SimpleNamespace` to store pod identity models and use the `@property` decorator to avoid overwriting.

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/azext_aks_preview/_podidentity.py
+++ b/src/aks-preview/azext_aks_preview/_podidentity.py
@@ -102,6 +102,7 @@ def _update_addon_pod_identity(
     instance.pod_identity_profile.user_assigned_identities = pod_identities or []
     instance.pod_identity_profile.user_assigned_identity_exceptions = pod_identity_exceptions or []
 
+
 def _ensure_managed_identity_operator_permission(cli_ctx, instance, scope):
     cluster_identity_object_id = None
     if instance.identity.type.lower() == 'userassigned':
--- a/src/aks-preview/azext_aks_preview/_podidentity.py
+++ b/src/aks-preview/azext_aks_preview/_podidentity.py
@@ -1,0 +1,150 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+import time
+
+from knack.log import get_logger
+from knack.util import CLIError
+
+from ._client_factory import get_auth_management_client
+from ._consts import (CONST_MANAGED_IDENTITY_OPERATOR_ROLE,
+                      CONST_MANAGED_IDENTITY_OPERATOR_ROLE_ID)
+from ._roleassignments import add_role_assignment
+
+logger = get_logger(__name__)
+
+
+def _is_pod_identity_addon_enabled(instance):
+    if not instance:
+        return False
+    if not instance.pod_identity_profile:
+        return False
+    return bool(instance.pod_identity_profile.enabled)
+
+
+def _ensure_pod_identity_addon_is_enabled(instance):
+    if not _is_pod_identity_addon_enabled(instance):
+        raise CLIError('The pod identity addon is not enabled for this managed cluster yet.\n'
+                       'To enable, run "az aks update --enable-pod-identity')
+
+
+def _ensure_pod_identity_kubenet_consent(network_profile, pod_identity_profile, customer_consent):
+    if not network_profile or not network_profile.network_plugin:
+        # invalid data
+        return
+    if network_profile.network_plugin.lower() != 'kubenet':
+        # not kubenet, no need to check
+        return
+
+    if customer_consent is None:
+        # no set this time, read from previous value
+        customer_consent = bool(
+            pod_identity_profile.allow_network_plugin_kubenet)
+
+    if not customer_consent:
+        raise CLIError(
+            "--enable-pod-identity-with-kubenet is required for "
+            "enabling pod identity addon when using Kubenet network plugin")
+    pod_identity_profile.allow_network_plugin_kubenet = True
+
+
+def _fill_defaults_for_pod_identity_exceptions(pod_identity_exceptions):
+    if not pod_identity_exceptions:
+        return
+
+    for exc in pod_identity_exceptions:
+        if exc.pod_labels is None:
+            # in previous version, we accidentally allowed user to specify empty pod labels,
+            # which will be converted to `None` in response. This behavior will break the extension
+            # when using 2021-10-01 version. As a workaround, we always back fill the empty dict value
+            # before sending to the server side.
+            exc.pod_labels = dict()
+
+
+def _fill_defaults_for_pod_identity_profile(pod_identity_profile):
+    if not pod_identity_profile:
+        return
+
+    _fill_defaults_for_pod_identity_exceptions(pod_identity_profile.user_assigned_identity_exceptions)
+
+
+def _update_addon_pod_identity(
+    instance,
+    enable,
+    pod_identities=None,
+    pod_identity_exceptions=None,
+    allow_kubenet_consent=None,
+    models=None,
+):
+    if not enable:
+        ManagedClusterPodIdentityProfile = models.get("ManagedClusterPodIdentityProfile")
+        # when disable, remove previous saved value
+        instance.pod_identity_profile = ManagedClusterPodIdentityProfile(
+            enabled=False)
+        return
+
+    _fill_defaults_for_pod_identity_exceptions(pod_identity_exceptions)
+
+    if not instance.pod_identity_profile:
+        ManagedClusterPodIdentityProfile = models.get("ManagedClusterPodIdentityProfile")
+        # not set before
+        instance.pod_identity_profile = ManagedClusterPodIdentityProfile(
+            enabled=enable,
+            user_assigned_identities=pod_identities,
+            user_assigned_identity_exceptions=pod_identity_exceptions,
+        )
+
+    _ensure_pod_identity_kubenet_consent(
+        instance.network_profile, instance.pod_identity_profile, allow_kubenet_consent)
+
+    instance.pod_identity_profile.enabled = enable
+    instance.pod_identity_profile.user_assigned_identities = pod_identities or []
+    instance.pod_identity_profile.user_assigned_identity_exceptions = pod_identity_exceptions or []
+
+def _ensure_managed_identity_operator_permission(cli_ctx, instance, scope):
+    cluster_identity_object_id = None
+    if instance.identity.type.lower() == 'userassigned':
+        for identity in instance.identity.user_assigned_identities.values():
+            cluster_identity_object_id = identity.principal_id
+            break
+    elif instance.identity.type.lower() == 'systemassigned':
+        cluster_identity_object_id = instance.identity.principal_id
+    else:
+        raise CLIError('unsupported identity type: {}'.format(
+            instance.identity.type))
+    if cluster_identity_object_id is None:
+        raise CLIError('unable to resolve cluster identity')
+
+    factory = get_auth_management_client(cli_ctx, scope)
+    assignments_client = factory.role_assignments
+    cluster_identity_object_id = cluster_identity_object_id.lower()
+    scope = scope.lower()
+
+    # list all assignments of the target identity (scope) that assigned to the cluster identity
+    filter_query = "atScope() and assignedTo('{}')".format(cluster_identity_object_id)
+    for i in assignments_client.list_for_scope(scope=scope, filter=filter_query):
+        if not i.role_definition_id.lower().endswith(CONST_MANAGED_IDENTITY_OPERATOR_ROLE_ID):
+            continue
+
+        # sanity checks to make sure we see the correct assignments
+        if i.principal_id.lower() != cluster_identity_object_id:
+            # assignedTo() should return the assignment to cluster identity
+            continue
+        if not scope.startswith(i.scope.lower()):
+            # atScope() should return the assignments in subscription / resource group / resource level
+            continue
+
+        # already assigned
+        logger.debug('Managed Identity Opereator role has been assigned to {}'.format(i.scope))
+        return
+
+    if not add_role_assignment(cli_ctx, CONST_MANAGED_IDENTITY_OPERATOR_ROLE, cluster_identity_object_id,
+                               is_service_principal=False, scope=scope):
+        raise CLIError(
+            'Could not grant Managed Identity Operator permission for cluster')
+
+    # need more time to propogate this assignment...
+    print()
+    print('Wait 30 seconds for identity role assignment propagation.')
+    time.sleep(30)
--- a/src/aks-preview/azext_aks_preview/_podidentity.py
+++ b/src/aks-preview/azext_aks_preview/_podidentity.py
@@ -3,6 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 import time
+from types import SimpleNamespace
 
 from knack.log import get_logger
 from knack.util import CLIError
@@ -75,10 +76,10 @@ def _update_addon_pod_identity(
     pod_identities=None,
     pod_identity_exceptions=None,
     allow_kubenet_consent=None,
-    models=None,
+    models: SimpleNamespace = None,
 ):
     if not enable:
-        ManagedClusterPodIdentityProfile = models.get("ManagedClusterPodIdentityProfile")
+        ManagedClusterPodIdentityProfile = models.ManagedClusterPodIdentityProfile
         # when disable, remove previous saved value
         instance.pod_identity_profile = ManagedClusterPodIdentityProfile(
             enabled=False)
@@ -87,7 +88,7 @@ def _update_addon_pod_identity(
     _fill_defaults_for_pod_identity_exceptions(pod_identity_exceptions)
 
     if not instance.pod_identity_profile:
-        ManagedClusterPodIdentityProfile = models.get("ManagedClusterPodIdentityProfile")
+        ManagedClusterPodIdentityProfile = models.ManagedClusterPodIdentityProfile
         # not set before
         instance.pod_identity_profile = ManagedClusterPodIdentityProfile(
             enabled=enable,
--- a/src/aks-preview/azext_aks_preview/custom.py
+++ b/src/aks-preview/azext_aks_preview/custom.py
@@ -4,12 +4,10 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
-from __future__ import print_function
-
+import base64
 import binascii
 import datetime
 import errno
-import io
 import json
 import os
 import os.path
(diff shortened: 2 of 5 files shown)


src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -608,23 +655,29 @@ def _get_enable_pod_identity_with_kubenet(self, enable_validation: bool = False,
...
         # validation
         if enable_validation:
-            if self.mc and self.mc.network_profile and safe_lower(self.mc.network_profile.network_plugin) == "kubenet":
-                if not enable_pod_identity_with_kubenet and self._get_enable_pod_identity(enable_validation=False):
-                    raise RequiredArgumentMissingError(
-                        "--enable-pod-identity-with-kubenet is required for enabling pod identity addon "
-                        "when using Kubenet network plugin"
-                    )
+            if self.decorator_mode == DecoratorMode.CREATE:
+                if (
```
bcho, line 671: nit: can we extract this kubenet validation with L608 to a helper method?

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -1583,7 +1634,7 @@ def set_up_pod_identity_profile(self, mc: ManagedCluster) -> ManagedCluster:
         enable_pod_identity = self.context.get_enable_pod_identity()
         enable_pod_identity_with_kubenet = self.context.get_enable_pod_identity_with_kubenet()
         if enable_pod_identity:
-            pod_identity_profile = self.models.ManagedClusterPodIdentityProfile(
+            pod_identity_profile = self.models.pod_identity_models.get("ManagedClusterPodIdentityProfile")(
```
bcho, line 1637: is it type safe? This looks easy to have typo here

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -1887,6 +1938,27 @@ def update_nat_gateway_profile(self, mc: ManagedCluster) -> ManagedCluster:
...
+            if not _is_pod_identity_addon_enabled(mc):
+                # we only rebuild the pod identity profile if it's disabled before
+                _update_addon_pod_identity(
+                    mc,
+                    enable=True,
+                    allow_kubenet_consent=self.context.get_enable_pod_identity_with_kubenet(),
+                    models=self.models.pod_identity_models
+                )
+
+        if self.context.get_disable_pod_identity():
```
bcho, line 1958: nit: use `elseif` here? Since we don't allow to enable & disable at the same time

bcho: also I think we can move the `_ensure_mc` call into the branch to save some repeated calls

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -135,6 +140,29 @@ def init_nat_gateway_models(self) -> None:
...
+
+        The models are stored in a dictionary, the key is the model name and the value is the model type.
+
+        :return: None
+        """
+        pod_identity_models = {}
+        pod_identity_models["ManagedClusterPodIdentityProfile"] = self.__cmd.get_models(
+            "ManagedClusterPodIdentityProfile",
+            resource_type=self.resource_type,
+            operation_group="managed_clusters",
```
bcho, line 154: nit: can we make this `managed_clusters` operation group as a const?

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -135,6 +140,29 @@ def init_nat_gateway_models(self) -> None:
         # for model_name, model_type in nat_gateway_models.items():
         #     setattr(self, model_name, model_type)
 
+    def init_pod_identity_models(self) -> None:
```
bcho, line 143: question(non-blocking): is it possible to decorate it with `@property` and make it a lazy initialization? This could make the caller side easier to use.

For example:

```python
@property
def pod_identity_models(self):
    if self._pod_identity_models not None:
      return self._pod_identity_models
    # otherwise, initialize the model

# in caller side:
profile = self.pod_identity_models.ManagedClusterPodIdentityProfile()
```

But since I don't know how the whole thing works in the new implementation, I might miss some contexts here.

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -111,6 +114,8 @@ def __init__(self, cmd: AzCommandsLoader, resource_type: ResourceType):
         )
         # init nat gateway models
         self.init_nat_gateway_models()
+        # init pod identity models
+        self.init_pod_identity_models()
```
bcho, line 118: question: what's the differences between L118 and L100?

```python
        self.ManagedClusterPodIdentityProfile = self.__cmd.get_models(
            "ManagedClusterPodIdentityProfile",
            resource_type=self.resource_type,
            operation_group="managed_clusters",
        )
```

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -3293,10 +3216,15 @@ def aks_pod_identity_add(cmd, client, resource_group_name, cluster_name,
...
 
+    from azext_aks_preview.decorator import AKSPreviewModels
+
+    # store all the models used by pod identity
+    pod_identity_models = AKSPreviewModels(cmd, CUSTOM_MGMT_AKS_PREVIEW).pod_identity_models
     _update_addon_pod_identity(
         instance, enable=True,
         pod_identities=pod_identities,
         pod_identity_exceptions=instance.pod_identity_profile.user_assigned_identity_exceptions,
+        models=pod_identity_models
```
bcho, line 3227: question: since we are repeating the `pod_identity_models` construction here, why not just let the `_update_addon_pod_identity` to construct it?

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -560,27 +588,33 @@ def _get_enable_pod_identity(self, enable_validation: bool = False, **kwargs) ->
...
         # this parameter does not need dynamic completion
         # validation
         if enable_validation:
-            if enable_pod_identity and not self._get_enable_managed_identity(enable_validation=False):
-                raise RequiredArgumentMissingError(
-                    "--enable-pod-identity can only be specified when --enable-managed-identity is specified"
-                )
-            if self.mc and self.mc.network_profile and safe_lower(self.mc.network_profile.network_plugin) == "kubenet":
-                if enable_pod_identity and not self._get_enable_pod_identity_with_kubenet(enable_validation=False):
+            if self.decorator_mode == DecoratorMode.CREATE:
```
bcho, line 603: question: what will happen if `decorator_mode` is not `CREATE`? Do we simply skip the else branch?

src/aks-preview/azext_aks_preview/_podidentity.py
```diff
@@ -0,0 +1,151 @@
...
+    _fill_defaults_for_pod_identity_exceptions(pod_identity_profile.user_assigned_identity_exceptions)
+
+
+def _update_addon_pod_identity(
+    instance,
+    enable,
+    pod_identities=None,
+    pod_identity_exceptions=None,
+    allow_kubenet_consent=None,
+    models=None,
```
bcho, line 78: nit: we should properly annotate the type for the `models` parameter here

src/aks-preview/azext_aks_preview/_podidentity.py
```diff
@@ -0,0 +1,151 @@
...
+):
+    if not enable:
+        ManagedClusterPodIdentityProfile = models.get("ManagedClusterPodIdentityProfile")
```
bcho, line 81: question: as the `models` is a dict, why use `get` here? But I still think it's better to define a typed struct (https://docs.python.org/3/library/dataclasses.html) here for better type checking.

bcho (changes requested):
Thanks for refactoring the pod identity sub-commands! But I have a few questions about the new implementation here.

src/aks-preview/azext_aks_preview/_podidentity.py
bcho: It's the same question as in: https://github.com/Azure/azure-cli-extensions/pull/4243#discussion_r772105207

src/aks-preview/azext_aks_preview/decorator.py
bcho: yeah looks like we only expect to initialize once, I think it's better to rename the method to something like `def _init_pod_identity_models` to avoid referencing outside current class.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Good point, working on a fix.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: This should be safe. `self.models` refers to an instance of class `AKSPreviewModels`, which is a holder of the model definitions. This class will help load all the models of a certain api version through the `resource_type` parameter, which avoids hard-coding the import model of a certain api version in each file.

Further for this case, we store the two models of `ManagedClusterPodIdentityProfile` and `ManagedClusterPodIdentityException` in the `pod_identity_models` property of this class. This attribute is a dictionary and will be set within the initialization process of this class.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Yeah, for other pairs of `enable_xxx` and `disable_xxx` parameters, there is usually a mutual exclusion check on the cli side. Here I just moved the original implementation, if necessary, I will add a check in the getter functions.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Good point, I'll wrap the dictionary with the @property decorator and I suppose without setting a setter function, this would also help avoid accidentally setting values.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Will try the @property decorator.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: No substantial changes. Just put the attribute of the original class into a dictionary. In those functions that need to use model definitions, we will pass the dictionary as a parameter to it to avoid hard-coding the import statement (e.g. from xxxx_xx_xx import xxxx).

src/aks-preview/azext_aks_preview/custom.py
FumingZhang: `pod_identity_models` is a dictionary stores the definition of the pod identity related models. Before that, the model definition was passed to the function through global variables (e.g. from xxxx_xx_xx import xxxx).

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Yes, we will skip all the checks in `UPDATE` mode (which is `aks update`). I suppose the validation is handled by `_update_addon_pod_identity`.

src/aks-preview/azext_aks_preview/_podidentity.py
FumingZhang: Sure.

src/aks-preview/azext_aks_preview/_podidentity.py
FumingZhang: I have no preference for using `get` or `[]` to get the value, but using `get` can avoid throwing an exception when the value does not exist. Of course, adding an attribute to `None` after getting a non-existent object with `get` will also report an error.

So I think it should be covered by unit tests to ensure that no one is getting unexpected value from the dictionary.

As for `dataclasses`, the feature was introduced in python 3.7. Before that, it was a third-party library that needed to be installed explicitly. Since azure-cli still supports python 3.6, as a plug-in, it may be difficult for us to request cli-core to add this dependency.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Updated in commit [#059868e](https://github.com/Azure/azure-cli-extensions/pull/4243/commits/059868ed2acd69dfd246b17785c35a4630f21ff1).

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Updated in commit [#eaeaa6b](https://github.com/Azure/azure-cli-extensions/pull/4243/commits/eaeaa6b408a905c9b25edbdc5f127ea3c15ed7f5).

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Changed in commit [#4243 9e1827e](https://github.com/Azure/azure-cli-extensions/pull/4243/commits/9e1827e8f5150a8f3e3884c433c9d9e35f668c3d).

src/aks-preview/azext_aks_preview/_podidentity.py
FumingZhang: Updated in commit [#9e1827e](https://github.com/Azure/azure-cli-extensions/pull/4243/commits/9e1827e8f5150a8f3e3884c433c9d9e35f668c3d).

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Updated in commit [#9e1827e](https://github.com/Azure/azure-cli-extensions/pull/4243/commits/9e1827e8f5150a8f3e3884c433c9d9e35f668c3d).

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -135,6 +141,59 @@ def init_nat_gateway_models(self) -> None:
         # for model_name, model_type in nat_gateway_models.items():
         #     setattr(self, model_name, model_type)
 
+    def _get_pod_identity_models(self) -> Dict:
```
bcho, line 144: nit: why not just return a `SimpleNamespace` from here?

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -135,6 +141,59 @@ def init_nat_gateway_models(self) -> None:
...
+        The models are stored in a SimpleNamespace object, could be accessed by the dot operator like
+        `pod_identity_models.ManagedClusterPodIdentityProfile`.
+
+        :return: SimpleNamespace
+        """
+        if self.__pod_identity_models is None:
+            self.__pod_identity_models = SimpleNamespace(**self._get_pod_identity_models())
+        return self.__pod_identity_models
+
+    def flattern_to_class_attributes(self, models: Union[Dict, SimpleNamespace]) -> None:
```
bcho, line 177: question: what is the function for?

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -549,90 +627,133 @@ def _get_enable_managed_identity(
...
+                    )
+                self.validate_pod_identity_with_kubenet(
+                    self.mc,
+                    enable_pod_identity,
+                    self._get_enable_pod_identity_with_kubenet(
+                        enable_validation=False
+                    ),
+                )
+            elif self.decorator_mode == DecoratorMode.UPDATE:
+                if enable_pod_identity and self._get_disable_pod_identity(enable_validation=False):
```
bcho, line 666: nit: we better to extract this check out along with L654

bcho: also, we are throwing `RequiredArgumentMissingError` in the CRATE code path, while using `MutuallyExclusiveArgumentError ` here

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -549,90 +627,133 @@ def _get_enable_managed_identity(
...
+        "validate_pod_identity_with_kubenet" for verification. In update mode, if both enable_pod_identity and
+        disable_pod_identity are specified, raise a MutuallyExclusiveArgumentError.
 
         :return: bool
         """
 
         return self._get_enable_pod_identity(enable_validation=True)
 
+    # pylint: disable=unused-argument
+    def _get_disable_pod_identity(self, enable_validation: bool = False, **kwargs) -> bool:
```
bcho, line 687: question: why do we want to use `**kwargs` for this helper method? Is this a convention or just a catch all usage? I think it's better to explicitly specify the parameter we want to use here.

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -1887,6 +2006,27 @@ def update_nat_gateway_profile(self, mc: ManagedCluster) -> ManagedCluster:
...
         return mc
 
+    def update_pod_identity_profile(self, mc: ManagedCluster) -> ManagedCluster:
+        """Update pod identity profile for the ManagedCluster object.
+
+        :return: the ManagedCluster object
+        """
+        self._ensure_mc(mc)
+
+        if self.context.get_enable_pod_identity():
```
bcho, line 2016: question: what's the differences between `self._get_disable/enable_pod_identity(enable_validation=True)` and `context. get_enable_pod_identity` here?

bcho (changes requested):

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: In case we want to get the model directly from an instance of `AKSPreviewModel`. But this is not used anywhere currently.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: No problem for me, but I am not sure if we need to keep a dictionary.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Ah, I suppose you might need to take another look. We only have the `disable_pod_identity` available in **update** mode and I suppose it's unnecessary to check the mutual exclusion in **create** mode.

Besides, the two validations are not the same, one for mutual exclusion between `enable_pod_identity` and `disable_pod_identity` in **update** mode, the other for if `enable_pod_identity` is specified, `enable_managed_identity` must be specified too in **create** mode.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: I've added this additional parameter `**kwargs` for all the internal version of the getter functions. The original intention is to pass additional parameters in some special cases, but I have not used it so far. For the getter functions with complex logic, I have prepared both "public" function and "internal" function (function name starts with `_`) for them.
- The "public" function basically has no parameter and should be used in the decorator class (e.g. `AKSCreateDecorator`/`AKSPreviewCreateDecorator`), where users do not need to consider the details of getter implementation.
- The "internal" function has many parameters, which are used to control whether to perform validation or dynamic completion, etc. It should be used in the context class (e.g. `AKSContext`/`AKSPreviewContext`). With proper parameter settings, infinite loop calls during cross-checking would be avoided.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Briefly, no difference.

But I need to remind that both functions are defined in the context class (e.g. `AKSContext`/`AKSPreviewContext`), and to call them in the decorator class (e.g. `AKSCreateDecorator`/`AKSPreviewCreateDecorator`), it should be `self.context._get_enable_pod_identity(enable_validation=True)  # not recommended` or `self.context.get_enable_pod_identity()`.

yonzhan:
AKS

src/aks-preview/azext_aks_preview/decorator.py
bcho: Can we remove this if it's not being used anywhere? Why we want to copy a field to the class? IMO, it would cause more confuse to other developers, especially under a dynamic language context.

src/aks-preview/azext_aks_preview/decorator.py
bcho: I think we can simplify thing by returning a SimpleNamespace

src/aks-preview/azext_aks_preview/decorator.py
bcho: yes you are right, the diff looks the same before

src/aks-preview/azext_aks_preview/decorator.py
bcho: > The original intention is to pass additional parameters in some special cases, but I have not used it so far

I don't this is the pythonic way. By adding `**kwargs` here it's easy to invite people to put many things inside it, which eventually become a large bag without typing information. This would be hard to refactor and test in the future. I think this is also the case before your refactor - we don't know what are we referencing in each sub command.

> With proper parameter settings, infinite loop calls during cross-checking would be avoided.

Could you provide examples? I think if calling without setting special parameter settings could cause inf. loop, that means something wrong in the code path because it's easy to write it wrong. We should properly rethink the design.

FumingZhang: For the first part, I agree with you and I would remove the `**kwargs`.

For the second question, take the `enable_validation` parameter as an example, the default value is `False`, which means if you call the internal function without any parameter, it will skip the validation part and this would avoid loop calls.

So don’t worry, for almost all situations, we could avoid loop calls by the default values. The situations that require additional parameter settings are usually related to dynamic completion. At present, all these cases are in azure-cli. Some of the parameters involved are sp and managed identity, fqdn subdomain and dns prefix.

> For a getter function, the internal is mainly composed of three parts, obtaining the value, dynamic modification, and verification. In the verification part, we conduct cross-validation between multiple parameters (the verification of a single parameter is handled by the validator (defined in `_validators.py`) specified in the parameter declaration (defined in `_params.py`)). Take the getters for `enable_pod_identity` and `disable_pod_identity` as the example, When verifying whether `disable_pod_identity` is set at the same time in the getter of `enable_pod_identity`, we will call the getter of `disable_pod_identity` to obtain its value, and the verification part of `disable_pod_identity` will also obtain the value of `enable_pod_identity` by calling its getter to verify whether the two are assigned at the same time. If we do not use a "flag" to escape this verification, then we will fall into a loop call.

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Updated (removed) in commit [#91c0ca3](https://github.com/Azure/azure-cli-extensions/pull/4243/commits/91c0ca36c0f84ef93033e1de3e98cf5596dfb32b).

src/aks-preview/azext_aks_preview/decorator.py
FumingZhang: Updated (removed) in commit [#91c0ca3](https://github.com/Azure/azure-cli-extensions/pull/4243/commits/91c0ca36c0f84ef93033e1de3e98cf5596dfb32b).

bcho (approved):
LGTM, thanks for the refactor!

FumingZhang:
@bcho Thanks a lot for helping review!

wangzelin007 (approved):

PR #4308, opened 2022-01-11
fix(spring-cloud): 🐛 App update warns when  active deployment not found

yuwzho:
---

Previously wrongly fatal the `az spring-cloud app update` when there is no active deployment under it. It should just set a warning if user just want to update the app properties

Will upgrade version in #4294 

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/spring-cloud/azext_spring_cloud/_params.py
+++ b/src/spring-cloud/azext_spring_cloud/_params.py
@@ -17,9 +17,9 @@ from ._validators import (validate_env, validate_cosmos_type, validate_resource_
 from ._validators_enterprise import (only_support_enterprise,
                                      validate_git_uri, validate_acs_patterns,
                                      validate_routes)
-from ._app_validator import (fulfill_deployment_param, active_deployment_exist, active_deployment_exist_under_app,
+from ._app_validator import (fulfill_deployment_param, active_deployment_exist, active_deployment_exist_under_app_or_warning,
                              ensure_not_active_deployment, validate_deloy_path, validate_deloyment_create_path,
-                             validate_cpu, validate_memory)
+                             validate_cpu, validate_memory, fulfill_deployment_param_or_warning, active_deployment_exist_or_warning)
 from ._utils import ApiType
 
 from .vendored_sdks.appplatform.v2022_01_01_preview.models._app_platform_management_client_enums import SupportedRuntimeValue, TestKeyType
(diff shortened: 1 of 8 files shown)


yonzhan:
spring-cloud

src/spring-cloud/azext_spring_cloud/_params.py
```diff
@@ -160,6 +160,9 @@ def load_arguments(self, _):
                    help='A json file path for the persistent storages to be mounted to the app')
         c.argument('loaded_public_certificate_file', type=str, options_list=['--loaded-public-certificate-file', '-f'],
                    help='A json file path indicates the certificates which would be loaded to app')
+        c.argument('deployment', options_list=['--deployment', '-d'],
+                   help='Name of an existing deployment of the app. Default to the production deployment if not specified.',
+                   validator=fulfill_deployment_param_or_warning)
```
zhoxing-ms, line 165: I personally think it would be more appropriate to name it `deployment_name`, because this name can be consistent with the `az deployment` commands

src/spring-cloud/azext_spring_cloud/app.py
```diff
@@ -171,6 +172,12 @@ def app_update(cmd, client, resource_group, service, name,
         'https_only': https_only,
     }
 
+    if deployment is None:
+        updated_deployment_kwargs = {k: v for k, v in deployment_kwargs.items() if v}
+        if updated_deployment_kwargs:
+            raise InvalidArgumentValueError('{} cannot be set when there is no active deployment.'
```
zhoxing-ms, line 178: Is `ArgumentUsageError` a more appropriate error type for this usage case?

src/spring-cloud/azext_spring_cloud/_params.py
yuwzho: Here just update the validator for the argument. The name has been used for years.

src/spring-cloud/azext_spring_cloud/app.py
yuwzho: updated

src/spring-cloud/azext_spring_cloud/_params.py
zhoxing-ms: OK

zhoxing-ms (approved):

PR #4318, opened 2022-01-14
Update data plane client from swagger

ricardo-espinoza:
This change includes the following:

- New data plane generated client.
- Updates to show the "CostEstimate" field when representing job details as table.
- Added a script to facilitate updating the generated client in the future.


yonzhan:
quantum

PR #4381, opened 2022-02-01
add message of the day for aks agentpools

alexeldeib:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/HISTORY.md
+++ b/src/aks-preview/HISTORY.md
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+0.5.56
+++++++
+* Add `--message-of-the-day` flag for `az aks create` and `az aks nodepool add` for Linux message of the day.
+
 0.5.55
 +++++
 * Add option `none` to `--network-plugin` parameter to skip CNI installation during cluster creation.
@@ -19,11 +23,6 @@ Release History
   * `az aks create --crg-id`
 * Add support for nodepool adding with Capacity Reservation Group.
   * `az aks nodepool add --crg-id`
-
-0.5.53
-++++++
-
-* Add `--message-of-the-day` flag for `az aks create` and `az aks nodepool add` for Linux message of the day.
 
 0.5.52
 ++++++
--- a/src/aks-preview/HISTORY.md
+++ b/src/aks-preview/HISTORY.md
@@ -20,6 +20,11 @@ Release History
 * Add support for nodepool adding with Capacity Reservation Group.
   * `az aks nodepool add --crg-id`
 
+0.5.53
+++++++
+
+* Add `--message-of-the-day` flag for `az aks create` and `az aks nodepool add` for Linux message of the day.
+
 0.5.52
 ++++++
 * Add yaml template files to package data to fix issue [#148](https://github.com/Azure/aks-periscope/issues/148).
--- a/src/aks-preview/azext_aks_preview/_params.py
+++ b/src/aks-preview/azext_aks_preview/_params.py
@@ -162,6 +162,7 @@ def load_arguments(self, _):
         c.argument('enable_oidc_issuer', action='store_true', is_preview=True)
         c.argument('host_group_id', validator=validate_host_group_id, is_preview=True)
         c.argument('crg_id', validator=validate_crg_id, is_preview=True)
+        c.argument('message_of_the_day', type=str)
 
     with self.argument_context('aks update') as c:
         c.argument('enable_cluster_autoscaler', options_list=["--enable-cluster-autoscaler", "-e"], action='store_true')
@@ -266,6 +267,7 @@ def load_arguments(self, _):
             c.argument('snapshot_id', type=str, validator=validate_snapshot_id)
             c.argument('host_group_id', validator=validate_host_group_id, is_preview=True)
             c.argument('crg_id', validator=validate_crg_id, is_preview=True)
+            c.argument('message_of_the_day', type=str)
 
     for scope in ['aks nodepool show', 'aks nodepool delete', 'aks nodepool scale', 'aks nodepool upgrade', 'aks nodepool update']:
         with self.argument_context(scope) as c:
--- a/src/aks-preview/azext_aks_preview/_params.py
+++ b/src/aks-preview/azext_aks_preview/_params.py
@@ -206,7 +206,7 @@ def load_arguments(self, _):
         c.argument('host_group_id',
                    validator=validate_host_group_id, is_preview=True)
         c.argument('crg_id', validator=validate_crg_id, is_preview=True)
-        c.argument('message_of_the_day', type=str)
+        c.argument('message_of_the_day', type=str) # no validation for aks create because it already only supports Linux.
 
     with self.argument_context('aks update') as c:
         c.argument('enable_cluster_autoscaler', options_list=[
--- a/src/aks-preview/azext_aks_preview/_params.py
+++ b/src/aks-preview/azext_aks_preview/_params.py
@@ -23,7 +23,8 @@ from ._validators import (
     validate_taints, validate_priority, validate_eviction_policy, validate_spot_max_price, validate_acr, validate_user,
     validate_load_balancer_outbound_ports, validate_load_balancer_idle_timeout, validate_nat_gateway_idle_timeout, validate_nodepool_tags, validate_addon,
     validate_nodepool_labels, validate_vnet_subnet_id, validate_pod_subnet_id, validate_max_surge, validate_assign_identity, validate_addons,
-    validate_pod_identity_pod_labels, validate_pod_identity_resource_name, validate_pod_identity_resource_namespace, validate_assign_kubelet_identity, validate_host_group_id)
+    validate_pod_identity_pod_labels, validate_pod_identity_resource_name, validate_pod_identity_resource_namespace, validate_assign_kubelet_identity,
+    validate_host_group_id, validate_message_of_the_day)
 from ._consts import CONST_OUTBOUND_TYPE_LOAD_BALANCER, CONST_OUTBOUND_TYPE_USER_DEFINED_ROUTING, CONST_OUTBOUND_TYPE_MANAGED_NAT_GATEWAY, \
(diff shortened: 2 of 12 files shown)


src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -1344,6 +1344,44 @@ def test_aks_nodepool_add_with_ossku(self, resource_group, resource_group_locati
...
+
+        # nodepool get-upgrades
+        self.cmd('aks nodepool add '
+                 '--resource-group={resource_group} '
+                 '--cluster-name={name} '
+                 '--name={node_pool_name_second} '
+                 '--message-of-the-day="YmFyCg=="',
+                 checks=[
+                    self.check('provisioningState', 'Succeeded'),
+                    self.check('messageOfTheDay', 'YmFyCg==')
```
alexeldeib, line 1379: this is somewhat similar to http proxy config with trusted CA, we could read the content from a file instead, as it could be large. open to suggestions.

yonzhan:
aks

wangzelin007:
@alexeldeib,
Since the `--message-of-the-day` is only valid for linux nodes, may I ask why not add a validation for this ?

alexeldeib:
> may I ask why not add a validation for this

added some validation 👍

alexeldeib:
@wangzelin007 any thoughts after adding some validation?

wangzelin007 (approved):

wangzelin007:
> @wangzelin007 Zelin Wang FTE any thoughts after adding some validation?

LGTM

src/aks-preview/azext_aks_preview/_validators.py
```diff
@@ -221,6 +221,14 @@ def validate_spot_max_price(namespace):
...
 
 
+def validate_message_of_the_day(namespace):
+    """Validates message of the pool can only be used on Linux."""
+    if namespace.message_of_the_day is not None and namespace.message_of_the_day != "":
+        if hasattr(namespace, 'os_type'):
+            if namespace.os_type is not None and namespace.os_type != "Linux":
+                raise CLIError('--message-of-the-day can only be set for linux nodepools')
+
+
```
FumingZhang, line 231: I suppose this is not a good place to validate multiple options. Especially for command lie `aks create`, there's no such option (`--os-type`).

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
FumingZhang: I find it a bit odd to let the user input an encoded string, maybe we can let the user input a normal string and then encode it in the cli?

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -370,6 +370,29 @@ def get_gpu_instance_profile(self) -> Union[str, None]:
         # this parameter does not need validation
         return gpu_instance_profile
 
+    def get_message_of_the_day(self) -> Union[str, None]:
```
FumingZhang, line 373: Please add unit test for this getter function.

FumingZhang (changes requested):

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
alexeldeib: so the content is usually expected to be a blob of multi-line text. I think that would be more difficult to pass to CLI without encoding. I was thinking the user would either have it as-is, and invoke as e.g. `--message-of-the-day=$(cat foo.txt | base64)` or else we could do `--message-of-the-day foo.txt` and read the file as plain text + encode. 

WDYT?

src/aks-preview/azext_aks_preview/_validators.py
alexeldeib: yeah, I tried to approximate with `hasattr`, which works but is not very idiomatic/pythonic. 

would you prefer I revert this?

alexeldeib: suppose I could split the validation func and have it different for create / nodepool add

alexeldeib: ah actually, I didn't realize you can't do windows nodepools at all on az aks create. so validation isn't required there.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
FumingZhang: I prefer normal string then encode it in cli.

We could support both file name and raw input. This way, we first take the user's input as a filename, and if no such file exists, treat it as a normal string. I'm not sure if we need a PM to participate in the discussion though.

src/aks-preview/azext_aks_preview/_validators.py
FumingZhang: At first I thought we should do some validation against windows profile in the getter function of this option, but then I realized that we can still use this feature on linux node pools.

Never mind, I think adding verification to this option for `nodepool add` will do. But we may need to leave a note for `aks create`.

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -161,6 +162,7 @@ def load_arguments(self, _):
         c.argument('snapshot_id', type=str, validator=validate_snapshot_id)
         c.argument('enable_oidc_issuer', action='store_true', is_preview=True)
         c.argument('crg_id', validator=validate_crg_id, is_preview=True)
+        c.argument('message_of_the_day', type=str)
```
FumingZhang, line 165: Better to leave a note here, explain that the windows nodepool cannot be created (at this time) so verification is not needed.

FumingZhang:
A reminder, we have adopted the 2022-01-02-preview api for aks-preview and need to update the recording file.

justindavies:
@FumingZhang - can we get this merged please?

FumingZhang:
> A reminder, we have adopted the 2022-01-02-preview api for aks-preview and need to update the recording file.

Hi @justindavies, I suppose we need to address the failed checks before merging this PR. For unit test related pipelines, please update the api version to 2022-01-02-preview in recording file. For live test related pipelines, it seems to be a transient issue and rerun would work.

src/aks-preview/azext_aks_preview/_validators.py
```diff
@@ -221,6 +221,12 @@ def validate_spot_max_price(namespace):
                 "default price to be up-to on-demand")
 
 
+def validate_message_of_the_day(namespace):
+    """Validates message of the day can only be used on Linux."""
+    if namespace.message_of_the_day is not None and namespace.message_of_the_day != "":
+        if namespace.os_type is not None and namespace.os_type != "Linux":
+            raise CLIError('--message-of-the-day can only be set for linux nodepools')
```
zhoxing-ms, line 228: Please use the specific error type `ArgumentUsageError` instead of `CLIError`

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
alexeldeib: for context, the real world use-case is expected to be something like quoted below. I am not sure passing it as raw input is very convenient for users. I will add file-based input and see what you think?

"
UNAUTHORIZED ACCESS TO THIS DEVICE IS PROHIBITED

You must have explicit, authorized permission to access or configure this device. Unauthorized attempts and actions to access or use this system may result in civil and/or criminal penalties. All activities performed on this device are logged and monitored.
"

src/aks-preview/HISTORY.md
```diff
@@ -20,6 +20,11 @@ Release History
 * Add support for nodepool adding with Capacity Reservation Group.
   * `az aks nodepool add --crg-id`
 
+0.5.53
+++++++
+
+* Add `--message-of-the-day` flag for `az aks create` and `az aks nodepool add` for Linux message of the day.
+
```
FumingZhang, line 27: Need to update the version. Maybe 0.5.56? Please also update `setup.py`.

FumingZhang (changes requested):

src/aks-preview/HISTORY.md
alexeldeib: yep. also still reworking the encoding, not quite done yet.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
FumingZhang: I fully support adding file-based input. But if the user provides the file name, we will convert the file content in the cli? Or does it require that the user-provided file has already been encoded?

alexeldeib: I was thinking the content above as plain-text in the file, and then we encode it for consumption in the api.

alexeldeib: something like that?

src/aks-preview/azext_aks_preview/decorator.py
```diff
@@ -370,6 +371,41 @@ def get_gpu_instance_profile(self) -> Union[str, None]:
...
+
+        :return: string or None
+        """
+        # read the original value passed by the command
+        message_of_the_day = None
+        message_of_the_day_file_path = self.raw_param.get("message_of_the_day")
+
+        if message_of_the_day_file_path:
+            if not os.path.isfile(message_of_the_day_file_path):
+                raise InvalidArgumentValueError(
```
FumingZhang, line 385: Missing unit test coverage for this exception. See pipeline [log](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=52206529&view=logs&j=1c778e8e-596b-5a0a-22c2-460364725d7c&t=3e82742e-7cc5-5f7d-8ec2-cd9417b56eb7&l=386).

FumingZhang (approved):
LGTM :shipit:

zhoxing-ms (approved):

justindavies:
Hey @FumingZhang @zhoxing-ms the publish pipeline seems to have failed after this PR was merged, would you be able to take a look please

PR #4428, opened 2022-02-16
Customcert and fix for non escaping values in helm

Anumita:
---
[Anumita]
Added custom cert support 

[Siri]
Fixed issue where because of reuse-values
Case:
 1. Onboard using proxy
 2. Now run update command passing only one of the proxy params/other params
 3. It will fail with error because we send  –reuse-values to helm  which gets the other proxy parameters that  you would have set previously but helm does not escape those parameters hence the “—set data, key “l” has no value,  ” error comes up
 
Tests done (all the below should happen successfully)
1. Onboard with proxy with cert 
2. Update only proxy-skip-range or proxy cert
3. Disable proxy
4. Update only proxycert (custom cert scenario)

1. Onboard without proxy
2. Provide only proxycert

### General Guidelines

For new extensions:

### About Extension Publish


mouliaddaguduru:
What tests were run to validate the fix? can you please capture that information ?

yonzhan:
connectedk8s

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -879,9 +879,25 @@ def update_agents(cmd, client, resource_group_name, cluster_name, https_proxy=""
     # Get Helm chart path
     chart_path = utils.get_chart_path(registry_path, kube_config, kube_context, helm_client_location)
 
+    cmd_helm_values = [helm_client_location, "get", "values", "azure-arc", "--namespace", release_namespace]
+    if kube_config:
+        cmd_helm_values.extend(["--kubeconfig", kube_config])
+    if kube_context:
+        cmd_helm_values.extend(["--kube-context", kube_context])
+    
+    existing_user_values = open('userValues.txt', 'w+')
```
akashkeshari, line 888: Instead of using a file, we can directly use an object to store the output
https://github.com/Azure/azure-cli-extensions/blob/6da4c7bfc76e7ec37339ec3c864a1edc96f9476b/src/connectedk8s/azext_connectedk8s/custom.py#L1056

sirireddy12: Makes sense. Done

zhoxing-ms:
@Anumita Could you add some scenario tests in `test_connectedk8s_scenario.py` ?

sirireddy12:
> @Anumita Could you add some scenario tests in `test_connectedk8s_scenario.py` ?

@zhoxing-ms We will track this as a work item and will be adding more test coverage in the coming days. Is it possible to go ahead with this PR ? 
cc: @Anumita

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -777,6 +777,8 @@ def update_agents(cmd, client, resource_group_name, cluster_name, https_proxy=""
 
     # Send cloud information to telemetry
     send_cloud_telemetry(cmd)
+    # Fetch OS details
+    operating_system = platform.system()
```
akashkeshari, line 781: why is this reqd?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -879,9 +881,34 @@ def update_agents(cmd, client, resource_group_name, cluster_name, https_proxy=""
     # Get Helm chart path
     chart_path = utils.get_chart_path(registry_path, kube_config, kube_context, helm_client_location)
 
+    cmd_helm_values = [helm_client_location, "get", "values", "azure-arc", "--namespace", release_namespace]
+    if kube_config:
+        cmd_helm_values.extend(["--kubeconfig", kube_config])
+    if kube_context:
+        cmd_helm_values.extend(["--kube-context", kube_context])
+
+    if(operating_system == 'Windows'):
```
akashkeshari, line 890: Instead of doing this, use python apis which are generic for all OS
https://github.com/Azure/azure-cli-extensions/blob/284295daaf4e266fcad57bb997293bc74afb8b09/src/connectedk8s/azext_connectedk8s/_utils.py#L77

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: Removed it

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: Done. Thanks

zhoxing-ms (approved):

PR #4438, opened 2022-02-18
{CI} Add fabric config-as-code to support auto assign

wangzelin007:
---

Add fabric config-as-code to supprot auto assign.

- Fabricbotservices team are planning to move everyone to config-as-code in next quarter.

- All rules need to commit in .github/fabricbot.json.

- If we have the fabricbot.json file in our repo that will always take precedence.  

- So while add rules and click "save" in the GUI, that will do nothing unless export the config and do a PR to repo for it.


yonzhan:
CI

jiasli (approved):
How can we pre-check if the JSON `.github/fabricbot.json` is correct before merging the PR?

wangzelin007:
> How can we pre-check if the JSON `.github/fabricbot.json` is correct before merging the PR?

Very good question.
We cannot pre-check the fabric rule, and we already send a email to fabricbot service team.
image
image

PR #4463, opened 2022-02-25
[Quantum] Output error information in table when Quantum job fails

warren-jones:
Modified the Quantum CLI extension code to output error information when a job fails and the CLI output format parameter "-o table" has been specified.  The error message was nested too deeply in the Quantum provider's response for normal table formatting to display it.

The new output looks like this:
```
Status    Error Code             Error Message              Target          Job ID                                Submission Time
--------  ---------------------  -------------------------  --------------  ------------------------------------  --------------------------------
Failed    InsufficientResources  Too many qubits requested  ionq.simulator  19f19add-4617-4cd6-8ec6-66d0b055d332  2022-02-25T18:56:53.275035+00:00
```
---


yonzhan:
Quantum

warren-jones:
@vxfield, Please review

src/quantum/azext_quantum/commands.py
```diff
@@ -94,6 +98,30 @@ def one(key, value):
...
 
+    elif 'errorData' in results:
+        notFound = 'Not found'
+        status = notFound
+        errorCode = notFound
+        errorMessage = notFound
+        target = notFound
+        jobId = notFound
+        submissionTime = notFound
+        if 'status' in results:
```
vxfield, line 109: Might be better to use [inline ternary operator](https://pythonexamples.org/python-ternary-operator/) instead of these `if` statements

vxfield (approved):
Approved with a suggestion

src/quantum/azext_quantum/commands.py
```diff
@@ -94,6 +98,30 @@ def one(key, value):
...
+        if 'code' in errorData:
+            errorCode = errorData['code']
+        if 'message' in errorData:
+            errorMessage = errorData['message']
+        if 'target' in results:
+            target = results['target']
+        if 'id' in results:
+            jobId = results['id']
+        if 'creationTime' in results:
+            submissionTime = results['creationTime']
```
wangzelin007, line 121:
```suggestion
    elif 'errorData' in results:
        notFound = 'Not found'
        errorData = results['errorData']
        status = results.get('status', notFound)
        errorCode = errorData.get('code', notFound)
        errorMessage = errorData.get('message', notFound)
        target = results.get('target', notFound)
        jobId = results.get('id', notFound)
        submissionTime = results.get('creationTime', notFound)
```

src/quantum/azext_quantum/commands.py
warren-jones: Thanks, XField, I really appreciate you pointing out the Python ternary operator, and I will definitely keep that construct in mind for future Python adventures, but in this instance I’d like to go with Zelin’s suggestion below to use the dict.get() function.

src/quantum/azext_quantum/commands.py
warren-jones: Good call Zelin -- My original 18 lines are condensed into 6, and it still protects against raising a KeyError like my more verbose series of “if…in” statements would.  Thanks! 👍

PR #4468, opened 2022-02-25
ADAL to MSAL move for cli>=2.30.0

sirireddy12:
PR Description:

2.30 version az cli has moved from ADAL to MSAL and as a part of this, "accesstoken.json" file which used to store logged in user access token and refresh token has been removed. Az connectedk8s proxy scenario has a dependency on same to read refresh token and pass it to our Go application to generate PoP token. 

We have worked with MSAL team for an alternative approach to support this migration .  Our PoP contract expects a token with CNF claim. According to MSAL team's suggestion, we are retrieving JWT token with ‘cnf’ claim using [MSAL API](https://github.com/Azure/azure-cli/blob/c6f89e9f5c9f222cfa59e3a27e0efd08828fad60/src/azure-cli-core/azure/cli/core/_profile.py#L300) . We have made changes in our Go application(which generates the PoP token) to accept this Jwt token and further generate PoP token using that token.


yonzhan:
connectedK8S

src/connectedk8s/azext_connectedk8s/_constants.py
```diff
@@ -113,7 +113,12 @@
 CC_Provider_Namespace_Not_Registered_Fault_Type = "Connected Cluster Provider MS.K8 namespace not registered"
 Default_Namespace_Does_Not_Exist_Fault_Type = "The default namespace defined in the kubeconfig doesn't exist on the kubernetes cluster."
 ClusterConnect_Not_Present_Fault_Type = "cluster-connect-feature-unavailable"
-CLIENT_PROXY_VERSION = '1.3.018101'
+KAP_1P_Server_App_Scope = "6256c85f-0aad-4d50-b960-e6e9b21efe35/.default"
+KAP_1P_Server_AppId = "6256c85f-0aad-4d50-b960-e6e9b21efe35"
+Get_PublicKey_Info_Fault_Type = 'Error while fetching the PoP publickey information from CSP'
+PoP_Public_Key_Expried_Fault_Type = 'The PoP public key used to generate the at has expired'
+AZ_CLI_ADAL_TO_MSAL_MIGRATE_VERSION = '2.30.0'
+CLIENT_PROXY_VERSION = '3'
```
sirireddy12, line 121: update it after csp team releases prod version

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1915,20 +1917,30 @@ def client_side_proxy(cmd,
...
-            # Needed to prevent skip tls warning from printing to the console
-            original_stderr = sys.stderr
-            f = open(os.devnull, 'w')
-            sys.stderr = f
-
-            make_api_call_with_retries(identity_uri, identity_data, False, consts.Post_RefreshToken_Fault_Type,
-                                       'Unable to post refresh token details to clientproxy',
-                                       "Failed to pass refresh token details to proxy.", clientproxy_process)
-            sys.stderr = original_stderr
+        if not utils.is_cli_using_msal_auth():  # refresh token approach if cli is ADAL auth. This is for cli < 2.30.0
```
sirireddy12, line 1920: typo

zhoxing-ms:
@sirireddy12 Could you sync code from the remote main branch, so that this PR can skip the test issues of other modules

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: done

sirireddy12:
> @sirireddy12 Could you sync code from the remote main branch, so that this PR can skip the test issues of other modules

done, thanks

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1915,20 +1917,30 @@ def client_side_proxy(cmd,
...
+                original_stderr = sys.stderr
+                f = open(os.devnull, 'w')
+                sys.stderr = f
+
+                make_api_call_with_retries(identity_uri, identity_data, "post", False, consts.Post_RefreshToken_Fault_Type,
+                                           'Unable to post refresh token details to clientproxy',
+                                           "Failed to pass refresh token details to proxy.", clientproxy_process)
+                sys.stderr = original_stderr
+
+    if utils.is_cli_using_msal_auth():  # jwt token approach if cli is using MSAL. This is for cli >= 2.30.0
```
krdhruva, line 1936: can be moved to else?

src/connectedk8s/azext_connectedk8s/_constants.py
```diff
@@ -113,7 +113,12 @@
 CC_Provider_Namespace_Not_Registered_Fault_Type = "Connected Cluster Provider MS.K8 namespace not registered"
 Default_Namespace_Does_Not_Exist_Fault_Type = "The default namespace defined in the kubeconfig doesn't exist on the kubernetes cluster."
 ClusterConnect_Not_Present_Fault_Type = "cluster-connect-feature-unavailable"
-CLIENT_PROXY_VERSION = '1.3.018101'
+KAP_1P_Server_App_Scope = "6256c85f-0aad-4d50-b960-e6e9b21efe35/.default"
+KAP_1P_Server_AppId = "6256c85f-0aad-4d50-b960-e6e9b21efe35"
+Get_PublicKey_Info_Fault_Type = 'Error while fetching the PoP publickey information from CSP'
```
krdhruva, line 118: client proxy?

sirireddy12: Since we will be doing the GET call on CSP's endpoint - "https://localhost:{api_server_port}/identity/poppublickey to get the publickey" to get the publickey, I mentioned CSP in the fault type also

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1882,7 +1884,7 @@ def client_side_proxy(cmd,
                       clientproxy_process=None):
 
     subscription_id = get_subscription_id(cmd.cli_ctx)
-
+    tenantId = _graph_client_factory(cmd.cli_ctx).config.tenant_id
```
krdhruva, line 1887: I think we made this change in earlier PR.

sirireddy12: We concluded that we won't make that change for proxy command right. Since there can be customers like backup team, who are using the additional values.yml  to set KAP tenantID. Due to this, KAP will contain user's tenantID and not resource's tenantID.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1915,20 +1917,30 @@ def client_side_proxy(cmd,
...
+
+                make_api_call_with_retries(identity_uri, identity_data, "post", False, consts.Post_RefreshToken_Fault_Type,
+                                           'Unable to post refresh token details to clientproxy',
+                                           "Failed to pass refresh token details to proxy.", clientproxy_process)
+                sys.stderr = original_stderr
+
+    if utils.is_cli_using_msal_auth():  # jwt token approach if cli is using MSAL. This is for cli >= 2.30.0
+        kid = fetch_pop_publickey_kid(api_server_port, clientproxy_process)
+        post_at_response = fetch_and_post_at_to_csp(cmd, api_server_port, tenantId, kid, clientproxy_process)
+        if post_at_response.status_code == 500 and "public key expired" in post_at_response.text:  # pop public key must have been rotated
```
krdhruva, line 1939: what will be the experience for other error scenarios?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1989,6 +2001,51 @@ def make_api_call_with_retries(uri, data, tls_verify, fault_type, summary, cli_e
                 close_subprocess_and_raise_cli_error(clientproxy_process, cli_error + str(e))
 
 
+def fetch_pop_publickey_kid(api_server_port, clientproxy_process):
```
krdhruva, line 2004: better to move this to pop specific class than overloading custom.py

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1915,20 +1917,30 @@ def client_side_proxy(cmd,
...
+                identity_data['refreshToken'] = creds
+                identity_uri = f'https://localhost:{api_server_port}/identity/rt'
+
+                # Needed to prevent skip tls warning from printing to the console
+                original_stderr = sys.stderr
+                f = open(os.devnull, 'w')
+                sys.stderr = f
+
+                make_api_call_with_retries(identity_uri, identity_data, "post", False, consts.Post_RefreshToken_Fault_Type,
+                                           'Unable to post refresh token details to clientproxy',
```
krdhruva, line 1932: any reason on different wording for both msgs?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1989,6 +2001,51 @@ def make_api_call_with_retries(uri, data, tls_verify, fault_type, summary, cli_e
...
+    requestbody = {}
+    poppublickey_uri = f'https://localhost:{api_server_port}/identity/poppublickey'
+    # Needed to prevent skip tls warning from printing to the console
+    original_stderr = sys.stderr
+    f = open(os.devnull, 'w')
+    sys.stderr = f
+
+    get_publickey_response = make_api_call_with_retries(poppublickey_uri, requestbody, "get", False, consts.Get_PublicKey_Info_Fault_Type,
+                                                        'Unable to fetch public key info from clientproxy',
+                                                        "Failed to fetch public key info from client proxy", clientproxy_process)
```
krdhruva, line 2014: same comment on wording as above.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1989,6 +2001,51 @@ def make_api_call_with_retries(uri, data, tls_verify, fault_type, summary, cli_e
...
+
+    sys.stderr = original_stderr
+    publickey_info = json.loads(get_publickey_response.text)
```
krdhruva, line 2017: Do we get TP also as part of response which we can reuse?

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: I don't think there's any specific reason. It was like this earlier also, so I didn't modify it while moving it to "if" condition

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1989,6 +2001,51 @@ def make_api_call_with_retries(uri, data, tls_verify, fault_type, summary, cli_e
...
+    req_cnfJSON = {"kid": kid, "xms_ksl": "sw"}
+    req_cnf = base64.urlsafe_b64encode(json.dumps(req_cnfJSON).encode('utf-8')).decode('utf-8')
+
+    # remove padding '=' character
+    if req_cnf[len(req_cnf) - 1] == '=':
+        req_cnf = req_cnf[:-1]
+
+    token_data = {"token_type": "pop", "key_id": kid, "req_cnf": req_cnf}
+    profile = Profile(cli_ctx=cmd.cli_ctx)
+    credential, _, _ = profile.get_login_credentials(subscription_id=profile.get_subscription()["id"], resource=consts.KAP_1P_Server_App_Scope)
```
krdhruva, line 2033: what are other two return objects? Can this method throw exception?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1989,6 +2001,51 @@ def make_api_call_with_retries(uri, data, tls_verify, fault_type, summary, cli_e
...
+    accessToken = credential.get_token(consts.KAP_1P_Server_App_Scope, data=token_data)
+    jwtToken = accessToken.token
+
+    jwtTokenData = {"accessToken": jwtToken, "serverId": consts.KAP_1P_Server_AppId, "tenantID": tenantId, "kid": kid}
+    post_at_uri = f'https://localhost:{api_server_port}/identity/at'
+    # Needed to prevent skip tls warning from printing to the console
+    original_stderr = sys.stderr
+    f = open(os.devnull, 'w')
+    sys.stderr = f
+    post_at_response = make_api_call_with_retries(post_at_uri, jwtTokenData, "post", False, consts.PublicKey_Export_Fault_Type, 'pass', "pass", clientproxy_process)
```
krdhruva, line 2043: Is it assumption that this call will never fail? Didn't get the meaning of `pass` as error message.

src/connectedk8s/azext_connectedk8s/custom.py
krdhruva, line 1939: For my clarity: Will this code executed as we are calling `close_subprocess_and_raise_cli_error` in the end of `make_api_call_with_retries`?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1975,10 +1987,10 @@ def client_side_proxy(cmd,
     return expiry, clientproxy_process
 
 
-def make_api_call_with_retries(uri, data, tls_verify, fault_type, summary, cli_error, clientproxy_process):
+def make_api_call_with_retries(uri, data, method, tls_verify, fault_type, summary, cli_error, clientproxy_process):
```
krdhruva, line 1990: nit: this can be moved to utils

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: close_subprocess_and_raise_cli_error will happen only when there is an exception while making the CSP api call. That is, if at all CSP process is not up and took longer to start, then the API call which we're making will not find the endpoint and it will throw an exception. In that case we'll retry and close the process after retries exhaust. But, in this case of "500" response, we will get a proper error response and it is not an exception. Which is why, this code will be executed and we will refresh the PoP public key also

sirireddy12: I have the tested it also by setting publickey expiry to 10 mins

src/connectedk8s/azext_connectedk8s/_constants.py
sirireddy12: just realized that you meant to rename CSP to client proxy

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: Also, I don't think we should modify it now. Reason- we have been dumping telemetry with this summary since so long, and modifying it now might bring some confusion which fetching telemetry for this particular case. Let me know what you think

krdhruva (changes requested):
🕐

src/connectedk8s/azext_connectedk8s/custom.py
krdhruva: Yes, didn't understand why this reflected as change. Are we reverting that changes in this PR?

src/connectedk8s/azext_connectedk8s/_constants.py
sirireddy12: Done. renamed it to client proxy

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: Updated it to handle other errors like non 200/500 responses

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: Sure, created seperate one named "clientproxyutils" and moved all CSP related utils to that

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: Modified the wording to be similar since it's a new telemetry summary

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: subID and tenantID are the two other objects. get_login_creds won't throw an exception but it's a good idea to capsule the AT fetch part in try catch . Thanks for pointing out. Updated it to handle any exceptions

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: My bad. It was set to 'pass' while testing the dev build . Corrected it now

src/connectedk8s/azext_connectedk8s/_constants.py
sirireddy12: Updated the csp version with prod version number. Yet to test with prod version

src/connectedk8s/azext_connectedk8s/_constants.py
krdhruva: CSP is still present in error message. Have you pushed the change?

src/connectedk8s/azext_connectedk8s/custom.py
krdhruva: I meant `req_cnfJSON = {"kid": kid, "xms_ksl": "sw"}` part which we create on our side. So was curious about response.

krdhruva:
Not related to this PR but I observed we have added all error messages and summary in code itself. Does we need to consider localization of those messages? If so, we should move all messages to separate file and load from there. We can have work item to track this work separately.

krdhruva (approved):

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: Got it. We don't get that as part of the response

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -444,3 +446,26 @@ def validate_node_api_response(api_instance, node_api_response):
...
+def is_cli_using_msal_auth():
+        response_cli_version = az_cli("version --output json")
+        try:
+            cli_version = response_cli_version['azure-cli']
+        except Exception as ex:
+            raise CLIInternalError("Unable to decode the az cli version installed: {}".format(str(ex)))
+        if version.parse(cli_version) >= version.parse(consts.AZ_CLI_ADAL_TO_MSAL_MIGRATE_VERSION):
+            return True
+        else:
+            return False
```
zhoxing-ms, line 471:
```suggestion
def is_cli_using_msal_auth():
    response_cli_version = az_cli("version --output json")
    try:
        cli_version = response_cli_version['azure-cli']
    except Exception as ex:
        raise CLIInternalError("Unable to decode the az cli version installed: {}".format(str(ex)))
    if version.parse(cli_version) >= version.parse(consts.AZ_CLI_ADAL_TO_MSAL_MIGRATE_VERSION):
        return True
    else:
        return False
```

src/connectedk8s/azext_connectedk8s/azext_metadata.json
```diff
@@ -1,6 +1,5 @@
 {
     "name": "connectedk8s",
     "version": "1.2.1",
-    "azext.minCliCoreVersion": "2.16.0",
-    "azext.maxCliCoreVersion": "2.29.0"
+    "azext.minCliCoreVersion": "2.16.0"
```
zhoxing-ms, line 4: The modification here is because migrating `MSAL` can be compatible with the higher version of `azure-cli-core`, right?

zhoxing-ms:
Could you add some test cases for those changes?

src/connectedk8s/azext_connectedk8s/azext_metadata.json
sirireddy12: Yes that's right.

sirireddy12:
> Could you add some test cases for those changes?

We are tracking a separate WI to author test cases. We will be increasing the test coverage with that. Can we go ahead with this PR?

zhoxing-ms:
> We are tracking a separate WI to author test cases. We will be increasing the test coverage with that. Can we go ahead with this PR?

OK

src/connectedk8s/azext_connectedk8s/_clientproxyutils.py
```diff
@@ -0,0 +1,135 @@
...
+    # remove padding '=' character
+    if req_cnf[len(req_cnf) - 1] == '=':
+        req_cnf = req_cnf[:-1]
+
+    token_data = {"token_type": "pop", "key_id": kid, "req_cnf": req_cnf}
+    profile = Profile(cli_ctx=cmd.cli_ctx)
+    try:
+        credential, _, _ = profile.get_login_credentials(subscription_id=profile.get_subscription()["id"], resource=consts.KAP_1P_Server_App_Scope)
+        accessToken = credential.get_token(consts.KAP_1P_Server_App_Scope, data=token_data)
+        jwtToken = accessToken.token
```
jiasli, line 97: Is this supported by MSAL? Could you please add some PR description explaining how this works?

src/connectedk8s/azext_connectedk8s/_clientproxyutils.py
```diff
@@ -0,0 +1,135 @@
...
+    sys.stderr = original_stderr
+    publickey_info = json.loads(get_publickey_response.text)
+    kid = publickey_info['publicKey']['kid']
+
+    return kid
+
+
+def fetch_and_post_at_to_csp(cmd, api_server_port, tenantId, kid, clientproxy_process):
+    req_cnfJSON = {"kid": kid, "xms_ksl": "sw"}
+    req_cnf = base64.urlsafe_b64encode(json.dumps(req_cnfJSON).encode('utf-8')).decode('utf-8')
```
jiasli, line 86: VM SSH doesn't require `req_cnf` to be base64-encoded:

https://github.com/Azure/azure-cli-extensions/blob/2ce2a17ed87b194ab1088705ba31a2c267d79c50/src/ssh/azext_ssh/custom.py#L187-L193

Is this necessary?

jiasli: This is an example of real-life SSH `data`:

{'token_type': 'ssh-cert', 'req_cnf': '{"kty": "RSA", "n": "ALy4dqBiZKiT3EPSo...", "e": "AQAB", "kid": "5a0ecc0b859302c966b63087c6ade42b44ab52101afb8894d9ba26fd1978318d"}', 'key_id': '5a0ecc0b859302c966b63087c6ade42b44ab52101afb8894d9ba26fd1978318d'}

krdhruva: PoP support contract is bit different than ssh cert. We need `req_cnf` to generate PoP token.

src/connectedk8s/azext_connectedk8s/_clientproxyutils.py
sirireddy12: Done, Added the detailed description

src/connectedk8s/azext_connectedk8s/_clientproxyutils.py
jiasli: No. I am not saying VM SSH doesn't require `req_cnf`. It's just VM SSH doesn't require `req_cnf` to be **base64-encoded**.

src/connectedk8s/azext_connectedk8s/_clientproxyutils.py
jiasli: > We have made changes in our Go application(which generates the PoP token) to accept this Jwt token and further generate PoP token using that token.

Is this implementation open-source? If so, could you link to it?

sirireddy12: It's not open-source. But I believe you'll have access - https://msazure.visualstudio.com/One/_git/AzureArc-GuestProxy?path=%2Fpkg%2Fadal-with-pop&version=GBmaster&_a=contents

src/connectedk8s/azext_connectedk8s/_clientproxyutils.py
sirireddy12: In our PoP contract our Go app expects it to be base64 encoded - https://msazure.visualstudio.com/One/_git/AzureArc-GuestProxy?path=/pkg/adal-with-pop/popkey.go&version=GBmaster&line=96&lineEnd=98&lineStartColumn=1&lineEndColumn=71&lineStyle=plain&_a=contents

jiasli (approved):

zhoxing-ms (approved):

src/connectedk8s/azext_connectedk8s/_clientproxyutils.py
rayluo: First of all, congrats to @sirireddy12 and @krdhruva  that this experiment we discussed in January worked out nicely. :-)

I did not notice this PR earlier, and I know it is now merged. Just want to share my 2 cents on this topic.

> > `req_cnf` to be base64-encoded.
> > Is this necessary?
>
> In our PoP contract, [our Go app expects it to be base64 encoded](https://msazure.visualstudio.com/One/_git/AzureArc-GuestProxy?path=/pkg/adal-with-pop/popkey.go&version=GBmaster&line=96&lineEnd=98&lineStartColumn=1&lineEndColumn=71&lineStyle=plain&_a=contents)

The link above is the Go app's implementation which expects the `req_cnf` to be base64-encoded. But do we have that "our PoP contract"? Is the base64 encoding requirement defined in that contract? I got an impression from [this open document](https://datatracker.ietf.org/doc/html/draft-ietf-ace-oauth-params-16#section-3) and its subsequent examples that the `req_cnf` payload does not have base64 encoding.

You don't have to make any changes, if the current base64 encoding behavior is what your internal design requires, and you do not want to adjust it. It is just that the long term maintenance burden would be on the Azure CLI extensions code base: future maintainers would somehow need to know "extension Foo needs base64 encoding, extension Bar does not", and this kind of conversation would provide some context for that future maintainer.

jiasli: This is also my understanding and thoughts. @sirireddy12, I still think it would be better to draft some design spec in the extension source code for future developers or maintainers to understand how "your PoP contract" works.

sirireddy12: Hey @jiasli / @rayluo , missed this comment. Sure, let me work with my partner team and draft up a one pager about the PoP contract and details

PR #4474, opened 2022-03-01
SSH Extension v.1.0.1

vthiebaut10:
1. Fix scenarios where paths with spaces don't work.
2. Fix scenarios where paths with non-english characters don't work.
3. Remove "preview" message after installation.
4. Ensure relative paths work for all parameters that expect paths.
5. Add --ssh-client-folder parameter to allow users to decide which ssh and ssh-keygen executables are used.
6. Fix scenario where 32bit machines can't find System32 folder.
7. Print error messages from the client logs to the console.
8. Add "-v" to ssh arguments if interactive session is run in debug mode.
9. Fix bug for when users use * as IP on config
10. Warn users when they are using incompatible OpenSSH versions.
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
SSH Extension

src/ssh/azext_ssh/ssh_info.py
```diff
@@ -0,0 +1,91 @@
...
+        if not self.ip:
+            self.ip = "*"
+        lines = lines + self._get_ip_entry()
+        return lines
+
+    def _get_rg_and_vm_entry(self):
+        lines = []
+        lines.append("Host " + self.resource_group_name + "-" + self.vm_name)
+        lines.append("\tUser " + self.local_user)
+        lines.append("\tHostName " + self.ip)
```
bagajjal, line 72: self.ip is always non null? Better we have a safe check before adding it to config file.
This will help arc scenario as well.

src/ssh/azext_ssh/ssh_info.py
```diff
@@ -0,0 +1,91 @@
...
+        self.public_key_file = os.path.abspath(public_key_file) if public_key_file else None
+        self.private_key_file = os.path.abspath(private_key_file) if private_key_file else None
+        self.cert_file = os.path.abspath(cert_file) if cert_file else None
+        self.ssh_client_folder = os.path.abspath(ssh_client_folder) if ssh_client_folder else None
+
+    def get_config_text(self):
+        lines = [""]
+        if self.resource_group_name and self.vm_name:
+            lines = lines + self._get_rg_and_vm_entry()
+        # default to all hosts for config
```
bagajjal, line 62: line 62-64 should be moved before line 60 to take care of self.ip null scenario.

src/ssh/azext_ssh/ssh_info.py
```diff
@@ -0,0 +1,91 @@
...
+        if not self.ip:
+            self.ip = "*"
+        lines = lines + self._get_ip_entry()
```
bagajjal, line 65: We end up having two entries. This isn't the right behavior?

Host rg-vmname
    ....
    .....

Host * (Assuming ip is null)
    ....
    .....

src/ssh/azext_ssh/ssh_info.py
bagajjal, line 62: Why should we default to all hosts?

src/ssh/azext_ssh/ssh_info.py
vthiebaut10: Yes, that is the original behavior.

src/ssh/azext_ssh/_params.py
```diff
@@ -37,10 +40,16 @@ def load_arguments(self, _):
                    help='Folder where new generated keys will be stored.')
         c.argument('cert_file', options_list=['--certificate-file', '-c'], help='Path to certificate file')
         c.argument('port', options_list=['--port'], help='SSH port')
+        c.argument('ssh_client_folder', options_list=['--ssh-client-folder'],
```
bagajjal, line 43: We may need to update the documentation page (if any). please check with sandeep.

src/ssh/azext_ssh/_params.py
```diff
@@ -19,6 +19,9 @@ def load_arguments(self, _):
         c.argument('cert_file', options_list=['--certificate-file', '-c'],
                    help='Path to a certificate file used for authentication when using local user credentials.')
         c.argument('port', options_list=['--port'], help='SSH port')
+        c.argument('ssh_client_folder', options_list=['--ssh-client-folder'],
+                   help='Path to folder that contains ssh executables (ssh.exe, ssh-keygen.exe, etc). '
```
bagajjal, line 23: Folder path that contains ssh executables ...

src/ssh/azext_ssh/ssh_info.py
vthiebaut10: That was inherited from the original code before we picked it up. That condition never really happens though, because if the user doesn't provide an IP address and the _get_ssh_ip logic fails in custom._do_ssh_ip, then the command will fail.

src/ssh/azext_ssh/_params.py
vthiebaut10: I'll confirm, but it seems that the documentation for "az ssh" gets updated automatically once the changes get merged. https://docs.microsoft.com/en-us/cli/azure/ssh?view=azure-cli-latest

src/ssh/azext_ssh/_params.py
```diff
@@ -19,6 +19,9 @@ def load_arguments(self, _):
...
+                   'Default to ssh pre-installed if not provided.')
         c.positional('ssh_args', nargs='*', help='Additional arguments passed to OpenSSH')
 
     with self.argument_context('ssh config') as c:
```
bagajjal, line 27: Mismatch.
Line 28, config_path is defined as file path but ssh_config() treats it as a folder.

vthiebaut10: Config path is not treated as a folder. There is some logic in ssh_config() that checks if the folder exists, but that takes the dirname of the config_path. Is that what you are referring to?

src/ssh/azext_ssh/custom.py
```diff
@@ -15,87 +14,125 @@
...
+                       "Please delete once this certificate is no longer being used.", keys_folder)
 
-def _do_ssh_op(cmd, resource_group, vm_name, ssh_ip, public_key_file, private_key_file, use_private_ip,
-               username, cert_file, credentials_folder, op_call):
+    try:
+        cert_expiration = ssh_utils.get_certificate_start_and_end_times(cert_file, ssh_client_folder)[1]
+        logger.warning("Generated SSH certificate %s is valid until %s in local time.", cert_file, cert_expiration)
+    except Exception as e:
+        logger.warning("Couldn't determine certificate validity. Error: %s", str(e))
+        print(cert_file + "\n")
```
bagajjal, line 102: line 102 can be deleted?

vthiebaut10: I'm leaving that there for now because even if we fail to get the certificate validity, we still want to show the user the path of the certificate, and it felt weird to print that as a "warning".

src/ssh/azext_ssh/ssh_info.py
bagajjal: If it doesn't happen then please remove it, add an assertion and comments why we think it should be populated all the time.

src/ssh/azext_ssh/ssh_info.py
```diff
@@ -0,0 +1,91 @@
...
+        self.local_user = local_user
+        self.port = port
+        self.ssh_args = ssh_args
+        self.public_key_file = os.path.abspath(public_key_file) if public_key_file else None
+        self.private_key_file = os.path.abspath(private_key_file) if private_key_file else None
+        self.cert_file = os.path.abspath(cert_file) if cert_file else None
+        self.ssh_client_folder = os.path.abspath(ssh_client_folder) if ssh_client_folder else None
+
+    def get_host(self):
+        return self.local_user + "@" + self.ip
```
bagajjal, line 25: Please add a safe check local_user and IP as non-null.. I'm more worried about IP.

src/ssh/setup.py
```diff
@@ -7,7 +7,7 @@
 
 from setuptools import setup, find_packages
 
-VERSION = "1.0.0"
+VERSION = "1.0.1"
 
 CLASSIFIERS = [
     'Development Status :: 4 - Beta',
```
bagajjal, line 13: We must remove beta here.

bagajjal (changes requested):
.

src/ssh/azext_ssh/ssh_info.py
vthiebaut10: I know. In arc scenarios we don't add the "Host IP" block, but since this version doesn't include Arc changes, we don't need to check for that.

src/ssh/azext_ssh/custom.py
vthiebaut10: In a perfect world we would have the warning in yellow, and the message with the certificate file in green or something.

src/ssh/azext_ssh/ssh_utils.py
```diff
@@ -19,32 +20,35 @@
...
         # Create a new process that will wait until the connection is established and then delete keys.
-        cleanup_process = mp.Process(target=_do_cleanup, args=(delete_keys, delete_cert, cert_file, private_key_file,
-                                     log_file, True))
+        cleanup_process = mp.Process(target=_do_cleanup, args=(delete_keys, delete_cert, ssh_info.cert_file,
+                                     ssh_info.private_key_file, ssh_info.public_key_file, log_file, True))
         cleanup_process.start()
 
-    command = [_get_ssh_path(), _get_host(username, ip)]
-    command = command + _build_args(cert_file, private_key_file, port) + ssh_arg_list
+    command = [_get_ssh_client_path(ssh_client_folder=ssh_info.ssh_client_folder), ssh_info.get_host()]
```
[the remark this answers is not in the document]
vthiebaut10: That is looking a little confusing, but the arguments are correct. The ssh_command argument has "ssh" as the default value, so we don't need to provide it in this case, so we are only providing ssh_client_folder. "user@ip" is outside of the parenthesis, so it's not being taken as an argument. In the end "command" is a list with the path to ssh and the host: ['ssh', 'host@ip'].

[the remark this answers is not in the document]
bagajjal: My bad, I misread the code. For better readability, takeout the default value and always pass the "ssh" wherever its required.

src/ssh/setup.py
vthiebaut10: @jiasli Now this extension is no longer in preview, what should the Development Status be set to? I looked at the Development Status of other extensions, and it looks like they are all set to "4 - Beta".

bagajjal (approved):

src/ssh/azext_ssh/custom.py
```diff
@@ -62,16 +62,18 @@ def ssh_config(cmd, config_path, resource_group_name=None, vm_name=None, ssh_ip=
                                                    "does not exist.")
 
     # Default credential location
+    # Add logic to test if credentials folder is valid
     if not credentials_folder:
-        folder_name = config_session.ip
+        # * is not a valid name for a folder in Windows. Treat this as a special case.
+        folder_name = config_session.ip if config_session.ip != "*" else "all_ips"
```
bagajjal, line 68: some where you have to reinterpret all_ips?

jiasli (approved):

src/ssh/setup.py
```diff
@@ -7,7 +7,7 @@
 
 from setuptools import setup, find_packages
 
-VERSION = "1.0.1"
+VERSION = "1.0.0"
```
jiasli, line 10: Why it was changed back?

bagajjal: We still have some changes to make. We want to do it in a different PR.
Changing the version here will do a new release of "az ssh" cli extension. We want to change the version once we have all the changes.

src/ssh/setup.py
jiasli: I have no idea what it means. You may remove it.

bagajjal (approved):
.

PR #4524, opened 2022-03-14
[Quantum] Handle more storage types during workspace creation

warren-jones:
If you create a workspace with a storage account that is not “Standard_LRS”, the role assignment fails.  **`az workspace create`** needs to work with more storage account SKUs.  

Restrictions imposed by the new code: 
Only the "Standard" SKU tier is supported.  The storage account Kind is restricted to either "Storage" or "StorageV2".

---


warren-jones:
Adding test coverage

yonzhan:
Quantum

vxfield (approved):

warren-jones:
**Reverting to DRAFT status because of new comment on bug report:** 

> Per discussion, we still need to restrict certain storage account types due to known limitations.

PR #4536, opened 2022-03-16
Add 'packaging' as a dependency in the extension

sirireddy12:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/connectedk8s/setup.py
+++ b/src/connectedk8s/setup.py
@@ -17,7 +17,7 @@ except ImportError:
 # TODO: Confirm this is the right version number you want and it matches your
 # HISTORY.rst entry.
 
-VERSION = '1.2.5'
+VERSION = '1.2.6'
 
 # The full list of classifiers is available at
 # https://pypi.python.org/pypi?%3Aaction=list_classifiers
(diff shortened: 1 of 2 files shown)


zhoxing-ms (approved):

src/connectedk8s/setup.py
```diff
@@ -37,6 +37,7 @@
 DEPENDENCIES = [
     'kubernetes==11.0.0',
     'pycryptodome==3.9.8',
+    'packaging'
```
zhoxing-ms, line 40: In fact, I see that the dependency of `packaging>=20.9,<22.0` has been added in `azure-cli-core`. Why do you need to add it here?
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/setup.py#L57

In addition, don't we need to pin the version of `packaging`?

sirireddy12: azure-cli-core 2.16.0 doesn't have "packaging" , due to which it is unable to pull it

zhoxing-ms: This is because the CLI version used by the user is too old. Therefore, we recommend that the user upgrade the version used by CLI instead of adding duplicate dependencies in the CLI extension. Because adding duplicate dependencies can easily lead to package conflicts

jiasli: @zhoxing-ms is absolutely right. I really don't like the idea of adding `packaging` dependency because customer is using Azure CLI 2.16.0 which is just too old!

PR #4589, opened 2022-03-28
[Quantum][DRAFT -- Do not merge] Correct provider case

warren-jones:
When creating a workspace with the CLI, if a provider ID's case is not correct, that provider will not appear in the portal.  To ensure that the provider ID case is correct, this PR substitutes the provider ID from the offerings API for the provider ID specified in the `az quantum create workspace` command. 

During testing with multiple providers specified in a single command, a timeout occurred, so the workspace-creation timeout has been increased.  

---


yonzhan:
Quantum

vxfield (approved):

warren-jones:
Changes were made in the Quantum portal extension.  No changes are required in the Quantum CLI extension.

PR #4599, opened 2022-03-30
{CI} Update issue and pr template

wangzelin007:
Now we use keyword to route issue and pr.
However, as more and more modules are managed, the misjudgment rate is also getting higher.
The misjudgment rate is many times larger than the omission rate.

The reasons for misjudgment:
1. The information in the Environment Summary can easily lead to misjudgment
https://github.com/Azure/azure-cli/issues/21751
https://github.com/Azure/azure-cli-extensions/issues/4562
2. Users description contains multiple keywords
https://github.com/Azure/azure-cli/issues/21822
3. The log pasted by the user contains multiple keywords
https://github.com/Azure/azure-cli/issues/21793
4. Fabric bot supports neither complex regex nor if/else conditional

In the future, az $command will be used for matching.
But this will also lead to an increase in the omission rate.
https://github.com/Azure/azure-cli/issues/21726
So we should also modify template to guide user to fill in the related command.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,5 +1,9 @@
 - If the issue is to do with Azure CLI 2.0 in-particular, create an issue here at [Azure/azure-cli](https://github.com/Azure/azure-cli/issues)
 
+### Related command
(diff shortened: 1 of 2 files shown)


yonzhan:
CI

jsntcy (approved):

PR #4628, opened 2022-04-04
[Containerapp] Add new command `az containerapp ssh`

StrawnSC:
---

### General Guidelines

For new extensions:

### About Extension Publish


src/containerapp/azext_containerapp/commands.py
```diff
@@ -49,6 +50,11 @@ def load_command_table(self, _):
         g.custom_command('create', 'create_containerapp', supports_no_wait=True, exception_handler=ex_handler_factory(), table_transformer=transform_containerapp_output)
         g.custom_command('update', 'update_containerapp', supports_no_wait=True, exception_handler=ex_handler_factory(), table_transformer=transform_containerapp_output)
         g.custom_command('delete', 'delete_containerapp', supports_no_wait=True, confirmation=True, exception_handler=ex_handler_factory())
+        g.custom_command('exec', 'containerapp_ssh', is_preview=True, validator=validate_ssh)
```
panchagnula, line 53: we can remove is_preview here btw

panchagnula, line 53: @StrawnSC have we decided to call this exec? instead of SSH - ssh seems to be self-explanatory. "exec - can be execute or executive? Can we at least use execute (full word) if not ssh

src/containerapp/azext_containerapp/_validators.py
```diff
@@ -86,3 +89,68 @@ def validate_ingress(namespace):
         if namespace.ingress:
             if not namespace.target_port:
                 raise ValidationError("Usage error: must specify --target-port with --ingress")
+
+
+def _set_ssh_defaults(cmd, namespace):
```
panchagnula, line 94: NIT: since this technically a validator should the name of the method be like _validate_set (if this doing both)

src/containerapp/azext_containerapp/_ssh_utils.py
```diff
@@ -0,0 +1,171 @@
...
+
+def _decode_and_output_to_terminal(connection: WebSocketConnection, response, encodings):
+    for i, encoding in enumerate(encodings):
+        try:
+            print(response[2:].decode(encoding), end="", flush=True)
+            break
+        except UnicodeDecodeError as e:
+            if i == len(encodings) - 1:  # ran out of encodings to try
+                connection.disconnect()
+                logger.info("Proxy Control Byte: %s", response[0])
```
panchagnula, line 96: assuming logge.info is intentional over here - since these won't be visible to the user until using --verbose / --debug

src/containerapp/azext_containerapp/_ssh_utils.py
```diff
@@ -0,0 +1,171 @@
...
+def _send_stdin(connection: WebSocketConnection, getch_fn):
+    while connection.is_connected:
+        _resize_terminal(connection)
+        ch = getch_fn()
+        _resize_terminal(connection)
+        if connection.is_connected:
+            connection.send(b"".join([SSH_INPUT_PREFIX, ch]))
+
+
+def _resize_terminal(connection: WebSocketConnection):
```
panchagnula, line 134: Nice 👍

src/containerapp/azext_containerapp/_validators.py
panchagnula: ah! never mind just realized this is a private function & is called inside the validate.

src/containerapp/azext_containerapp/_ssh_utils.py
StrawnSC: Yes I thought it would be an excessive amount of logging to show automatically in this case

src/containerapp/azext_containerapp/commands.py
StrawnSC: During the meeting today, Tomer/Anthony seemed to be under the impression we should call it exec (like similar commands `docker exec` and `kubectl exec` which let you execute commands in docker containers/k8s pods)
 @anthonychu looped in the product owner for this feature on an ADO item to ask about what the naming should be

src/containerapp/azext_containerapp/_ssh_utils.py
StrawnSC: The ones I intend to display without `--debug` are done with `logger.warning` (like the direction to use `ctrl + D` to exit)

StrawnSC:
@panchagnula, I open a PR against Calvin's main [here](https://github.com/calvinsID/azure-cli-extensions/pull/56) so @calvinsID, @haroonf, and I can coordinate our containerapp releases together

PR #4643, opened 2022-04-07
[k8s-configuration] Bump `pycryptodome` to support Python 3.10

jiasli:
Fix https://github.com/Azure/azure-cli/issues/21964

MSI's bundled Python has been updated to 3.10 (https://github.com/Azure/azure-cli/pull/21746). `k8s-configuration` and `connectedk8s` depend on `pycryptodome==3.9.8` which doesn't release wheels for Python 3.10:

https://pypi.org/project/pycryptodome/3.9.8/#files

image

The latest `pycryptodome` 3.14.1 works with Python 3.10 using ABIs:

https://pypi.org/project/pycryptodome/#files

image

### DIFF
--- a/src/k8s-configuration/HISTORY.rst
+++ b/src/k8s-configuration/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+1.5.1
+++++++++++++++++++
+* Bump pycryptodome to 3.14.1 to support Python 3.10
+
 1.5.0
 ++++++++++++++++++
 * Update models to 2022-03-01 for GA
--- a/src/k8s-configuration/setup.py
+++ b/src/k8s-configuration/setup.py
@@ -16,7 +16,7 @@ except ImportError:
 
     logger.warn("Wheel is not available, disabling bdist_wheel hook")
 
-VERSION = "1.5.0"
+VERSION = "1.5.1"
 
 # The full list of classifiers is available at
 # https://pypi.python.org/pypi?%3Aaction=list_classifiers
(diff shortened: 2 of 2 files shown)


src/k8sconfiguration/setup.py
```diff
@@ -33,7 +33,7 @@
 ]
 
 # TODO: Add any additional SDK dependencies here
-DEPENDENCIES = ["pycryptodome~=3.9.8"]
+DEPENDENCIES = ["pycryptodome~=3.14.1"]
```
jiasli, line 36: If `k8sconfiguration` is deprecated:

https://github.com/Azure/azure-cli-extensions/blob/891004631627ec3e348198382e829b59803826ce/src/k8sconfiguration/HISTORY.rst#L6-L8

We should remove it.

yonzhan:
Bump pycryptodome

yonzhan (approved):

zhoxing-ms (approved):

src/k8sconfiguration/setup.py
zhoxing-ms: @ponatara Could you please confirm whether we can remove this extension?

zhoxing-ms: In fact, the description here should mean to deprecate/hide the `k8sconfiguration` command group and replace it with `k8s-configuration` command group.
@jonathan-innis Could please you confirm it? If so, please add specific information to the description to avoid ambiguity and confusion

jiasli:
`connectedk8s`'s `pycryptodome` has been bumped by https://github.com/Azure/azure-cli-extensions/pull/4607.

src/k8sconfiguration/setup.py
jonathan-innis: Yes, `k8sconfiguration` has been fully deprecated and we would like to remove it. @zhoxing-ms, do you want me to remove the extension? For this pycryptodome change, this change looks good and can be merged

jonathan-innis (approved):
🚀

src/k8sconfiguration/setup.py
zhoxing-ms: OK, if you are sure that most users have been migrated from `k8sconfiguration` extension to `k8s-configuration` extension, please remove it. 
In addition, the [Build Event (May 24-26)](https://mybuild.microsoft.com/en-US/home) the window period for CLI to launch batch breaking changes, which is a good opportunity for removing this extension

jiasli: I only mean removing the source code. The released extension whl won't be deleted.

zhoxing-ms: If the original extension is no longer used by users, I personally suggest that we can also consider removing them from `index.json`, so as to avoid too much contents in the `index.json` file

PR #4659, opened 2022-04-08
Flags override yaml

calvinsID:
- Flags should take precedence over the --yaml file supplied in create, update and copy-revision commands


yonzhan:
containerapp

PR #4660, opened 2022-04-09
Containerapps 4/11 Release [ Supporting MSI for ACA & some bug fixes]

calvinsID:
- P1 & P2 fixes
- MSI feature

### DIFF
--- a/src/containerapp/HISTORY.rst
+++ b/src/containerapp/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+0.1.1
+++++++
+* Subgroup commands for managed identities: az containerapp identity
+
 0.1.0
 ++++++
 * Initial release for Container App support with Microsoft.App RP.
--- a/src/containerapp/HISTORY.rst
+++ b/src/containerapp/HISTORY.rst
@@ -3,7 +3,7 @@
 Release History
 ===============
 
-0.1.1
+0.3.0
 ++++++
 * Subgroup commands for managed identities: az containerapp identity
 
--- a/src/containerapp/azext_containerapp/_help.py
+++ b/src/containerapp/azext_containerapp/_help.py
@@ -184,13 +184,13 @@ helps['containerapp env create'] = """
     - name: Create an environment with an auto-generated Log Analytics workspace.
       text: |
           az containerapp env create -n MyContainerappEnvironment -g MyResourceGroup \\
-              --location "Canada Central"
+              --location eastus2
     - name: Create an environment with an existing Log Analytics workspace.
       text: |
           az containerapp env create -n MyContainerappEnvironment -g MyResourceGroup \\
               --logs-workspace-id myLogsWorkspaceID \\
               --logs-workspace-key myLogsWorkspaceKey \\
-              --location "Canada Central"
+              --location eastus2
 """
 
 
--- a/src/containerapp/azext_containerapp/_help.py
+++ b/src/containerapp/azext_containerapp/_help.py
@@ -29,6 +29,12 @@ helps['containerapp create'] = """
               --image nginx --environment MyContainerappEnv \\
               --cpu 0.5 --memory 1.0Gi \\
               --min-replicas 4 --max-replicas 8
+    - name: Create a container app with secrets and environment variables.
+      text: |
+          az containerapp create -n MyContainerapp -g MyResourceGroup \\
+              --image my-app:v1.0 --environment MyContainerappEnv \\
+              --secrets mysecret=secretvalue1 anothersecret="secret value 2" \\
+              --env-vars GREETING="Hello, world" SECRETENV=secretref:anothersecret
     - name: Create a container app using a YAML configuration. Example YAML configuration - https://aka.ms/azure-container-apps-yaml
       text: |
           az containerapp create -n MyContainerapp -g MyResourceGroup \\
--- a/src/containerapp/azext_containerapp/_help.py
+++ b/src/containerapp/azext_containerapp/_help.py
@@ -29,12 +29,6 @@ helps['containerapp create'] = """
               --image nginx --environment MyContainerappEnv \\
               --cpu 0.5 --memory 1.0Gi \\
               --min-replicas 4 --max-replicas 8
-    - name: Create a container app with secrets and environment variables.
-      text: |
-          az containerapp create -n MyContainerapp -g MyResourceGroup \\
-              --image my-app:v1.0 --environment MyContainerappEnv \\
-              --secrets mysecret=secretvalue1 anothersecret="secret value 2" \\
-              --env-vars GREETING="Hello, world" SECRETENV=secretref:anothersecret
     - name: Create a container app using a YAML configuration. Example YAML configuration - https://aka.ms/azure-container-apps-yaml
       text: |
           az containerapp create -n MyContainerapp -g MyResourceGroup \\
--- a/src/containerapp/azext_containerapp/_help.py
+++ b/src/containerapp/azext_containerapp/_help.py
@@ -271,13 +271,13 @@ helps['containerapp identity assign'] = """
     examples:
     - name: Assign system identity.
       text: |
-          az containerapp identity assign
+          az containerapp identity assign -n myContainerapp -g MyResourceGroup --system-assigned
     - name: Assign user identity.
       text: |
-          az containerapp identity assign --identities myAssignedId
+          az containerapp identity assign -n myContainerapp -g MyResourceGroup --user-assigned myAssignedId
     - name: Assign system and user identity.
       text: |
-          az containerapp identity assign --identities [system] myAssignedId
+          az containerapp identity assign -n myContainerapp -g MyResourceGroup --system-assigned --user-assigned myAssignedId
 """
 
 helps['containerapp identity remove'] = """
@@ -286,15 +286,25 @@ helps['containerapp identity remove'] = """
     examples:
     - name: Remove system identity.
       text: |
-          az containerapp identity remove --identities [system]
+          az containerapp identity remove -n myContainerapp -g MyResourceGroup --system-assigned
     - name: Remove system and user identity.
       text: |
-          az containerapp identity remove --identities [system] myAssignedId
+          az containerapp identity remove -n myContainerapp -g MyResourceGroup --system-assigned --user-assigned myAssignedId
+    - name: Remove all user identities.
+      text: |
+          az containerapp identity remove -n myContainerapp -g MyResourceGroup --user-assigned
+    - name: Remove system identity and all user identities.
+      text: |
+          az containerapp identity remove -n myContainerapp -g MyResourceGroup --system-assigned --user-assigned
 """
 
 helps['containerapp identity show'] = """
     type: command
     short-summary: Show managed identities of a container app.
+    examples:
+    - name: Show managed identities.
+      text: |
+          az containerapp identity show -n myContainerapp -g MyResourceGroup
 """
 
 # Ingress Commands
@@ -520,7 +530,7 @@ helps['containerapp dapr enable'] = """
 
 helps['containerapp dapr disable'] = """
     type: command
-    short-summary: Disable Dapr for a container app.
+    short-summary: Disable Dapr for a container app. Removes existing values.
     examples:
     - name: Disable Dapr for a container app.
       text: |
--- a/src/containerapp/azext_containerapp/_help.py
+++ b/src/containerapp/azext_containerapp/_help.py
@@ -162,6 +162,14 @@ helps['containerapp revision copy'] = """
 
 """
 
+helps['containerapp revision copy'] = """
+    type: command
+    short-summary: Create a revision based on a previous revision.
+    examples:
+    - name: Create a revision based on a previous revision.
+      text: |
+          az containerapp revision copy -n MyContainerapp -g MyResourceGroup --cpu 0.75 --memory 1.5Gi
+"""
 
 # Environment Commands
 helps['containerapp env'] = """
@@ -254,6 +262,45 @@ helps['containerapp env dapr-component remove'] = """
     - name: Remove a Dapr component from a Container Apps environment.
       text: |
           az containerapp env dapr-component remove -g MyResourceGroup --dapr-component-name MyDaprComponentName --name MyEnvironment
+"""
+
+# Identity Commands
+helps['containerapp identity'] = """
+    type: group
+    short-summary: Commands to manage managed identities.
+"""
+
+helps['containerapp identity assign'] = """
+    type: command
+    short-summary: Assign managed identity to a container app.
+    long-summary: Managed identities can be user-assigned or system-assigned.
+    examples:
+    - name: Assign system identity.
+      text: |
+          az containerapp identity assign
+    - name: Assign user identity.
+      text: |
+          az containerapp identity assign --identities myAssignedId
+    - name: Assign system and user identity.
+      text: |
+          az containerapp identity assign --identities [system] myAssignedId
+"""
+
+helps['containerapp identity remove'] = """
+    type: command
+    short-summary: Remove a managed identity from a container app.
+    examples:
+    - name: Remove system identity.
+      text: |
+          az containerapp identity remove --identities [system]
+    - name: Remove system and user identity.
+      text: |
+          az containerapp identity remove --identities [system] myAssignedId
+"""
+
+helps['containerapp identity show'] = """
+    type: command
+    short-summary: Show managed identities of a container app.
 """
 
 # Ingress Commands
@@ -470,7 +517,7 @@ helps['containerapp dapr'] = """
 
 helps['containerapp dapr enable'] = """
     type: command
-    short-summary: Enable Dapr for a container app.
+    short-summary: Enable Dapr for a container app. Updates existing values.
     examples:
     - name: Enable Dapr for a container app.
       text: |
--- a/src/containerapp/azext_containerapp/_help.py
+++ b/src/containerapp/azext_containerapp/_help.py
@@ -280,10 +280,13 @@ helps['containerapp identity assign'] = """
           az containerapp identity assign -n myContainerapp -g MyResourceGroup --system-assigned
     - name: Assign user identity.
       text: |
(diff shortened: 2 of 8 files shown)


yonzhan:
Containerapps

panchagnula:
@zhoxing-ms can we get your review on this as well. We want to release this version on 4/11 (PST morning is better). Thanks!

src/containerapp/azext_containerapp/_help.py
```diff
@@ -162,6 +162,14 @@
 
 """
 
+helps['containerapp revision copy'] = """
```
panchagnula, line 165: Is this because we missed this In the existing help?

src/containerapp/azext_containerapp/_help.py
```diff
@@ -256,6 +264,45 @@
...
+    type: command
+    short-summary: Assign managed identity to a container app.
+    long-summary: Managed identities can be user-assigned or system-assigned.
+    examples:
+    - name: Assign system identity.
+      text: |
+          az containerapp identity assign
+    - name: Assign user identity.
+      text: |
+          az containerapp identity assign --identities myAssignedId
```
panchagnula, line 283: NIT: just to clarify this is user-assigned can we call this 'myUserAssignedId'

panchagnula: This comment is for all examples in help. There is no -n (containerapp name) in the help commands. I would expect this to be required

src/containerapp/azext_containerapp/_help.py
```diff
@@ -256,6 +264,45 @@
...
+      text: |
+          az containerapp identity remove --identities [system]
+    - name: Remove system and user identity.
+      text: |
+          az containerapp identity remove --identities [system] myAssignedId
+"""
+
+helps['containerapp identity show'] = """
+    type: command
+    short-summary: Show managed identities of a container app.
```
panchagnula, line 303: No examples for this help?

src/containerapp/azext_containerapp/_help.py
```diff
@@ -470,7 +517,7 @@
 
 helps['containerapp dapr enable'] = """
     type: command
-    short-summary: Enable Dapr for a container app.
+    short-summary: Enable Dapr for a container app. Updates existing values.
```
panchagnula, line 520: Curious given we added we update existing values here. What happens with disable? Do we remove existing values?

src/containerapp/azext_containerapp/_help.py
panchagnula: Not related  to this change but Line 156-157 of help " az containerapp revision copy -n MyContainerapp -g MyResourceGroup \\
              --cpu 0.75 --memory 1.5Gi" Does the --memory need the value with Units? what happens if one gives just the value? we should clarify this in examples.

panchagnula (changes requested):
Some generic help changes (not updated in this PR)
  text: |
          az containerapp env create -n MyContainerappEnvironment -g MyResourceGroup \\
              --location "Canada Central" 
Is there a command to get the list of regions supported on Container apps? If so  I would not use an location value here as <location of choice> See: command to get list of supported locations. If not change this to East US2 (Canada Central has some capacity constraints I believe)

In general, avoid using "Canada Central" across help examples.

panchagnula:
@haroonf FYI, incase you plan to make the changes here before the Calvin can get to it, Thanks!

panchagnula:
@calvinsID , @haroonf have you tried this example as is on your individual tests?  Mainly this line can cause confusion - since based on a shell this can be parsed differently  --secrets mysecret=secretvalue1 anothersecret="secret value 2" 

I would expect this to be --secrets "sec1=val1 sec2='spaced val2'" Ex. see this https://docs.microsoft.com/en-us/cli/azure/use-cli-effectively#use-quotation-marks-in-arguments. In general I would remove this example from our help, since most users will try to copy paste a help command & expect it to work.

 az containerapp create -n MyContainerapp -g MyResourceGroup \\
              --image my-app:v1.0 --environment MyContainerappEnv \\
              --secrets mysecret=secretvalue1 anothersecret="secret value 2" \\
              --env-vars GREETING="Hello, world" SECRETENV=secretref:anothersecret

panchagnula:
All the examples for --location has the "region name" instead of location. I assume we support both on the API, for consistency purposes use the location name instead of region name (avoiding the need to quotations around the region name) Ex. az group create --name RGlocalContext --location westeurope

src/containerapp/azext_containerapp/_params.py
```diff
@@ -28,7 +28,7 @@ def load_arguments(self, _):
     with self.argument_context('containerapp') as c:
         c.argument('tags', arg_type=tags_type)
         c.argument('managed_env', validator=validate_managed_env_name_or_id, options_list=['--environment'], help="Name or resource ID of the container app's environment.")
-        c.argument('yaml', type=file_type, help='Path to a .yaml file with the configuration of a container app. All other parameters will be ignored. For an example, see  https://docs.microsoft.com/azure/container-apps/azure-resource-manager-api-spec#examples')
+        c.argument('yaml', type=file_type, help='Path to a .yaml file with the configuration of a container app. Parameters passed in as flags will take precedence over the yaml. For an example, see  https://docs.microsoft.com/azure/container-apps/azure-resource-manager-api-spec#examples')
```
panchagnula, line 31: Update to : "Parameters used as arguments will override same the parameters used in yaml file." Flags can be misleading

src/containerapp/azext_containerapp/_params.py
```diff
@@ -144,14 +151,11 @@ def load_arguments(self, _):
     with self.argument_context('containerapp ingress traffic') as c:
         c.argument('traffic_weights', nargs='*', options_list=['--traffic-weight'], help="A list of revision weight(s) for the container app. Space-separated values in 'revision_name=weight' format. For latest revision, use 'latest=weight'")
 
-    with self.argument_context('containerapp secret set') as c:
+    with self.argument_context('containerapp secret') as c:
```
panchagnula, line 154: Should this be 'list' ? i.e containerapp secret list?

src/containerapp/azext_containerapp/custom.py
```diff
@@ -175,17 +211,288 @@ def update_containerapp_yaml(cmd, name, resource_group_name, file_name, from_rev
         handle_raw_exception(e)
 
 
-def create_containerapp_yaml(cmd, name, resource_group_name, file_name, no_wait=False):
-    yaml_containerapp = process_loaded_yaml(load_yaml_file(file_name))
+def _override_parameters_into_containerapp(cmd,
```
panchagnula, line 214: Can you add a comment to explain purpose of the helper method. With the definition of the method here , whenever a new property is added to the ContaineApp object we need to update this? Is there a way to make this generic or is this a limitation because we don't use SDK to get the Model object defintion?

src/containerapp/azext_containerapp/custom.py
```diff
@@ -175,17 +211,288 @@ def update_containerapp_yaml(cmd, name, resource_group_name, file_name, from_rev
...
+                                           dapr_enabled=None,
+                                           dapr_app_port=None,
+                                           dapr_app_id=None,
+                                           dapr_app_protocol=None,
+                                           revision_suffix=None,
+                                           startup_command=None,
+                                           args=None,
+                                           tags=None):
+    # Containers
+    container_needs_update = (image is not None or container_name is not None or set_env_vars is not None or
```
panchagnula, line 244: Same here add what is this for? especially if we are adding special handling on client side which is different from how the API behaves (making the behavior inconsistent from other clients). Thanks!

src/containerapp/azext_containerapp/custom.py
```diff
@@ -175,17 +211,288 @@ def update_containerapp_yaml(cmd, name, resource_group_name, file_name, from_rev
...
+                              remove_env_vars is not None or replace_env_vars is not None or remove_all_env_vars or
+                              cpu is not None or memory is not None or startup_command is not None or args is not None)
+
+    if container_needs_update:
+        if not container_name:
```
panchagnula, line 249: Honestly all this special code per argument type is making me nervous. Is this because we are trying to override the yaml arguments with the parameters from the command? Code comments will help here - so that the next person knows when they touch the code. Please add test cases for this in the next PR like we discussed. That will help ensure we have 0 regressions here.

panchagnula (changes requested):
@calvinsID let discuss these YAML specific changes on Monday, while we wait for review from @zhoxing-ms as well & target releasing this 4/11 evening PST. I will make sure to work with @zhoxing-ms

panchagnula:
@calvinsID we also need to update the version on this extension (please do a major version update because the argument vs YAML is a breaking change behavior)

panchagnula:
Lets make the non warning logger.warning - logger.info & add the name of the resource i.e instead of saying updating app say updating app "xyz".

src/containerapp/azext_containerapp/_help.py
```diff
@@ -256,6 +264,45 @@
...
+    examples:
+    - name: Assign system identity.
+      text: |
+          az containerapp identity assign
+    - name: Assign user identity.
+      text: |
+          az containerapp identity assign --identities myAssignedId
+    - name: Assign system and user identity.
+      text: |
+          az containerapp identity assign --identities [system] myAssignedId
```
zhoxing-ms, line 286: Please use `--system-assigned` to enable system assigned identity and `--user-assigned` with space separated resource IDs to add user assigned identities.
For more detail, please refer to the guideline [managed_identity_command_guideline.md](https://github.com/Azure/azure-cli/blob/dev/doc/managed_identity_command_guideline.md)

src/containerapp/azext_containerapp/_help.py
```diff
@@ -256,6 +264,45 @@
...
+helps['containerapp identity remove'] = """
+    type: command
+    short-summary: Remove a managed identity from a container app.
+    examples:
+    - name: Remove system identity.
+      text: |
+          az containerapp identity remove --identities [system]
+    - name: Remove system and user identity.
+      text: |
+          az containerapp identity remove --identities [system] myAssignedId
```
zhoxing-ms, line 298: Same as above

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -532,7 +532,7 @@ def _is_valid_weight(weight):
 
 
 def _update_traffic_weights(containerapp_def, list_weights):
-    if "traffic" not in containerapp_def["properties"]["configuration"]["ingress"] or list_weights and len(list_weights):
+    if "traffic" not in containerapp_def["properties"]["configuration"]["ingress"] or not containerapp_def["properties"]["configuration"]["ingress"]["traffic"] or list_weights and len(list_weights):
```
zhoxing-ms, line 535: > containerapp_def["properties"]["configuration"]["ingress"]["traffic"]

I want to confirm that you can ensure that there is no case where a parent attribute is `None`?

src/containerapp/azext_containerapp/custom.py
```diff
@@ -274,9 +608,9 @@ def create_containerapp(cmd,
                         min_replicas=None,
                         max_replicas=None,
                         target_port=None,
-                        transport="auto",
+                        transport=None,
                         ingress=None,
-                        revisions_mode="single",
+                        revisions_mode=None,
```
zhoxing-ms, line 613: May I ask will the modification of the default value here cause breaking change to users?

zhoxing-ms:
Could you add some tests for those changes?

src/containerapp/azext_containerapp/_help.py
calvinsID: There are error messages to handle this, I'll add additional helptext to clarify 
image

src/containerapp/azext_containerapp/_help.py
panchagnula: @zhoxing-ms thanks for pointing this out - we enabled these commands based on az acr [az acr identity | Microsoft Docs](https://docs.microsoft.com/en-us/cli/azure/acr/identity?msclkid=50244b51b9ba11ec9918d53ebe5b3292&view=azure-cli-latest) Is there a reason these are different from the guidelines?

src/containerapp/azext_containerapp/custom.py
calvinsID: Removed this change

panchagnula:
> Could you add some tests for those changes?

@zhoxing-ms we are reverting some major changes & will add tests as part of upcoming changes. Thanks!

src/containerapp/azext_containerapp/_help.py
panchagnula: Surface the API error message - we can do this later.

calvinsID: fixed

calvinsID: Made the `--memory` helptext more clear

src/containerapp/azext_containerapp/_help.py
calvinsID: Added

src/containerapp/azext_containerapp/_params.py
calvinsID: Removed yaml changes from this release

src/containerapp/azext_containerapp/_help.py
calvinsID: Added

src/containerapp/azext_containerapp/_params.py
calvinsID: Generalized it for `containerapp secret` so that all secret commands with these parameters will have the helptext

src/containerapp/azext_containerapp/custom.py
calvinsID: Removed yaml from this release

src/containerapp/azext_containerapp/custom.py
calvinsID: Removed yaml for this release

src/containerapp/azext_containerapp/_utils.py
panchagnula: remove this a part of this release

calvinsID: removed

src/containerapp/azext_containerapp/_params.py
```diff
@@ -75,6 +75,7 @@ def load_arguments(self, _):
         c.argument('transport', arg_type=get_enum_type(['auto', 'http', 'http2']), help="The transport protocol used for ingress traffic.")
 
     with self.argument_context('containerapp create') as c:
+        c.argument('assign_identity', nargs='+', help="Space-separated identities. Use '[system]' to refer to the system assigned identity.")
```
panchagnula, line 78: @haroonf update this to match the guidelines as well https://github.com/Azure/azure-cli/blob/dev/doc/managed_identity_command_guideline.md#enable-managed-identity-during-resource-creation

src/containerapp/azext_containerapp/_help.py
```diff
@@ -29,12 +29,6 @@
               --image nginx --environment MyContainerappEnv \\
               --cpu 0.5 --memory 1.0Gi \\
               --min-replicas 4 --max-replicas 8
-    - name: Create a container app with secrets and environment variables.
-      text: |
-          az containerapp create -n MyContainerapp -g MyResourceGroup \\
```
panchagnula, line 34: Did one or both of you try this example btw to check this works?

calvinsID: It works on powershell 7.x and bash 5.x

panchagnula: If this works then I am ok keeping this btw.

src/containerapp/azext_containerapp/_help.py
```diff
@@ -256,6 +258,55 @@
...
+    type: command
+    short-summary: Assign managed identity to a container app.
+    long-summary: Managed identities can be user-assigned or system-assigned.
+    examples:
+    - name: Assign system identity.
+      text: |
+          az containerapp identity assign -n myContainerapp -g MyResourceGroup --system-assigned
+    - name: Assign user identity.
+      text: |
+          az containerapp identity assign -n myContainerapp -g MyResourceGroup --user-assigned myAssignedId
```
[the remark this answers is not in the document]
haroonf: If the identity is in the same resource group as the CA, you don't need the entire resource id, otherwise you do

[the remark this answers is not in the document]
panchagnula: Might be good to another example clarifying for cross sub or cross RG support to include the full resource ID

src/containerapp/azext_containerapp/_help.py
```diff
@@ -176,13 +178,13 @@
     - name: Create an environment with an auto-generated Log Analytics workspace.
       text: |
           az containerapp env create -n MyContainerappEnvironment -g MyResourceGroup \\
-              --location "Canada Central"
```
panchagnula, line 179: can you check if there are other examples this needs to be updated as well?

panchagnula (approved):

src/containerapp/azext_containerapp/_help.py
calvinsID: re-added

src/containerapp/azext_containerapp/_help.py
calvinsID: Fixed by @haroonf

src/containerapp/HISTORY.rst
```diff
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+0.3.0
```
calvinsID, line 6: Bumped to 0.3.0 since the old containerapp private extension whl files we have been giving customers in the past went up to 0.2.4. We want to avoid confusion by bumping this

src/containerapp/azext_containerapp/_help.py
zhoxing-ms: > Is there a reason these are different from the guidelines?

@panchagnula This is because these commands were created before our guide existed. PR link: https://github.com/Azure/azure-cli/pull/12381
For these old commands, we will also consider migrating them to new commands with the same style in the future

zhoxing-ms (approved):

PR #4689, opened 2022-04-15
ContainerApps 04/19 Release

calvinsID:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
ContainerApps

StrawnSC:
@calvinsID can you un-draft this when you get the chance -- the GH actions update is good to go

The GitHub action didn't run successfully when I used a non-ACR registry (though ACR works great). I ran into the same issue when I tried with the previous API version so there wasn't a regression in the API

panchagnula:
@zhoxing-ms can we get an initial review from you please, so we can get this ready for 4/19 release after further testing on our end.

panchagnula:
@StrawnSC can you update the .github/CODEOWNERS to include your alias to the owners list as well please. Thanks!

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -527,7 +527,7 @@ class GitHubActionClient():
     @classmethod
     def create_or_update(cls, cmd, resource_group_name, name, github_action_envelope, headers, no_wait=False):
         management_hostname = cmd.cli_ctx.cloud.endpoints.resource_manager
-        api_version = NEW_API_VERSION
+        api_version = "2022-03-01"
```
zhoxing-ms, line 530: Why hard code the version here instead of upgrading the version of `NEW_API_VERSION`?  We usually use one version according to the RP dimension, because it is easy to maintain

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -556,7 +556,7 @@ def create_or_update(cls, cmd, resource_group_name, name, github_action_envelope
     @classmethod
     def show(cls, cmd, resource_group_name, name):
         management_hostname = cmd.cli_ctx.cloud.endpoints.resource_manager
-        api_version = NEW_API_VERSION
+        api_version = "2022-03-01"
```
zhoxing-ms, line 559: Same as above

src/containerapp/azext_containerapp/_clients.py
panchagnula: @zhoxing-ms We will have 2 versions for a while until we test all commands with stable api-version '2022-03-01'
The NEW_API_VERSION points to the preview version & we will move out of this in time for GA. Once we have all tests covered we will make the change as a separate PR. A lot of the nee features we are adding support for will be with the non-preview version, like this updates to GitHub actions. Hope that explains.

src/containerapp/azext_containerapp/_clients.py
panchagnula: @StrawnSC we should probably name these preview-version & stable-version, so the swap when we need to do is easy. "NEW_API_VERSION"  name here is misleading, since this is the not the new version.

StrawnSC: Just resolved this @panchagnula @zhoxing-ms

src/containerapp/HISTORY.rst
```diff
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+0.3.1
+++++++
+* Update "az containerapp github-action add" parameters
```
panchagnula, line 8: lets explicitly call out docker-path replaced by context-path

src/containerapp/azext_containerapp/tests/latest/test_containerapp_scenario.py
```diff
@@ -8,35 +8,176 @@
...
 
 
 TEST_DIR = os.path.abspath(os.path.join(os.path.abspath(__file__), '..'))
 
 
-@unittest.skip("Managed environment flaky")
+@live_only()
 class ContainerappScenarioTest(ScenarioTest):
     @AllowLargeResponse(8192)
     @ResourceGroupPreparer(location="centraluseuap")
```
panchagnula, line 21: not all subs have permissions to EUAP regions, & these tests don't run on our subs so avoid using this region. Use our non-canary region for testing.

panchagnula (approved):
One comment - otherwise LGTM

src/containerapp/azext_containerapp/tests/latest/test_containerapp_scenario.py
StrawnSC: Changed this to "eastus2"

src/containerapp/HISTORY.rst
StrawnSC: Just fixed this

StrawnSC:
> @calvinsID can you un-draft this when you get the chance -- the GH actions update is good to go
> 
> The GitHub action didn't run successfully when I used a non-ACR registry (though ACR works great). I ran into the same issue when I tried with the previous API version so there wasn't a regression in the API

@panchagnula update on this: I was erroneously using `--registry-url docker.io/<username>` instead of `--registry-url docker.io`. It works correctly now. That's my bad; the API is functional here.

src/containerapp/azext_containerapp/_params.py
```diff
@@ -131,10 +131,11 @@ def load_arguments(self, _):
         c.argument('registry_url', help='The container registry server, e.g. myregistry.azurecr.io')
         c.argument('registry_username', help='The username of the registry. If using Azure Container Registry, we will try to infer the credentials if not supplied')
         c.argument('registry_password', help='The password of the registry. If using Azure Container Registry, we will try to infer the credentials if not supplied')
-        c.argument('docker_file_path', help='The dockerfile location, e.g. ./Dockerfile')
+        c.argument('context_path', help='Path in the repo from which to run the docker build. Defaults to "./"')
         c.argument('service_principal_client_id', help='The service principal client ID. ')
         c.argument('service_principal_client_secret', help='The service principal client secret.')
         c.argument('service_principal_tenant_id', help='The service principal tenant ID.')
+        c.argument('image', type=str, options_list=['--image', '-i'], help="Container image name that the Github Action should use. Defaults to the Container App name.")
```
zhoxing-ms, line 138: The default value of `type` is `str`, so there is no need to define it repeatedly
```suggestion
        c.argument('image', options_list=['--image', '-i'], help="Container image name that the Github Action should use. Defaults to the Container App name.")
```

zhoxing-ms (approved):

PR #4721, opened 2022-04-22
[rdbms-connect] Use compatible release for `setproctitle` to support Python 3.10

jiasli:
Similar to #4643

Windows MSI package's bundled Python has been updated to 3.10 (https://github.com/Azure/azure-cli/pull/21746). `rdbms-connect` depends on `setproctitle==1.2.2` which doesn't release wheels for Python 3.10.

`setproctitle` starts to support Python 3.10 in 1.2.3. We use [compatible release](https://peps.python.org/pep-0440/#compatible-release) for `setproctitle` so that 1.2.3 can be installed.

Note: It is not a good practice to use a pinned version in `setup.py`.

### DIFF
--- a/src/rdbms-connect/HISTORY.rst
+++ b/src/rdbms-connect/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+1.0.2
+++++++
++ Use compatible release for `setproctitle` to support Python 3.10
+
 1.0.1
 ++++++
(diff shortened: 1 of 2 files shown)


yonzhan:
rdbms-connect

calvinhzy (approved):

DaeunYim (approved):

PR #4726, opened 2022-04-22
Azure Container Apps - 4/28 release

haroonf:
Adds support for
az containerapp up
az containerapp logs show (Log-stream support)
az containerapp exec (SSH support)
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
Container

panchagnula:
@zhoxing-ms please give us your initial review feedback & we will address this on Monday (PST hours) for a 4/26 release. Thank you.

src/containerapp/azext_containerapp/.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+ignore =
+     W503   # line break before binary operator, not compliant with PEP 8
+     E203   # whitespace before ':', not compliant with PEP 8
```
zhoxing-ms, line 4: Why don't you fix the issue of the code style directly but ignore the flake8 checks?

src/containerapp/azext_containerapp/_acr_run_polling.py
```diff
@@ -0,0 +1,112 @@
...
+    return LROPoller(
+        client=client,
+        initial_response=client.get(
+            resource_group_name, registry_name, run_id, cls=lambda x, y, z: x),
+        deserialization_callback=deserialize_run,
+        polling_method=RunPolling(
+            cmd=cmd,
+            registry_name=registry_name,
+            run_id=run_id
+        ))
```
zhoxing-ms, line 36: The logic of serializing and calling `LROPoller` is usually implemented in the SDK, why don't you use SDK?

src/containerapp/azext_containerapp/_params.py
```diff
@@ -81,6 +106,12 @@ def load_arguments(self, _):
         c.argument('user_assigned', nargs='+', help="Space-separated user identities to be assigned.")
         c.argument('system_assigned', help="Boolean indicating whether to assign system-assigned identity.")
 
+    with self.argument_context('containerapp create', arg_group='Container') as c:
+        c.argument('image', type=str, options_list=['--image', '-i'], help="Container image, e.g. publisher/image-name:tag.")
```
zhoxing-ms, line 110: > type=str

`str` is the default for `type`, so there is no need to declare it again

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -2,21 +2,218 @@
...
+from msrestazure.tools import parse_resource_id, is_valid_resource_id, resource_id
 
 from ._clients import ContainerAppClient
 from ._client_factory import handle_raw_exception, providers_client_factory, cf_resource_groups, log_analytics_client_factory, log_analytics_shared_key_client_factory
 
 logger = get_logger(__name__)
 
 
+# original implementation at azure.cli.command_modules.role.custom.create_service_principal_for_rbac
+# reimplemented to remove unnecessary warning statements
```
zhoxing-ms, line 27: > \# logger.warning("Creating '%s' role assignment under scope '%s'", role, scope) 
> \# retry till server replication is done

I'm not sure whether it's reasonable to reimplement the whole `create_service_principal_for_rbac` method just to remove unnecessary warning statements. It has a disadvantage that it will cause the same code duplication and increase the maintenance cost. @jiasli Do you have any suggestions for this?

src/containerapp/azext_containerapp/_params.py
panchagnula: @haroonf can you fix this.

src/containerapp/azext_containerapp/.flake8
panchagnula: @haroonf can you remove these changes as well - will re-visit this once Silas is back.

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -189,7 +189,10 @@ def await_github_action(cmd, token, repo, branch, name, resource_group_name, tim
     animation.flush()
     animation.tick()
     animation.flush()
-    run = workflow.get_runs()[0]
+    runs = workflow.get_runs()
+    while runs is None or runs.totalCount < 1:
```
panchagnula, line 193: While this works @StrawnSC to give you some context this issue is that, the call the check list & retrieve the [0] once the GitHub action add is completed. Talking to Sanchit the GitHub actions add Async response returns an operationId & we can poll on the operationId, to validate this is 200 & successful, before we call the list of workflows.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -2068,6 +2068,9 @@ def containerapp_up_logic(cmd, resource_group_name, name, managed_env, image, en
     if containerapp_def:
         ca_exists = True
 
+    if image is None:
+        image = "mcr.microsoft.com/azuredocs/containerapps-helloworld:latest"
```
panchagnula, line 2071: NIT: Can we add comment explaining the need for this - so we don't accidentally delete / remove this?

src/containerapp/azext_containerapp/custom.py
```diff
@@ -1996,7 +1996,7 @@ def containerapp_up(cmd,
                     logs_key=None,
                     repo=None,
                     token=None,
-                    branch=None,
+                    branch="main",
```
panchagnula, line 1998: @StrawnSC add some exceptional handling for if the branch info is incorrect or missing, so a user can run the command with --branch (See Jeff's email with more details as well)

src/containerapp/azext_containerapp/custom.py
```diff
@@ -2016,6 +2016,11 @@ def containerapp_up(cmd,
         ingress = "external" if not ingress else ingress
         target_port = 80 if not target_port else target_port
 
+    if image:
+        if ingress and not target_port:
+            target_port = 80
+            logger.warning("No ingress provided, defaulting to port 80.")
```
panchagnula, line 2022: Also add the command to change this later. i.e may be containerapp up with --ingress & --target-port & the update command they can run post app is created.

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -704,7 +704,7 @@ def up_output(app):
         url = f"http://{url}"
 
     logger.warning(
-        f"\nYour container app ({app.name}) has been created a deployed! Congrats! \n"
+        f"\nYour container app ({app.name}) has been created and deployed! Congrats! \n"
```
panchagnula, line 707: Nit: remove () around the app name.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -2012,7 +2012,7 @@ def containerapp_up(cmd,
     image = _reformat_image(source, repo, image)
     token = None if not repo else get_github_access_token(cmd, ["admin:repo_hook", "repo", "workflow"], token)
 
-    if image and "mcr.microsoft.com/azuredocs/containerapps-helloworld:latest" in image.lower():
+    if image and "mcr.microsoft.com/azuredocs/containerapps-helloworld" in image.lower():
```
panchagnula, line 2015: lets add the quick start image as a constant & reference here so if this changes we can easily update this in the constants.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -2012,7 +2012,7 @@ def containerapp_up(cmd,
...
         ingress = "external" if not ingress else ingress
```
panchagnula, line 2016: Good to see some defaults added but can we also make sure the Output response includes the FQDN, currently not seeing it.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -1021,6 +1031,36 @@ def show_managed_identity(cmd, name, resource_group_name):
...
+                                  "the --token argument. Run 'az webapp deployment github-actions add --help' "
+                                  "for more information.") from e
+        except GithubException as e:
+            error_msg = "Encountered GitHub error when accessing {} repo".format(repo)
+            if e.data and e.data['message']:
+                error_msg += " Error: {}".format(e.data['message'])
+            raise CLIInternalError(error_msg) from e
+
+
 def create_or_update_github_action(cmd,
```
panchagnula, line 1064: Fix the optional branch value here as well from None to main. GitHub actions add is broken with the same behavior as well. @StrawnSC can the existing tests for GitHub-actions be improved to catch the lack of defaults like for branch?

src/containerapp/azext_containerapp/_utils.py
StrawnSC: To be more specific, I reimplemented the method to eliminate a warning statement that said: 
```
The output includes credentials that you must protect. 
Be sure that you do not include these credentials in your code or check the credentials into your source control.
For more information, see https://aka.ms/azadsp-cli
```

In this case, the warning is incorrect because these service principal credentials are not outputted (they are passed to the containerapp GitHub action API)

jiasli: I agree with @zhoxing-ms. Vendoring `create_service_principal_for_rbac` is a bad idea. `create_service_principal_for_rbac` is updated frequently due to design change or security requirements. Once vendored, you also need to redo Microsoft Graph migration (https://github.com/Azure/azure-cli/issues/22174). For example, `credential_description` and `password` are removed from `create_application`.

If you want the warning not to be shown, I can add a switch.

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -2,21 +2,218 @@
...
 
+# original implementation at azure.cli.command_modules.role.custom.create_service_principal_for_rbac
+# reimplemented to remove unnecessary warning statements
+def create_service_principal_for_rbac(  # pylint:disable=too-many-statements,too-many-locals, too-many-branches, unused-argument, inconsistent-return-statements
+        cmd, name=None, years=None, create_cert=False, cert=None, scopes=None, role=None,
+        show_auth_for_sdk=None, skip_assignment=False, keyvault=None):
+    from azure.cli.command_modules.role.custom import (_graph_client_factory, TZ_UTC, _process_service_principal_creds,
+                                                       _validate_app_dates, create_application,
+                                                       _create_service_principal, _create_role_assignment,
+                                                       _error_caused_by_role_assignment_exists)
```
jiasli, line 34: Importing and using private method is definitely a no-go!

jiasli (changes requested):
Shouldn't copy the code of `create_service_principal_for_rbac`.

src/containerapp/azext_containerapp/custom.py
StrawnSC: I should have called this out in a comment, my bad. The reason the help text says the GitHub action defaults to the main branch, but this parameter is None by default is that the API assumes the branch when it is passed as None. When I tested on two of my own repos, the action is created on the main branch when no `--branch` is supplied

src/containerapp/azext_containerapp/custom.py
haroonf: @panchagnula should this be logger.info or warning?

src/containerapp/azext_containerapp/_utils.py
panchagnula: @jiasli this is something we need for the initial release of the extension for 'up' support &  we have to copy this over since we want some additional logger.warning we want to add. We will move to the new graph calls with the next release, Could you point us to the new Microsoft.Graph migration changes as well, so we can change it to that & do testing for 'up' command? Thanks!

src/containerapp/azext_containerapp/tests/latest/test_containerapp_up.py
```diff
@@ -0,0 +1,29 @@
...
+TEST_DIR = os.path.abspath(os.path.join(os.path.abspath(__file__), '..'))
+
+
+@live_only()
+class ContainerAppUpImageTest(ScenarioTest):
+    @ResourceGroupPreparer(location="eastus2")
+    def test_containerapp_up_image_e2e(self, resource_group):
+        env_name = self.create_random_name(prefix='env', length=24)
+        self.cmd(f'containerapp env create -g {resource_group} -n {env_name}')
+        image = "mcr.microsoft.com/azuredocs/containerapps-helloworld:latest"
```
panchagnula, line 26: @haroonf in the next release lets update this test to not include the tag - since we had a bug on this a while back would be good to make that the test case. Will file a bug on you for this - not a blocker. Thanks!

panchagnula (approved):
LGTM

panchagnula:
> Shouldn't copy the code of `create_service_principal_for_rbac`

@jiasli / @zhoxing-ms we need to copy & maintain this code for "up" scenarios, since we want some specific logging & error handling for up. We don't to modify the generic command code, for our purposes.  We add this mainly because with az containerapp up with GitHub-actions we do the best effort to have all the required non ACA resources ready/set up, so they don't need to run multiple separate commands. Hope this answers why we need this here. Thanks!

src/containerapp/azext_containerapp/_utils.py
panchagnula: I found this https://docs.microsoft.com/en-us/cli/azure/microsoft-graph-migration?tabs=powershell#az-ad-app-create @StrawnSC I will add this to our ADO work item so we understand the exact changes we need to make next to ensure we use Microsft.Graph implementation to add SP.

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -339,7 +341,7 @@ def _create_service_principal(cmd, resource_group_name, env_resource_group_name)
         )
     sp = create_service_principal_for_rbac(cmd, scopes=scopes, role="contributor")
 
-    logger.info(f"Created service principal: {sp['displayName']}")
+    logger.warning(f"Created service principal: {sp['displayName']} with ID {sp['appId']}")
```
panchagnula, line 344: Thanks for adding this is useful for investigations when GitHub workflow is failing with SP access permissions error. Please add the same when we are fetching & re-using an existing SP. Thanks!

src/containerapp/azext_containerapp/_acr_run_polling.py
zhoxing-ms: OK, please try to use Python SDK instead of manually implementing these codes in the future

zhoxing-ms:
In fact, I personally think the main problem of blocking this PR merge is to copy the code of `create_service_principal_for_rbac`, which is really not a good design. 
@panchagnula Personally, I think maybe we can approve and merge this PR first to meet your release schedule, then please submit a PR for refactoring these codes as soon as possible, what do you think of this plan? @panchagnula @jiasli

src/containerapp/azext_containerapp/_utils.py
jiasli: This extension will break with `azure-cli-core` 2.37.0. Please see the upper bound of `azure-cli-core` to 2.36.0.

jiasli (approved):
Approve as long as `maxCliCoreVersion` is set to 2.36.0.

StrawnSC:
> Approve as long as `maxCliCoreVersion` is set to 2.36.0.

Done, thank you

panchagnula:
> Approve as long as `maxCliCoreVersion` is set to 2.36.0.

Thank you @jiasli & @zhoxing-ms - can we get this merged & released for 4/29 early hours PST .

zhoxing-ms (approved):

PR #4740, opened 2022-04-26
[AKS] Support update kubelet identity

norshtein:
---

### General Guidelines

For new extensions:

### About Extension Publish


src/aks-preview/HISTORY.md
```diff
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+0.5.64
+++++++
+
+* Add support for updating kubelet identity. Command is
+  * `az aks create --cluster-snapshot-id <snapshot-id>`
```
PixelRobots, line 10: Are we sure this is the right command?

FumingZhang (approved):
LGTM

src/aks-preview/HISTORY.md
```diff
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+0.5.64
+++++++
+
+* Add support for updating kubelet identity. Command is
+  * `az aks update --assign-kubelet-identity <kubelelt-identity-resource-id>`
```
zhoxing-ms, line 10: May I ask what's the difference between `kubelet identity` and `managed identity`?

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -354,6 +354,8 @@ def load_arguments(self, _):
         c.argument('enable_managed_identity', action='store_true')
         c.argument('assign_identity', type=str,
                    validator=validate_assign_identity)
+        c.argument('assign_kubelet_identity', type=str,
+                   validator=validate_assign_kubelet_identity)
```
zhoxing-ms, line 358: `str` is the default value of `assign_kubelet_identity`, so there is no need to add it repeatedly
```suggestion
        c.argument('assign_kubelet_identity',
                   validator=validate_assign_kubelet_identity)
```

src/aks-preview/HISTORY.md
norshtein: kubelet identity is used by kubelet in agent node, its main purpose is authenticating to ACR to pull image from ACR;
`--enable-managed-identity` or `--assign-identity` refers to control plane identity, which is used by AKS managed master components such as kube-controller-manager, its main purpose to authenticating to Azure to manage Azure resource, for example, when kube-controller-manager find it needs to update load balancer, it will authenticate using control plane identity  to do the operation.

FumingZhang: As far as I know, there is a (managed or user assigned) identity used by the control plane (i.e., managed cluster) and another one (this kubelet identity) used by the nodes (mainly used for authentication when pulling image from acr).

norshtein:
The live test failed because RP change is not rolled out to all regions

norshtein:
Updated live test region to centraluseuap. Previous live test is done in westus2 staging env, prod env westus2 does not have this change yet.

zhoxing-ms (approved):

PR #4784, opened 2022-05-06
[Quantum] Add "Credits for all" support to the CLI

warren-jones:
Automatically add providers participating in the "Credits for all" promotion when a workspace is created.

Because a new "autoAdd" flag was added in the Quantum Service API, the CLI's control-plane client requires an update.  A script that generates either the data-plane client or the control-plane client is included in this PR, along with the new AutoRest-generated client files that support API version 2022-01-10-preview.

Note: The files in src\quantum\azext_quantum\vendored_sdks were generated by the Microsoft AutoRest Code Generator according to parameters in src\quantum\eng\Generate-Client.ps1

---

### Related commands
az quantum workspace create
az quantum offerings list

### General Guidelines


yonzhan:
Quantum

wangzelin007:
Hi [warren-jones](https://github.com/warren-jones),
Do these features need to be released before Microsoft Build 2022 (2022-05-24)
If not, can we move this pr to next milestone Jun 2022 (2022-07-05) ?

src/quantum/eng/Generate-Client.ps1
```diff
@@ -0,0 +1,100 @@
+# --------------------------------------------------------------------------------------------
```
ricardo-espinoza, line 1: Comment only: I'm glad that this got added to the repo. While it's not a part of this PR, I'd recommend a follow up change adding a Readme.md in this directory explaining how to use it as well as the other script.

ricardo-espinoza, line 1: Is this file a replacement for Generate-DataPlane-Client.ps1, or will we only use this to generate the control plane one?

ricardo-espinoza (approved):

ricardo-espinoza:
LGTM, recommend changing it to 'ready for review'.

src/quantum/eng/Generate-Client.ps1
warren-jones: Yes, a Readme.md is a good idea.  I planned to write a wiki page about this procedure, but a Readme.md would be easier to find.

kairu-ms (approved):

PR #4794, opened 2022-05-09
containerapp Version 0.3.4 Release 

StrawnSC:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
containerapp

src/containerapp/azext_containerapp/_params.py
```diff
@@ -179,14 +187,20 @@ def load_arguments(self, _):
...
+
     with self.argument_context('containerapp ingress') as c:
         c.argument('allow_insecure', help='Allow insecure connections for ingress traffic.')
         c.argument('type', validator=validate_ingress, arg_type=get_enum_type(['internal', 'external']), help="The ingress type.")
         c.argument('transport', arg_type=get_enum_type(['auto', 'http', 'http2']), help="The transport protocol used for ingress traffic.")
         c.argument('target_port', type=int, validator=validate_target_port, help="The application port used for ingress traffic.")
 
     with self.argument_context('containerapp ingress traffic') as c:
-        c.argument('traffic_weights', nargs='*', options_list=['--traffic-weight'], help="A list of revision weight(s) for the container app. Space-separated values in 'revision_name=weight' format. For latest revision, use 'latest=weight'")
+        c.argument('revision_weights', nargs='+', options_list=['--revision-weight'], help="A list of revision weight(s) for the container app. Space-separated values in 'revision_name=weight' format. For latest revision, use 'latest=weight'")
```
zhoxing-ms, line 202: Could we use the `deprecate_info` flag to the `traffic_weights` parameter to set deprecation information first to avoid directly causing breaking change to users?

src/containerapp/azext_containerapp/custom.py
```diff
@@ -1071,7 +1075,7 @@ def create_or_update_github_action(cmd,
                                    registry_url=None,
                                    registry_username=None,
                                    registry_password=None,
-                                   branch="main",
+                                   branch=None,
```
zhoxing-ms, line 1078: Will the change of this default value cause breaking change to users?

src/containerapp/azext_containerapp/_params.py
StrawnSC: @zhoxing-ms essentially this renamed to `--revision-weight` to reduce confusion with the new parameter `--traffic-weight`. To get around the need for a deprecation, I can add `--traffic-weight` as an option for `--revision-weight`

src/containerapp/azext_containerapp/custom.py
StrawnSC: The change we made here was to default the branch to the GitHub repository's default branch -- so if a user was leaving `--branch` empty *and* their default branch is `main`, there will not be a breaking change for them. Thus I don't think many users will run into a problem with this. Just in case, I will go ahead and add "BREAKING CHANGE" to the history.rst entry

StrawnSC:
@zhoxing-ms just made the two changes in response to your comments

src/containerapp/azext_containerapp/_params.py
```diff
@@ -199,7 +199,7 @@ def load_arguments(self, _):
         c.argument('target_port', type=int, validator=validate_target_port, help="The application port used for ingress traffic.")
 
     with self.argument_context('containerapp ingress traffic') as c:
-        c.argument('revision_weights', nargs='+', options_list=['--revision-weight'], help="A list of revision weight(s) for the container app. Space-separated values in 'revision_name=weight' format. For latest revision, use 'latest=weight'")
+        c.argument('revision_weights', nargs='+', options_list=['--revision-weight', '--traffic-weight'], help="A list of revision weight(s) for the container app. Space-separated values in 'revision_name=weight' format. For latest revision, use 'latest=weight'")
```
zhoxing-ms, line 202: I suggest using the `deprecated_info` for `--traffic-weight` if the `--traffic-weight` parameter is no longer recommended in the future

src/containerapp/azext_containerapp/_params.py
zhoxing-ms: I suggest using the `deprecated_info` for `--traffic-weight` if the `--traffic-weight` parameter is no longer recommended in the future, what do you think of it?

StrawnSC: Sounds good I just added as a deprecated argument option

zhoxing-ms (approved):

PR #4837, opened 2022-05-16
Containerapp Version 0.3.5 : Final release for //BUILD

StrawnSC:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
Containerapp

panchagnula:
@zhoxing-ms can we get a review on this for release on Monday?
@StrawnSC please include this change to this as well, so we can release this right after core-cli release for new graph-api support, https://github.com/calvinsID/azure-cli-extensions/pull/110

yonzhan:
Add @jiasli @zhoxing-ms  for awareness

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -12,145 +12,85 @@
...
+    vnet_client = network_client_factory(cmd.cli_ctx)
+    location = vnet_client.virtual_networks.get(resource_group_name=parsed_rid.get("resource_group"),
+                                                virtual_network_name=parsed_rid.get("name")).location
+    return _normalize_location(cmd, location)
+
+
+def create_service_principal_for_github_action(cmd, scopes=None, role="contributor"):
+    from azure.cli.command_modules.role.custom import (create_application, create_service_principal,
+                                                       create_role_assignment, show_service_principal)
+    from azure.cli.command_modules.role.msgrpah._graph_client import GraphClient
```
jiasli, line 65: Please don't import from private module. Consider using `azure.cli.command_modules.role.graph_client_factory`: https://github.com/Azure/azure-cli/blob/cac3f06cf483d8c81f93c9cd556116e27aff2182/src/azure-cli/azure/cli/command_modules/role/__init__.py#L44

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -12,145 +12,85 @@
...
+def get_vnet_location(cmd, subnet_resource_id):
+    parsed_rid = parse_resource_id(subnet_resource_id)
+    vnet_client = network_client_factory(cmd.cli_ctx)
+    location = vnet_client.virtual_networks.get(resource_group_name=parsed_rid.get("resource_group"),
+                                                virtual_network_name=parsed_rid.get("name")).location
+    return _normalize_location(cmd, location)
+
+
+def create_service_principal_for_github_action(cmd, scopes=None, role="contributor"):
+    from azure.cli.command_modules.role.custom import (create_application, create_service_principal,
```
jiasli, line 63: Methods like `create_application` are for `role` modules internal private use and don't maintain its backward compatibility. Is it possible to directly use `GraphClient` retrieved from `azure.cli.command_modules.role.graph_client_factory`?

src/containerapp/azext_containerapp/_utils.py
StrawnSC: Sounds good; just pushed this change

src/containerapp/azext_containerapp/_utils.py
StrawnSC: My bad, I assumed that these functions were safe to use since they're tied to CLI commands (`az ad sp app create` etc.) and appeared to be public. 

I just removed the dependencies on all functions in `azure.cli.command_modules.role.custom`. Our extension is now using the SDK/client for all of these instead

StrawnSC: Just to clarify, this change is also in the latest commit I pushed

src/containerapp/HISTORY.rst
```diff
@@ -3,6 +3,15 @@
...
 ===============
 
+0.3.5
+++++++
+* Add parameter --zone-redundant to 'az containerapp env create'
+* Added 'az containerapp env certificate' to manage certificates in a container app environment
+* Added 'az containerapp hostname' to manage hostnames in a container app
+* Added 'az containerapp ssl upload' to upload a certificate, add a hostname and the binding to a container app
+* Added 'az containerapp auth' to manage AuthConfigs for a containerapp
+* BREAKING CHANGE: require Azure CLI version of at least 2.37.0
```
zhoxing-ms, line 13: In fact, modifying `azext.minCliCoreVersion` is not breaking change, it just makes users who using the old version of CLI unable to install this version of extension

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -415,6 +415,22 @@ def get_auth_token(cls, cmd, resource_group_name, name):
...
+        request_url = url_fmt.format(
+            management_hostname.strip('/'),
+            sub_id,
+            resource_group_name,
+            name,
+            STABLE_API_VERSION,
+            hostname)
+
+        r = send_raw_request(cmd.cli_ctx, "POST", request_url)
+        return r.json()
```
zhoxing-ms, line 432: Why are these requests not considered to be implemented in the Python SDK method, but directly request the REST service

src/containerapp/HISTORY.rst
StrawnSC: @zhoxing-ms just updated the history entry to remove that

src/containerapp/azext_containerapp/_params.py
```diff
@@ -255,3 +273,59 @@ def load_arguments(self, _):
...
         c.argument('service_principal_client_secret', help='The service principal client secret. Used by Github Actions to authenticate with Azure.', options_list=["--service-principal-client-secret", "--sp-sec"])
         c.argument('service_principal_tenant_id', help='The service principal tenant ID. Used by Github Actions to authenticate with Azure.', options_list=["--service-principal-tenant-id", "--sp-tid"])
+
+    with self.argument_context('containerapp auth') as c:
+        # subgroup update
+        c.argument('client_id', options_list=['--client-id'], help='The Client ID of the app used for login.')
+        c.argument('client_secret', options_list=['--client-secret'], help='The client secret.')
+        c.argument('client_secret_setting_name', options_list=['--client-secret-name'], help='The app secret name that contains the client secret of the relying party application.')
+        c.argument('issuer', options_list=['--issuer'], help='The OpenID Connect Issuer URI that represents the entity which issues access tokens for this application.')
```
zhoxing-ms, line 282: The `options_list` defaults to the argument name, so when if the parameter has no other alias, please do not define `options_list` repeatedly
```suggestion
        c.argument('client_id', help='The Client ID of the app used for login.')
        c.argument('client_secret', help='The client secret.')
        c.argument('client_secret_setting_name', help='The app secret name that contains the client secret of the relying party application.')
        c.argument('issuer', help='The OpenID Connect Issuer URI that represents the entity which issues access tokens for this application.')
```

StrawnSC: In the third case, there's a reason since the default option name would be `--client-secret-setting-name` not `--client-secret-name`

But other than that I'll correct those

src/containerapp/azext_containerapp/_params.py
```diff
@@ -255,3 +273,59 @@ def load_arguments(self, _):
...
+
+    with self.argument_context('containerapp hostname bind') as c:
+        c.argument('hostname', help='The custom domain name.')
+        c.argument('thumbprint', options_list=['--thumbprint', '-t'], help='Thumbprint of the certificate.')
+        c.argument('certificate', options_list=['--certificate', '-c'], help='Name or resource id of the certificate.')
+        c.argument('environment', options_list=['--environment', '-e'], help='Name or resource id of the Container App environment.')
+
+    with self.argument_context('containerapp hostname list') as c:
+        c.argument('name', id_part=None)
+        c.argument('location', configured_default='location')
```
zhoxing-ms, line 328: Do we need to consider adding `arg_type=get_location_type(self.cli_ctx)` for `location` argument?

zhoxing-ms: Could we consider using `validator=get_default_location_from_resource_group` to make the location of resource group as the default location?

src/containerapp/azext_containerapp/commands.py
```diff
@@ -126,3 +131,45 @@ def load_command_table(self, _):
...
+
+    with self.command_group('containerapp auth apple') as g:
+        g.custom_show_command('show', 'get_apple_settings')
+        g.custom_command('update', 'update_apple_settings', exception_handler=ex_handler_factory())
+
+    with self.command_group('containerapp auth openid-connect') as g:
+        g.custom_show_command('show', 'get_openid_connect_provider_settings')
+        g.custom_command('add', 'add_openid_connect_provider_settings', exception_handler=ex_handler_factory())
+        g.custom_command('update', 'update_openid_connect_provider_settings', exception_handler=ex_handler_factory())
+        g.custom_command('remove', 'remove_openid_connect_provider_settings')
```
zhoxing-ms, line 167: Do we need to consider adding `confirmation=True` for the `remove` command?

src/containerapp/azext_containerapp/tests/latest/test_containerapp_commands.py
```diff
@@ -9,13 +9,14 @@
...
 from azure.cli.testsdk import (ScenarioTest, ResourceGroupPreparer, JMESPathCheck)
-
+from msrestazure.tools import parse_resource_id
 
 TEST_DIR = os.path.abspath(os.path.join(os.path.abspath(__file__), '..'))
 
 class ContainerappIdentityTests(ScenarioTest):
     @AllowLargeResponse(8192)
     @ResourceGroupPreparer(location="eastus2")
+    @live_only()  # encounters 'CannotOverwriteExistingCassetteException' only when run from recording (passes when run live)
```
zhoxing-ms, line 19: > 'CannotOverwriteExistingCassetteException'
Will it fail to execute this test locally in record mode? Or it failed only in CI pipeline?
Could you please show us the specific error message?

StrawnSC: Early this test was passing when run live locally but failing when run from the recording locally. I just tried it again and it seems to work.

I've added the recording back and removed `@live_only()`

StrawnSC: Actually it looks like these are failing the CI again with `CannotOverwriteExistingCassetteException` (despite passing locally). Can I go ahead and mark these as `@live_only` again @zhoxing-ms?

src/containerapp/azext_containerapp/_clients.py
StrawnSC: @zhoxing-ms our plan with the `az containerapp` commands was to use raw requests while in the extension, and then use the SDK when we put this in a core CLI command module -- similar to the `appservice-kube` commands. This has helped us keep the extension install size down and add features independently of their support in the SDK. We were under the impression that using raw requests is not allowed in the core CLI but permissible in CLI extensions

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -12,147 +12,126 @@
...
-            cert_file)
-        result['fileWithCertAndPrivateKey'] = cert_file
+    try:
+        result = client.application_create(body)
+    except GraphError as ex:
+        if 'insufficient privileges' in str(ex).lower():
+            link = 'https://docs.microsoft.com/azure/azure-resource-manager/resource-group-create-service-principal-portal'  # pylint: disable=line-too-long
+            raise ValidationError("Directory permission is needed for the current user to register the application. "
+                                  "For how to configure, please refer '{}'. Original error: {}".format(link, ex))
+        raise
```
jiasli, line 74: The implementation in main Azure CLI is too generic for your use case. I am glad to see your implementation is much simpler.

jiasli (approved):

src/containerapp/azext_containerapp/_utils.py
panchagnula: @StrawnSC looks like we have approval but build pipeline is failing  once we have this working - we can get this build deployed & released. Thank you for all the efforts here.

panchagnula (approved):
fix the failing tests

src/containerapp/azext_containerapp/tests/latest/test_containerapp_commands.py
StrawnSC: per @panchagnula's comment I'm going to go ahead and mark this test as `@live_only`

panchagnula: @StrawnSC yes mark them live for now & lets figure out these flaky tests with upcoming changes.

src/containerapp/azext_containerapp/_clients.py
panchagnula: @zhoxing-ms , @StrawnSC is right - the swagger & Python SDK changes were too delayed for us to take a dependency on this. POST GA we plan to move this to using SDK completely + maybe move to core-cli as well

StrawnSC:
> fix the failing tests

@panchagnula the CI tests are passing now

src/containerapp/azext_containerapp/_clients.py
zhoxing-ms: Got it, thanks~

src/containerapp/azext_containerapp/_params.py
```diff
@@ -255,3 +274,59 @@ def load_arguments(self, _):
...
+        c.argument('proxy_convention', options_list=['--proxy-convention'], arg_type=get_enum_type(FORWARD_PROXY_CONVENTION), help='The convention used to determine the url of the request made.')
+        c.argument('proxy_custom_host_header', options_list=['--proxy-custom-host-header', '--custom-host-header'], help='The name of the header containing the host of the request.')
+        c.argument('proxy_custom_proto_header', options_list=['--proxy-custom-proto-header', '--custom-proto-header'], help='The name of the header containing the scheme of the request.')
+        c.argument('excluded_paths', options_list=['--excluded-paths'], help='The list of paths that should be excluded from authentication rules.')
+        c.argument('enabled', options_list=['--enabled'], arg_type=get_three_state_flag(), help='true if the Authentication / Authorization feature is enabled for the current app; otherwise, false.')
+        c.argument('runtime_version', options_list=['--runtime-version'], help='The RuntimeVersion of the Authentication / Authorization feature in use for the current app.')
```
zhoxing-ms, line 312: Could you remove all the duplicate `options_list`?

StrawnSC: Just pushed the fix for this

jiasli (approved):

zhoxing-ms (approved):

PR #4956, opened 2022-06-08
[Spring] Fix the incorrect split of byoc commands and args.

LGDoor:
---
Fixed a regression bug introduced in the previous refactoring that the commands and arguments aren't split correctly.

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/spring-cloud/azext_spring_cloud/_params.py
+++ b/src/spring-cloud/azext_spring_cloud/_params.py
@@ -364,9 +364,9 @@ def load_arguments(self, _):
             c.argument(
                 'registry_password', help='The password of the container registry.', arg_group='Custom Container')
             c.argument(
-                'container_command', help='The command of the container image.', nargs='*', arg_group='Custom Container')
+                'container_command', help='The command of the container image.', arg_group='Custom Container')
             c.argument(
-                'container_args', help='The arguments of the container image.', nargs='*', arg_group='Custom Container')
+                'container_args', help='The arguments of the container image.', arg_group='Custom Container')
             c.argument(
                 'build_env', build_env_type)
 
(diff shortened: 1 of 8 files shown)


yonzhan:
spring cloud

src/spring-cloud/azext_spring_cloud/_params.py
```diff
@@ -364,9 +364,9 @@ def prepare_logs_argument(c):
             c.argument(
                 'registry_password', help='The password of the container registry.', arg_group='Custom Container')
             c.argument(
-                'container_command', help='The command of the container image.', nargs='*', arg_group='Custom Container')
+                'container_command', help='The command of the container image.', arg_group='Custom Container')
             c.argument(
-                'container_args', help='The arguments of the container image.', nargs='*', arg_group='Custom Container')
+                'container_args', help='The arguments of the container image.', arg_group='Custom Container')
```
zhoxing-ms, line 369: I would like to ask whether the change will cause breaking change to users?

zhoxing-ms:
Could you add some tests for those change?

zhoxing-ms:
Please add description of those change to the `HISTORY.rst` file. By the way, do you need to release a new version of extension for this PR? If so, please also modify the `setup.py` file

src/spring-cloud/azext_spring_cloud/_params.py
LGDoor: No. Actually the correct type is `string`. It's changed to the array accidentally in a previous refactoring.

zhoxing-ms: OK

zhoxing-ms (approved):

PR #5010, opened 2022-06-17
Fix kollect bug with inconsistent container name for private clusters

peterbom:
This fixes a bug in which the storage container used to retrieve Periscope logs is not the same as the one they are uploaded to.

The change is to put all the logic for determining the container name into one function, rather than spreading it throughout the kollect function.

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/HISTORY.rst
+++ b/src/aks-preview/HISTORY.rst
@@ -14,6 +14,7 @@ Pending
 
 * Update the minimum required cli core version to `2.37.0`.
 * Enable v2 decorator pattern.
+* Fix container name inconsistency for private clusters in kollect command.
 
 0.5.82
 +++++++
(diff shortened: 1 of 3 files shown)


yonzhan:
AKS

FumingZhang (approved):
LGTM

zhoxing-ms:
Please add a description to the history notes `HISTORY.rst`. In addition, if you need to release a new version of extension for this change, please also modify `setup.py`

peterbom:
> Please add a description to the history notes `HISTORY.rst`. In addition, if you need to release a new version of extension for this change, please also modify `setup.py`

Thanks, @zhoxing-ms! I've updated the history file. This bug has been around forever and there's no hurry for a release, so I'm leaving `setup.py` as it is, unless instructed otherwise.

zhoxing-ms (approved):

PR #5026, opened 2022-06-21
Containerapp 0.3.8 Release

StrawnSC:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/scripts/ci/credscan/CredScanSuppressions.json
+++ b/scripts/ci/credscan/CredScanSuppressions.json
@@ -144,7 +144,8 @@
         "src\\containerapp\\azext_containerapp\\tests\\latest\\cert.pfx",
         "src\\containerapp\\azext_containerapp\\tests\\latest\\test_containerapp_commands.py",
         "src\\containerapp\\azext_containerapp\\tests\\latest\\test_containerapp_env_commands.py",
-        "src\\containerapp\\azext_containerapp\\tests\\latest\\recordings\\test_containerapp_registry_msi.yaml"
+        "src\\containerapp\\azext_containerapp\\tests\\latest\\recordings\\test_containerapp_registry_msi.yaml",
+        "src\\containerapp\\azext_containerapp\\tests\\latest\\recordings\\test_containerapp_update_containers.yaml"
       ],
       "_justification": "Dummy resources' keys left during testing Microsoft.App (required for log-analytics to create managedEnvironments)"
     },
--- a/src/containerapp/HISTORY.rst
+++ b/src/containerapp/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+0.3.8
+++++++
+* TODO
+
 0.3.7
 ++++++
 * Fixed bug with 'az containerapp up' where --registry-server was ignored
--- a/src/containerapp/HISTORY.rst
+++ b/src/containerapp/HISTORY.rst
@@ -5,8 +5,10 @@ Release History
 
 0.3.8
 ++++++
-* Fixed bug with 'az containerapp update' where --yaml would error out due to secret values
-* Fixed bug with 'az containerapp logs show' and 'az containerapp exec' where they would fail on apps with networking restrictions 
+* 'az containerapp update': Fix bug where --yaml would error out due to secret values
+* 'az containerapp update': use PATCH API instead of GET and PUT
+* 'az containerapp up': Fix bug where using --source with an invalid name parameter causes ACR build to fail
+* 'az containerapp logs show'/'az containerapp exec': Fix bug where ssh/logstream they would fail on apps with networking restrictions
 
 0.3.7
 ++++++
--- a/src/containerapp/HISTORY.rst
+++ b/src/containerapp/HISTORY.rst
@@ -6,6 +6,7 @@ Release History
 0.3.8
 ++++++
 * Fixed bug with 'az containerapp update' where --yaml would error out due to secret values
+* Fixed bug with 'az containerapp logs show' and 'az containerapp exec' where they would fail on apps with networking restrictions 
 
 0.3.7
 ++++++
--- a/src/containerapp/HISTORY.rst
+++ b/src/containerapp/HISTORY.rst
@@ -5,7 +5,7 @@ Release History
 
 0.3.8
 ++++++
-* TODO
+* Fixed bug with 'az containerapp update' where --yaml would error out due to secret values
 
 0.3.7
 ++++++
--- a/src/containerapp/azext_containerapp/_up_utils.py
+++ b/src/containerapp/azext_containerapp/_up_utils.py
@@ -679,7 +679,7 @@ def _validate_containerapp_name(name):
     is_valid = is_valid and name.isalnum()
     is_valid = is_valid and name[0].isalpha()
     if not is_valid:
-        raise ValidationError(f"Invalid Container App name {name}. A name must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character and cannot have '--'. The length must not be more than {MAXIMUM_CONTAINER_APP_NAME_LENGTH} characters.")
+        raise ValidationError(f"Invalid Container App name {name}. A name must consist of lower case alphanumeric characters or '-', start with a letter, end with an alphanumeric character, cannot have '--', and must be less than {MAXIMUM_CONTAINER_APP_NAME_LENGTH} characters.")
 
 
 # attempt to populate defaults for managed env, RG, ACR, etc
--- a/src/containerapp/azext_containerapp/_up_utils.py
+++ b/src/containerapp/azext_containerapp/_up_utils.py
@@ -670,6 +670,18 @@ def _get_registry_details(cmd, app: "ContainerApp", source):
     )
 
 
+def _validate_containerapp_name(name):
+    is_valid = True
+    is_valid = is_valid and name.lower() == name
+    is_valid = is_valid and len(name) <= 32
+    is_valid = is_valid and '--' not in name
+    name = name.replace('-', '')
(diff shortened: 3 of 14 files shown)


panchagnula:
@zhoxing-ms when you are online today could you review this please, we would like to get this payload released soon. Thanks!

scripts/ci/credscan/CredScanSuppressions.json
```diff
@@ -144,7 +144,8 @@
         "src\\containerapp\\azext_containerapp\\tests\\latest\\cert.pfx",
         "src\\containerapp\\azext_containerapp\\tests\\latest\\test_containerapp_commands.py",
         "src\\containerapp\\azext_containerapp\\tests\\latest\\test_containerapp_env_commands.py",
-        "src\\containerapp\\azext_containerapp\\tests\\latest\\recordings\\test_containerapp_registry_msi.yaml"
```
panchagnula, line 147: these don't seem to be new tests, but maybe re-recorded? Is there a reason we need to add these to CredSupressions with the new changes? as in have there been changes in API that is adding some secrets recently?

src/containerapp/HISTORY.rst
```diff
@@ -3,6 +3,11 @@
 Release History
 ===============
 
+0.3.8
+++++++
+* Fixed bug with 'az containerapp update' where --yaml would error out due to secret values
```
panchagnula, line 8: @haroonf this doesn't seem to have the full of list issues you fixed called out in history. Can you update please

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -670,6 +670,18 @@ def _get_registry_details(cmd, app: "ContainerApp", source):
     )
 
 
+def _validate_containerapp_name(name):
+    is_valid = True
+    is_valid = is_valid and name.lower() == name
+    is_valid = is_valid and len(name) <= 32
+    is_valid = is_valid and '--' not in name
```
panchagnula, line 677: Hmm instead of replacing these unsupported characters with spaces? Why do we not throw an error to the user to not these characters?

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -670,6 +670,18 @@ def _get_registry_details(cmd, app: "ContainerApp", source):
...
+    name = name.replace('-', '')
+    is_valid = is_valid and name.isalnum()
+    is_valid = is_valid and name[0].isalpha()
+    if not is_valid:
+        raise ValidationError(f"Invalid Container App name {name}. A name must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character and cannot have '--'. The length must not be more than 32 characters.")
```
panchagnula, line 682: The name should have alpha numeric characters? Can you run this message via a PM ? seems long. I would add this requirements in the parameters.txt / help.txt as well, so users are aware of the requirements when checking the  help docs.

panchagnula:
Unrelated to changes on this PR , however, want to understand this code.
what is the purpose of this? Maintaining this means we keep track of any new read_only_attributes added & update this list?
def _remove_readonly_attributes(containerapp_def):
    unneeded_properties = [
        "id",
        "name",
        "type",
        "systemData",
        "provisioningState",
        "latestRevisionName",
        "latestRevisionFqdn",
        "customDomainVerificationId",
        "outboundIpAddresses",
        "fqdn"
    ]

panchagnula (changes requested):
a few Qs/ comments in the chat. @haroonf / @StrawnSC  please check. Thanks!

panchagnula:
@calvinsID please review the changes as well, esp. the update using PATCH. Thanks!

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -788,6 +788,35 @@ def _remove_readonly_attributes(containerapp_def):
             del containerapp_def['properties'][unneeded_property]
 
 
+def clean_null_values(d):
```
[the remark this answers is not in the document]
StrawnSC: I believe we're doing this because null values in the PATCH payload make the API delete those properties. The API expects values to not be in the payload at all if we don't want to update them. The SDK takes care of removing null values for us, but for now we have to do that ourselves. I'll add a comment for this 

@haroonf correct me that's incorrect

src/containerapp/azext_containerapp/_up_utils.py
StrawnSC: Are you referring to the line with `name = name.replace('-', '')`? 
Since the dash ("-") is an allowed character, and the only other allowed characters are alphanumeric ones, we eliminate the dashes from the name string so we can check if the rest of the name is alphanumeric by calling `name.isalnum()`. Note that we're just changing a local variable here so we're not actually changing what the containerapp name will be 

@haroonf correct me if I'm wrong on that

scripts/ci/credscan/CredScanSuppressions.json
StrawnSC: We had two tests with the same name so we had to rename one and rerecord it. All the containerapp tests need a credscan suppression because of the log analytics keys that are used during the creation of containerapp environments

StrawnSC:
> Unrelated to changes on this PR , however, want to understand this code.
> what is the purpose of this? Maintaining this means we keep track of any new read_only_attributes added & update this list?
> def _remove_readonly_attributes(containerapp_def):
> unneeded_properties = [
> "id",
> "name",
> "type",
> "systemData",
> "provisioningState",
> "latestRevisionName",
> "latestRevisionFqdn",
> "customDomainVerificationId",
> "outboundIpAddresses",
> "fqdn"
> ]

@panchagnula looks like this is needed since we're not using the SDK. The SDK automatically ignores readonly attributes that are set on a model before serializing it. @calvinsID is that right? (I believe you authored that part if I'm not mistaken since it's in the yaml command)

calvinsID (approved):

src/containerapp/azext_containerapp/_up_utils.py
panchagnula: I see - I am wondering why don't we just error out if there is a - & throw an error with the message on what the expected format is?

panchagnula: Can we update help text for this parameter to display what the supported format it? Don't see that change.

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -673,13 +673,13 @@ def _get_registry_details(cmd, app: "ContainerApp", source):
...
     is_valid = is_valid and name.lower() == name
-    is_valid = is_valid and len(name) <= 32
+    is_valid = is_valid and len(name) <= MAXIMUM_CONTAINER_APP_NAME_LENGTH
     is_valid = is_valid and '--' not in name
     name = name.replace('-', '')
     is_valid = is_valid and name.isalnum()
     is_valid = is_valid and name[0].isalpha()
     if not is_valid:
-        raise ValidationError(f"Invalid Container App name {name}. A name must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character and cannot have '--'. The length must not be more than 32 characters.")
+        raise ValidationError(f"Invalid Container App name {name}. A name must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character and cannot have '--'. The length must not be more than {MAXIMUM_CONTAINER_APP_NAME_LENGTH} characters.")
```
panchagnula, line 683: seems like the message is copied from portal ;) but "alphabetic character" seems incorrect English to me.

panchagnula:
@StrawnSC - some of my comments are still not completely addressed - ex. adding comments to the helper library, explaining the need for those . & update help for container app name parameter, so that the format requirements are displayed in help docs?

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -788,6 +788,35 @@ def _remove_readonly_attributes(containerapp_def):
...
+            k: v
+            for k, v in ((k, clean_null_values(v)) for k, v in d.items())
+            if v
+        }
+    if isinstance(d, list):
+        return [v for v in map(clean_null_values, d) if v]
+    return d
+
+
+def _populate_secret_values(containerapp_def, secret_values):
```
panchagnula, line 803: @haroonf , don't need to change this logic for now, since we need to release the fix.
Can you help me understand the behavior in CLI with YAML here , when CLI parses the YAML with secretName & no value, does the payload to the API translate this object in JSON payload  just name property & no value i.e{ name: xyz} or. {name: xyz, value: null}. if it is the former is the PATCH API, setting the value to null or is it cleaning these up because CLI explicitly passes null?

StrawnSC:
> @StrawnSC - some of my comments are still not completely addressed - ex. adding comments to the helper library, explaining the need for those . & update help for container app name parameter, so that the format requirements are displayed in help docs?

@panchagnula Sorry about that - I must have misread your comment on the help text. Just pushed those changes. 

@haroonf can you take a look at Sisira's comment above

panchagnula (approved):
LGTM - please do update the comments if possible in this one. & answer the pending Qs

src/containerapp/azext_containerapp/_utils.py
haroonf: The deserialization process leaves the json payload to {name: xyz, value: null}. The API returns an error "cannot set secret values to null."

src/containerapp/azext_containerapp/_up_utils.py
haroonf: This message is copied from the API

StrawnSC:
@zhoxing-ms could you take a look at this when you get the chance? We've had some customers asking about these fixes for a while

zhoxing-ms (approved):

PR #5045, opened 2022-06-27
Update _params.py

bjqian:
remove hard restriction on sku. Leave this to server side for future Premium_P1, Standard_S2

---

### Related command

az webpubsub

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/webpubsub/azext_webpubsub/_params.py
+++ b/src/webpubsub/azext_webpubsub/_params.py
@@ -32,7 +32,7 @@ def load_arguments(self, _):
         c.argument('webpubsub_name', webpubsub_name_type, options_list=['--name', '-n'])
 
     with self.argument_context('webpubsub create') as c:
-        c.argument('sku', help='The sku name of the webpubsub service. E.g. Standard_S1')
+        c.argument('sku', help='The sku name of the webpubsub service. Allowed values: Free_F1, Standard_S1')
         c.argument('unit_count', help='The number of webpubsub service unit count', type=int)
 
     with self.argument_context('webpubsub update') as c:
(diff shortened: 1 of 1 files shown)


yonzhan:
webpubsub

src/webpubsub/azext_webpubsub/_params.py
```diff
@@ -33,12 +32,12 @@ def load_arguments(self, _):
         c.argument('webpubsub_name', webpubsub_name_type, options_list=['--name', '-n'])
 
     with self.argument_context('webpubsub create') as c:
-        c.argument('sku', arg_type=get_enum_type(SKU_TYPE), help='The sku name of the signalr service.')
-        c.argument('unit_count', help='The number of signalr service unit count', type=int)
+        c.argument('sku', help='The sku name of the webpubsub service. E.g. Standard_S1')
```
kairu-ms, line 35: why diable the choice options of sku argument?

bjqian: We have SKUs that are internal only. Curretly we don't hope customers outside Microsoft to see those SKUs as they are not allowed to create them. But we hope az cli is available for our internal customers. So we need to remove this restriction from client side and leave it to server side.

kairu-ms: The sku parameter is required for `az webpubsub create`. Did you mean the customers outside Microsoft are not supposed to run the create command?

zackliu: I think we can put all `sku` available for public in `help`. But as it's a server-side parameter, I think it doesn't make sense to make restrictions in cli side. We can pass `sku` parameter to RP side for checking.

kairu-ms: > I think we can put all `sku` available for public in `help`. But as it's a server-side parameter, I think it doesn't make sense to make restrictions in cli side. We can pass `sku` parameter to RP side for checking.

I totally agree

bjqian: Fixed

kairu-ms (approved):

kairu-ms:
@bjqian If you want to release a new version, please create a new PR with extension version updated.

PR #5052, opened 2022-06-27
[Fix] Log warning rather than throw exception when we by default create app insights

jiec-msft:
### Background
When customer create Azure Spring Apps service instance with CLI, if customer didn't specify anything for application insights, we'll by default create an Application Insights, and associate it to the service instance. But create application insights can fail, we don't want to block customer's flow to create service instance, so log warning rather than throw exception and fail the command, just as was done in Basic and Standard tiers.

#### Warning when create application insights failed
image

### ADO workitem
[Bug 13041269](https://msazure.visualstudio.com/AzureDMSS/_workitems/edit/13041269): [CLI][All tiers][Pilot]:Create service instance successfully，but application insights create failed

---

### Related command

```
az spring create xxxxxxx
```

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/spring/azext_spring/buildpack_binding.py
+++ b/src/spring/azext_spring/buildpack_binding.py
@@ -83,22 +83,24 @@ def _build_buildpack_binding_resource(binding_type, properties_dict, secrets_dic
 
 def _get_buildpack_binding_properties(cmd, resource_group, service_name, location,
                                       app_insights_key, app_insights, sampling_rate):
-
-    sampling_rate = sampling_rate or 10
-    connection_string = app_insights_key or \
-        _get_connection_string_from_app_insights(cmd, resource_group, app_insights) or \
-        _create_app_insights_and_get_connection_string(cmd, resource_group, service_name, location)
+    connection_string = _get_connection_string(cmd, resource_group, service_name, location, app_insights_key, app_insights)
 
     if not connection_string:
-        raise InvalidArgumentValueError('Error while trying to get the ConnectionString of Application Insights for the Azure Spring Apps. '
-                                        'Please use the Azure Portal to create and configure the Application Insights, if needed.')
+        return None
 
+    sampling_rate = sampling_rate or 10
     launch_properties = models.BuildpackBindingLaunchProperties(properties={
         "connection-string": connection_string,
         "sampling-percentage": sampling_rate,
     })
 
     return models.BuildpackBindingProperties(binding_type="ApplicationInsights", launch_properties=launch_properties)
+
+
+def _get_connection_string(cmd, resource_group, service_name, location, app_insights_key, app_insights):
+    return app_insights_key or \
+        _get_connection_string_from_app_insights(cmd, resource_group, app_insights) or \
+        _create_app_insights_and_get_connection_string(cmd, resource_group, service_name, location)
 
 
 def _create_app_insights_and_get_connection_string(cmd, resource_group, service_name, location):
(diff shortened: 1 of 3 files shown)


src/spring/azext_spring/buildpack_binding.py
```diff
@@ -64,13 +65,27 @@ def buildpack_binding_delete(cmd, client, resource_group, service, name, builder
 
 def create_default_buildpack_binding_for_application_insights(cmd, client, resource_group, name, location,
                                                               app_insights_key, app_insights, sampling_rate):
+    try:
```
yuwzho, line 68: I don't get why should catch the exception at the very outter place and decide what to do for an internal exception. Can you make the code more comprehensive?

jiec-msft: At inner code, `_create_app_insights_and_get_connection_string` its purpose is to purely create application insights, it should throw exception when it failed to do it. And it depends on the upper business to decide how to handle such exception.

So here, `create_default_buildpack_binding_for_application_insights`, when it sees such exception, it happens when customer didn't specify anything for application insights, and we create one for cx. The business requirement guides us that we should skip this. So I suppose this method is the place to know how to handle this exception.

src/spring/azext_spring/buildpack_binding.py
```diff
@@ -107,10 +121,8 @@ def _create_app_insights_and_get_connection_string(cmd, resource_group, service_
         created_app_insights = _try_create_application_insights(cmd, resource_group, service_name, location)
         if created_app_insights:
             return created_app_insights.connection_string
-    except Exception:  # pylint: disable=broad-except
-        logger.warning(
-            'Error while trying to create and configure an Application Insights for the Azure Spring Apps. '
-            'Please use the Azure Portal to create and configure the Application Insights, if needed.')
+    except Exception as ex:  # pylint: disable=broad-except
```
yuwzho, line 124: How about making the logic like this? Because the creation AI is just a fallback solution for fallback solution to get the connection string. It is wired the properties contruct method to handle the creation failure error.

```
def get_properties(sampling_rate, **kwargs):
   conn = get_connection_string(**kwargs)
   if not conn:
        logger.warning('No connection string blablabla')
        return None
   return models.BuildpackBindingProperties(.....)

def get_connection_string(app_insights_key, **kwargs):
    return app_insights_key or get_ai_connection_string(**kwargs) or create_ai_and_get_connection_string(**kwargs)

def create_ai_and_get_connection_string(**kwargs):
    try:
         return create_ai(**kwargs).connection_string
   except Exception as ex:
        logger.warning('failed to create ai blablabla')
        return None
```

src/spring/azext_spring/buildpack_binding.py
```diff
@@ -130,8 +132,10 @@ def _get_connection_string_from_app_insights(cmd, resource_group, app_insights):
     else:
         connection_string = _get_app_insights_connection_string(cmd.cli_ctx, resource_group, app_insights)
 
+    # Customer has specify the resourceId or application insights name.
+    # Raise exception when connection string not found in this scenario.
     if not connection_string:
-        logger.warning(
+        raise InvalidArgumentValueError(
```
jiec-msft, line 138: We could further enhance this case, move it to validator, so we can fail fast, rather than fail after service instance is created. We can enhance this in another PR.

src/spring/azext_spring/buildpack_binding.py
jiec-msft: Yeah, it makes it clearer, updated.

yuwzho (approved):

src/spring/azext_spring/buildpack_binding.py
zhoxing-ms: Will this cause breaking change after raising the `InvalidArgumentValueError`?

zhoxing-ms (approved):

PR #5130, opened 2022-07-19
[Quantum] Show required parameters in CLI Reference documentation

warren-jones:
Currently, almost all quantum CLI command parameters in the [`az quantum` CLI reference documentation](https://docs.microsoft.com/en-us/cli/azure/quantum?view=azure-cli-latest) appear under “Optional Parameters”, which gives the impression that none of these parameters are required parameters.  This has been criticized as confusing and misleading. 

The [`az quantum` CLI Reference documentation](https://docs.microsoft.com/en-us/cli/azure/quantum?view=azure-cli-latest) is automatically generated from source files like _params.py, _help.py, and the command function  signatures, so it is not directly editable like our tutorials and how-to guides.  The source code dictates which parameters are documented as required, and which parameters are optional.

A positive side effect of making these code changes is a dramatic improvement in the detail level of error messages about missing parameters, since we're leveraging the framework's error message generation feature that adds help examples.

The required vs. optional treatment of `--location` and `--resource-group` is not consistent throughout the Azure CLI, but for example, Azure Virtual Machines generally treats `--resource-group` as a required parameter, see https://docs.microsoft.com/en-us/cli/azure/vm/aem?view=azure-cli-latest

I believe the least user confusion will arise from designating them Required Parameters to leverage the CLI framework’s validation and detailed error message generation.  For `--location`, the help text always includes `You can configure the default location using az configure --defaults location=<location>` and a similar note appears for `--resource-group`.

---

### Related command
All `az quantum` commands

### General Guidelines

### DIFF
--- a/src/quantum/azext_quantum/_help.py
+++ b/src/quantum/azext_quantum/_help.py
@@ -86,15 +86,15 @@ helps['quantum job submit'] = """
       - name: Submit the Q# program from the current folder.
         text: |-
             az quantum job submit -g MyResourceGroup -w MyWorkspace -l MyLocation \\
-               --job-name MyJob
+               -t MyTarget --job-name MyJob 
       - name: Submit the Q# program from the current folder with job parameters for a target.
         text: |-
             az quantum job submit -g MyResourceGroup -w MyWorkspace -l MyLocation \\
-               --job-name MyJob --job-params param1=value1 param2=value2
+               -t MyTarget --job-name MyJob --job-params param1=value1 param2=value2
       - name: Submit the Q# program with program parameters (e.g. n-qubits = 2).
         text: |-
             az quantum job submit -g MyResourceGroup -w MyWorkspace -l MyLocation \\
-               --job-name MyJob -- --n-qubits=2
+               -t MyTarget --job-name MyJob -- --n-qubits=2
       - name: Submit a Q# program from the current folder with a target-capability parameter.
         text: |-
             az quantum job submit -g MyResourceGroup -w MyWorkspace -l MyLocation -t MyTarget \\
@@ -236,7 +236,7 @@ helps['quantum workspace delete'] = """
             az quantum workspace delete -g MyResourceGroup -w MyWorkspace
       - name: Delete and clear the default Azure Quantum workspace (if one has been set).
         text: |-
-            az quantum workspace delete
+            az quantum workspace delete -g MyResourceGroup -w MyWorkspace
 """
 
 helps['quantum workspace list'] = """
--- a/src/quantum/azext_quantum/_help.py
+++ b/src/quantum/azext_quantum/_help.py
@@ -253,7 +253,7 @@ helps['quantum workspace quotas'] = """
     type: command
     short-summary: List the quotas for the given (or current) Azure Quantum workspace.
     examples:
-      - name: List the quota information of a specified Azure Quantum workspace.
+      - name: List the quota information of a specified Azure Quantum workspace. If a default workspace has been set, the -g, -w, and -l parameters are not required.
         text: |-
             az quantum workspace quotas -g MyResourceGroup -w MyWorkspace -l MyLocation
 """
--- a/src/quantum/azext_quantum/_help.py
+++ b/src/quantum/azext_quantum/_help.py
@@ -190,7 +190,7 @@ helps['quantum target set'] = """
 
 helps['quantum target show'] = """
     type: command
-    short-summary: Get the details of the given (or current) target to use when submitting jobs to Azure Quantum.
+    short-summary: Get the Target ID of the current default target to use when submitting jobs to Azure Quantum.
     examples:
       - name: Show the currently selected default target
         text: |-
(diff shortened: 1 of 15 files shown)


yonzhan:
Quantum

shenyingjun (approved):

src/quantum/setup.py
```diff
@@ -17,7 +17,7 @@
 # This version should match the latest entry in HISTORY.rst
 # Also, when updating this, please review the version used by the extension to
 # submit requests, which can be found at './azext_quantum/__init__.py'
-VERSION = '0.16.0'
```
ricardo-espinoza, line 20: Is this version just used during development? I don't know if this will trigger a release. What is the intention here? I assume that the real release will happen when the version goes to 0.17.0, but I'm not sure about the consequences of merging as is. Have you discussed it with the CLI team?

src/quantum/azext_quantum/tests/latest/utils.py
```diff
@@ -53,3 +53,22 @@ def all_providers_are_in_capabilities(provider_sku_string, capabilities_string):
         if provider not in capabilities_string:
             return False
     return True
+
+# import pytest
```
ricardo-espinoza, line 57: If you merge with this, please open an item to track fixing this properly.

src/quantum/azext_quantum/tests/latest/utils.py
```diff
@@ -53,3 +53,22 @@ def all_providers_are_in_capabilities(provider_sku_string, capabilities_string):
...
+# # See "TODO" in except block below
+
+# TEST_ERROR_MESSAGE_PREAMBLE = "the following arguments are required: "
+
+def issue_cmd_with_param_missing(calling_object, command, help_example):
+    try:
+        calling_object.cmd(command)
+        assert False    # Fail the test if we DON'T get an exception
+    except:
+        # TODO: Figure out why this works locally, but not in the Azure CLI CI/CD checks pipeline.  Is there an alternative way to capture the error message?
```
ricardo-espinoza, line 69: Are you planning to fix this TODO before or after the next CLI release?

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -75,10 +71,11 @@ def clear(self):
     def save(self, cmd):
         from azure.cli.core.util import ConfiguredDefaultSetter
 
+        # Save in the global [defaults] section of the .azure\config file
         with ConfiguredDefaultSetter(cmd.cli_ctx.config, False):
-            cmd.cli_ctx.config.set_value('quantum', 'group', self.resource_group)
-            cmd.cli_ctx.config.set_value('quantum', 'workspace', self.name)
-            cmd.cli_ctx.config.set_value('quantum', 'location', self.location)
+            cmd.cli_ctx.config.set_value(cmd.cli_ctx.config.defaults_section_name, 'group', self.resource_group)
```
ricardo-espinoza, line 76: Thanks for fixing this!

src/quantum/azext_quantum/_help.py
```diff
@@ -236,7 +236,7 @@
             az quantum workspace delete -g MyResourceGroup -w MyWorkspace
       - name: Delete and clear the default Azure Quantum workspace (if one has been set).
```
ricardo-espinoza, line 237: The example text and description don't match. The explanation is for usage without a default workspace.

ricardo-espinoza:
Overall looks good to me, but we need to verify that the expectations from users when setting a default workspace (and its resource group) have not changed. Many of the commands were written with the idea that users would not need to specify -w/-g/-l if they already did an `azure quantum workspace set`, just in the same way that a target could also be set.

If usage has changed as a result of this PR, we need to discuss it in more detail.

src/quantum/setup.py
warren-jones: I'll restore the 0.16.0 version number before I click "Ready for review".

src/quantum/azext_quantum/_help.py
warren-jones: I corrected the help example.

src/quantum/azext_quantum/tests/latest/utils.py
warren-jones: A work item has been created.  The unit tests could be more robust if this pytest fixture could be used.

src/quantum/azext_quantum/tests/latest/utils.py
warren-jones: No, but a work item has been created.  The unit tests could be more robust if this pytest fixture could be used, so I'm leaving the TODO in the code.

src/quantum/azext_quantum/operations/workspace.py
warren-jones: You're welcome

warren-jones:
@ricardo-espinoza,
I responded to your "Overall looks good to me, but we need to verify that the expectations from users..." comment in email, but I'll reiterate it here to keep the conversation together:

> `az quantum workspace set` has the same effect as before -- so I think we’re OK.  The difference is internal:  Workspace name is saved in the [defaults] section of the .azure/config file instead or [quantum], along with group and location.  Do you see a problem there?

ricardo-espinoza (approved):
Approved, but I hope that by the next release (after this one) the commented code is fixed.

shenyingjun (approved):

kairu-ms (approved):

PR #5162, opened 2022-07-28
{rdbms-connect} Add support to read UTF-8 files with and without BOM

alanenriqueo:
If a sql file with UTF-8 BOM encoding is passed to the command `az postgres flexible-server execute` it will fail because its contents are read assuming that the encoding is `utf-8`. We need to change to `utf-8-sig` to be able to handle both cases (UTF-8 with/without BOM).

---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
rdbms-connect

evelyn-ys (approved):

evelyn-ys:
If you want to release a new version, please update `setup.py` and `history.rst`

PR #5183, opened 2022-08-03
{CI} Fix Azure CLI Extensions Sync to MicrosoftDocs error

wangzelin007:
The pipeline (Azure CLI Extensions Sync to MicrosoftDocs）didn't work since July 5th.
image
Here is the error message:
image
~~Since MarkupSafe remove soft_unicode in version 2.1.0~~
~~So we use MarkupSafe == 2.0.1 to fix this error.~~
And we use jinja2 == 3.0.3 in azure-cli which already drops the soft_unicode.
https://github.com/Azure/azure-cli/blob/75f7048672cf8198bb8f073262caa9e706b68699/scripts/ci/test_ref_doc.sh#L17
So we also upgrade jinja2 to 3.0.3 in azure-cli-extensions.
Test result:
image

### DIFF
--- a/scripts/ci/avail-ext-doc/requirements.txt
+++ b/scripts/ci/avail-ext-doc/requirements.txt
@@ -1,3 +1,4 @@
 Jinja2~=2.10.1
-wheel==0.31.1
+MarkupSafe==2.0.1
 requests
+wheel==0.31.1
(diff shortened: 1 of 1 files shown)


kairu-ms (approved):

scripts/ci/avail-ext-doc/requirements.txt
```diff
@@ -1,3 +1,4 @@
 Jinja2~=2.10.1
```
kairu-ms, line 1: Azure-Cli is using Jinja2~=3.0.3

kairu-ms: We need to update Jinja2 to 3.0.3

jsntcy (approved):

yonzhan:
CI

jiasli (approved):

PR #5202, opened 2022-08-04
[AKS][TA] remove source_resource_id of rolebinding update command

YitongFeng-git:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/HISTORY.rst
+++ b/src/aks-preview/HISTORY.rst
@@ -11,7 +11,7 @@ To release a new version, please select a new version number (usually plus 1 to 
 
 Pending
 +++++++
-* Remove source_resource_id of rolebinding update command
+*  [BREAKING CHANGE] Since the service no longer supports updating source resource id for role binding, so remove --source-resource-id of `aks trustedaccess rolebinding update` command
 
 0.5.92
 ++++++
--- a/src/aks-preview/HISTORY.rst
+++ b/src/aks-preview/HISTORY.rst
@@ -11,6 +11,7 @@ To release a new version, please select a new version number (usually plus 1 to 
 
 Pending
 +++++++
+* Remove source_resource_id of rolebinding update command
 
 0.5.92
 ++++++
--- a/src/aks-preview/azext_aks_preview/_help.py
+++ b/src/aks-preview/azext_aks_preview/_help.py
@@ -1850,9 +1850,6 @@ helps['aks trustedaccess rolebinding update'] = """
         - name: --roles
           type: string
           short-summary: Specify the space-separated roles.
-        - name: --source-resource-id -s
-          type: string
-          short-summary: Specify the source resource id of the binding.
 """
 
 helps['aks trustedaccess rolebinding delete'] = """
--- a/src/aks-preview/azext_aks_preview/custom.py
+++ b/src/aks-preview/azext_aks_preview/custom.py
@@ -2218,14 +2218,26 @@ def aks_trustedaccess_role_binding_get(cmd, client, resource_group_name, cluster
     return client.get(resource_group_name, cluster_name, role_binding_name)
 
 
-def aks_trustedaccess_role_binding_create_or_update(cmd, client, resource_group_name, cluster_name, role_binding_name,
-                                                    source_resource_id, roles):
+def aks_trustedaccess_role_binding_create(cmd, client, resource_group_name, cluster_name, role_binding_name,
+                                          source_resource_id, roles):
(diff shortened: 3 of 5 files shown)


yonzhan:
AKS

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -2229,5 +2230,17 @@ def aks_trustedaccess_role_binding_create_or_update(cmd, client, resource_group_
     return client.create_or_update(resource_group_name, cluster_name, role_binding_name, roleBinding)
 
 
+def aks_trustedaccess_role_binding_update(cmd, client, resource_group_name, cluster_name, role_binding_name, roles):
+    TrustedAccessRoleBinding = cmd.get_models(
+        "TrustedAccessRoleBinding",
+        resource_type=CUSTOM_MGMT_AKS_PREVIEW,
+        operation_group="trusted_access_role_bindings",
+    )
+    existedBinding = client.get(resource_group_name, cluster_name, role_binding_name)
```
karataliu, line 2239: what happens if the binding doesn't exist? should handle None case here?

FumingZhang: A similar situation is that in the `aks update` command CLI side does not handle None case. RP will return a `ResourceNotFound` error, and this error will be raised directly.

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -9,6 +9,7 @@
 import os.path
 import platform
 import re
+import resource
```
FumingZhang, line 12: useless?

src/aks-preview/azext_aks_preview/custom.py
YitongFeng-git: Yes I noticed the `client.get` will raise NotFound error if the binding is notfound

FumingZhang (approved):
LGTM

karataliu (approved):

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -1850,9 +1850,6 @@
         - name: --roles
           type: string
           short-summary: Specify the space-separated roles.
-        - name: --source-resource-id -s
-          type: string
-          short-summary: Specify the source resource id of the binding.
```
zhoxing-ms, line 1855: May I ask deleting the `--source-resource-id` parameter for the `aks trustedaccess rolebinding update` command will cause breaking change to users?

YitongFeng-git: What does this "breaking change" means? From server side we already deny the request of updating source-resource-id, so if users use old version CLI and specify the --source-resource-id, it will get error unless he remove --source-resource-id option.

zhoxing-ms: Got it, thanks~

src/aks-preview/HISTORY.rst
```diff
@@ -11,6 +11,7 @@ To release a new version, please select a new version number (usually plus 1 to
 
 Pending
 +++++++
+* Remove source_resource_id of rolebinding update command
```
zhoxing-ms, line 14: In fact, this is an inevitable breaking change caused by REST service. Therefore, it is recommended to mark it as a breaking change in the history notes and explain the reason
```suggestion
*  [BREAKING CHANGE] Since the service no longer supports updating source resource id for role binding, so remove --source-resource-id of `aks trustedaccess rolebinding update` command
```

src/aks-preview/azext_aks_preview/_help.py
zhoxing-ms: For CLI breaking change, please refer to this doc: [breaking_changes.md](https://msazure.visualstudio.com/One/_git/AzureCliTools-Wiki?version=GBmain&path=/src/teams_docs/azcli_docs/breaking_changes.md)

karataliu:
@zhoxing-ms Thanks for the comment.

Regarding removal of `source_resource_id` parameter:
This feature is still in *private preview* so only internal team has access to the API. I can confirm there is no external impact.
Do you think it's ok to proceed? Thank you.

src/aks-preview/HISTORY.rst
karataliu: see comment below, this feature is in private preview and not accessible by external yet.

zhoxing-ms (approved):

PR #5218, opened 2022-08-09
{CI} drop python 3.6

wangzelin007:
---
By default, the end-of-life is scheduled 5 years after the first release.

image

And the python SDK has already dropped python 3.6.
But we still use python 3.6 in our test enviroment, that will cause some error like these:

image

The root cause is that azure-mgmt-monitor no longer supports python3.6, so the latest azure-mgmt-monitor cannot be installed in the python3.6 test environment.

image
This pull request will drop python 3.6 in our pipeline.

And I will bump pylint to 2.11.1 in this [PR](https://github.com/Azure/azure-cli-extensions/pull/4645) when I have time.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -71,9 +71,9 @@ jobs:
     vmImage: 'ubuntu-20.04'
   steps:
     - task: UsePythonVersion@0
-      displayName: 'Use Python 3.8'
+      displayName: 'Use Python 3.6'
       inputs:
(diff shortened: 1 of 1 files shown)


yonzhan:
drop python 3.6

swells:
NOTE: this is blocking patch Tuesday releases for arcdata today. https://github.com/Azure/azure-cli-extensions/pull/5176

yonzhan (approved):

PR #5241, opened 2022-08-16
Containerapp 0.3.10 Release

StrawnSC:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/containerapp/azext_containerapp/_params.py
+++ b/src/containerapp/azext_containerapp/_params.py
@@ -6,11 +6,9 @@
 
 from knack.arguments import CLIArgumentType
 
-from azure.cli.core.commands.validators import get_default_location_from_resource_group
 from azure.cli.core.commands.parameters import (resource_group_name_type, get_location_type,
                                                 file_type,
                                                 get_three_state_flag, get_enum_type, tags_type)
-# from azure.cli.core.commands.validators import get_default_location_from_resource_group
 
 from ._validators import (validate_memory, validate_cpu, validate_managed_env_name_or_id, validate_registry_server,
                           validate_registry_user, validate_registry_pass, validate_target_port, validate_ingress)
@@ -338,7 +336,7 @@ def load_arguments(self, _):
 
     with self.argument_context('containerapp hostname list') as c:
         c.argument('name', id_part=None)
-        c.argument('location', arg_type=get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
+        c.argument('location', arg_type=get_location_type(self.cli_ctx))
 
     with self.argument_context('containerapp hostname delete') as c:
         c.argument('hostname', help='The custom domain name.')
--- a/src/containerapp/azext_containerapp/tests/latest/test_containerapp_commands.py
+++ b/src/containerapp/azext_containerapp/tests/latest/test_containerapp_commands.py
@@ -561,6 +561,34 @@ class ContainerappRevisionTests(ScenarioTest):
         self.assertEqual(len([w for w in traffic_weight if "label" in w]), 0)
 
 
+class ContainerappAnonymousRegistryTests(ScenarioTest):
+    @AllowLargeResponse(8192)
+    @ResourceGroupPreparer(location="northeurope")
+    @live_only()  # encounters 'CannotOverwriteExistingCassetteException' only when run from recording (passes when run live)
+    def test_containerapp_anonymous_registry(self, resource_group):
+        import requests
+
+        env = self.create_random_name(prefix='env', length=24)
+        app = self.create_random_name(prefix='aca', length=24)
+        image = "mcr.microsoft.com/azuredocs/containerapps-helloworld:latest"
+
+        self.cmd(f'containerapp env create -g {resource_group} -n {env}')
+
+        containerapp_env = self.cmd(f'containerapp env show -g {resource_group} -n {env}').get_output_in_json()
+
+        while containerapp_env["properties"]["provisioningState"].lower() == "waiting":
+            time.sleep(5)
+            containerapp_env = self.cmd(f'containerapp env show -g {resource_group} -n {env}').get_output_in_json()
+
+        self.cmd(f'containerapp create -g {resource_group} -n {app} --image {image} --ingress external --target-port 80 --environment {env}')
+
+        url = self.cmd(f'containerapp show -g {resource_group} -n {app}').get_output_in_json()["properties"]["configuration"]["ingress"]["fqdn"]
+        url = f"https://{url}"
+        resp = requests.get(url)
+        self.assertTrue(resp.ok)
+        self.assertEqual(resp.status_code, 200)
+
+
 class ContainerappRegistryIdentityTests(ScenarioTest):
     @AllowLargeResponse(8192)
     @ResourceGroupPreparer(location="westeurope")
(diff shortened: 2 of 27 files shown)


yonzhan:
Containerapp

panchagnula:
@StrawnSC / @haroonf when is this version going to be ready for release? Seems like it missing some other changes we discussed should be part of the release?

StrawnSC:
> @StrawnSC / @haroonf when is this version going to be ready for release? Seems like it missing some other changes we discussed should be part of the release?

@panchagnula is this the change you were talking about: https://github.com/calvinsID/azure-cli-extensions/pull/145 (Rune's fix for the `--location` in `az containerapp hostname list`)? 
The only other outstanding change is the "no dockerfile" feature, but I thought we wanted to give some time for @anthonychu to try it out. I wouldn't be opposed to merging that with this release though

panchagnula:
> > @StrawnSC / @haroonf when is this version going to be ready for release? Seems like it missing some other changes we discussed should be part of the release?
> 
> @panchagnula is this the change you were talking about: [calvinsID#145](https://github.com/calvinsID/azure-cli-extensions/pull/145) (Rune's fix for the `--location` in `az containerapp hostname list`)? The only other outstanding change is the "no dockerfile" feature, but I thought we wanted to give some time for @anthonychu to try it out. I wouldn't be opposed to merging that with this release though

if we need to wait on testing/ sign-off from Anthony that is ok - but we need to get the regression fix deployed soon, so lets prioritize that accordingly. Thanks!

panchagnula:
@zhoxing-ms could you review this today & share your initial comments!

zhoxing-ms (approved):

src/containerapp/azext_containerapp/tests/latest/test_containerapp_commands.py
```diff
@@ -561,6 +561,34 @@ def test_containerapp_revision_label_e2e(self, resource_group):
         self.assertEqual(len([w for w in traffic_weight if "label" in w]), 0)
 
 
+class ContainerappAnonymousRegistryTests(ScenarioTest):
+    @AllowLargeResponse(8192)
+    @ResourceGroupPreparer(location="northeurope")
+    @live_only()  # encounters 'CannotOverwriteExistingCassetteException' only when run from recording (passes when run live)
+    def test_containerapp_anonymous_registry(self, resource_group):
```
panchagnula, line 568: In general can we work with core-cli to figure out getting these tests run without live mode. Given extension tests don't run in nightly or release builds, or PR builds, these don't help until a dev runs all tests locally.
The containerapp-compose is supposed to have running these kind of tests, using a monkey patch of some sort, can we do something similar?

src/containerapp/azext_containerapp/tests/latest/test_containerapp_commands.py
```diff
@@ -561,6 +561,34 @@ def test_containerapp_revision_label_e2e(self, resource_group):
...
+        containerapp_env = self.cmd(f'containerapp env show -g {resource_group} -n {env}').get_output_in_json()
+
+        while containerapp_env["properties"]["provisioningState"].lower() == "waiting":
+            time.sleep(5)
+            containerapp_env = self.cmd(f'containerapp env show -g {resource_group} -n {env}').get_output_in_json()
+
+        self.cmd(f'containerapp create -g {resource_group} -n {app} --image {image} --ingress external --target-port 80 --environment {env}')
+
+        url = self.cmd(f'containerapp show -g {resource_group} -n {app}').get_output_in_json()["properties"]["configuration"]["ingress"]["fqdn"]
+        url = f"https://{url}"
```
panchagnula, line 586: Curious instead this sort of validation, why can't we do a GET / show command & verify the app exists & some property for FDQN/ ingress or is as expected

panchagnula (changes requested):
Please update the validation logic of the test. Create a work item at the min to investigate & work with core-cli how to make the "live only" tests (espcially for not "up") work as true scenario tests.

src/containerapp/azext_containerapp/tests/latest/test_containerapp_commands.py
smurawski: The problem we bumped into was the generated name for the log analytics workspace.

We worked around this by adding [a hidden/suppressed parameter](https://github.com/Azure/azure-cli-extensions/blob/main/src/containerapp-compose/azext_containerapp_compose/_params.py#L24) and then using the `self.create_random_name` to pass in a known value (known to the test harness) for the name of the log analytics workspace.

This required a monkey patch of `_generate_log_analytics_workspace_name` by https://github.com/Azure/azure-cli-extensions/blob/main/src/containerapp-compose/azext_containerapp_compose/_monkey_patch.py#L57-L60

I'd be happy to grab some time to pair to see if that'd be helpful for your tests.

panchagnula: Thanks @smurawski - given this a create command that can take an existing log-analytics as an input I am curious why this failing. but to your point yes, happy to engage with you as needed.

smurawski: @panchagnula - in this test (see line 575), it's doing a `containerapp env create` which generates a log analytics workspace with a random name.  This will fail on any attempt to replay the recorded test.

StrawnSC: Thanks for the suggestion @smurawski! I'm testing this right now. Thankfully, we can pass in log analytics workspace names to `az containerapp env create` without needing to add hidden parameters to any of our commands, so this might be an easy fix for us

StrawnSC:
@panchagnula I made the change to testing validation you mentioned above. The fix @smurawski suggested didn't work for all our tests (but I changed it for the ones it worked for) so I went ahead and made a work item to investigate this further so it won't block this release

panchagnula:
> @panchagnula I made the change to testing validation you mentioned above. The fix @smurawski suggested didn't work for all our tests (but I changed it for the ones it worked for) so I went ahead and made a work item to investigate this further so it won't block this release

Thanks a lot @StrawnSC , good to see the Diff show more test cases are not live mode anymore. For UP , we need to figure out a better way to make these not live. We can work with core-cli as well during our investigation. These might be tests we need more eyes, so we will still go our manual test pass route.

panchagnula (approved):
Approved.

zhoxing-ms:
@StrawnSC Could you please address those CI issues?

src/containerapp/azext_containerapp/_params.py
```diff
@@ -338,7 +336,7 @@ def load_arguments(self, _):
 
     with self.argument_context('containerapp hostname list') as c:
         c.argument('name', id_part=None)
-        c.argument('location', arg_type=get_location_type(self.cli_ctx), validator=get_default_location_from_resource_group)
+        c.argument('location', arg_type=get_location_type(self.cli_ctx))
```
zhoxing-ms, line 339: I want to confirm that deleting the default location from resource group will cause a breaking change to users?

StrawnSC: Yes, this is not a breaking change. The API accepts a `null` value for location, and the default location that was being chosen before was causing errors for users

panchagnula: Correct - RG supports a lot more locations than ACA does @zhoxing-ms , hence this change. Thanks for calling this out & checking.

zhoxing-ms (approved):

PR #5272, opened 2022-08-25
Adding custom token support

akashkeshari:
---

### General Guidelines

For new extensions:

### About Extension Publish

Changes in the PR:

1. Adding custom AAD token support which does not require az login. User can directly pass the AAD access token to the CLI and execute the connectedk8s commands
2. Adding a flag for overriding container log path in the connect and update command


yonzhan:
connectedk8s

src/connectedk8s/azext_connectedk8s/_params.py
```diff
@@ -43,6 +46,7 @@ def load_arguments(self, _):
         c.argument('onboarding_timeout', options_list=['--onboarding-timeout'], arg_group='Timeout', help='Time required (in seconds) for the arc-agent pods to be installed on the kubernetes cluster. Override this value if the hardware/network constraints on your cluster requires more time for installing the arc-agent pods.')
         c.argument('no_wait', options_list=['--no-wait'], arg_group='Timeout', help="Do not wait for the long-running operation to finish.")
         c.argument('correlation_id', options_list=['--correlation-id'], help='A guid that is used to internally track the source of cluster onboarding. Please do not modify it unless advised', validator=override_client_request_id_header)
+        c.argument('container_log_path', options_list=['--container-log-path'], help='Override the default container log path to enable fluent-bit logging')
```
zhoxing-ms, line 49: In fact, `--container-log-path` is the default option name for parameter `container_log_path`, so there is no need to define `options_list=['--container-log-path']` repeatedly

akashkeshari: fixed

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1277,6 +1286,9 @@ def enable_features(cmd, client, resource_group_name, cluster_name, features, ku
                     azrbac_client_id=None, azrbac_client_secret=None, azrbac_skip_authz_check=None, cl_oid=None):
     logger.warning("This operation might take a while...\n")
 
+    # Validate custom token operation
+    custom_token_passed, _ = utils.validate_custom_token(cmd, resource_group_name, "dummyLocation")
```
sirireddy12, line 1290: can you clarify why we've added "dummyLocation" here?

akashkeshari: this is done since if the location is passed as 'None', we would unnecessarily execute the location fetch step in the function 'validate_custom_token'. This step is required only in case of connect and not enable features

sirireddy12: can we add small comment to explain that?

sirireddy12 (approved):

zhoxing-ms:
Just out of curiosity, may I ask is it safe to put tokens into the environment variables?

akashkeshari:
> Just out of curiosity, may I ask is it safe to put tokens into the environment variables? + CLI Authentication expert @jiasli help review this PR

@zhoxing-ms , wanted to clarify one thing from our side. the custom token support in connectedk8s CLI will only be consumed by internal partners. They have a use case where they want to use the managed identity token for executing the connectedk8s operations. Also, in these scenarios, the connectedk8s CLI will be running inside a container and not on user's machine, so we're fine with setting it as environment variable

zhoxing-ms (approved):

src/connectedk8s/azext_connectedk8s/_client_factory.py
```diff
@@ -2,16 +2,31 @@
...
+import requests
+import azext_connectedk8s._constants as consts
+from collections import namedtuple
+
+AccessToken = namedtuple("AccessToken", ["token", "expires_on"])
+
 
 def cf_connectedk8s(cli_ctx, *_):
     from azext_connectedk8s.vendored_sdks import ConnectedKubernetesClient
+    if os.getenv('AZURE_ACCESS_TOKEN'):
```
jiasli, line 25: The name `AZURE_ACCESS_TOKEN` is never finalized. If https://github.com/Azure/azure-cli/issues/16459 is implemented in the future, it may change.

src/connectedk8s/azext_connectedk8s/_client_factory.py
```diff
@@ -2,16 +2,31 @@
...
+        validate_custom_token()
+        credential = AccessTokenCredential(access_token=os.getenv('AZURE_ACCESS_TOKEN'))
+        return ConnectedKubernetesClient(credential=credential, subscription_id=os.getenv('AZURE_SUBSCRIPTION_ID'),
+                                         base_url=cli_ctx.cloud.endpoints.resource_manager, credential_scopes=[cli_ctx.cloud.endpoints.resource_manager + '/.default'])
```
jiasli, line 29: The best practice is to instantiate `ConnectedKubernetesClient` with `get_mgmt_service_client`, as shown in https://github.com/jiasli/azure-cli/pull/15, instead of by yourself.

Otherwise, as what is done in this PR, you won't have any features provided by `_prepare_mgmt_client_kwargs_track2`, such as CAE and cross-tenant support.

But this does require changes made on `_get_mgmt_service_client` (from https://github.com/jiasli/azure-cli/pull/15) to be merged.

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -69,6 +69,29 @@ def validate_location(cmd, location):
             break
 
 
+def validate_custom_token(cmd, resource_group_name, location):
+    if os.getenv('AZURE_ACCESS_TOKEN'):
+        if os.getenv('AZURE_SUBSCRIPTION_ID') is None:
```
jiasli, line 74: Like `AZURE_ACCESS_TOKEN`, `AZURE_SUBSCRIPTION_ID` is also not finalized.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -66,19 +66,26 @@
...
     # Setting subscription id and tenant Id
-    subscription_id = get_subscription_id(cmd.cli_ctx)
-    account = Profile().get_subscription(subscription_id)
-    onboarding_tenant_id = account['homeTenantId']
+    subscription_id = os.getenv('AZURE_SUBSCRIPTION_ID') if custom_token_passed is True else get_subscription_id(cmd.cli_ctx)
+    if custom_token_passed is True:
+        onboarding_tenant_id = os.getenv('AZURE_TENANT_ID')
+    else:
+        account = Profile().get_subscription(subscription_id)
+        onboarding_tenant_id = account['homeTenantId']
```
jiasli, line 82: I don't have too much knowledge of `onboarding_tenant_id`, but I would like to call out that `tenantId` is the tenant ID of the access token, `homeTenantId` is the tenant ID of the subscription. Be aware! 😉

src/connectedk8s/azext_connectedk8s/_client_factory.py
```diff
@@ -30,36 +50,74 @@ def cf_connected_cluster_prev_2022_05_01(cli_ctx, _):
...
+    def get_token(self, *arg, **kwargs):
+        import time
+        # Assume the access token expires in 60 minutes
+        return AccessToken(self.access_token, int(time.time()) + 3600)
+
+    def signed_session(self, session=None):
+        session = session or requests.Session()
+        header = "{} {}".format('Bearer', self.access_token)
+        session.headers['Authorization'] = header
+        return session
```
jiasli, line 116: This is only for Track 1 SDK. Most likely you won't need it. :)

src/connectedk8s/azext_connectedk8s/_client_factory.py
akashkeshari: This functionality is introduced for supporting internal partners only and hence we read the token via environment variables. If in future this gets implemented in az cli, we can use/expose the same parameters as defined by az cli

src/connectedk8s/azext_connectedk8s/custom.py
akashkeshari: onboarding_tenant_id is something we have defined for arc scenario. it corresponds to 'homeTenantId' only.

src/connectedk8s/azext_connectedk8s/_client_factory.py
akashkeshari: I think we require this since the 'GraphRbacManagementClient' still calls this function. When this function was not implemented , I saw the following error : ''AccessTokenCredential' object has no attribute 'signed_session''

jiasli: `GraphRbacManagementClient` is the old Track 1 SDK for AD Graph. Consider using MS Graph client: https://github.com/Azure/azure-cli/blob/dev/doc/microsoft_graph_client.md

jiasli (approved):

zhoxing-ms (approved):

src/connectedk8s/azext_connectedk8s/_client_factory.py
```diff
@@ -30,36 +48,72 @@ def cf_connected_cluster_prev_2022_10_01(cli_ctx, _):
...
 
 def _graph_client_factory(cli_ctx, **_):
-    profile = Profile(cli_ctx=cli_ctx)
-    cred, _, tenant_id = profile.get_login_credentials(
-        resource=cli_ctx.cloud.endpoints.active_directory_graph_resource_id)
-    client = GraphRbacManagementClient(cred, tenant_id,
-                                       base_url=cli_ctx.cloud.endpoints.active_directory_graph_resource_id)
+    if os.getenv('AZURE_ACCESS_TOKEN'):
+        credential = AccessTokenCredential(access_token=os.getenv('AZURE_ACCESS_TOKEN'))
+        client = GraphRbacManagementClient(credential, os.getenv('AZURE_TENANT_ID'),
```
jiasli, line 80: Wait, the same token for ARM won't work for Graph as the resource/scope of the token is incorrect. You will need to provide a separate access token for Graph invocations.

akashkeshari: That is correct. And that's why the partners who are going to use this feature will not be calling this api. They will be passing the '--custom-location-oid' parameter which will skip the calling of this api. So, its okay for us to merge this change :)

PR #5337, opened 2022-09-12
{AKS}: add enable-cluster-snapshot to aks create command

zqingqing1:
---

### Related command

az aks create --enable-managed-cluster-snapshot

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
AKS

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -678,6 +679,12 @@ def aks_create(
     workload_runtime=None,
     enable_custom_ca_trust=False,
 ):
+    if enable_cluster_snapshot:
+        header = "AKSHTTPCustomFeatures=AddonManagerV2Preview,AKSHTTPCustomFeatures=CSIControllersV2Preview"
+        if aks_custom_headers is None:
```
andyliuliming, line 684: @FumingZhang any comment function can used to update the custom header?

andyliuliming: I think the customer header should be like:
"AKSHTTPCustomFeatures=AddonManagerV2Preview,CSIControllersV2Preview"

if no commen function to call, then there would be issues like this

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2037,6 +2054,12 @@ def set_up_addon_profiles(self, mc: ManagedCluster) -> ManagedCluster:
             addon_profiles[
                 CONST_GITOPS_ADDON_NAME
             ] = self.build_gitops_addon_profile()
+
+        if self.context.get_enable_cluster_snapshot():
+            supported = ["coredns", "overlay-upgrade-data", "kube-proxy", "tunnelfront", "metrics-server", "csi-azurefile-node", "csi-azuredisk-node"]
```
andyliuliming, line 2059: how people manage the list?

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2037,6 +2054,12 @@ def set_up_addon_profiles(self, mc: ManagedCluster) -> ManagedCluster:
             addon_profiles[
                 CONST_GITOPS_ADDON_NAME
             ] = self.build_gitops_addon_profile()
+
+        if self.context.get_enable_cluster_snapshot():
```
andyliuliming, line 2058: why "get_enable_cluster_snapshot" is true, then the addonv2 will be enabled for the items in the list? what's the relation here.

src/aks-preview/HISTORY.rst
```diff
@@ -12,8 +12,13 @@ To release a new version, please select a new version number (usually plus 1 to
 Pending
 +++++++
 
+0.5.101
+++++++
```
FumingZhang, line 16: Per rst syntax, this should be at least the same length of the section name?

src/aks-preview/azext_aks_preview/custom.py
zqingqing1: No, it is not. I tried "AKSHTTPCustomFeatures=AddonManagerV2Preview,CSIControllersV2Preview", it returned error saying CSIControllersV2Preview doesn't have value associate with the key.

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
zqingqing1: They are core addons. Right now, we only snapshot core addons

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
zqingqing1: There is a toggle to control adding the addonv2 config to use addonManagerV2(with this, we can snapshot addon versions), but that takes quite long time to roll out to everywhere, So after syncing with Tingting, she mentioned that we can directly put the addon profile instead of waiting toggle. So to use snapshot feature, we need to have cluster to running addonManagerV2, in other words, put addonprofile for core addons to enable the addonv2.

src/aks-preview/azext_aks_preview/custom.py
zqingqing1: I didn't see the place to update custom header, only for managed cluster.

FumingZhang: The custom headers are usually specified by users, i.e., using this feature at their own risk. If the sub doesn't have `AKSHTTPCustomFeatures` registered, the error message might be strange to the user, since they may have no idea why the feature is required.

FumingZhang: Yes, there's no common function for this, I'll add one later.

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
FumingZhang: Maybe it's better to put this list in `_consts.py`?

src/aks-preview/azext_aks_preview/custom.py
FumingZhang: The helper functions are implemented in PR [#5342](https://github.com/Azure/azure-cli-extensions/pull/5342).

zqingqing1:
It turns out that RP can work on a fix to add addonprofile, close this PR

PR #5367, opened 2022-09-19
[ssh] Add support to "Microsoft.ConnectedVMwarevSphere/virtualMachines" resource type

vthiebaut10:
- Refactored how we check the resource type of the target to facilitate adding support for more resource types in the future.
- Changed how we check the OS of the target resources to make it optional in the future.
- Added connectedvmware sdk
- Fixed the supported resource types names. (Extension was using resource provider names as resource types). Not a breaking change, we still support resource providers.
----

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
ssh

src/ssh/azext_ssh/_client_factory.py
```diff
@@ -29,3 +29,13 @@ def cf_connectedmachine_cl(cli_ctx, *_):
 
 def cf_machine(cli_ctx, *_):
     return cf_connectedmachine_cl(cli_ctx).machines
+
+
+def cf_connectedvmware_cl(cli_ctx, *_):
+    from azext_ssh.vendored_sdks.connectedvmware import AzureArcVMwareManagementServiceAPI
+    return get_mgmt_service_client(cli_ctx,
+                                   AzureArcVMwareManagementServiceAPI)
```
zhoxing-ms, line 37: Please define a custom `CustomResourceType` for  `AzureArcVMwareManagementServiceAPI`

vthiebaut10: @zhoxing-ms Why do we need to use this custom resource type? do I need to use it for the other clients in the same file as well? (hybrid connectivity and connected machine clients)

vthiebaut10: when I try to use that I get this error: `azure.cli.core.profiles._shared.APIVersionException: Unable to get API version for type '<azure.cli.core.profiles._shared.CustomResourceType object at 0x00000267F5ADE250>' in profile 'latest`

vthiebaut10: @zhoxing-ms I pushed the changes you requested, but this error still happens. Do you know what might cause that?

zhoxing-ms: @ vthiebaut10 I recommended you define `CustomResourceType` because the parameter of method `get_mgmt_service_client` should be passed in `client_or_resource_type`, I personally think it will be more in line with the code specification.
But it seems that `AzureArcVMwareManagementServiceAPI` is `# Get the non-versioned client`, so it is not suitable to define as `CustomResourceType`. So please use `AzureArcVMwareManagementServiceAPI` directly as before , thanks~

https://github.com/Azure/azure-cli/blob/07608421fa34cc13c8bbcc1164df08a292b7002a/src/azure-cli-core/azure/cli/core/commands/client_factory.py#L60-L69

zhoxing-ms:
@vthiebaut10 Please resolve those conflicts

src/ssh/azext_ssh/_client_factory.py
vthiebaut10: @zhoxing-ms I reverted the changes.

zhoxing-ms (approved):

PR #5452, opened 2022-10-14
Adding Software Assurance support on connected clusters for AKS HCI

akashkeshari:
---

### General Guidelines

For new extensions:

### About Extension Publish


akashkeshari:
Adding tests for new api is still pending

yonzhan:
connectedk8s

src/connectedk8s/azext_connectedk8s/_params.py
```diff
@@ -35,14 +35,18 @@ def load_arguments(self, _):
         c.argument('no_proxy', options_list=['--proxy-skip-range'], arg_group='Proxy', help='List of URLs/CIDRs for which proxy should not to be used.')
         c.argument('proxy_cert', options_list=['--proxy-cert', '--custom-ca-cert'], arg_group='Proxy', type=file_type, completer=FilesCompleter(), help='Path to the certificate file for proxy or custom Certificate Authority')
         c.argument('distribution', options_list=['--distribution'], help='The Kubernetes distribution which will be running on this connected cluster.', arg_type=get_enum_type(Distribution_Enum_Values))
+        c.argument('distribution_version', help='The Kubernetes distribution version of the connected cluster.')
         c.argument('infrastructure', options_list=['--infrastructure'], help='The infrastructure on which the Kubernetes cluster represented by this connected cluster will be running on.', arg_type=get_enum_type(Infrastructure_Enum_Values))
+        c.argument('azure_hybrid_benefit', help='Flag to enable/disable Azure Hybrid Benefit feature.', arg_type=get_enum_type(AHB_Enum_Values))
```
sirireddy12, line 40: should we mention the supported distros here in the help as well? @anraghun

akashkeshari:
> Adding tests for new features is still pending

Tests added & re-recorded

adkanetk (approved):

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -66,9 +66,20 @@
 
 def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlation_id=None, https_proxy="", http_proxy="", no_proxy="", proxy_cert="", location=None,
                         kube_config=None, kube_context=None, no_wait=False, tags=None, distribution='auto', infrastructure='auto',
-                        disable_auto_upgrade=False, cl_oid=None, onboarding_timeout="600", enable_private_link=None, private_link_scope_resource_id=None):
+                        disable_auto_upgrade=False, cl_oid=None, onboarding_timeout="600", enable_private_link=None, private_link_scope_resource_id=None,
+                        distribution_version=None, azure_hybrid_benefit=None, yes=False, container_log_path=None):
     logger.warning("This operation might take a while...\n")
 
+    # Prompt for confirmation for few parameters
+    if enable_private_link is True:
```
KarthikK123, line 74: We need to check for the SKIP_PROMPT env variable here, without that the runners will be broken.

akashkeshari: we don't require the SKIP_PROMPT env var anymore. We can just use the --yes flag to skip the prompt message for private link. Let's take this change in the runners too

sirireddy12 (approved):

src/connectedk8s/azext_connectedk8s/_params.py
anraghun: I'm okay to skip that. We will call that out in our public docs.

zhoxing-ms, line 40: Why don't we define it as `action='store_true'` but as an enumeration `AHB_Enum_Values`? What is the role of `NotApplicable`?

akashkeshari: The 'AzureHybridBenefit' feature is currently only available for selected type of kubernetes (AKS on HCI) and only those customers are allowed to toggle the value to 'True' or 'False'. For all the remaining kubernetes clusters, the value should be 'NotApplicable' since they are not eligible to claim this benefit. Hence we need to support all three values for this enum

zhoxing-ms: Got it, thanks~

zhoxing-ms (approved):

PR #5469, opened 2022-10-19
{AKS} Add delete step for all test cases

bingosummer:
This PR is to add the step to delete cluster for most test cases.


yonzhan:
AKS

FumingZhang (approved):
LGTM

zhoxing-ms:
Could we uniformly add the `aks delete --yes` command by overriding the `remove_resource` method in `AKSCustomResourceGroupPreparer`?

zhoxing-ms:
By the way, could we uniformly add the `aks delete --yes` command by overriding the `remove_resource` method in `AKSCustomResourceGroupPreparer`?

bingosummer:
@zhoxing-ms 
I didn't know ResourceGroupPreparer already has the logic of resource deletion. Then we don't need this PR, I think. Will close.

PR #5476, opened 2022-10-21
{rdbms-connect} Update mycli to 1.26.1 and pgcli to 3.4.1

alanenriqueo:
- When trying to connect to a mysql server using `az mysql flexible-server connect --interactive`, we get an error message saying `module 'click' has no attribute 'get_terminal_size'`, this is fixed in `mycli==1.26.1`. This version change requires explicitely to set `autocommit=true` for executing queries.
- When trying to connect to a postgres server using `az postgres flexible-server connect --interactive` in Windows 32-bit we could get an error message saying `ImportError: no pq wrapper available` if we don't have library `libpq.dll` installed, that is because current `pgcli==3.0.0` depends on `pgspecial>=1.11.8` which in turns depends on `psycopg` 3, which is only compatible for 64-bit. So, we need to use specific version `pgcli==3.4.1` which depends on exactly `pgspecial=1.13.1` which in turn depends on `psycopg2`, and that is compatible for 32-bit also since it's older. https://github.com/Azure/azure-cli/issues/22889
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
rdbms-connect

evelyn-ys (approved):

PR #5548, opened 2022-11-15
[Container App]: Remove quota check for 'az containerapps env create'

Zijian-Ju:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/containerapp/azext_containerapp/_up_utils.py
+++ b/src/containerapp/azext_containerapp/_up_utils.py
@@ -206,10 +206,10 @@ class ContainerAppEnvironment(Resource):
     def create(self):
         register_provider_if_needed(self.cmd, LOG_ANALYTICS_RP)
 
-        res_locations = list_environment_locations(self.cmd)
-        for loc in res_locations:
-            try:
-                env = create_managed_environment(
+        self.location = validate_environment_location(self.cmd, self.location)
+
+        if self.location: 
+            env = create_managed_environment(
                     self.cmd,
                     self.name,
                     location=self.location,
@@ -218,17 +218,32 @@ class ContainerAppEnvironment(Resource):
                     logs_customer_id=self.logs_customer_id,
                     disable_warnings=True,
                 )
+            self.exists = True 
 
-                self.exists = True 
-                self.location = loc
+            return env
+        else: 
+            res_locations = list_environment_locations(self.cmd)
+            for loc in res_locations:
+                try:
+                    env = create_managed_environment(
+                        self.cmd,
+                        self.name,
+                        location=loc,
+                        resource_group_name=self.resource_group.name,
+                        logs_key=self.logs_key,
+                        logs_customer_id=self.logs_customer_id,
+                        disable_warnings=True,
+                    )
 
-                return env
-            except Exception as ex:
-                logger.info(
-                    f"Failed to create ManagedEnvironment in {loc} due to {ex}"
-                )
-        
-        raise ValidationError("Can not find a region with quota to create ManagedEnvironment")
+                    self.exists = True 
+                    self.location = loc
+
+                    return env
+                except Exception as ex:
+                    logger.info(
+                        f"Failed to create ManagedEnvironment in {loc} due to {ex}"
+                    )
+            raise ValidationError("Can not find a region with quota to create ManagedEnvironment")
 
     def get_rid(self):
         rid = self.name
--- a/src/containerapp/azext_containerapp/_up_utils.py
+++ b/src/containerapp/azext_containerapp/_up_utils.py
@@ -206,9 +206,9 @@ class ContainerAppEnvironment(Resource):
     def create(self):
         register_provider_if_needed(self.cmd, LOG_ANALYTICS_RP)
 
-        self.location = validate_environment_location(self.cmd, self.location)
-
         if self.location:
+            self.location = validate_environment_location(self.cmd, self.location)
+            
             env = create_managed_environment(
                 self.cmd,
                 self.name,
--- a/src/containerapp/azext_containerapp/_up_utils.py
+++ b/src/containerapp/azext_containerapp/_up_utils.py
@@ -46,7 +46,8 @@ from ._utils import (
     trigger_workflow,
     _ensure_location_allowed,
     register_provider_if_needed,
-    validate_environment_location
+    validate_environment_location,
+    list_environment_locations
 )
 
 from ._constants import (MAXIMUM_SECRET_LENGTH,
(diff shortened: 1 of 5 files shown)


src/containerapp/azext_containerapp/_utils.py
```diff
@@ -1506,24 +1506,7 @@ def is_registry_msi_system(identity):
 
 
 def validate_environment_location(cmd, location):
-    from ._constants import MAX_ENV_PER_LOCATION
```
Juliehzl, line 1509: should we also remove MAX_ENV_PER_LOCATION in constants file?

Zijian-Ju: Removed. Also remove the last ValidationError which reference this variable. Can you please have a look again?

yonzhan:
Container App

Juliehzl (approved):

haroonf:
This will break `az containerapp up` logic. Could you instead remove the `validate_environment_location` call from `az containerapp create` and leave `validate_environment_location` as is?

haroonf (changes requested):
Please leave `validate_environment_location` as-is and simply remove the call of `validate_managed_environment` from `create_managed_environment`.` containerapp up` auto creates environments and therefore needs to ensure the environment is created so it must be sure there is available quota on region using `validate_environment_logic`. Removing this would result in a worse customer experience.

Zijian-Ju:
>

zhoxing-ms:
Do you want to release a new extension version for this PR change? If so, please write the PR change into the `HISTORY.rst` and also upgrade the version defined in `setup.py`

Zijian-Ju:
> 

Currently we do allow our customers to create more than 5 managed environments in a region after they submitted a quota request. Otherwise we restrict some of the customers to create less than 5 environments according to their subscription offer type.  

In this case, will it still block customers to create environment using 'az container up' when they are allowed to create more?

Zijian-Ju:
> Currently we do allow our customers to create more than 5 managed environments in a region after they submitted a quota request. Otherwise we restrict some of the customers to create less than 5 environments according to their subscription offer type.
> 
> In this case, will it still block customers to create environment using 'az container up' when they are allowed to create more?

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -1533,15 +1516,10 @@ def validate_environment_location(cmd, location):
         except Exception as e:  # pylint: disable=broad-except
             raise ValidationError("You cannot create a Containerapp environment in location {}. List of eligible locations: {}.".format(location, allowed_locs)) from e
 
-    if len(res_locations) > 0:
-        if not location:
-            logger.warning("Creating environment on location %s.", res_locations[0])
-            return res_locations[0]
-        if location in disallowed_locations:
-            raise ValidationError("You have more than {} environments in location {}. List of eligible locations: {}.".format(MAX_ENV_PER_LOCATION, location, allowed_locs))
```
haroonf, line 1541: Perhaps change this validation error to mention increased quota. Something like: "You have more than {} environments in location {}. If you have an increased limit, create your containerapp using `az containerapp env create` and then pass your environment with --environment. List of eligible locations: {}."

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -203,19 +204,31 @@ def create_if_needed(self, app_name):
...
-        self.exists = True
-        return env
+
+        res_locations = list_environment_locations(self.cmd)
+        for loc in res_locations:
+            try:
+                env = create_managed_environment(
+                    self.cmd,
+                    self.name,
+                    location=self.location,
```
haroonf, line 215: Shouldn't this be location=loc?

haroonf: Looks like this code will keep trying the same location for every loc in res_locations, since we only assign loc to self.location if the env create is successful. Might make sense to keep location=self.location to honor user input, and have self.location = loc at the end of the for loop. So flow would be fail -> log failure -> iterate location.

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -1533,15 +1516,10 @@ def validate_environment_location(cmd, location):
...
-        return location
-    else:
-        raise ValidationError("You cannot create any more environments. Environments are limited to {} per location in a subscription. Please specify an existing environment using --environment.".format(MAX_ENV_PER_LOCATION))
+    if not location:
+        logger.warning("Creating environment on location %s.", res_locations[0])
```
haroonf, line 1520: Might want to just remove this util entirely, since auto-selecting a location and then failing would be a pretty poor user experience. Best to just require the parameter.

src/containerapp/azext_containerapp/_up_utils.py
haroonf: HIstorically, we honor user parameter. For instance, if user passes a location that has full environment quota, we do not try other locations, we simply show them the error message. Only if the user does not pass location do we want to auto-select location. I would reach out to @anthonychu for more details.

Zijian-Ju: I have updated the code to only create once when there is user-input location. When there is no location input, will iterate on res_locations, assign it to self.location first and try to create with selt.location

Thanks for reminding! Please check if this will still break the user experience

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -1533,15 +1516,10 @@ def validate_environment_location(cmd, location):
...
-    if len(res_locations) > 0:
-        if not location:
-            logger.warning("Creating environment on location %s.", res_locations[0])
-            return res_locations[0]
-        if location in disallowed_locations:
-            raise ValidationError("You have more than {} environments in location {}. List of eligible locations: {}.".format(MAX_ENV_PER_LOCATION, location, allowed_locs))
-        return location
-    else:
-        raise ValidationError("You cannot create any more environments. Environments are limited to {} per location in a subscription. Please specify an existing environment using --environment.".format(MAX_ENV_PER_LOCATION))
+    if not location:
```
haroonf, line 1519: Might want to remove this for `containerapp up`, we don't want to say "creating environment on location %s" since we do not know which location we will be creating it on.

Zijian-Ju: Added, thanks for addressing

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -1506,42 +1506,19 @@ def is_registry_msi_system(identity):
...
-    if len(res_locations) > 0:
-        if not location:
-            logger.warning("Creating environment on location %s.", res_locations[0])
-            return res_locations[0]
-        if location in disallowed_locations:
-            raise ValidationError("You have more than {} environments in location {}. List of eligible locations: {}.".format(MAX_ENV_PER_LOCATION, location, allowed_locs))
-        return location
     else:
-        raise ValidationError("You cannot create any more environments. Environments are limited to {} per location in a subscription. Please specify an existing environment using --environment.".format(MAX_ENV_PER_LOCATION))
+        return res_locations[0]
```
Juliehzl, line 1521: add null check

haroonf (approved):

Zijian-Ju:
Hi @zhoxing-ms, could you please help merge this PR and make a new release? The customer is urgently waiting the fix. Thank you

src/containerapp/azext_containerapp/_utils.py
Zijian-Ju: res_location will not be empty after the change as we don't filter by remaining quota anymore

zhoxing-ms (approved):

Juliehzl (approved):

PR #5566, opened 2022-11-21
Connectedk8s nightly testing

rohan-dassani:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
Connectedk8s

wangzelin007:
Hi @rohan-dassani,
```
No match for the request (<Request (PUT) [https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/conk8stest000001/providers/Microsoft.ContainerService/managedClusters/test-upgrade000002?api-version=2022-09-01>](https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/conk8stest000001/providers/Microsoft.ContainerService/managedClusters/test-upgrade000002?api-version=2022-09-01%3E)) was found.
```
**This error message indicates that the version of ContainerService found by the CLI should be 2022-09-01.**
https://github.com/Azure/azure-cli/blob/9ef066f0571b8c4f03af32c439e669a14b563b85/src/azure-cli-core/azure/cli/core/profiles/_shared.py#L252
image
**But the version of ContainerService you recorded is 2022-04-01.**
image

**The recommended steps are as follows:**
```
1. Use the latest azure-cli code: cd $(C:\code\azure-cli); git pull
2. Update azcli: azdev setup --cli $(C:\code\azure-cli) --repo $(C:\code\azure-cli-extensions)
3. Confirm that the local containerService version has been updated to 2022-09-01.  
   (local path: src/azure-cli-core/azure/cli/core/profiles/_shared.py:252)
4. Re-record all tests locally in live mode.
5. Re-upload all recording files
```

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -18,85 +28,506 @@ def _get_test_data_file(filename):
     return os.path.join(curr_dir, 'data', filename).replace('\\', '\\\\')
 
 
+def install_helm_client():
+    # Return helm client path set by user
+    if os.getenv('HELM_CLIENT_PATH'):
+        return os.getenv('HELM_CLIENT_PATH')
```
sirireddy12, line 34: is this needed in test env? no one's going to set the env variables right?

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -18,85 +28,506 @@ def _get_test_data_file(filename):
...
+        except Exception as e:
+            raise ClientRequestError("Failed to create helm executable." + str(e), recommendation="Please ensure that you delete the directory '{}' before trying again.".format(download_dir))
+
+    # Extract compressed helm binary
+    if not os.path.isfile(install_location):
+        try:
+            shutil.unpack_archive(download_location, download_dir)
+            os.chmod(install_location, os.stat(install_location).st_mode | stat.S_IXUSR)
+        except Exception as e:
+            raise ClientRequestError("Failed to extract helm executable." + str(e), recommendation="Please ensure that you delete the directory '{}' before trying again.".format(download_dir))
```
sirireddy12, line 88: let's check with @wangzelin007 / @zhoxing-ms if we should use these error types like "ClientRequestErro" in tests

zhoxing-ms: Good question! It is not recommended to use the `ClientRequestError` in tests because this error type is to collect the request error when customers using CLI. We do not want to confuse the traffic from tests with the traffic from CLI.

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
rohan-dassani: Right , resolved

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,532 @@
...
+            self.check('resourceGroup', '{rg}'),
+            self.check('tags.foo', 'doo')
         ])
 
-        self.cmd('aks delete -g akkeshar -n {} -y'.format(managed_cluster_name))
+        os.environ.setdefault('KUBECONFIG', kubeconfig)
+        helm_client_location = install_helm_client()
+        cmd = [helm_client_location, 'get', 'values', 'azure-arc', "--namespace", "azure-arc-release", "-ojson"]
+
+        # scenario-1 : custom loc off , custom loc on  (no dependencies)
```
sirireddy12, line 229: could you make these scenario descriptions a little more elaborate? It's not completely clear

rohan-dassani: Done , please check

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
+        self.cmd('aks get-credentials -g {rg} -n {managed_cluster_name} -f {kubeconfig}')
+        self.cmd('connectedk8s connect -g {rg} -n {name} -l eastus --tags foo=doo --kube-config {kubeconfig} --kube-context {managed_cluster_name}', checks=[
+            self.check('tags.foo', 'doo'),
+            self.check('resourceGroup', '{rg}'),
+            self.check('name', '{name}')
+        ])
 
-        managed_cluster_name = self.create_random_name(prefix='cli-test-aks-', length=24)
+        self.cmd('connectedk8s delete -g {rg} -n {name} --kube-config {kubeconfig} --kube-context {managed_cluster_name} -y')
+        self.cmd('aks delete -g {rg} -n {managed_cluster_name} -y')
```
sirireddy12, line 156: there's no self.check here , will it still show up in test failures if this command fails here?

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
-            self.check('distributionVersion', '1.0'),
+        # Simulating the condition in which the azure-arc namespace got deleted
+        # connectedk8s delete command fails in this case
+        kubectl_client_location = install_kubectl_client()
+        subprocess.run([kubectl_client_location, "delete", "namespace", "azure-arc","--kube-config", kubeconfig])
+
+        # Using the force delete command
+        # -y to supress the prompts
+        self.cmd('connectedk8s delete -g {rg} -n {name} --kube-config {kubeconfig} --kube-context {managed_cluster_name} --force -y')
+        self.cmd('aks delete -g {rg} -n {managed_cluster_name} -y')
```
sirireddy12, line 195: same comment as above, without self.check will it properly point the failure when this command fails?

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
         os.remove("%s" % (_get_test_data_file('pls-config.yaml')))
 
-    def test_forcedelete(self):
 
-        managed_cluster_name = self.create_random_name(prefix='test-force-delete', length=24)
+    @live_only()
+    @ResourceGroupPreparer(name_prefix='conk8stest', location='eastus2euap', random_name_length=16)
+    def test_upgrade(self,resource_group):
+
+        managed_cluster_name = self.create_random_name(prefix='test-upgrade', length=24)
```
sirireddy12, line 367: Is it mandatory to create the aks cluster and onboard it every time? Is there any way with which we can onboard once and run other commands/tests on top of it . If not all, atleast if we can reuse it for some tests so that we can save time for the overall test run

sirireddy12: AKS cluster creation everytime will take a lot of time. What is the avg time our current pipeline is taking?

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
+        helm_client_location = install_helm_client()
+        cmd = [helm_client_location, 'get', 'values', 'azure-arc', "--namespace", "azure-arc-release", "-ojson"]
+
+        # scenario - auto-upgrade is turned on
+        self.cmd('connectedk8s update -n {name} -g {rg} --auto-upgrade true --kube-config {kubeconfig} --kube-context {managed_cluster_name}')
+        cmd_output1 = subprocess.Popen(cmd, stdout=PIPE, stderr=PIPE)
+        _, error_helm_delete = cmd_output1.communicate()
+        assert(cmd_output1.returncode == 0)
+        changed_cmd1 = json.loads(cmd_output1.communicate()[0].strip())
+        assert(changed_cmd1["systemDefaultValues"]['azureArcAgents']['autoUpdate'] == bool(1))
```
sirireddy12, line 470: can we rename it to something more meaningful? instead of changed, may be updated?

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
rohan-dassani: all the tests actually run parallelly, the avg time is 50 mins but the point here is when the pipeline is triggered it just not only goes to our extension it sequentially goes to all of the extension spend around 2 mins each (again this is parallel too) and then executes our tests too. Pipeline ex - https://dev.azure.com/azclitools/internal/_build/results?buildId=33353&view=results

Having only one aks means , testing everything inside one function , this might actually increase the execution time of the tests and not sure if this is a cleaner approach

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
+
+        with self.assertRaisesRegexp(CLIError, "az connectedk8s upgrade to manually upgrade agents and extensions is only supported when auto-upgrade is set to false"):
+            self.cmd('connectedk8s upgrade -g {rg} -n {name} --kube-config {kubeconfig} --kube-context {managed_cluster_name}')
+
+        # scenario - auto upgrade is off ,then updating agnets to latest and check if the version of agents matches with latest version
```
sirireddy12, line 404: nit: "Turn off auto-upgrade, then updating agents .......so on". So that it is clear that we're turning off as part of this test only

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
sirireddy12: Are all the tests run parallely or sequentially? If parallel, then it makes sense to create everytime. But if sequential, then we should reuse

sirireddy12: why does it go to other extensions if we set the pipeline variable to run only our extension ?

sirireddy12: Also, do we know how many tests run in parallel at once? Even with parallel if it is taking 50 mins, we should get that clarity

sirireddy12: I mean, how many tests are run in parallel for a particular extension

rohan-dassani: thats how the pipeline is designed , will confirm with zelin but i think 8 can run parallely , connectedk8s execution time is around 37 mins

sirireddy12: Okay

rohan-dassani: when we set the pipeline paramter , there is a parameter called user_parallelism = 8 by default , and I have seen 8 or maybe more extensions that are run parallely

yonzhan:
Connectedk8s

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
+import requests
+import platform
+import stat
+from knack.util import CLIError
+import azext_connectedk8s._constants as consts
+import urllib.request
+import shutil
+from knack.log import get_logger
+from azure.cli.core import get_default_cli
+from azure.cli.core.azclierror import ManualInterrupt, InvalidArgumentValueError, UnclassifiedUserFault, CLIInternalError, FileOperationError, ClientRequestError, DeploymentError, ValidationError, ArgumentUsageError, MutuallyExclusiveArgumentError, RequiredArgumentMissingError, ResourceNotFoundError
```
sirireddy12, line 18: let's cleanup these imports if not needed

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
+        requestUri = f'{consts.HELM_STORAGE_URL}/helm/helm-{consts.HELM_VERSION}-{operating_system}-amd64.tar.gz'
+    else:
+        logger.warning(f'The {operating_system} platform is not currently supported for installing helm client.')
+        return
+
+    download_location = os.path.expanduser(os.path.join('~', download_location_string))
+    download_dir = os.path.dirname(download_location)
+    install_location = os.path.expanduser(os.path.join('~', install_location_string))
+
+    # Download compressed halm binary if not already present
```
sirireddy12, line 56: type - *helm

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
sirireddy12: That is number of extensions run parallely, I am asking about number of tests per extension parallely.

sirireddy12: If tests per extension don't run parallely, then there would be no point in create aks and onboarding it everytime. Also, does the pipeline provide an option to run selected tests per extension or all tests will be run always

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
+        os.environ.setdefault('KUBECONFIG', kubeconfig)
+        helm_client_location = install_helm_client()
+        cmd = [helm_client_location, 'get', 'values', 'azure-arc', "--namespace", "azure-arc-release", "-ojson"]
 
-        # Using the force delete command
-        # -y to supress the prompts
-        self.cmd('connectedk8s delete -g rohanazuregroup -n {name} --kube-config {kubeconfig} --force -y')
-        self.cmd('aks delete -g rohanazuregroup -n {} -y'.format(managed_cluster_name))
+        # scenario - auto-upgrade is true , so implicit upgrade commands dont work
+        self.cmd('connectedk8s update -n {name} -g {rg} --auto-upgrade true --kube-config {kubeconfig} --kube-context {managed_cluster_name}')
```
sirireddy12, line 394: should this be upgrade command?  why are we running update command with autoupgrade true?

sirireddy12: based on the scenario description, it looks like it should be upgrade command

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
+            if  repo_path[ind]==':':
+                break
+            index_value += 1
+
+        self.cmd('connectedk8s show -g {rg} -n {name}', checks=[
+            self.check('agentVersion', jsonData['repositoryPath'][index_value+1:]),
+        ])
+
+        # scenario : testing the onboarding timeout change
+        self.cmd('connectedk8s upgrade -g {rg} -n {name} --upgrade-timeout 650 --kube-config {kubeconfig} --kube-context {managed_cluster_name}')
```
sirireddy12, line 427: should the description say "changing upgrade timeout"?

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
+        managed_cluster_name = self.create_random_name(prefix='test-update', length=24)
+        kubeconfig="%s" % (_get_test_data_file(managed_cluster_name + '-config.yaml')) 
+        self.kwargs.update({
+            'name': self.create_random_name(prefix='cc-', length=12),
+            'kubeconfig': kubeconfig,
+            'rg':resource_group,
+            'managed_cluster_name': managed_cluster_name
+        })
+
+        self.cmd('aks create -g {rg} -n {managed_cluster_name} --generate-ssh-keys')
```
sirireddy12, line 448: do we need the "--generate-ssh-keys" flag here? we can remove it

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,521 @@
...
+        self.cmd('aks get-credentials -g {rg} -n {managed_cluster_name} -f {kubeconfig}')
```
sirireddy12, line 449: shouldn't we do --admin to get admin credentials?

rohan-dassani:
@wangzelin007 can you please review this PR

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -5,98 +5,512 @@
...
+
+    except Exception as e:
+        logger.warning("Unable to install kubectl. Error: " + str(e))
+        return
+
+
 class Connectedk8sScenarioTest(LiveScenarioTest):
 
-    def test_connectedk8s(self):
+    @live_only()
```
zhoxing-ms, line 132: May I ask why these tests marked as `@live_only()`?

wangzelin007: I also have the same problem.

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
wangzelin007: The live test was not designed to be run every time. It will only run automatically at specific times, such as every Thursday morning.
Each extension will assign an agent to run the test, which is why you will see so many extensions running at the same time, but they all run for only two minutes, because they are skipped because the running conditions are not met.
If you want to run the tests on every commit, please remove the @live_only decorator and we will run the recording tests after every commit. (Please remember to upload the recording file generated by the live test)

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
wangzelin007: - If you want to run recording tests on every commit, please remove the @live_only decorator and we will run the recording tests after every commit. (Please remember to upload the recording file generated by the live test)
- Otherwise none of the following tests will actually be executed：
image

rohan-dassani: @wangzelin007 if you remember we earlier attempted to try the recording mode but it didnt work out because of some issue in our connectedk8s connect code ( we are fetching tenant-id from cache because of which the tests were failing in recording mode) so for now we have decided to go with live mode only , we know this will not run as integration tests but till the time we find a solution for our command, we will make this a practice from our end that the member that raises the PR will trigger the CLI testing pipeline

@sirireddy12 if you have anything to add

wangzelin007 (approved):

zhoxing-ms (approved):

PR #5596, opened 2022-12-02
Least privilege private preview changes

sirireddy12:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
connectedk8s

src/connectedk8s/azext_connectedk8s/_help.py
```diff
@@ -30,6 +30,8 @@
       text: az connectedk8s connect -g resourceGroupName -n connectedClusterName --enable-private-link true --private-link-scope-resource-id pls/resource/arm/id
     - name: Onboard a connected kubernetes cluster with custom onboarding timeout.
       text: az connectedk8s connect -g resourceGroupName -n connectedClusterName --onboarding-timeout 600
+    - name: Onboard a connected kubernetes cluster with least privileges
+      text: az connectedk8s connect -g resourceGroupName -n connectedClusterName --config-settings "{\\"service-account-name\\":\\"azure-arc-operator-sa\\"}" --least-privilege true
```
akashkeshari, line 34: would be good to add this help example in upgrade & delete

src/connectedk8s/azext_connectedk8s/_params.py
```diff
@@ -47,6 +47,8 @@ def load_arguments(self, _):
         c.argument('correlation_id', options_list=['--correlation-id'], help='A guid that is used to internally track the source of cluster onboarding. Please do not modify it unless advised', validator=override_client_request_id_header)
         c.argument('container_log_path', help='Override the default container log path to enable fluent-bit logging')
         c.argument('yes', options_list=['--yes', '-y'], help='Do not prompt for confirmation.', action='store_true')
+        c.argument('least_privilege', options_list=['--least-privilege'], help='Flag to onboard kubernetes cluster to Arc with least privileges', arg_type=get_enum_type(["true", "false"]))
```
akashkeshari, line 50: options_list is not required. by default it will select the flags '--least-privilege' & '--config-settings'

src/connectedk8s/azext_connectedk8s/_help.py
akashkeshari, line 34: Cant we have this as multiple key value pairs instead of this complex string format? user might make mistakes when passing these. Also check how the --config flag works in k8s-extension cli extension

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -80,6 +80,25 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
...
 
+    if least_privilege is True:
+        # pre-checks specific to least privilege scenario
+        # check if config settings are passed
+        if config_settings is None:
+            telemetry.set_user_fault()
+            telemetry.set_exception(exception="Configuration settings are not passed", fault_type=consts.Config_Settings_Not_Passed_Least_Privileges_Fault_Type, summary="Configuration settings are not passed which are mandatory when onboarding with leastPrivileges")
+            raise ValidationError("Configuration settings are not passed", "Please pass required configuration settings using '--config-settings' flag while onboarding with least-privilege enabled.")
+        else:
+            logger.warning("You are onboarding your k8s cluster to Azure Arc in least privileges mode. Please ensure you have met all the pre-requisites.")
```
akashkeshari, line 91: should we give a link to the pre-reqs here?

sirireddy12: we're not doing public preview currently so we don't have a public aka.ms link as of now

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -80,6 +80,25 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
         confirmation_message = "I confirm I have an eligible Windows Server license with Azure Hybrid Benefit to apply this benefit to AKS on HCI or Windows Server. Visit https://aka.ms/ahb-aks for details"
         utils.user_confirmation(confirmation_message, yes)
 
+    if least_privilege is True:
+        # pre-checks specific to least privilege scenario
+        # check if config settings are passed
+        if config_settings is None:
+            telemetry.set_user_fault()
+            telemetry.set_exception(exception="Configuration settings are not passed", fault_type=consts.Config_Settings_Not_Passed_Least_Privileges_Fault_Type, summary="Configuration settings are not passed which are mandatory when onboarding with leastPrivileges")
+            raise ValidationError("Configuration settings are not passed", "Please pass required configuration settings using '--config-settings' flag while onboarding with least-privilege enabled.")
```
akashkeshari, line 89: should use ArgumentUsageError instead of ValidationError

akashkeshari: you should check out the error types supported by az cli: https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md
this file has some basic description of these error types: https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/azclierror.py

akashkeshari: also, setting ArgumentUsageError might already be setting it as user fault by default so we need not set it explicitly. But confirm this once

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -80,6 +80,25 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
...
+
+        # check if azure-arc ns is present - pre-req for least privilege
+        try:
+            api_instance = kube_client.CoreV1Api()
+            api_instance.read_namespace("azure-arc")
+        except Exception as ex:
+            if ex.status == 404:
+                telemetry.set_user_fault()
+                telemetry.set_exception(exception="Azure-arc namespace is not found on the cluster", fault_type=consts.Azure_Arc_Namespace_Not_Found_Least_Privileges_Fault_Type, summary="Azure-arc namespace is not found on the cluster while onboarding with leastPrivileges")
+
```
akashkeshari, line 101: raise ValidationError?

akashkeshari: you should use the utils.kubernetes_exception_handler while raising error here

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -288,13 +288,14 @@ def delete_arc_agents(release_namespace, kube_config, kube_context, helm_client_
         raise CLIInternalError("Error occured while cleaning up arc agents. " +
                                "Helm release deletion failed: " + error_helm_delete.decode("ascii") +
                                " Please run 'helm delete azure-arc' to ensure that the release is deleted.")
-    ensure_namespace_cleanup()
+    if not least_privilege:
+        ensure_namespace_cleanup()  # NS cleanup shouldn't be attempted when least_privilege=True since onboarding persona won't have sufficient access to delete NS. It's an additional admin operation on user end to cleanup NS
```
akashkeshari, line 292: shouldn't you wait for the cleanup of azure-arc-release namespace here?

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -353,6 +361,20 @@ def helm_install_release(chart_path, subscription_id, kubernetes_distro, kuberne
         raise CLIInternalError("Unable to install helm release: " + error_helm_install.decode("ascii"))
 
 
+def get_serviceaccount_name_from_configsettings(config_settings):
+    try:
+        data = json.loads(config_settings)
+    except Exception as e:
+        telemetry.set_exception(exception=e, fault_type=consts.Config_Settings_Load_Fault_Type, summary="Unable to load config settings. Please ensure the settings are passed in the right format")
```
akashkeshari, line 368: raise error here?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -947,12 +973,6 @@ def update_connected_cluster(cmd, client, resource_group_name, cluster_name, htt
         patch_cc_response = update_connected_cluster_internal(client, resource_group_name, cluster_name, tags, distribution, distribution_version, azure_hybrid_benefit)
 
     proxy_params_unset = (https_proxy == "" and http_proxy == "" and no_proxy == "" and proxy_cert == "" and not disable_proxy)
-
-    # Returning the ARM response if only AHB is being updated
```
akashkeshari, line 951: why is this check being removed? we should not do it since AKS HCI customers will see errors while enabling AHB on their clusters. (since they dont have LoS to the cluster)

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -983,6 +1003,12 @@ def update_connected_cluster(cmd, client, resource_group_name, cluster_name, htt
 
     release_namespace = validate_release_namespace(client, cluster_name, resource_group_name, kube_config, kube_context, helm_client_location)
 
+    helm_values = get_all_helm_values(release_namespace, kube_config, kube_context, helm_client_location)
+
+    if helm_values.get('global').get('isLeastPrivilegesMode') is True:
+        if auto_upgrade is True:
+            raise InvalidArgumentValueError("Your cluster is running in least privileges mode. Autoupdates are not supported in this mode")
```
akashkeshari, line 1010: collect exception telemetry for this

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -338,6 +339,13 @@ def helm_install_release(chart_path, subscription_id, kubernetes_distro, kuberne
         cmd_helm_install.extend(["--kubeconfig", kube_config])
     if kube_context:
         cmd_helm_install.extend(["--kube-context", kube_context])
+
+    if least_privilege:
+        platformServiceAccountName = get_serviceaccount_name_from_configsettings(config_settings)
```
akashkeshari, line 344: shouldn't you pass auto upgrade as false here?

src/connectedk8s/azext_connectedk8s/custom.py
akashkeshari: is the same check being done during onboarding?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -2215,6 +2251,21 @@ def get_custom_locations_oid(cmd, cl_oid):
...
+    release_namespace = validate_release_namespace(client, cluster_name, resource_group_name, kube_config, kube_context, helm_client_location)
+
+    # Checking the connection to kubernetes cluster.
+    # This check was added to avoid large timeouts when connecting to AAD Enabled AKS clusters
+    # if the user had not logged in.
+    check_kube_connection()
+    utils.try_list_node_fix()
+    helm_values = get_all_helm_values(release_namespace, kube_config, kube_context, helm_client_location)
+
+    if helm_values.get('global').get('isLeastPrivilegesMode') is True:
```
akashkeshari, line 2266: least priviledge clusters could also be prone to network/dns/proxy issues. can't we just run the troubleshoot as is and we  just give a warning instead of an error that all least priviledge scenarios are not covered in troubleshoot

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -80,6 +80,25 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
...
+            telemetry.set_user_fault()
+            telemetry.set_exception(exception="Configuration settings are not passed", fault_type=consts.Config_Settings_Not_Passed_Least_Privileges_Fault_Type, summary="Configuration settings are not passed which are mandatory when onboarding with leastPrivileges")
+            raise ValidationError("Configuration settings are not passed", "Please pass required configuration settings using '--config-settings' flag while onboarding with least-privilege enabled.")
+        else:
+            logger.warning("You are onboarding your k8s cluster to Azure Arc in least privileges mode. Please ensure you have met all the pre-requisites.")
+
+        # check if azure-arc ns is present - pre-req for least privilege
+        try:
+            api_instance = kube_client.CoreV1Api()
+            api_instance.read_namespace("azure-arc")
```
akashkeshari, line 96: should we also check the presence of 'azure-arc-release' namespace as well and throw error if its not present?

src/connectedk8s/azext_connectedk8s/_help.py
sirireddy12: Discussed offline, upgrade and delete commands usage will be same as regular mode and hence not adding examples for that

sirireddy12: Got it

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: We can add after we do public preview

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -67,7 +67,7 @@
 def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlation_id=None, https_proxy="", http_proxy="", no_proxy="", proxy_cert="", location=None,
```
akashkeshari, line 67: will this check get affected because of the least priviledge scenario?
https://github.com/Azure/azure-cli-extensions/blob/aca06ac34f5e94cc95e58c0fc86939636eb6c6eb/src/connectedk8s/azext_connectedk8s/custom.py#L147

akashkeshari:
There is a re-put scenario where helm operations also happen. (reput to enable private link). have you thought about the user experience for re-connect in least priviledge mode? are they expected to pass the --least-priviledge true every time in re-connect?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1411,6 +1441,12 @@ def enable_features(cmd, client, resource_group_name, cluster_name, features, ku
 
     release_namespace = validate_release_namespace(client, cluster_name, resource_group_name, kube_config, kube_context, helm_client_location)
 
+    helm_values = get_all_helm_values(release_namespace, kube_config, kube_context, helm_client_location)
+
+    if helm_values.get('global').get('isLeastPrivilegesMode') is True:
```
akashkeshari, line 1446: have you added this check in disable features?

akashkeshari:
are you planning to add integration tests for least priviledge ?

src/connectedk8s/azext_connectedk8s/_params.py
sirireddy12: done

src/connectedk8s/azext_connectedk8s/_help.py
sirireddy12: Done

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: done

src/connectedk8s/azext_connectedk8s/_utils.py
sirireddy12: actually we're blocking autoupdate for LP from agent side itself. But thanks for pointing out, would be good to keep config map up to date

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: I didn't remove this. It's probably because I started this branch sometime back when your change wasn't pushed yet. Adding it

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: Yep added the check. Thanks

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: To disable a feature user shouldn't need any role update and hence this warning is not necessary for disable features. In case of enable , a feature might need higher access and hence the warning

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: We're setting auto-upgrade to false by default whenever it is in least privileges mode, so we don't need a check here, right? And there's no enable-auto-upgrade flag in connect command anyway

src/connectedk8s/azext_connectedk8s/custom.py
sirireddy12: Actually we had added "create" access on clusterrolebindings previously, so this check was passing. But now that access is also not required. So removed it from role requirements and made this check conditional

PR #5620, opened 2022-12-09
{CI} Add check __init__.py file in all extensions

wangzelin007:
Check if the vendored_sdks directory contains `__init__.py` in all extensions.
Will exclude empty vendored_sdks dir, for example:
`azure-cli-extensions\src\containerapp-preview\azext_containerapp_preview\vendored_sdks` is a empty dir that didn't contain any files.

Test screenshot:
image
image

### DIFF
--- a/scripts/ci/test_init.py
+++ b/scripts/ci/test_init.py
@@ -37,10 +37,10 @@ def check_init_files():
 
 def check_init_recursive(root_dir):
     """ Check if a extension contains __init__.py
-    :param root_dir: azure-cli-extensions\\src\\account\\azext_account
-    :param dirpath: azure-cli-extensions\\src\\account\\azext_account
-    :param dirnames: ['generated', 'manual', 'tests', 'vendored_sdks']
-    :param filenames: ['.flake8', 'action.py', 'azext_metadata.json', 'custom.py', '__init__.py']
+    :param root_dir: azure-cli-extensions\src\{ext_name}\azext_{ext_name}
+    :param dirpath: azure-cli-extensions\src\{ext_name}\azext_{ext_name}
+    :param dirnames: all directories under dirpath, type: List[str]
+    :param filenames: all files under dirpath, type: List[str]
     """
     error_flag = False
     for (dirpath, dirnames, filenames) in os.walk(root_dir):
--- a/scripts/ci/test_init.py
+++ b/scripts/ci/test_init.py
@@ -5,9 +5,10 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
-import os
 from util import SRC_PATH
 import logging
+import os
+import sys
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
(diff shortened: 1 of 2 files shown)


yonzhan:
Add check for __init__.py file

scripts/ci/test_init.py
```diff
@@ -0,0 +1,65 @@
...
+                if d.startswith('azext_'):
+                    # root_dir: azure-cli-extensions\src\ext_name\azext_ext_name
+                    ref.append(check_init_recursive(os.path.join(src_d_full, d)))
+    return ref
+
+
+def check_init_recursive(root_dir):
+    """ Check if a extension contains __init__.py
+    :param root_dir: azure-cli-extensions\\src\\account\\azext_account
+    :param dirpath: azure-cli-extensions\\src\\account\\azext_account
```
zhoxing-ms, line 41:
```suggestion
    :param dirpath: azure-cli-extensions\\src\\account\\azext_account\\{dirname}
```

zhoxing-ms (approved):
Just a small comment, no other questions

jiasli:
I think an alternative solution is to use [`find_namespace_packages`](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-namespace-packages) to build a namespace package, but this is not an easy job.

scripts/ci/test_init.py
```diff
@@ -0,0 +1,65 @@
...
+    """ Check if the vendored_sdks directory contains __init__.py in all extensions """
+    ref = []
+    # SRC_PATH: azure-cli-extensions\src
+    for src_d in os.listdir(SRC_PATH):
+        # src_d: azure-cli-extensions\src\ext_name
+        src_d_full = os.path.join(SRC_PATH, src_d)
+        if os.path.isdir(src_d_full):
+            for d in os.listdir(src_d_full):
+                if d.startswith('azext_'):
+                    # root_dir: azure-cli-extensions\src\ext_name\azext_ext_name
```
jiasli, line 33: Nice comment! It makes the code very easy to understand! 👍

jiasli (approved):

wangzelin007:
> I think an alternative solution is to use [`find_namespace_packages`](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-namespace-packages) to build a namespace package, but this is not an easy job.

This is a very good suggestion, it solves the problem fundamentally, but we need to modify the setup.py files of all extensions

Test screenshot:
image
image

scripts/ci/test_init.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+check_path = 'vendored_sdks'
+
+
+def check_init_files():
+    """ Check if the vendored_sdks directory contains __init__.py in all extensions """
+    ref = []
+    # SRC_PATH: azure-cli-extensions\src
+    for src_d in os.listdir(SRC_PATH):
```
jsntcy, line 27: Is it better to only check the extension that the PR is working on?

wangzelin007: `os.walk` is very fast, it takes less than six seconds to check all the files, and the incremental check takes more time and effort.

PR #5674, opened 2023-01-03
[Compute] Fix #5663: `az image copy`: Fix subscription id input by user being override

### LINKED ISSUE #5663: Image Copy '--target-subscription' Argument Ignored
### Related command

```bash
az image copy --source-resource-group <source_rg_name> \
        --source-object-name <source_image_name> \
        --source-type image \
        --target-subscription <target_subscription_id> \
        --target-location <target_location> \
        --target-resource-group <target_rg_name>  \
        --target-name <target_image_name> \
        --temporary-resource-group-name <temp_rg_name>
```

image-copy-extension

### Description of issue (in as much detail as possible)

My organization is using the `image-copy-extension` and we noticed on recent workflows that the extension is now ignoring the `--target-subscription` argument and seems to use the authentication context subscription as the target subscription for all temporary resource and final image creation instead.

We noticed this change because previously the temporary resource group we used was stored within the target subscription and this meant that the temporary storage account, blob, and snapshot all got created within this temporary resource group but now the extension seems to be looking for the temporary resource group in the default subscription associated with the authentication context instead of the target subscription provided to the `--target-subscription` argument.

My Python experience is still limited but I have looked at the [recent commit](https://github.com/Azure/azure-cli-extensions/commit/3a6621cc0c96aab5ddd2388f4a2e6a81f4ed2f0a) and I see that the `target_subscription` variable is now being set without checking if the argument has been passed and I'm not sure if this was changed for a reason or if this is just a bug that can be fixed by only setting the `target_subscription` variable if the argument is null.

image

Link to the line in question.

https://github.com/Azure/azure-cli-extensions/blob/42102c3a81ad23407e7d929521ea212805b4688d/src/image-copy/azext_imagecopy/custom.py#L37

There is also a debug line below the variable declaration that we've seen continues to display the subscription ID for the authentication context and not the target subscription we're providing as an argument.

-----

yanzhudd:
---

### Related command

`az image copy`

### Description

Fix subscription id input by user being override
Issue - close: #5663

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/image-copy/azext_imagecopy/custom.py
+++ b/src/image-copy/azext_imagecopy/custom.py
@@ -33,8 +33,9 @@ def imagecopy(cmd, source_resource_group_name, source_object_name, target_locati
             raise CLIError('Don\'t specify an existing resource group in --temporary-resource-group-name '
                            'when --cleanup is set')
 
-    from azure.cli.core.commands.client_factory import get_subscription_id
-    target_subscription = get_subscription_id(cmd.cli_ctx)
+    if not target_subscription:
+        from azure.cli.core.commands.client_factory import get_subscription_id
+        target_subscription = get_subscription_id(cmd.cli_ctx)
     logger.debug('subscription id - %s', target_subscription)
 
     # get the os disk id from source vm/image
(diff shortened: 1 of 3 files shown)


yonzhan:
Compute

zhoxing-ms (approved):

PR #5723, opened 2023-01-04
{AKS} Remove the error prompt about "no argument specified" when `--enable-workload-identity=False` is specified

FumingZhang:
---

Remove the error prompt about "no argument specified" when `--enable-workload-identity=False` is specified.

### Related command

`az aks update --enable-workload-identity=False`

### General Guidelines

For new extensions:

### About Extension Publish


src/aks-preview/azext_aks_preview/tests/latest/test_managed_cluster_decorator.py
```diff
@@ -4068,6 +4068,29 @@ def test_check_raw_parameters(self):
         )
         self.assertIsNone(dec_2.check_raw_parameters())
 
+        # custom value
+        dec_2 = AKSPreviewManagedClusterUpdateDecorator(
```
bcho, line 4072: this should be dec_3? L4060 uses `dec_2` already. To make these validations easier to read and maintain (why rely on human to count the number?), I suggest we use a table test to list these assertions

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2199,6 +2199,47 @@ def init_agentpool_decorator_context(self) -> None:
...
+        is_changed = any(v for k, v in self.context.raw_param.items() if k not in excluded_keys)
+
+        # special cases
+        # some parameters support the use of empty string or dictionary to update/remove previously set values
+        # the default value of a three state flag is None, false is also manually specified by the user
+        is_default = (
+            self.context.get_cluster_autoscaler_profile() is None and
+            self.context.get_api_server_authorized_ip_ranges() is None and
+            self.context.get_nodepool_labels() is None and
+            self.context.raw_param.get("enable_workload_identity") is None
```
bcho, line 2226: we should extract this flag as a const to avoid forget updating this flag in the future

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2199,6 +2199,47 @@ def init_agentpool_decorator_context(self) -> None:
...
+        )
+
+        if not is_changed and is_default:
+            reconcilePrompt = 'no argument specified to update would you like to reconcile to current settings?'
```
bcho, line 2230: use snake case

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2199,6 +2199,47 @@ def init_agentpool_decorator_context(self) -> None:
...
+        # some parameters support the use of empty string or dictionary to update/remove previously set values
+        # the default value of a three state flag is None, false is also manually specified by the user
+        is_default = (
+            self.context.get_cluster_autoscaler_profile() is None and
+            self.context.get_api_server_authorized_ip_ranges() is None and
+            self.context.get_nodepool_labels() is None and
+            self.context.raw_param.get("enable_workload_identity") is None
+        )
+
+        if not is_changed and is_default:
```
bcho, line 2229: to make the code more readable, I suggest we use early return:

```python
if is_changed or not is_default:
  return

if prompt_y_n(xxx):
  return

# raise error here
```

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2199,6 +2199,47 @@ def init_agentpool_decorator_context(self) -> None:
...
+            reconcilePrompt = 'no argument specified to update would you like to reconcile to current settings?'
+            if not prompt_y_n(reconcilePrompt, default="n"):
+                # Note: Uncomment the followings to automatically generate the error message.
+                option_names = [
```
bcho, line 2233: sort names before joining them to get consistent output

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -312,7 +312,7 @@ def load_arguments(self, _):
         c.argument('enable_vmss', action='store_true', help='To be deprecated. Use vm_set_type instead.', deprecate_info=c.deprecate(redirect='--vm-set-type', hide=True))
         c.argument('zones', zones_type, options_list=['--zones', '-z'], help='Space-separated list of availability zones where agent nodes will be placed.')
         c.argument('ppg')
-        c.argument('enable_encryption_at_host', arg_type=get_three_state_flag(), help='Enable EncryptionAtHost.')
+        c.argument('enable_encryption_at_host', action='store_true')
```
bcho, line 315: why remove the help message?

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2199,6 +2199,47 @@ def init_agentpool_decorator_context(self) -> None:
...
+        If the values of all the parameters are the default values, the command execution will be terminated early and
+        raise a RequiredArgumentMissingError. Neither the request to fetch or update the ManagedCluster object will be
+        sent.
+
+        :return: None
+        """
+        # exclude some irrelevant or mandatory parameters
+        excluded_keys = ("cmd", "client", "resource_group_name", "name")
+        # check whether the remaining parameters are set
+        # the default value None or False (and other empty values, like empty string) will be considered as not set
```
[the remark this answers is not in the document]
bcho: and do we count `0` as not set value here? I feel like not all falsy values == not set value. Therefore, I suggest we explicitly layout all considered not set value in the helper function.

yonzhan:
AKS

src/aks-preview/azext_aks_preview/_params.py
FumingZhang: The declaration of this option in azure-cli/acs is an ordinary bool flag, and this option does not support update. So remove the confusing three state flag.

src/aks-preview/azext_aks_preview/tests/latest/test_managed_cluster_decorator.py
FumingZhang: Ah, I copied the previous test code by mistake which results in reusing the `_2` label. Would remove this piece of code.

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
FumingZhang: sure

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
FumingZhang: ack

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
FumingZhang: sure

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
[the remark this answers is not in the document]
FumingZhang: I think `0` may or may not be considered the default value, which all depends on the design of the feature. The approach taken here is to first screen whether any parameter is explicitly set to a non-false value through implicit conversion, and then determine whether any parameter is assigned a value through the secondary screening of special parameters.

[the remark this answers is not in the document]
FumingZhang: Now the data source of this secondary screening is split into a separate function `get_special_parameter_default_value_pairs_list`.

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
FumingZhang: These flags are now moved to a separate function.

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2199,6 +2194,59 @@ def init_agentpool_decorator_context(self) -> None:
...
+        is_changed = any(v for k, v in self.context.raw_param.items() if k not in excluded_keys)
+
+        # special cases
+        # some parameters support the use of empty string or dictionary to update/remove previously set values
+        # the default value of a three state flag is None, false is also manually specified by the user
+        is_different_from_special_default = False
+        for pair in self.get_special_parameter_default_value_pairs_list():
+            if pair[0] != pair[1]:
+                is_different_from_special_default = True
```
bcho, line 2231: break here?

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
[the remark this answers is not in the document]
bcho: then could you change the comment to explicitly call out `falsy` value? Like:

> the default falsy value will be considered as not set

And update the special cases comment as well:

> special cases
> Some parameters support using falsy value to update/remove previously set values.
> In this case, we need to declare the expected value pair in `get_special_parameter_default_value_pairs_list`

bcho (changes requested):
minor comment issue

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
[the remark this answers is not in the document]
bcho: then could you change the comment to explicitly call out `falsy` value? Like:

> the default falsy value will be considered as not set

And update the special cases comment as well:

> special cases
> Some parameters support using falsy value to update/remove previously set values.
> In this case, we need to declare the expected value pair in `get_special_parameter_default_value_pairs_list`

bcho (changes requested):

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
FumingZhang: good catch

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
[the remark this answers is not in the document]
FumingZhang: Updated, thanks for the suggestions.

bcho (approved):

src/aks-preview/azext_aks_preview/_params.py
zhoxing-ms: I'd like to confirm whether changing `get_three_state_flag()` to `action='store_true'` will cause a breaking change to users? Did any users use `--enable-encryption-at-host False` before?

FumingZhang: Hi @zhoxing-ms, I think this wouldn't be a breaking change to users.

This property does not support being updated, it is only allowed to be used when creating an aks cluster, and the default value is false, only need to be specified when this feature needs to be enabled.

In the [official doc](https://learn.microsoft.com/en-us/azure/aks/enable-host-encryption#use-host-based-encryption-on-new-clusters), the example given in the guide is in the form of `az aks create xxx --enable-encryption-at-host`.

Besides the option is also supported in azure-cli/acs, and the [same parameter](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acs/_params.py#L259) declared there is a normal bool flag.

zhoxing-ms (approved):

PR #5751, opened 2023-01-12
Add NRG Profile + Restriction Level to the Preview CLI Update + Create paths

charliedmcb:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
AKS

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -3002,6 +3002,38 @@ def test_aks_create_with_auto_upgrade_channel(self, resource_group, resource_gro
...
+                     '--enable-managed-identity ' \
+                     '--nrg-lockdown-restriction-level ReadOnly ' \
+                     '--ssh-key-value={ssh_key_value} ' \
+                     '--aks-custom-headers AKSHTTPCustomFeatures=Microsoft.ContainerService/NRGLockdownPreview'
+        self.cmd(create_cmd, checks=[
+            self.check('provisioningState', 'Succeeded'),
+            self.check('nodeResourceGroupProfile.restrictionLevel', 'ReadOnly')
+        ])
+
+        # update upgrade channel
```
FumingZhang, line 3027: > update upgrade channel
misleading comment, updating nrg lockdown

src/aks-preview/setup.py
```diff
@@ -9,7 +9,7 @@
 
 from setuptools import setup, find_packages
 
-VERSION = "0.5.122"
+VERSION = "0.5.123"
```
FumingZhang, line 12: Please also update history.rst to describe your changes.

FumingZhang (commented):
Queued a live test pipeline to test newly added test case `test_aks_create_and_update_with_nrg_restriction_level`.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
charliedmcb: Fixed

src/aks-preview/setup.py
charliedmcb: Added

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -292,6 +292,10 @@
         - name: --node-resource-group
           type: string
           short-summary: The node resource group is the resource group where all customer's resources will be created in, such as virtual machines.
+        - name: --nrg-restriction
```
charliedmcb, line 295: @FumingZhang curious why there is a `22` character limit on the option lengths:
https://dev.azure.com/azclitools/public/_build/results?buildId=25902&view=logs&j=ccdac6d3-465a-5ada-8819-d5b4a6b3a631&t=ffdbba7d-e523-5908-c7bf-0703bafb81b0&l=472

Would have preferred to use `--nrg-lockdown-restriction-level` which was approved in the API review. However, due to the character limit I had to switch it to `--nrg-restriction` as seen currently in code

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -467,6 +467,26 @@ def get_load_balancer_backend_pool_type(self) -> str:
...
 
+    def get_nrg_restriction(self) -> Union[str, None]:
+        """Obtain the value of nrg_restriction.
+        :return: string or None
+        """
+        # read the original value passed by the command
+        nrg_restriction = self.raw_param.get("nrg_restriction")
+
+        # In create mode, try to read the property value corresponding to the parameter from the `mc` object.
+        if self.decorator_mode == DecoratorMode.CREATE:
```
charliedmcb, line 478: @FumingZhang, I'm curious if you know the reasoning for this line.

I found it in `get_auto_upgrade_channel` method:
https://github.com/Azure/azure-cli/blob/faccf0b590f42b87163c393b7cdfc25bd73640d0/src/azure-cli/azure/cli/command_modules/acs/managed_cluster_decorator.py#L3839

src/aks-preview/azext_aks_preview/_help.py
FumingZhang: Hey @charliedmcb, you can use the option names designed and approved before. To bypass such CI checks, just update file `src/aks-preview/linter_exclusions.yml` and add an entry about this new option.

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
FumingZhang: We have two code paths, one for create and the other for update. In both of them, the parameter value is obtained through the same `get_xxx` function, but the behavior of obtaining the value is expected to be different.

In the create code path, we want to obtain the option value from the target mc first (when creating the resource, if the option value has been written into the target mc, it's considered as immutable), if it's not set yet, we'll obtain the value from user input.

In the update code path, we just obtain the value from user input.

Many options do not support being updated (can only be set at create time), so for their getter function, it doesn't matter whether to add this line (`if self.decorator_mode == DecoratorMode.CREATE:`) or not.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -3025,6 +3025,38 @@ def test_aks_create_with_auto_upgrade_channel(self, resource_group, resource_gro
         self.cmd(
             'aks delete -g {resource_group} -n {name} --yes --no-wait', checks=[self.is_empty()])
 
+    @AllowLargeResponse()
+    @AKSCustomResourceGroupPreparer(random_name_length=17, name_prefix='clitest', location='westus2')
+    def test_aks_create_and_update_with_nrg_restriction_level(self, resource_group, resource_group_location):
```
FumingZhang, line 3030: Queued a [pipeline](https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=67101641&view=results) last Friday, the test passed!

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
charliedmcb: Is there an example case where its expected to possibly set the value on the mc before obtaining it? Possibly an interaction between a profile being set both in the core cli + the preview cli?

src/aks-preview/azext_aks_preview/_help.py
charliedmcb: Ok, thanks! :)

charliedmcb: I'll look at updating this tomorrow, and ping you for a final review when ready

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
FumingZhang: When getting the parameter value for the very first time, the corresponding value in mc shouldn't be set yet (there are cases where SDK models provide default values, but that's actually handled by some other hack). However, some cross-checks between profiles will obtain the property value ​​in the previously set profile again, and at this time the value will be read from mc (instead of user's raw input). One such example would be the validation between dns prefix and fqdn subdomain.

src/aks-preview/azext_aks_preview/_help.py
charliedmcb: Updated

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
charliedmcb: Ok, interesting.

zhoxing-ms (approved):

PR #5832, opened 2023-02-02
add model mapping to support add/update of initContainers to Containe…

anandanthony:
…rApps via Yaml

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/containerapp/azext_containerapp/_sdk_models.py
+++ b/src/containerapp/azext_containerapp/_sdk_models.py
@@ -3217,6 +3217,7 @@ class Template(Model):
     _attribute_map = {
         'revision_suffix': {'key': 'revisionSuffix', 'type': 'str'},
         'containers': {'key': 'containers', 'type': '[Container]'},
+        'initContainers': {'key': 'initContainers', 'type': '[Container]'},
         'scale': {'key': 'scale', 'type': 'Scale'},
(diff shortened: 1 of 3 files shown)


yonzhan:
containerapp

src/containerapp/azext_containerapp/_sdk_models.py
```diff
@@ -3217,6 +3217,7 @@ class Template(Model):
     _attribute_map = {
         'revision_suffix': {'key': 'revisionSuffix', 'type': 'str'},
         'containers': {'key': 'containers', 'type': '[Container]'},
+        'initContainers': {'key': 'initContainers', 'type': '[Container]'},
```
anthonychu, line 3220: In the spec, it looks like initContainers are different from app containers because [app containers have probes](https://github.com/Azure/azure-rest-api-specs/blob/9f33c53e4b6d09e22329df29ac2ef404fe2149de/specification/app/resource-manager/Microsoft.App/stable/2022-10-01/CommonDefinitions.json#L115). Would this be a problem here?

anandanthony: @anthonychu if we set probes for a InitContainer, it will not be processed. We handle it in the actual API create/update API.

zhoxing-ms (approved):

PR #5841, opened 2023-02-03
Containerapp - Add CLI changes for sticky sessions ingress 

tdaroly:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
Containerapp

tdaroly:
Lets NOT merge this PR yet, there are a few discussions regarding api version's we would like to get done first. With that being said we would appreciate the necessary approvals.

zhoxing-ms:
Could you add some tests for this PR?

src/containerapp/azext_containerapp/custom.py
```diff
@@ -2162,6 +2162,46 @@ def show_ip_restrictions(cmd, name, resource_group_name):
...
+    except:
+        pass
+
+    if not containerapp_def:
+        raise ResourceNotFoundError("The containerapp '{}' does not exist".format(name))
+
+    try:
+        return containerapp_def["properties"]["configuration"]["ingress"]
+    except Exception as e:
+        raise ValidationError("Ingress must be enabled to enable sticky sessions. Try running `az containerapp ingress -h` for more info.") from e
```
p-bouchon, line 2202: you should handle_raw_exception(e) and check for ingress being enabled seperately

src/containerapp/HISTORY.rst
```diff
@@ -5,6 +5,7 @@ Release History
 0.3.26
 +++++++
 * 'az containerapp exec': fix bugs for consumption workload based environment
+* Add CLI support for ingress sticky-sessions'
```
anandanthony, line 8: would it be better to add the exact command group that we have added for ingress sticky sessions here? 'containerapp ingress sticky-sessions'

src/containerapp/azext_containerapp/_help.py
```diff
@@ -778,6 +778,32 @@
           az containerapp ingress access-restriction list -n MyContainerapp -g MyResourceGroup
 """
 
+helps['containerapp ingress sticky-sessions'] = """
+    type: group
+    short-summary: Commands to set Sticky affinity for a container app .
```
anandanthony, line 783: NIT: remove space before .

src/containerapp/azext_containerapp/_models.py
```diff
@@ -131,6 +131,10 @@
     "latestRevision": False
 }
 
+StickySessions = {
```
anandanthony, line 134: do we need to define this model explicitly here? Can't it be used like 'ipSecurityRestrictions' that does not need to be added separately here, we just have a mapping to it in the sdk_models.py.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -2162,6 +2162,46 @@ def show_ip_restrictions(cmd, name, resource_group_name):
         return []
 
 
+def set_ingress_sticky_session(cmd, name, resource_group_name, affinity, no_wait=False):
+    _validate_subscription_registered(cmd, CONTAINER_APPS_RP)
```
anandanthony, line 2166: do we need validation here? We would add sticky sessions to a Container App, which means that the sub would already have a validated subscription?

src/containerapp/HISTORY.rst
p-bouchon: also I believe this should be added in an Upcoming section instead of under the 0.3.26 notes

src/containerapp/azext_containerapp/custom.py
tdaroly: no but it is a good idea to have it anyways

src/containerapp/azext_containerapp/_models.py
tdaroly: that was my reasoning as well but there is a `TrafficWeight` model in the same file `_models.py` as well hence i added StickySession struct

src/containerapp/HISTORY.rst
tdaroly: what do you mean `command group` ? can you elaborate a little more ?

anandanthony: the command that has been added for [sticky sessions](https://github.com/Azure/azure-cli-extensions/blob/81e5104f8aeb2267b5f3c5cac1e90fb0c45a05c2/src/containerapp/azext_containerapp/commands.py#L125).

src/containerapp/azext_containerapp/_models.py
anandanthony: ok, if that struct is being used in the custom.py file it makes sense to have it, else remove it. 
Ref: https://github.com/Azure/azure-cli-extensions/blob/81e5104f8aeb2267b5f3c5cac1e90fb0c45a05c2/src/containerapp/azext_containerapp/custom.py#L1767

src/containerapp/azext_containerapp/custom.py
anandanthony: if you don't need it, remove it. It seems to be making an API call, which is in addition to the call to get the Container App, which in itself is an API call. If you get a Container App for a subscription, that implies that the subscription already has the RP registered right?

tdaroly: Other ingress settings are using this validation and therefore i will keep it to follow pattern 
image

src/containerapp/azext_containerapp/_models.py
tdaroly: done

src/containerapp/azext_containerapp/_help.py
tdaroly: done

src/containerapp/HISTORY.rst
tdaroly: done

tdaroly:
We CAN merge PR, asking mods to remove do not merge label and merge the PR. Thank you

anandanthony (approved):

p-bouchon (approved):

zhoxing-ms (approved):

PR #5843, opened 2023-02-04
Containerapp 0.3.22 Release

lil131:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
Containerapp

src/containerapp/HISTORY.rst
```diff
@@ -2,8 +2,15 @@
 
 Release History
 ===============
-Upcoming
-+++++++
+0.3.22
+++++++
+* BREAKING CHANGE: 'az containerapp env certificate list' returns [] if certificate not found, instead of raising an error.
```
zhoxing-ms, line 7: Why do we need to do this breaking change?

src/containerapp/azext_containerapp/_help.py
```diff
@@ -500,14 +500,23 @@
...
+
 helps['containerapp env certificate list'] = """
     type: command
     short-summary: List certificates for an environment.
     examples:
     - name: List certificates for an environment.
       text: |
           az containerapp env certificate list -g MyResourceGroup --name MyEnvironment
-    - name: List certificates by certificate id.
+    - name: List certificate by certificate id.
```
zhoxing-ms, line 519: Why use the singular nouns?

src/containerapp/azext_containerapp/_params.py
```diff
@@ -176,6 +176,12 @@ def load_arguments(self, _):
     with self.argument_context('containerapp env show') as c:
         c.argument('name', name_type, help='Name of the Container Apps Environment.')
 
+    with self.argument_context('containerapp env certificate create') as c:
+        c.argument('hostname', options_list=['--hostname'], help='The custom domain name.')
```
zhoxing-ms, line 180: Please do not explicitly define the default option name
```suggestion
        c.argument('hostname', help='The custom domain name.')
```

src/containerapp/azext_containerapp/_params.py
```diff
@@ -176,6 +176,12 @@ def load_arguments(self, _):
...
+        c.argument('certificate_name', options_list=['--certificate-name', '-c'], help='Name of the managed certificate which should be unique within the Container Apps environment.')
+        c.argument('location', options_list=['--location'], help='Location of the managed certificate which can be different from the location of the Container Apps environment.')
```
zhoxing-ms, line 182:
```suggestion
        c.argument('location', get_location_type(self.cli_ctx), help='Location of the managed certificate which can be different from the location of the Container Apps environment.')
```

src/containerapp/HISTORY.rst
lil131: Previously, we were only querying for private-key certificates and showing whatever the API returned. Now with managed certificates added, we have to query two separate APIs to look for the matching certificates. So we can't show the API error directly. Considering it's a `list` command instead of `show` command, I think it makes more sense to return `[]` than raising an error.

src/containerapp/azext_containerapp/_help.py
lil131: If certificate id is provided, there will be only one certificate returned.

src/containerapp/azext_containerapp/_params.py
lil131: The default option name for hostname is `--custom-domain-dns-suffix`/`--dns-suffix`, which is for environment-level custom domain.
This hostname here is for app-level hostname, which is the same concept as `--hostname` in `containerapp hostname` commands.

src/containerapp/HISTORY.rst
zhoxing-ms: Got it, thanks~

src/containerapp/azext_containerapp/_help.py
zhoxing-ms: OK, but in fact, this operation will be more appropriate in the show command instead of list command

zhoxing-ms, line 519:
```suggestion
    - name: Show certificate by certificate id.
```

src/containerapp/azext_containerapp/_params.py
zhoxing-ms: > The default option name for hostname is --custom-domain-dns-suffix/--dns-suffix, which is for environment-level custom domain.

Sorry, I don't get what you mean. Could you please show me the relevant code links or explain the reasons?

src/containerapp/azext_containerapp/_help.py
lil131: I see. We will consider having a show command for certificate. Thanks!

src/containerapp/azext_containerapp/_params.py
lil131: Sure!
In this `containerapp env certificate create` command, if we don't explicitly define hostname, it will inherit the definition from `containerapp env`, which will be `--custom-domain-dns-suffix` or `--dns-suffix`, and it means "The DNS suffix for the **environment**\'s custom domain.":
https://github.com/Azure/azure-cli-extensions/blob/2a4fceb2302cc9f0ff13e002ed8a7e2fc11de35f/src/containerapp/azext_containerapp/_params.py#L161-L162
But the hostname here in `containerapp env certificate create` is actually a different concept. It stands for the **app**-level custom domain. This concept is also used in `containerapp hostname` commands:
https://github.com/Azure/azure-cli-extensions/blob/2a4fceb2302cc9f0ff13e002ed8a7e2fc11de35f/src/containerapp/azext_containerapp/_params.py#L364-L365
So we want to use the name `--hostname` here to keep it consistent with `containerapp hostname` commands.

zhoxing-ms: Got it, thanks for your detailed explanation

zhoxing-ms (approved):

src/containerapp/azext_containerapp/_params.py
lil131: Np! Thanks for approving the PR!

PR #5867, opened 2023-02-08
[devcenter] Add devcenter extension

am-lim:
---

### Related command

az devcenter

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
devcenter

kairu-ms:
Hi @am-lim, Thanks to onborading your extension to azure-cli. I found the code are generated by `autorest.az` which is not maintained. Could use [aaz-dev-tools](https://github.com/Azure/aaz-dev-tools) to generate azure-cli commands? Thanks.

am-lim:
Hi @kairu-ms, could we check in these commands using autorest for this API version? Currently, our service is in public preview and we would like to make this available to customers. We will soon have our GA API version finalized and will use aaz-dev-tools to generate this API.

linter_exclusions.yml
```diff
@@ -623,6 +623,26 @@ desktopvirtualization workspace update:
     application_group_references:
       rule_exclusions:
       - option_length_too_long
+devcenter admin catalog update:
+  parameters:
+    dev_center_name:
+      rule_exclusions:
+      - no_parameter_defaults_for_update_commands
```
kairu-ms, line 630: Why do you need to use default value for update commands?

src/devcenter/azext_devcenter/generated/_params.py
```diff
@@ -0,0 +1,730 @@
...
+from azext_devcenter.action import (
+    AddGitHub,
+    AddImageReference,
+    AddSku
+)
+
+
+def load_arguments(self, _):
+
+    with self.argument_context('devcenter dev-center list') as c:
```
kairu-ms, line 36: Why the command name contains double devcenter?

kairu-ms:
> Hi @kairu-ms, could we check in these commands using autorest for this API version? Currently, our service is in public preview and we would like to make this available to customers. We will soon have our GA API version finalized and will use aaz-dev-tools to generate the GA API.

Hi @am-lim, I found you copy the generated folder into manual and change code here instead of creating configuration in autorest.az to achieve the modification. That is not the recommended way to use autorest.az. But we know the pain point that autorest.az is hard to config. That is the reason we developed  aaz-dev-tools for azure-cli developers. You can have a try. I've reviewed your code. Most of the commands can be easily generated by using aaz-dev-tools. And I think the migration of your extension from autorest.az to aaz-dev-tools can be done in less than 3 days for the new user. 

Another two reasons to use aaz-dev-tools are that
- The update commands generated by autorest.az are not correct.
- The complicated arguments generated by autorest.az are not easy to use. For example `--parameters` in your commands. But by using aaz-dev-tools, we have shorthand syntax and hints to tell customers use these arguemnts.

I can scheduler a demo meeting for you if you want.

src/devcenter/azext_devcenter/generated/_params.py
am-lim: Our RP name is Microsoft.DevCenter and we have a resource named "DevCenter" as well. We have gotten feedback about this "double" naming and this still under review.

linter_exclusions.yml
am-lim: We would like to use "az configure" for this parameter because it appears in a lot of our commands and it would improve the usability of the commands for the users. You cannot update the dev_center/dev_center_name name for any of these resources - they are child resources of dev_center and is only used to help identify the resource to perform the update operation on (similar to resource group).

yonzhan:
devcenter

src/devcenter/azext_devcenter/custom.py
```diff
@@ -0,0 +1,1117 @@
...
+        return args_schema
+
+    def _execute_operations(self):
+        self.pre_operations()
+        condition_0 = has_value(self.ctx.args.dev_center_name)
+        condition_1 = has_value(self.ctx.args.project_name)
+        if condition_0:
+            self.AttachedNetworksListByDevCenter(ctx=self.ctx)()
+        elif condition_1:
+            AttachedNetworkListByProject.AttachedNetworksListByProject(ctx=self.ctx)()
```
kairu-ms, line 147: aaz-dev-tools have a way to merge those two apis in one command.

src/devcenter/azext_devcenter/custom.py
```diff
@@ -0,0 +1,1034 @@
...
+        if arg_name == "dev_center_name":
+            arg.configured_default = "dev-center"
+    return args
+
+
+class AttachedNetworkCreate(_AttachedNetworkCreate):
+    @classmethod
+    def _build_arguments_schema(cls, *args, **kwargs):
+        args_schema = super()._build_arguments_schema(*args, **kwargs)
+        args_schema.network_connection_id._required = True
```
kairu-ms, line 109: For those arguments which is not used in commands, you should hide them in aaz-dev-tools 
image

kairu-ms (approved):

PR #5515, opened 2022-11-04
{CI} Support min version test for extensions

wangzelin007:
### This PR will support test extension with min cli version:

- **detect which extension need to test.**
`ALL_TESTS: [('azext_healthcareapis', '/mnt/vss/_work/1/s/src/healthcareapis')]`

- **find min_version in azext_metadata.json**
`find minCliCoreVersion: 2.15.0`

- **find python version that is compatible with the cli min version**
For example: 
If minCliCoreVersion is 2.15.0, according to [azure-cli-core/setup.py](https://github.com/Azure/azure-cli/blob/b59c67a96959021a8b414d6ff2a304b210bec0c7/src/azure-cli-core/setup.py#L38-L41), 
Supported python version is 3.6~3.8, we will use all python versions that have not reached the end-of-life for testing.
Which is 3.7 and 3.8

- **use python version found in previous step**
Use pipeline conditions to control which versions should be used for testing.
`condition: and(succeeded(), contains(variables.PythonVersion, ${{version.value}}))`
[Use output variables across jobs](https://github.com/Azure/azure-cli-extensions/pull/5515#issuecomment-1315347619)

- **check out to min cli version**
```
['az', '--version']
WARNING: You have 2 updates available. Consider updating your CLI installation with 'az upgrade'
azure-cli                         2.15.0 *

core                              2.15.0 *
telemetry                          1.0.6

Extensions:
azure-devops                      0.25.0
```

- **install old testsdk**
Use old testsdk's setup.py file to install DEPENDENCIES
[Why do we need to install the old testsdk?](https://github.com/Azure/azure-cli-extensions/pull/5515#issuecomment-1303550066)
```
Processing azure_cli_testsdk-0.2.4-py3.7.egg
creating /mnt/vss/_work/1/s/env/lib/python3.7/site-packages/azure_cli_testsdk-0.2.4-py3.7.egg
Extracting azure_cli_testsdk-0.2.4-py3.7.egg to /mnt/vss/_work/1/s/env/lib/python3.7/site-packages
Removing azure-cli-testsdk 0.3.0 from easy-install.pth file
Adding azure-cli-testsdk 0.2.4 to easy-install.pth file

Installed /mnt/vss/_work/1/s/env/lib/python3.7/site-packages/azure_cli_testsdk-0.2.4-py3.7.egg
Processing dependencies for azure-cli-testsdk==0.2.4
Searching for azure-devtools==1.2.0
Reading https://pypi.org/simple/azure-devtools/
[...]

PR #5877, opened 2023-02-13
Revert "{CI} Support min version test for extensions"

wangzelin007:
Reverts Azure/azure-cli-extensions#5515
Our recording files only use the latest api version.
If we test extension with min version, we also need to provide the corresponding recording file for min version or run live test.
image


bebound (approved):

yonzhan:
CI

PR #5891, opened 2023-02-17
Fix app create help text - it wraps incorrectly

duglin:
Fixes: https://github.com/microsoft/azure-container-apps/issues/496 by splitting one help text option item into two.

Even with this I think the help text probably still starts too far to the right but with this PR it's at least better than today (21 chars wide).

---

### Related command
az conainerapp create

### General Guidelines


yonzhan:
containerapp

duglin:
@anthonychu @yonzhan any chance of this one getting in?  No idea why this is taking so long

src/containerapp/azext_containerapp/_params.py
```diff
@@ -85,7 +85,8 @@ def load_arguments(self, _):
         c.argument('max_replicas', type=int, help="The maximum number of replicas.")
         c.argument('scale_rule_name', options_list=['--scale-rule-name', '--srn'], help="The name of the scale rule.")
         c.argument('scale_rule_type', options_list=['--scale-rule-type', '--srt'], help="The type of the scale rule. Default: http.")
-        c.argument('scale_rule_http_concurrency', type=int, options_list=['--scale-rule-http-concurrency', '--srhc', '--srtc', '--scale-rule-tcp-concurrency'], help="The maximum number of concurrent requests before scale out. Only supported for http and tcp scale rules.")
+        c.argument('scale_rule_http_concurrency', type=int, options_list=['--scale-rule-http-concurrency', '--srhc'], help="The maximum number of concurrent http requests before scale out.")
+        c.argument('scale_rule_tcp_concurrency', type=int, options_list=['--scale-rule-tcp-concurrency', '--srtc'], help="The maximum number of concurrent tcp requests before scale out.")
```
zhoxing-ms, line 89: The newly added parameter `scale_rule_tcp_concurrency` should also have logic similar to `scale_rule_http_concurrency` in the custom method

duglin:
@yonzhan can you help me see what the errors are? Clicking on the "Details" doesn't show me anything interesting.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -516,6 +517,9 @@ def create_containerapp(cmd,
         if scale_rule_http_concurrency:
             if scale_rule_type in ('http', 'tcp'):
                 curr_metadata["concurrentRequests"] = str(scale_rule_http_concurrency)
+        if scale_rule_tcp_concurrency:
+            if scale_rule_type in ('http', 'tcp'):
+                curr_metadata["concurrentRequests"] = str(scale_rule_tcp_concurrency)
```
zhoxing-ms, line 522: 1. Why can parameter `--scale-rule-tcp-concurrency` be passed in as `scale_rule_type==http`?
2. When both `--scale-rule-tcp-concurrency` and `--scale-rule-http-concurrency` parameters are passed in simultaneously, `--scale-rule-tcp-concurrency` will overwrite `--scale-rule-http-concurrency`. Is this expected?

zhoxing-ms: If `scale_rule_http_concurrency` and `scale_rule_tcp_concurrency` are essentially the same property on server side, it is recommended to use a more general new parameter `--scale-rule-concurrency` and deprecate the original parameter `--scale-rule-http-concurrency`. Otherwise, using two different parameters to set the same property can easily confuse customers

bri4tub: I'm going to close this issue and someone else can open up a new one to deal with it.
I was hoping it was a simple text change but clearly it's not and it requires someone who understand the code to make the change. @anthonychu maybe you can find someone to work on this because right now the help text is totally unusable for someone (like me) who has an 80 column window.

bri4tub:
Apparently I can't close the PR. Can one of you do it for me?

duglin:
Sorry - I was logged in with the wrong ID before. I'll close it now

PR #5925, opened 2023-02-24
{CI} Optimize the extension test: Use `azdev test` for testing, consistent with the main repo

wangzelin007:
The old test uses python unittest, and the test log is as follows:
image
The new test uses azdev test, and the test log is as follows:
image


bebound (approved):

yonzhan:
CI

scripts/ci/test_source.py
```diff
@@ -62,38 +61,26 @@
                f'ALL_TESTS: {ALL_TESTS}.')
 
 
-class TestExtensionSourceMeta(type):
```
jiasli, line 65: The intention of this metaclass is to add test methods to `TestExtensionSource` during the initialization of `TestExtensionSource`.

To verify that, we can add `print(TestExtensionSource.__dict__.keys())` to L97 and check the output:

```
dict_keys(['__module__', 'setUp', 'tearDown', 'test_azext_account', 'test_azext_acrtransfer', 'test_azext_ad', 'test_azext_adp', 'test_azext_aem', 'test_azext_ai_examples', 'test_azext_aks_preview', 'test_azext_alertsmanagement', 
```

It generates a bunch of methods starting with `test_` for each extension so that `unittest` can run them one by one.

Question is: what is the point of it, when it can simply be replaced by a `for` loop and invoke `unittest` or `pytest` on one extension in each iteration?

scripts/ci/test_source.py
```diff
@@ -62,38 +61,26 @@
...
-                env['AZURE_CORE_USE_COMMAND_INDEX'] = 'false'
-                check_call(unittest_args, env=env)
-            return test
+    def run_command(self, cmd, check_return_code=False, cwd=None):
+        logger.info(f'cmd: {cmd}')
+        out = run(cmd, check=True, cwd=cwd)
+        if check_return_code and out.returncode:
+            raise RuntimeError(f"{cmd} failed")
 
+    def test_extension(self):
```
jiasli, line 72: This method can automatically be run by `unittest.main()` because it has `test_` prefix. How about we refactor this whole `scripts/ci/test_source.py` and get rid of `unittest`?

wangzelin007: In the future, we need to find a way to manage all the test scripts under the scripts directory, which is too messy now.

PR #5926, opened 2023-02-24
[AKS] Allow updating the ssh key value if cluster was created without ssh key

bingosummer:
If the cluster was created with `az aks create --name $CLUSTER_NAME --resource-group $RG_NAME --no-ssh-key`, this PR allows the customer to update ssh public key with `az aks update --name $CLUSTER_NAME --resource-group $RG_NAME --ssh-key-value $YOUR_SSH_KEY_VAULE`.

### DIFF
--- a/src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
+++ b/src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
@@ -2793,14 +2793,24 @@ class AKSPreviewManagedClusterUpdateDecorator(AKSManagedClusterUpdateDecorator):
 
         if ssh_key_value:
             if mc.linux_profile is None:
-                raise InvalidArgumentValueError("Updating the cluster with no ssh key set at creation is not supported")
-            mc.linux_profile.ssh = self.models.ContainerServiceSshConfiguration(
-                public_keys=[
-                    self.models.ContainerServiceSshPublicKey(
-                        key_data=ssh_key_value
+                mc.linux_profile = self.models.ContainerServiceLinuxProfile(
+                    admin_username="azureuser",
+                    ssh=self.models.ContainerServiceSshConfiguration(
+                        public_keys=[
+                            self.models.ContainerServiceSshPublicKey(
+                                key_data=ssh_key_value
+                            )
+                        ]
                     )
-                ]
-            )
+                )
+            else:
+                mc.linux_profile.ssh = self.models.ContainerServiceSshConfiguration(
+                    public_keys=[
+                        self.models.ContainerServiceSshPublicKey(
+                            key_data=ssh_key_value
+                        )
+                    ]
+                )
         return mc
 
     def update_node_resource_group_profile(self, mc: ManagedCluster) -> ManagedCluster:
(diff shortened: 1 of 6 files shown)


yonzhan:
AKS

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2793,14 +2793,24 @@ def update_linux_profile(self, mc: ManagedCluster) -> ManagedCluster:
 
         if ssh_key_value:
             if mc.linux_profile is None:
-                raise InvalidArgumentValueError("Updating the cluster with no ssh key set at creation is not supported")
-            mc.linux_profile.ssh = self.models.ContainerServiceSshConfiguration(
-                public_keys=[
-                    self.models.ContainerServiceSshPublicKey(
-                        key_data=ssh_key_value
+                mc.linux_profile = self.models.ContainerServiceLinuxProfile(
```
FumingZhang, line 2796: Will rp handle the backfill if admin_username not set previsouly? If so, may skip the backfill here

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2793,14 +2793,24 @@ def update_linux_profile(self, mc: ManagedCluster) -> ManagedCluster:
...
+                            self.models.ContainerServiceSshPublicKey(
+                                key_data=ssh_key_value
+                            )
+                        ]
                     )
-                ]
-            )
+                )
+            else:
+                mc.linux_profile.ssh = self.models.ContainerServiceSshConfiguration(
```
FumingZhang, line 2807: still need to check if .ssh is None or not?

FumingZhang (commented):
Better add a new live test case (or update existing one) to test the change

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
bingosummer: If I don't set "admin_username" here, I'll get an error like:

```
  File "/home/binxi/.azure/cliextensions/aks-preview/azext_aks_preview/managed_cluster_decorator.py", line 2796, in update_linux_profile
    mc.linux_profile = self.models.ContainerServiceLinuxProfile(
TypeError: ContainerServiceLinuxProfile.__init__() missing 1 required keyword-only argument: 'admin_username'
```

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
bingosummer: no matter it's none or not, we need to update `.ssh`.

bingosummer:
> Better add a new live test case (or update existing one) to test the change

added.

bingosummer:
Queued a [live test pipeline](https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=69211902&view=results) to verify the change

FumingZhang (approved):
LGTM

zhoxing-ms (approved):

PR #5953, opened 2023-03-07
[Azure Spring Apps] Add StandardGen2 tier

ShichaoQiu:
---

### General Guidelines

For new extensions:

### About Extension Publish


ShichaoQiu:
Hi @zhoxing-ms , would you pls help to review this PR? We plan to release this PR on 3/15, pls **don't merge** this PR before that.

src/spring/azext_spring/app.py
```diff
@@ -78,6 +87,17 @@ def app_create(cmd, client, resource_group, service, name,
     logger.warning(LOG_RUNNING_PROMPT)
     _ensure_app_not_exist(client, resource_group, service, name)
     sku = get_spring_sku(client, resource_group, service)
+
+    if sku.tier.upper() == 'STANDARDGEN2':
+        if cpu is None and memory is None:
+            cpu = '500m'
+            memory = '1Gi'
```
zhoxing-ms, line 94: Why isn't this logic like this?
```
       if cpu is None:
            cpu = '500m'
       if memory is None:
            memory = '1Gi'
```

src/spring/azext_spring/app.py
```diff
@@ -560,5 +607,10 @@ def _fulfill_deployment_creation_options(skip_clone_settings, client, resource_g
             options.update(deployment_source_options_from_resource(active_deployment))
     if not options.get('sku', None):
         options['sku'] = get_spring_sku(client, resource_group, service)
+
+    if options['sku'].tier.upper() == 'STANDARDGEN2' and skip_clone_settings and kwargs['cpu'] is None and kwargs['memory'] is None:
```
zhoxing-ms, line 611: Could you please split this complex judgment statement into multiple variables with better readability?

src/spring/azext_spring/_marketplace.py
```diff
@@ -14,8 +14,8 @@
 def _spring_list_marketplace_plan(cmd, client):
     # return get_mgmt_service_client(cli_ctx, AppPlatformManagementClient_20220501preview)
     from azure.cli.core.commands.client_factory import get_mgmt_service_client
-    from .vendored_sdks.marketplace.v2022_11_01_preview import MarketplaceRPService
-    from .vendored_sdks.marketplace.v2022_11_01_preview.models import Offer
+    from .vendored_sdks.marketplace.v2023_01_01_preview import MarketplaceRPService
+    from .vendored_sdks.marketplace.v2023_01_01_preview.models import Offer
```
zhoxing-ms, line 18: I have a small suggestion: if the SDK path we import does not contain the api-version part, then when upgrading the SDK version in the future, we only need to replace the SDK folder, not all the import paths.

src/spring/azext_spring/app.py
ShichaoQiu: Good question, actually, the background is that there is a hard limit for new tier - StandardGen2 that the value of Memory must be double value of CPU, it means that, for example, if the cpu is 500m (0.5 cpu), the memory must be 0.5 * 2 = 1 Gi. 

The logic here is that if customer don't input any value for cpu and memory, we will use the default value (500m, 1Gi)，but if they input any value of CPU or Memory, we won't do some special logic to handle the other one, which will be handled by our backend service.

src/spring/azext_spring/app.py
ShichaoQiu: I have added a comment to explain the logic here

src/spring/azext_spring/_marketplace.py
ShichaoQiu: Very very good idea. As you can see, there are multi api version in our CLI extension, we plan to only use the latest API version in the future. There are 2 steps here.
1. Make sure we only use 1 API version in the code
2. Remove unused the API versions and remove the version info from the SDK file path.

We will create another PR to do the step 2.

zhoxing-ms: OK, sounds good!

src/spring/azext_spring/app.py
zhoxing-ms: Got it, thanks~

zhoxing-ms (approved):

PR #5987, opened 2023-03-14
[confcom] bugfixes for using sha values and tar files

SethHollandsworth:
---

### Related command
confcom

### General Guidelines

### Description
Bugfix to make sha hashes work better. Fixed a bug where it would error if you specified a hash but not a tag name. Also moved the error checking to make sure you're not trying to use a hash on a tar file.

Right now, caching only works if you're doing multiple of the same image within the same container group. This PR will make it so it's template-wide which is a more reasonable and helpful mechanism.


hgarvison (approved):

stevendongatmsft (approved):

yonzhan:
confcom

PR #5989, opened 2023-03-14
{AKS} Update cli tests disable upgrade node image related tests

yewmsft:
---
Current cli tests which involves node image upgrades are not meaningful. These tests is relying on a buggish behavior in upgradeNodeImageVersion API to function. If we are upgrading the node image of an agent pool which is already on the latest node image version, upgradeNodeImageVersion API will exit silently after very long validation on the async side.

Existing nodeimageupgrade related cli tests are all trying to upgrade the node image of clusters which is already on the latest image version. And it only checks if the agent pool goes into UpgradingNodeImageVersion provisioningState. It never verifies whether the upgrade succeeded or not, which make these tests meaningless. 

After changing upgradeNodeImageVersion API to return 304 when agent pool is already on the latest version. These tests will stop function. Temporarily disable node image upgrade tests to unblock RP release. We will provide a short term solution in RP to re-enable these tests first. And after we have the API to create clusters with an older image, we will redo these tests to make real node image upgrades.

### Related command

N/A, tests only

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
+++ b/src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
@@ -1992,7 +1992,7 @@ class AzureKubernetesServiceScenarioTest(ScenarioTest):
                                                '-n {nodepool_name2} ' \
                                                '--node-image-only ' \
                                                '--snapshot-id {snapshot_resource_id} -o json'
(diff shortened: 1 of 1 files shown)


yonzhan:
AKS

zhoxing-ms (approved):

PR #6053, opened 2023-03-22
{AKS} `az aks update`: Remove the error that could not create a role assignment error when the role assignment already exists

FumingZhang:
---

> In version 2.46.0 of azure-cli, due to the [update of the API](https://github.com/Azure/azure-cli/pull/25452) used by the role assignment-related operations, when the role assignment already exists, the returned error message is changed, and the code that previously performed this type of error checking can no longer handle such error, so that the command finally raise an error similar to `Cannot create a role assignment for xxx...`.
> 
> The error message of adding exiting role assignment is (sending request with API version `2022-04-01`)
> ```
> (RoleAssignmentExists) The role assignment already exists.
> Code: RoleAssignmentExists
> Message: The role assignment already exists.
> ```
> 
> Previously the error message is (sending request with API version `2020-04-01-preview`)
> ```
> The role assignment already exists.
> ```
> 
> - A simple and **recommended** workaround is to use azure-cli of version 2.45.0.
> - A complex workaround is to run the command (`az aks update`) with an extra option `--verbose` and check if the error is caused by _role assignment already exists_, if so, please ignore the error as cluster should have been updated successfully.
> 
> This issue should only occur in version 2.46.0, and the fix will be [released in 2.47.0](https://github.com/Azure/azure-cli/milestone/128) (available since 04/04). Having aks-preview installed or not won't mitigate the problem.

Sync change from PR [#25908](https://github.com/Azure/azure-cli/pull/25908).

### Related command

`az aks update`

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/azext_aks_preview/_roleassignments.py
+++ b/src/aks-preview/azext_aks_preview/_roleassignments.py
@@ -5,11 +5,16 @@
 
 import time
 import uuid
+
+from azext_aks_preview._client_factory import (
+    get_auth_management_client,
+    get_graph_rbac_management_client,
+)
+from azure.core.exceptions import HttpResponseError, ResourceExistsError
 from azure.graphrbac.models import GetObjectsParameters
 from knack.log import get_logger
 from knack.util import CLIError
 from msrestazure.azure_exceptions import CloudError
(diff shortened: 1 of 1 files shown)


yonzhan:
AKS

zhoxing-ms (approved):

PR #6080, opened 2023-03-27
[Draft] Adding the new flag for enabling oidc issuer for workload identity

deeksha345:
Adding the new flag for enabling oidc issuer creation for workload identity on connected cluster

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/connectedk8s/azext_connectedk8s/_help.py
+++ b/src/connectedk8s/azext_connectedk8s/_help.py
@@ -30,6 +30,8 @@ helps['connectedk8s connect'] = """
       text: az connectedk8s connect -g resourceGroupName -n connectedClusterName --enable-private-link true --private-link-scope-resource-id pls/resource/arm/id
     - name: Onboard a connected kubernetes cluster with custom onboarding timeout.
       text: az connectedk8s connect -g resourceGroupName -n connectedClusterName --onboarding-timeout 600
+    - name: Onboard a connected kubernetes cluster with oidc issuer creation for workload identity enabled.
+      text: az connectedk8s connect -g resourceGroupName -n connectedClusterName --enable-oidc-issuer
 
 """
 
@@ -43,6 +45,8 @@ helps['connectedk8s update'] = """
       text: az connectedk8s update -g resourceGroupName -n connectedClusterName --disable-proxy
     - name: Disable auto-upgrade of agents
       text: az connectedk8s update -g resourceGroupName -n connectedClusterName --auto-upgrade false
+    - name: Onboard a connected kubernetes cluster with oidc issuer creation for workload identity enabled.
+      text: az connectedk8s connect -g resourceGroupName -n connectedClusterName --enable-oidc-issuer
 """
 
 helps['connectedk8s upgrade'] = """
--- a/src/connectedk8s/azext_connectedk8s/_params.py
+++ b/src/connectedk8s/azext_connectedk8s/_params.py
@@ -49,6 +49,7 @@ def load_arguments(self, _):
         c.argument('no_wait', options_list=['--no-wait'], arg_group='Timeout', help="Do not wait for the long-running operation to finish.")
         c.argument('correlation_id', options_list=['--correlation-id'], help='A guid that is used to internally track the source of cluster onboarding. Please do not modify it unless advised', validator=override_client_request_id_header)
         c.argument('container_log_path', help='Override the default container log path to enable fluent-bit logging')
+        c.argument('enable_oidc_issuer', options_list=['--enable-oidc-issuer'], help="Enable creation of OIDC issuer endpoint used for workload identity")
         c.argument('yes', options_list=['--yes', '-y'], help='Do not prompt for confirmation.', action='store_true')
 
     with self.argument_context('connectedk8s update') as c:
@@ -66,6 +67,7 @@ def load_arguments(self, _):
         c.argument('disable_proxy', options_list=['--disable-proxy'], arg_group='Proxy', action='store_true', help='Disables proxy settings for agents')
         c.argument('auto_upgrade', options_list=['--auto-upgrade'], help='Flag to enable/disable auto upgrade of arc agents. By default, auto upgrade of agents is enabled.', arg_type=get_enum_type(["true", "false"]))
         c.argument('container_log_path', help='Override the default container log path to enable fluent-bit logging')
+        c.argument('enable_oidc_issuer', options_list=['--enable-oidc-issuer'], help="Enable creation of OIDC issuer endpoint used for workload identity")
         c.argument('yes', options_list=['--yes', '-y'], help='Do not prompt for confirmation.', action='store_true')
 
     with self.argument_context('connectedk8s upgrade') as c:
--- a/src/connectedk8s/azext_connectedk8s/_utils.py
+++ b/src/connectedk8s/azext_connectedk8s/_utils.py
@@ -526,7 +526,7 @@ def helm_install_release(chart_path, subscription_id, kubernetes_distro, kuberne
                          location, onboarding_tenant_id, http_proxy, https_proxy, no_proxy, proxy_cert, private_key_pem,
                          kube_config, kube_context, no_wait, values_file_provided, values_file, cloud_name, disable_auto_upgrade,
                          enable_custom_locations, custom_locations_oid, helm_client_location, enable_private_link, onboarding_timeout="600",
-                         container_log_path=None):
+                         container_log_path=None, enable_oidc_issuer=False):
     cmd_helm_install = [helm_client_location, "upgrade", "--install", "azure-arc", chart_path,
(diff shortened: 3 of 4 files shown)


anagg929 (approved):

yonzhan:
connectedk8s

zhoxing-ms:
@deeksha345 Could you please resolve the linter issues?

src/connectedk8s/azext_connectedk8s/_help.py
```diff
@@ -43,6 +45,8 @@
       text: az connectedk8s update -g resourceGroupName -n connectedClusterName --disable-proxy
     - name: Disable auto-upgrade of agents
       text: az connectedk8s update -g resourceGroupName -n connectedClusterName --auto-upgrade false
+    - name: Onboard a connected kubernetes cluster with oidc issuer creation for workload identity enabled.
```
akashkeshari, line 48: Update instead of Onboard

src/connectedk8s/azext_connectedk8s/_help.py
```diff
@@ -43,6 +45,8 @@
...
+      text: az connectedk8s connect -g resourceGroupName -n connectedClusterName --enable-oidc-issuer
```
akashkeshari, line 49: update instead of connect

src/connectedk8s/azext_connectedk8s/_params.py
```diff
@@ -49,6 +49,7 @@ def load_arguments(self, _):
         c.argument('no_wait', options_list=['--no-wait'], arg_group='Timeout', help="Do not wait for the long-running operation to finish.")
         c.argument('correlation_id', options_list=['--correlation-id'], help='A guid that is used to internally track the source of cluster onboarding. Please do not modify it unless advised', validator=override_client_request_id_header)
         c.argument('container_log_path', help='Override the default container log path to enable fluent-bit logging')
+        c.argument('enable_oidc_issuer', options_list=['--enable-oidc-issuer'], help="Enable creation of OIDC issuer endpoint used for workload identity")
```
akashkeshari, line 52: options_list array is not required here

src/connectedk8s/azext_connectedk8s/_params.py
```diff
@@ -66,6 +67,7 @@ def load_arguments(self, _):
         c.argument('disable_proxy', options_list=['--disable-proxy'], arg_group='Proxy', action='store_true', help='Disables proxy settings for agents')
         c.argument('auto_upgrade', options_list=['--auto-upgrade'], help='Flag to enable/disable auto upgrade of arc agents. By default, auto upgrade of agents is enabled.', arg_type=get_enum_type(["true", "false"]))
         c.argument('container_log_path', help='Override the default container log path to enable fluent-bit logging')
+        c.argument('enable_oidc_issuer', options_list=['--enable-oidc-issuer'], help="Enable creation of OIDC issuer endpoint used for workload identity")
```
akashkeshari, line 70: aren't we supposed to keep this in enable/disable features cmdlet? how will the customer disable this feature?

src/connectedk8s/azext_connectedk8s/_params.py
akashkeshari: should be a boolean parameter. set arg_type to get_three_state_flag().

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1101,6 +1102,8 @@ def update_connected_cluster(cmd, client, resource_group_name, cluster_name, htt
         cmd_helm_upgrade.extend(["--set", "global.isCustomCert={}".format(True)])
     if container_log_path is not None:
         cmd_helm_upgrade.extend(["--set", "systemDefaultValues.fluent-bit.containerLogPath={}".format(container_log_path)])
+    if enable_oidc_issuer:
```
akashkeshari, line 1105: how are we supporting disabling of this feature?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -949,7 +949,8 @@ def update_connected_cluster_internal(client, resource_group_name, cluster_name,
 
 def update_connected_cluster(cmd, client, resource_group_name, cluster_name, https_proxy="", http_proxy="", no_proxy="", proxy_cert="",
                              disable_proxy=False, kube_config=None, kube_context=None, auto_upgrade=None, tags=None,
-                             distribution=None, distribution_version=None, azure_hybrid_benefit=None, yes=False, container_log_path=None):
+                             distribution=None, distribution_version=None, azure_hybrid_benefit=None, yes=False, container_log_path=None,
+                             enable_oidc_issuer=False):
```
akashkeshari, line 953: For brownfield scenario, there could be some customers who have agent auto update disabled and might not have latest helm chart changes for oidcenabled flag. we should add this check and give warning to those customers to upgrade their agent version to latest if they want to enable this feature

akashkeshari (changes requested):
left comments

src/connectedk8s/azext_connectedk8s/_params.py
anagg929: by explicitly giving --enable-oidc-issuer false in the update

src/connectedk8s/azext_connectedk8s/custom.py
deeksha345: We do not currently have plans to support disabling of this feature as far as I am aware.

src/connectedk8s/azext_connectedk8s/_params.py
```diff
@@ -49,6 +49,7 @@ def load_arguments(self, _):
         c.argument('no_wait', options_list=['--no-wait'], arg_group='Timeout', help="Do not wait for the long-running operation to finish.")
         c.argument('correlation_id', options_list=['--correlation-id'], help='A guid that is used to internally track the source of cluster onboarding. Please do not modify it unless advised', validator=override_client_request_id_header)
         c.argument('container_log_path', help='Override the default container log path to enable fluent-bit logging')
+        c.argument('enable_oidc_issuer', arg_type=get_three_state_flag(), help="Enable creation of OIDC issuer endpoint used for workload identity")
```
sirireddy12, line 52: we should add is_preview=True similar to how 'enable_private_link' argument is marked

src/connectedk8s/azext_connectedk8s/_utils.py
```diff
@@ -567,6 +567,8 @@ def helm_install_release(chart_path, subscription_id, kubernetes_distro, kuberne
         cmd_helm_install.extend(["--set", "global.isProxyEnabled={}".format(True)])
     if container_log_path is not None:
         cmd_helm_install.extend(["--set", "systemDefaultValues.fluent-bit.containerLogPath={}".format(container_log_path)])
+    if enable_oidc_issuer:
```
anagg929, line 570: if it is false,, can you explicitly set to false for the update to false scenario

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1101,6 +1102,11 @@ def update_connected_cluster(cmd, client, resource_group_name, cluster_name, htt
         cmd_helm_upgrade.extend(["--set", "global.isCustomCert={}".format(True)])
     if container_log_path is not None:
         cmd_helm_upgrade.extend(["--set", "systemDefaultValues.fluent-bit.containerLogPath={}".format(container_log_path)])
+    if enable_oidc_issuer:
+        if not auto_upgrade:
+            logger.warning("Please enable auto_upgrade to upgrade to latest agent version in order to enable this oidc issuer feature")
+        else:
+            cmd_helm_upgrade.extend(["--set", "systemDefaultValues.signingkeycontroller.oidcenabled={}".format(True)])
```
akashkeshari, line 1109: we also need to handle the scenario of disabling the oidc issuer feature on day N. so, we should set 'systemDefaultValues.signingkeycontroller.oidcenabled' to false as well when the flag enable_oidc_issuer is passed as False.
If passed as None, then it should be kept as is

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1101,6 +1102,11 @@ def update_connected_cluster(cmd, client, resource_group_name, cluster_name, htt
         cmd_helm_upgrade.extend(["--set", "global.isCustomCert={}".format(True)])
     if container_log_path is not None:
         cmd_helm_upgrade.extend(["--set", "systemDefaultValues.fluent-bit.containerLogPath={}".format(container_log_path)])
+    if enable_oidc_issuer:
+        if not auto_upgrade:
```
akashkeshari, line 1106: this is not the right way to check if a cluster has auto_upgrade enabled. you would have to read this from the configmap and accordingly raise this warning

PR #6144, opened 2023-03-29
[confcom] changing version number naming scheme and bugfix for 32bit python

SethHollandsworth:
---

### Related command

az confcom

### General Guidelines


yonzhan:
confcom

yonzhan:
confcom

src/confcom/samples/sample-policy-output.rego
```diff
@@ -3,190 +3,200 @@ package policy
...
-]
+api_version := "0.10.0"
+
+framework_version := "0.2.3"
+
+fragments := [{
+	"feed": "mcr.microsoft.com/aci/aci-cc-infra-fragment",
+	"includes": ["containers"],
+	"issuer": "did:x509:0:sha256:I__iuL25oXEVFdTP_aBLx_eT1RPHbCQ_ECBQfYZpt9s::eku:1.3.6.1.4.1.311.76.59.1.3",
+	"minimum_svn": 1,
```
matajoh, line 14: The value should be an integer string, not a raw integer

src/confcom/azext_confcom/config.py
```diff
@@ -132,5 +132,9 @@
 SIDECAR_REGO_FILE = "./data/sidecar_rego_policy.txt"
 SIDECAR_REGO_FILE_PATH = f"{script_directory}/{SIDECAR_REGO_FILE}"
 SIDECAR_REGO_POLICY = os_util.load_str_from_file(SIDECAR_REGO_FILE_PATH)
+# api svn file
+SVN_API_FILE = "./data/svn_api"
```
matajoh, line 136: Not quite sure I understand all the context for this file, but this seems confusingly named. I believe it should be `version_api`?

src/confcom/setup.py
```diff
@@ -47,17 +47,44 @@
...
+            # save the file to the bin folder
+            with open(bin_folder + asset["name"], "wb") as f:
+                f.write(r.content)
+    # break out of the loop if we have both files
+    if bin_flag and exe_flag:
+        # get the download url for the dmverity-vhd file
+        api_svn_url = release["html_url"]
+        # update the url to get framework svn file
+        api_svn_url = api_svn_url.replace("releases/tag", "raw")
+        api_svn_url += "/pkg/securitypolicy/svn_api"
```
matajoh, line 81: This file will be called `version_api` now, but should be very stable. You can safely hardcode it in instead of pulling it down like this.

src/confcom/azext_confcom/custom.py
```diff
@@ -40,27 +40,26 @@ def acipolicygen_confcom(
...
         )
-        sys.exit(1)
+    elif save_to_file and arm_template and not (print_policy_to_terminal or outraw or outraw_pretty_print):
+        error_out("Must print policy to terminal when saving to file")
 
     if print_existing_policy:
-        if not arm_template:
-            logger.error("Can only print existing policy from ARM Template")
-            sys.exit(1)
-        else:
```
kairu-ms, line 61: Why removed this check?

src/confcom/azext_confcom/rootfs_proxy.py
```diff
@@ -9,29 +9,62 @@
...
+    @staticmethod
+    def download_binaries():
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+
+        bin_folder = os.path.join(dir_path, "bin")
+        if not os.path.exists(bin_folder):
+            os.makedirs(bin_folder)
+
+        # get the most recent release artifacts from github
+        r = requests.get("https://api.github.com/repos/microsoft/hcsshim/releases")
```
kairu-ms, line 33: That request may fail when the user cannot access that url. Please add additional logic to handle that.

src/confcom/setup.py
```diff
@@ -39,25 +35,9 @@
...
-    r = requests.get("https://github.com/microsoft/hcsshim/releases/download/v0.10.0-rc.6/dmverity-vhd.exe")
-    with open(exe_path, "wb") as f:
-        f.write(r.content)
-
-bin_path = dir_path + "/bin/dmverity-vhd"
-if not os.path.exists(bin_path):
-    r = requests.get("https://github.com/microsoft/hcsshim/releases/download/v0.10.0-rc.6/dmverity-vhd")
-    with open(bin_path, "wb") as f:
-        f.write(r.content)
+SecurityPolicyProxy.download_binaries()
```
kairu-ms, line 40: Do you need to catch and handle the download error in setup.py

src/confcom/azext_confcom/custom.py
SethHollandsworth: We were hitting the "can't have more than x statements in a function" style check so I moved that error checking inside of the `print_existing_policy_from_arm_template` function here: https://github.com/SethHollandsworth/azure-cli-extensions/blob/834fba5034189dea4df21e499997c872e24806f1/src/confcom/azext_confcom/template_util.py#L817 .
So it's still in there, just a different place

src/confcom/azext_confcom/rootfs_proxy.py
SethHollandsworth: As mentioned in another comment, this function is only called in setup.py. Would it ever get called on a client machine?

src/confcom/setup.py
kairu-ms: That's correct. Setup.py runs only for building the .whl file.

kairu-ms (approved):

PR #6219, opened 2023-04-19
Remove connectedk8s version 1.3.0 from index.sjon

rohan-dassani:
This is done because connectedk8s 1.3.0 had some imports which can only be done using azure-cli >=2.30.0 and the minimum cli version in 1.3.0 is 2.16.0 causing no module found errors for users.

connectedk8s 1.3.1 is same as 1.3.0 just the minimum cli version is bumped up now , so users with older cli version will be able to use til 1.2.11 , we are currently on 1.3.16


yonzhan:

zhoxing-ms (approved):

PR #6313, opened 2023-05-23
Use provided target port in buildpack for 'containerapp up'

cormacpayne:
The current iteration of source-to-cloud using buildpacks in `containerapp up` is always listening to port 8080, independent of what the user is providing for the `--target-port` value.

This PR will propagate the `--target-port` value provided by the user to the buildpack, and ultimately the final run image, via the `PORT` environment variable that the buildpack is currently checking the value for (_i.e._, if `PORT` is set, used that as the port value in the final run image, otherwise use 8080).

EDIT: In addition to the above changes, this PR also introduces the following changes:
- In the `containerapp up` flow, when building a runnable application image using buildpacks, ensure that the user is authenticated against the created/provided ACR instance before pushing the produced image
- Add initial live-only tests for the `containerapp up` scenarios modified in this PR
  - _Note_: when an ACR Build/Task was executed by `containerapp up` (either when the user provided source with a Dockerfile or source without a Dockerfile that wasn't .NET 7), when using the Storage SDK to upload the source to a blob, there was a `utf8` error that would pop up with no way to resolve it. Looped in the Storage SDK for Python folks and their remediation was to update the Storage SDK used from V1 to V2, and doing so resolved the issue for one of the tests, however, the `upload_source_code` function that this logic is housed in was directly copied over from the `acr` module, and there is a flow in `containerapp up` that creates and runs an ACR Task which uses _their_ implementation of `upload_source_code` that still uses the old Storage SDK. I added the live-only test for the ACR Task scenario, but skipped it until we're able to update the `acr` module to use the new version of the SDK.

---

### Related command

`az containerapp up`

### General Guidelines

For new extensions:

~- [ ] My extension description/summary conforms to the [Extension Summary Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/extensions/extension_summary_guidelines.md).~

### About Extension Publish


yonzhan:

daniv-msft (approved):

zhoxing-ms:
Could you please add a test case for this PR change?

cormacpayne:
@zhoxing-ms -- I just force pushed a commit with the previous target port change, along with some other changes I outlined in the "EDIT" of the PR description. These changes include live-only tests for this scenario, but I added context around why Storage SDK changes were needed within the `containerapp` module and future changes we'd need in the `acr` module that `containerapp` depends on. Please let me know if there are any additional questions.

zhoxing-ms:
@cormacpayne Please add the PR description in the `HISTORY.rst`. Moreover, if you want release a new extension for this PR, please also upgrade the version defined in the `setup.py`.

cormacpayne:
@zhoxing-ms -- I pushed a new commit to update the `HISTORY.rst` for `containerapp`, but before we merge this PR, I'm waiting to get a review and/or approval from additional Container Apps folks to ensure these changes are OK. I'll follow up when we should be good to merge this PR. Thanks!

src/containerapp/azext_containerapp/_archive_utils.py
```diff
@@ -54,16 +53,10 @@ def upload_source_code(cmd, client,
...
-    BlockBlobService = get_sdk(cmd.cli_ctx, ResourceType.DATA_STORAGE, 'blob#BlockBlobService')
-    BlockBlobService(account_name=account_name,
-                     sas_token=sas_token,
-                     endpoint_suffix=endpoint_suffix,
-                     # Increase socket timeout from default of 20s for clients with slow network connection.
-                     socket_timeout=300).create_blob_from_path(
-                         container_name=container_name,
-                         blob_name=blob_name,
-                         file_path=tar_file_path)
+    BlobClient = get_sdk(cmd.cli_ctx, ResourceType.DATA_STORAGE_BLOB, '_blob_client#BlobClient')
```
daniv-msft, line 56: Could you please clarify where we're actually moving from V1 to V2? Here the code looks like what we need to update after we moved to the newer SDK, but I cannot find where the actual "Storage SDK update" happened. What am I missing?

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -470,6 +503,10 @@ def build_container_from_source_with_buildpack(self, image_name, source):  # pyl
             logger.debug(f"Determined the run image to use as {buildpack_run_image}.")
             command.extend(['--run-image', buildpack_run_image])
 
+        # If the user specifies a target port, pass it to the buildpack
+        if self.target_port:
+            command.extend(['--env', f"PORT={self.target_port}"])
```
daniv-msft, line 508: Just to validate: from your comment in the description of the PR it sound like the PORT env variable value is basically intercepted by our logic in the Buildpack and then used by us to set the port to expose on the final image. Is this correct?

src/containerapp/azext_containerapp/tests/latest/test_containerapp_up.py
```diff
@@ -37,3 +38,65 @@ def test_containerapp_up_image_e2e(self, resource_group):
         self.assertTrue(resp.ok)
 
         self.cmd(f"containerapp up --image {image} --environment {env_name} -g {resource_group} -n {app_name} -l {TEST_LOCATION.upper()}")
+
+    @ResourceGroupPreparer(location="eastus2")
+    def test_containerapp_up_source_with_buildpack_e2e(self, resource_group):
+        self.cmd('configure --defaults location={}'.format(TEST_LOCATION))
+
+        source_path = os.path.join(TEST_DIR, os.path.join("data", "source_built_using_buildpack"))
```
daniv-msft, line 46: Please consider creating a helper function to share some of the logic between the test cases, as it looks like a good part is duplicated.

src/containerapp/azext_containerapp/tests/latest/test_containerapp_up.py
```diff
@@ -37,3 +38,65 @@ def test_containerapp_up_image_e2e(self, resource_group):
...
+        target_port = '8080'
+        self.cmd('containerapp up --source "{}" --environment {} -g {} -n {} --ingress {} --target-port {}'.format(source_path, env_name, resource_group, app_name, ingress, target_port))
+
+        app = self.cmd(f"containerapp show -g {resource_group} -n {app_name}").get_output_in_json()
+        url = app["properties"]["configuration"]["ingress"]["fqdn"]
+        url = url if url.startswith("http") else f"http://{url}"
+        resp = requests.get(url)
+        self.assertTrue(resp.ok)
+
+        self.cmd('containerapp up --source "{}" --environment {} -g {} -n {} -l {} --ingress {} --target-port {}'.format(source_path, env_name, resource_group, app_name, TEST_LOCATION.upper(), ingress, target_port))
```
daniv-msft, line 60: Could you please clarify why we're not directly using a f-string rather than calling `format`?

src/containerapp/azext_containerapp/_archive_utils.py
Juliehzl: For resource type ResourceType.DATA_STORAGE_BLOB and DATA_STORAGE, they will correspond to different SDK.
https://github.com/Azure/azure-cli/blob/82358037c5af4a43c46facfcff054100a730c678/src/azure-cli-core/azure/cli/core/profiles/_shared.py#LL121-L126

cormacpayne: Yep, sorry, it's not obvious just from changing the `ResourceType` property, but the snippet that Zunli linked above is exactly what the Storage folks sent me when recommending to upgrade to their V2 SDK.

So we needed to change the `ResourceType` property so the `get_sdk` function targets the new SDK, and then we create a `BlobClient` object defined in the V2 Storage's `_blob_client.py` file.

src/containerapp/azext_containerapp/_up_utils.py
cormacpayne: That's correct -- the `--env` argument will propagate the given environment variables to the stack that the detect/build phases occur, so our .NET 7 buildpack will have access to this `PORT` value set

src/containerapp/azext_containerapp/tests/latest/test_containerapp_up.py
cormacpayne: I was following the pattern in place for the original test in this suite, but I'll update the tests to use f-strings

daniv-msft (approved):
Looks good to me! Thanks

daniv-msft (approved):
Looks good to me! Thanks

cormacpayne:
@zhoxing-ms -- We should be OK to merge this PR whenever someone from the CLI side has the opportunity to merge it.

zhoxing-ms (approved):

PR #6334, opened 2023-05-28
[Azure Load Testing] Adding Data-Plane commands for existing load extension.

Himanshu49:
We have added data plane commands for load testing service to the existing load cli extension which contains commands for control plane. 

In this implementation we will be keeping each command group and files related to it such as helps, params, custom commands, etc. in a separate directory for that particular command group itself.

The commands were not auto generated.
---

### Related command

az load test
az load test-run
each having multiple sub command group.

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

src/load/setup.py
```diff
@@ -26,7 +26,7 @@
     'License :: OSI Approved :: MIT License',
 ]
 
-DEPENDENCIES = []
+DEPENDENCIES = ["azure-core", "pyyaml", "azure-mgmt-network"]
```
kairu-ms, line 29: azure-mgmt-network is removed from azure-cli. Which function do you rely on?

kairu-ms: I think `azure-core` and `pyyaml` are added in `azure-cli`

src/load/azext_load/tests/latest/test_load_test.py
```diff
@@ -0,0 +1,500 @@
...
+            "az load test list "
+            "--load-test-resource {load_test_resource} "
+            "--resource-group {resource_group}"
+        ).get_output_in_json()
+
+        assert len(tests) > 0
+        assert "fake_test_id" not in [test["testId"] for test in tests]
+        assert self.kwargs["test_id"] in [test.get("testId") for test in tests]
+
+        delete_test(self)
```
kairu-ms, line 101: Recommend moving `delete_test(self)` in the `try... finally...`

src/load/azext_load/data_plane/load_test/help.py
```diff
@@ -0,0 +1,137 @@
...
+short-summary: Show details of test.
+long-summary: Show details of the test identified by given test-id.
+"""
+
+helps[
+    "load test update"
+] = """
+type: command
+short-summary: Update an existing load test.
+long-summary: Update an existing load test.
```
kairu-ms, line 40: `long-summary` is used to add more description for the command. If it's almost the same as `short-summary` you should remove the `long-summary`

src/load/azext_load/data_plane/utils/validators.py
```diff
@@ -0,0 +1,335 @@
...
+    allowed_metric_namespaces = utils.get_enum_values(AllowedMetricNamespaces)
+    if namespace.metric_namespace not in allowed_metric_namespaces:
+        raise InvalidArgumentValueError(
+            f"Invalid metric-namespace value: {namespace.metric_namespace}. "
+            f"Allowed values: {', '.join(allowed_metric_namespaces)}"
+        )
+
+
+def validate_dimension_filters(namespace):
+    """Extracts multiple space and comma-separated dimension filters in key1[=value1,value2] format"""
```
kairu-ms, line 313: Usually, we use `key1=value1 key1=value2` to pass multiple values.

src/load/azext_load/data_plane/utils/utils.py
```diff
@@ -0,0 +1,531 @@
...
+            file_type=file_type,
+            body=file,
+        )
+        response = (
+            upload_poller.result()
+            if wait
+            else upload_poller.polling_method().resource()
+        )
+        logger.debug(
+            "Upload result for file with --wait%s passed: %s",
```
kairu-ms, line 165: Usually we use `--no-wait` argument instead of `--wait`, because the default behavior for LRO commands is waiting.

src/load/azext_load/data_plane/utils/utils.py
```diff
@@ -0,0 +1,531 @@
...
+    return certificate
+
+
+def parse_secrets(secrets):
+    logger.debug("Parsing secrets")
+    secrets_dict = {}
+    for secret in secrets:
+        name, value = secret.get("name"), secret.get("value")
+        if name is None or value is None:
+            raise ValueError("Both name and value are required for secret")
```
kairu-ms, line 197: Please use error types in https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md#error-type

src/load/setup.py
Himanshu49: We are using azure-mgmt-network for completers of subnets.

src/load/azext_load/data_plane/utils/completers.py
```diff
@@ -0,0 +1,75 @@
...
+        return test_run_ids
+
+    return completer
+
+
+def get_subnet_resource_id_completion_list():
+    @Completer
+    def completer(cmd, prefix, namespace, **kwargs):  # pylint: disable=unused-argument
+        client = get_mgmt_service_client(
+            cmd.cli_ctx, ResourceType.MGMT_NETWORK, api_version="2015-06-15"
```
kairu-ms, line 67: You can reference this: https://github.com/Azure/azure-cli/blob/759bb130cc6736c80aab945a0902d13a75ce7b8f/src/azure-cli/azure/cli/command_modules/network/_completers.py#L12-L22

kairu-ms: The network sdk is removed from azure-cli because of its huge size.

src/load/azext_load/data_plane/utils/argtypes.py
```diff
@@ -0,0 +1,335 @@
...
+    completer=get_resource_name_completion_list("Microsoft.LoadTestService/LoadTests"),
+    help="Name or ARM resource ID of the load test resource.",
+)
+
+wait = CLIArgumentType(
+    options_list=["--wait"],
+    action="store_true",
+    default=False,
+    help="Wait for the Long-Running Operation to complete.",
+)
```
kairu-ms, line 33: We use `--no-wait` instead.

src/load/azext_load/data_plane/utils/utils.py
Himanshu49: There are few reasons to move to wait instead of no-wait.

1. no-wait's initial response is not getting returned to output CMD. 
2. Our portal behavior is similar to wait. We won't wait in create command for file upload. 
3. For test run command generally, it can be of any duration (in hours too), so if user wants to wait they should provide wait explicitly otherwise they might get stuck for hours unwillingly.

src/load/azext_load/data_plane/utils/argtypes.py
Himanshu49: There are few reasons to move to wait instead of no-wait.

1. no-wait's initial response is not getting returned to output CMD.
2. Our portal behavior is similar to wait. We won't wait in create command for file upload.
3. For test run command generally, it can be of any duration (in hours too), so if user wants to wait they should provide wait explicitly otherwise they might get stuck for hours unwillingly.

src/load/azext_load/data_plane/utils/argtypes.py
```diff
@@ -0,0 +1,335 @@
...
+
+dimension_filters = CLIArgumentType(
+    validator=validators.validate_dimension_filters,
+    options_list=["--dimension-filters"],
+    nargs="*",
+    help=(
+        "space and comma-separated dimension filters: key1[=value1,value2] [key2[=value3] ...]. "
+        "* is supported as a wildcard for both key and value. "
+        "Example: `--dimension-filters key1=value1,key2=*`, `--dimension-filters *`"
```
kairu-ms, line 333: we use space to separate by default instead of ','

src/load/azext_load/data_plane/load_test_run/custom.py
```diff
@@ -0,0 +1,549 @@
...
+    poller = client.begin_test_run(
+        test_run_id=test_run_id,
+        body=test_run_body,
+        old_test_run_id=existing_test_run_id,
+    )
+    response = poller.polling_method().resource()
+    if wait:
+        response = poller.result()
+    logger.info("Test run created with following response %s", response)
+    return response
```
kairu-ms, line 63: For LRO command, when it's no wait, we will not return the initial response.

Himanshu49: For few commands such as create test and create testrun, the command works in two parts. In both second part is a LRO so if its no wait we want to return response of first part.
For eg -> create test run will create a test run with given details and subsequently run it. test run execution is an LRO. So if a user creates a test run and does not want to wait for test execution duration (can be in hours) should we not return user the initial response of our API which in this case would be the test run object?

src/load/azext_load/data_plane/utils/validators.py
kairu-ms: The difference is that we use space to separate two equations __key1=value` `key1=value__. Please remove the second case.

src/load/setup.py
kairu-ms: Please drop the dependance on azure-mgmt-network.

src/load/azext_load/tests/latest/test_load_test.py
Himanshu49: we removed all the delete_test as code cleaner take care of that. Resolving this

src/load/setup.py
Himanshu49: Removed dependency on azure-mgmt-network

src/load/azext_load/data_plane/load_test/help.py
Himanshu49: Updated help as suggested

src/load/azext_load/data_plane/utils/validators.py
Himanshu49: We have updated the implementation

src/load/azext_load/data_plane/utils/utils.py
Himanshu49: have updated the commands, create and update test have custom no wait and all other commands follow the supports_no_wait flag as you suggested

src/load/azext_load/data_plane/utils/utils.py
Himanshu49: updated all the generic Error to specific error as mentioned in the guidelines

src/load/azext_load/data_plane/utils/argtypes.py
Himanshu49: resolved

src/load/azext_load/data_plane/utils/argtypes.py
Himanshu49: Updated as suggested

PR #6341, opened 2023-05-30
containerapp fix az containerapp create with --registry-identity system and --revision-suffix

Greedygre:
az containerapp create --registry-identity --revision-suffix

### DIFF
--- a/src/containerapp/azext_containerapp/custom.py
+++ b/src/containerapp/azext_containerapp/custom.py
@@ -648,7 +648,7 @@ def create_containerapp(cmd,
         container_def["volumeMounts"] = [volume_mount_def]
         template_def["volumes"] = [volume_def]
 
-    if revision_suffix is not None:
+    if revision_suffix is not None and not is_registry_msi_system(registry_identity):
         template_def["revisionSuffix"] = revision_suffix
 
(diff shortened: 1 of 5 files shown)


yonzhan:

src/containerapp/azext_containerapp/custom.py
```diff
@@ -648,7 +648,7 @@ def create_containerapp(cmd,
         container_def["volumeMounts"] = [volume_mount_def]
         template_def["volumes"] = [volume_def]
 
-    if revision_suffix is not None:
+    if revision_suffix is not None and not is_registry_msi_system(registry_identity):
```
zhenqxuMSFT, line 651: Will this break the scenario that only sets revision suffix?

Greedygre:
```
def is_registry_msi_system(identity):
    if identity is None:
        return False
    return identity.lower() == "system"
```

no, as shown code, for the scenario that only sets revision suffix, is_registry_msi_system return false, so the value will be set.

zhenqxuMSFT (approved):

zhoxing-ms (approved):

PR #6363, opened 2023-06-05
{interactive} Fix bugs of CTRL+C to cease the whole interactive mode

wu5bocheng:
Prevent users from exiting the entire az interactive by using Ctrl+C during command execution


yonzhan:
interactive

PR #6409, opened 2023-06-14
Fix oras target string

Cyclam:
Small fix for broken ORAS target string.

Original error seen:
```
The command failed with an unexpected error. Here is the traceback:
HTTPSConnectionPool(host='ac4mitestingac4mitestingacred058481d7.azurecr.io%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20', port=443):
```
Those `%20`s are spaces that have crept in.


yonzhan:

PR #6410, opened 2023-06-15
containerapp - Add integration tests for patch

snehapar9:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

src/containerapp/azext_containerapp/tests/latest/test_containerapp_patch.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # Verify that the Container App is running
+        app = self.cmd(f"containerapp show -g {resource_group} -n {app_name}").get_output_in_json()
+        url = app["properties"]["configuration"]["ingress"]["fqdn"]
+        url = url if url.startswith("http") else f"http://{url}"
+        resp = requests.get(url)
+        self.assertTrue(resp.ok)
+
+        # Re-run the 'az containerapp up' command with the location parameter if provided
+        if location:
```
harryli0108, line 201: I know this part is similar to the [up command test case](https://github.com/Azure/azure-cli-extensions/blob/main/src/containerapp/azext_containerapp/tests/latest/test_containerapp_up.py#LL92C1-L95C29)., but could you clarify the reason why we need to rerun the command with location argument instead of directly providing this as an argument?

src/containerapp/azext_containerapp/tests/latest/test_containerapp_patch.py
```diff
@@ -0,0 +1,204 @@
...
+        patch_cmd = f'containerapp patch list -g {resource_group} --environment {env_name} --show-all'
+        output = self.cmd(patch_cmd).get_output_in_json()
+        self.assertEquals(output[0]["targetImageName"],"mcr.microsoft.com/k8se/quickstart:latest")
+
+        # Execute and verify patch apply command
+        self.cmd(f'containerapp patch apply -g {resource_group} --environment {env_name} --show-all')
+        app = self.cmd(f"containerapp show -g {resource_group} -n {app_name}").get_output_in_json()
+        image = app["properties"]["template"]["containers"][0]["image"]
+        self.assertEquals(image,"mcr.microsoft.com/k8se/quickstart:latest")
+
```
harryli0108, line 104: I'm wondering if we could deploy 2 container apps (one with quick start image, one with an old patchable image) to make sure that the `--show-all` is working properly by adding assert to check how many results have been returned by the `patcher list` command.

harryli0108 (commented):
I think all looks good from my side. Meanwhile, commented a couple of questions.

src/containerapp/azext_containerapp/tests/latest/test_containerapp_patch.py
```diff
@@ -0,0 +1,156 @@
...
+    def _retag_image_to_older_version_and_push(self,resource_group,app_name):
+        client = docker.from_env()
+        app = self.cmd(f"containerapp show -g {resource_group} -n {app_name}").get_output_in_json()
+        image_name = app["properties"]["template"]["containers"][0]["image"]
+        current_image = client.images.get(image_name)
+        registry = app["properties"]["configuration"]["registries"][0]["server"]
+        repo = app["properties"]["template"]["containers"][0]["name"]
+
+        # Re-tag image to an older version to simulate patch list
+        old_tag = "run-dotnet-aspnet-7.0.1-cbl-mariner2.0"
```
cormacpayne, line 144: Let's make `old_tag` an argument of this function to allow tests to be flexible with the older version they want to use

src/containerapp/azext_containerapp/tests/latest/test_containerapp_patch.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python
```
cormacpayne, line 1: nit: do we need this line?

cormacpayne (approved):
Minor comments, otherwise LGTM :shipit:

src/containerapp/azext_containerapp/tests/latest/test_containerapp_patch.py
snehapar9: @harryli0108 In my opinion, it may be better to explicitly check for the image name than the count of container apps returned from patch list. The ordering may not be the same every time if we have multiple apps and would make it difficult to assert against the image name. I would think it may be okay to keep it simple with just a single container app. Thoughts?

src/containerapp/azext_containerapp/tests/latest/test_containerapp_patch.py
snehapar9: Updated.

src/containerapp/azext_containerapp/tests/latest/test_containerapp_patch.py
```diff
@@ -0,0 +1,159 @@
...
+from azure.cli.testsdk import (ScenarioTest, ResourceGroupPreparer, JMESPathCheck, live_only)
+from knack.util import CLIError
+
+from azext_containerapp.tests.latest.common import TEST_LOCATION
+from azext_containerapp import _utils
+from azext_containerapp.tests.latest.utils import create_and_verify_containerapp_up
+TEST_DIR = os.path.abspath(os.path.join(os.path.abspath(__file__), '..'))
+
+class ContainerAppPatchTest(ScenarioTest):
+    @live_only()
```
zhoxing-ms, line 23: May I ask why do you need to mark these new tests as `@live_only()`?

snehapar9: @zhoxing-ms, these tests are marked `@live_only()` because they require docker to be running. Since the CI does not have `docker` installed they would not work if they aren't run live.

zhoxing-ms: Got it, thanks~

zhoxing-ms (approved):

PR #6573, opened 2023-07-28
{CI} Fix command tree broken issue

wangzelin007:
---
When two extensions have the same command, the following error will be reported:
`Exception: Key: show already exists in containerapp. 2 extensions cannot have the same command!`
Temporarily skip the containerapp-preview extension to aviod this error.
Which will cause the containerapp-preview extension to be unable to use the dynamic load function.
That is, when using the unique command of containerapp-preview, the extension cannot be automatically prompted to install.

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/scripts/ci/build_ext_cmd_tree.sh
+++ b/scripts/ci/build_ext_cmd_tree.sh
@@ -20,7 +20,15 @@ export AZURE_EXTENSION_INDEX_URL=https://raw.githubusercontent.com/Azure/azure-c
 output=$(az extension list-available --query [].name -otsv)
 # azure-cli-ml is replaced by ml
 # disable alias which relies on Jinja2 2.10
-blocklist=("azure-cli-ml" "alias")
+# -----------------------------------------------
+# When two extensions have the same command, the following error will be reported:
+# `Exception: Key: show already exists in containerapp. 2 extensions cannot have the same command!`
+# Temporarily skip the containerapp-preview extension,
+# Which will cause the containerapp-preview extension to be unable to use the dynamic load function.
+# That is, when using the unique command of containerapp-preview, the extension cannot be automatically prompted to install.
+# TODO: remove this after support for building dependencies in command index between extensions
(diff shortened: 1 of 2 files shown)


yonzhan:
CI

kairu-ms:
It's better to update it here https://github.com/Azure/azure-cli-extensions/blob/2491e8d6c96e40b437a9e32783ed174bb9ec612b/scripts/ci/build_ext_cmd_tree.sh#L23

kairu-ms (approved):

evelyn-ys (approved):

scripts/ci/build_ext_cmd_tree.sh
```diff
@@ -20,7 +20,15 @@ export AZURE_EXTENSION_INDEX_URL=https://raw.githubusercontent.com/Azure/azure-c
...
-blocklist=("azure-cli-ml" "alias")
+# -----------------------------------------------
+# When two extensions have the same command, the following error will be reported:
+# `Exception: Key: show already exists in containerapp. 2 extensions cannot have the same command!`
+# Temporarily skip the containerapp-preview extension,
+# Which will cause the containerapp-preview extension to be unable to use the dynamic load function.
+# That is, when using the unique command of containerapp-preview, the extension cannot be automatically prompted to install.
+# TODO: remove this after support for building dependencies in command index between extensions
+# -----------------------------------------------
+blocklist=("azure-cli-ml" "alias" "containerapp-preview")
```
Juliehzl, line 31: we do not want to be a special case with such exclusion. Let's discuss for better solution next week.

Juliehzl (changes requested):
we do not want to be a special case with such exclusion. Let's discuss for better solution next week.

wangzelin007:
> we do not want to be a special case with such exclusion. Let's discuss for better solution next week.

But we can't block other extensions, so we need to merge this temporary solution anyway.

PR #6617, opened 2023-08-09
Migrate vmware

kairu-ms:
---

### Related command

- `az vmware workload-network segment create`: Removed the `--port-name` argument which is used by a read-only property.
- `az vmware privated-cloud identity-source`: New command group to manage identity source.
- `az vmware private-cloud add-identity-source/delete-identity-source`: Deprecated and redirect to `az vmware private-cloud identity-source create/delete`.
- `az vmware script-execution create`: Fix bug for `--output` argument, it should accept a list of string values.
- `az vmware placement-policy vm update`: Removed the `--affinity-strength` and `--azure-hybrid-benefit` arguments. They should only be used in `vm-host update`.
- Use `Get`+`Put` operations for all the `update` commands.

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

src/vmware/azext_vmware/_help.py
```diff
@@ -846,15 +846,15 @@
     short-summary: Create a segment by ID in a private cloud workload network.
     examples:
     - name: Create a segment by ID in a workload network.
-      text: az vmware workload-network segment create --resource-group group1 --private-cloud cloud1 --segment segment1 --display-name segment1 --connected-gateway /infra/tier-1s/gateway --revision 1 --dhcp-ranges 40.20.0.0 40.20.0.1 --gateway-address 40.20.20.20/16 --port-name port1
+      text: az vmware workload-network segment create --resource-group group1 --private-cloud cloud1 --segment segment1 --display-name segment1 --connected-gateway /infra/tier-1s/gateway --revision 1 --dhcp-ranges 40.20.0.0 40.20.0.1 --gateway-address 40.20.20.20/16
```
yanzhudd, line 849: May I ask why `--port-name` is removed in the example?

kairu-ms: Good question. Because the --port-name is read-only property in swagger. And this argument was not work.

src/vmware/azext_vmware/tests/latest/test_addon_scenario.py
```diff
@@ -81,8 +82,8 @@ def test_vmware_addon(self):
         # Create an Arc addon
         self.cmd('az vmware addon arc create -g {rg} -c {privatecloud} --vcenter vcenterId')
 
-        # Update an Arc addon
-        self.cmd('az vmware addon arc update -g {rg} -c {privatecloud} --vcenter vcenterId')
+        # # Update an Arc addon
+        # self.cmd('az vmware addon arc update -g {rg} -c {privatecloud} --vcenter vcenterId')
```
yanzhudd, line 86: May I ask why these tested commands are skipped?

kairu-ms: We cannot use live test to record test because of the fee is too high for us. The old recording only has 'PATCH' request for old `update` command and we use `GET`+`PUT` for update commands after migration, that why we cannot test update commands based on current recordings.

src/vmware/azext_vmware/commands.py
```diff
@@ -4,209 +4,97 @@
...
-        g.custom_show_command('show', 'authorization_show')
-
-    with self.command_group('vmware hcx-enterprise-site', vmware_sdk, client_factory=cf_vmware) as g:
-        g.custom_command('create', 'hcxenterprisesite_create')
-        g.custom_command('list', 'hcxenterprisesite_list')
-        g.custom_command('delete', 'hcxenterprisesite_delete')
-        g.custom_show_command('show', 'hcxenterprisesite_show')
-
-    with self.command_group('vmware location', vmware_sdk, client_factory=cf_vmware) as g:
+    with self.command_group('vmware location') as g:
```
yanzhudd, line 31: Just out of curiosity, if the below commands are not migrated, why would the client factory be removed from the context?

kairu-ms: Because they use aaz commands in custom function internally, so there's no need to use client_factory which is designed for python SDK usage.

yanzhudd (approved):

PR #6658, opened 2023-08-17
[WIP] [containerapp] Add support to initialize Dapr components

shubham1172:
---
This PR adds support to initialize dapr components in a containerapp environment for development purpose, and introduces a new command - `az containerapp env dapr-component init`.

image

### Related command

`az containerapp env dapr-component`

### General Guidelines

For new extensions:

### About Extension Publish


src/containerapp/azext_containerapp/_constants.py
```diff
@@ -77,3 +77,10 @@
 HELLO_WORLD_IMAGE = "mcr.microsoft.com/k8se/quickstart:latest"
 
 LOGS_STRING = '[{"category":"ContainerAppConsoleLogs","categoryGroup":null,"enabled":true,"retentionPolicy":{"days":0,"enabled":false}},{"category":"ContainerAppSystemLogs","categoryGroup":null,"enabled":true,"retentionPolicy":{"days":0,"enabled":false}}]'  # pylint: disable=line-too-long
+
+DAPR_REDIS_PREFIX = "dapr-redis"
+DAPR_REDIS_SECRET_NAME = "redis-config"
+DAPR_REDIS_CONFIG_PASSWORD_KEY = "requirepass"
+DAPR_REDIS_PORT = 6379
+DAPR_COMPONENT_REDIS_STATESTORE_NAME = "redis-statestore"
+DAPR_COMPONENT_REDIS_PUBSUB_NAME = "redis-pubsub"
```
anthonychu, line 86: We should make these match `dapr init`, which I think creates them as `statestore` and `pubsub`. That way you can go from local to ACA without code change.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3656,6 +3660,72 @@ def disable_dapr(cmd, name, resource_group_name, no_wait=False):
     except Exception as e:
         handle_raw_exception(e)
 
+def init_dapr_component(cmd, resource_group_name, environment_name):
+    _validate_subscription_registered(cmd, CONTAINER_APPS_RP)
+
+    # Create the Redis service
+    redis_capp_name = DAPR_REDIS_PREFIX +'-' + ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(5))
```
anthonychu, line 3667: What happens when you run `dapr init` multiple times? I think the ACA behavior should be similar. My expectation would be that it's a no-op if it's been run already on the ACA environment (or it could confirm components are correctly installed and repair). Perhaps we can create the Redis service using a fixed name that is unlikely to collide with something the customer would create. We could then check for its existence. Having a fixed naming scheme could make it easier to implement an "un-init" command too.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3656,6 +3660,72 @@ def disable_dapr(cmd, name, resource_group_name, no_wait=False):
     except Exception as e:
         handle_raw_exception(e)
 
+def init_dapr_component(cmd, resource_group_name, environment_name):
```
anthonychu, line 3663: What should we do if the Redis service already exists or there are already components created with the names we want to use? Wonder if we should ask for confirmation before proceeding (not sure what the CLI guidelines are for this, might need to add a `--yes` flag or something to bypass).

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3656,6 +3660,72 @@ def disable_dapr(cmd, name, resource_group_name, no_wait=False):
...
+    try:
+        redis_capp_def = create_redis_service(cmd, redis_capp_name, environment_name, resource_group_name)
+    except Exception as e:
+        raise ValidationError("Failed to create Redis service.") from e
+    
+    if redis_capp_def is None:
+        raise ValidationError("Failed to create Redis service, did not receive a response.")
+
+    # Look up the Redis secret
+    _get_existing_secrets(cmd, resource_group_name, redis_capp_name, redis_capp_def)
```
anthonychu, line 3678: @ahmelsayed Is there a better way to retrieve the Redis connection string from a service without binding it to another app?

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3656,6 +3660,72 @@ def disable_dapr(cmd, name, resource_group_name, no_wait=False):
...
+    for line in redis_secret_value.splitlines():
+        if line.startswith(DAPR_REDIS_CONFIG_PASSWORD_KEY):
+            redis_password = line.split(" ")[1]
+            break
+
+    if redis_password is None:
+        raise ValidationError(f"Failed to setup Dapr components, {DAPR_REDIS_SECRET_NAME} secret does not contain a password for Redis service.")
+    
+    # Create the Dapr components
+    redis_host = f"{redis_capp_name}:{DAPR_REDIS_PORT}"
```
anthonychu, line 3704: I think technically the port should be read from the `exposedPort` in `redis_capp_def` although that's unlikely to be different.

src/containerapp/azext_containerapp/custom.py
ahmelsayed: Sorry, what is this supposed to do here? you should be able to bind a redis dev service to a dapr component directly without needing to retrieve secrets.

anthonychu: Can you share how to do it and which API version? Couldn't find service bind in the spec but maybe looking in the wrong place.

This is a command that performs the equivalent of `dapr init` in ACA.

halspang: When you do a default `dapr init -k` for a standard cluster, it would use the k8s secretstore to reference the secret redis creates when it is provisioned. Since that's disabled, we'd need to fetch the secrets this way and either apply them directly to the component, or, go through the secret management layer that Container Apps provides if you don't want to touch them directly, though I think they'd still need to be fetched here.

yonzhan:
containerapp

src/containerapp/azext_containerapp/custom.py
Greedygre, line 3663: please add test for the new command.

src/containerapp/azext_containerapp/_utils.py
```diff
@@ -2083,3 +2083,37 @@ def parse_oryx_mariner_tag(tag: str) -> OryxMarinerRunImgTagProperty:
...
+                "value": redis_host
+            },
+            {
+                "name": "redisPassword",
+                "value": redis_password
+            }]
+        }
+    }
+
+def get_dapr_redis_pubsub_component(redis_host, redis_password):
```
Greedygre, line 2104: There is `DaprComponent` in _models.py.
Please use it.

Greedygre:
Please add the description in history.rst.

src/containerapp/azext_containerapp/custom.py
shubham1172: Right, let us call it `dapr-redis` only, and all the operations are `PATCH`, so it becomes a no-op.

src/containerapp/azext_containerapp/custom.py
shubham1172: Yes, I initially wrote it but then it's unlikely to change. BUT, now that redis can *pre-exist*, let me add it. This will allow users to create their own Redis with the name `dapr-redis` and continue to use it here.

src/containerapp/azext_containerapp/custom.py
shubham1172: image

shubham1172: Added!

zhoxing-ms (approved):

src/containerapp/azext_containerapp/tests/latest/test_containerapp_env_commands.py
```diff
@@ -198,6 +198,17 @@ def test_containerapp_env_dapr_components(self, resource_group):
             JMESPathCheck('length(@)', 0),
         ])
 
+        self.cmd('containerapp env dapr-component init -n {} -g {}'.format(env_name, resource_group), checks=[
```
Greedygre, line 201: Please rerun this test lively with `--live`:
`azdev test test_containerapp_env_dapr_components --live`
https://github.com/Azure/azure-cli-dev-tools#style-linter-check-and-testing

src/containerapp/azext_containerapp/custom.py
shubham1172: @ahmelsayed could you share more details here please?

shubham1172: Making these changes from our offline discussion!

shubham1172:
Closing this PR, will open a new one.

PR #6677, opened 2023-08-23
enforce valid custom locations SP app object id

TheOnlyWei:
enforce valid custom locations SP app object id
add .gitignore and test configuration file

---

### Related command
az connectedk8s

### General Guidelines
Style errors from the command were already there. Didn't seem style guide was previously enforced, so won't fix these errors here.

### DIFF
--- a/src/connectedk8s/.gitignore
+++ b/src/connectedk8s/.gitignore
@@ -1,0 +1,2 @@
+# Ignore user-specific test configuration
+azext_connectedk8s/tests/latest/config.json
--- a/src/connectedk8s/HISTORY.rst
+++ b/src/connectedk8s/HISTORY.rst
@@ -2,6 +2,11 @@
 
 Release History
 ===============
+
+1.5.0
+++++++
+* Enforce valid custom locations service principal application object id passed in by the user for enabling custom locations feature.
+
 1.4.2
 ++++++
 * Fix reference error.
@@ -16,7 +21,6 @@ Release History
 
 1.3.20
 ++++++
-
 * Bug fix in parsing logs for outbound connectivity check for troubleshoot command
 
 1.3.19
--- a/src/connectedk8s/azext_connectedk8s/custom.py
+++ b/src/connectedk8s/azext_connectedk8s/custom.py
@@ -404,15 +404,18 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
     put_cc_response = create_cc_resource(client, resource_group_name, cluster_name, cc, no_wait)
     put_cc_response = LongRunningOperation(cmd.cli_ctx)(put_cc_response)
     print("Azure resource provisioning has finished.")
-    # Checking if custom locations rp is registered and fetching oid if it is registered
-    enable_custom_locations, custom_locations_oid = check_cl_registration_and_get_oid(cmd, cl_oid, subscription_id)
+
+    # Checking if custom locations rp is registered and verify passed-in cl_oid.
+    test_custom_location_requirements(cmd, cl_oid, subscription_id)
 
     print("Starting to install Azure arc agents on the Kubernetes cluster.")
+
     # Install azure-arc agents
     utils.helm_install_release(cmd.cli_ctx.cloud.endpoints.resource_manager, chart_path, subscription_id, kubernetes_distro, kubernetes_infra, resource_group_name, cluster_name,
                                location, onboarding_tenant_id, http_proxy, https_proxy, no_proxy, proxy_cert, private_key_pem, kube_config,
-                               kube_context, no_wait, values_file, azure_cloud, disable_auto_upgrade, enable_custom_locations,
-                               custom_locations_oid, helm_client_location, enable_private_link, arm_metadata, onboarding_timeout, container_log_path)
+                               kube_context, no_wait, values_file, azure_cloud, disable_auto_upgrade, cl_oid, helm_client_location,
+                               enable_private_link, arm_metadata, onboarding_timeout, container_log_path)
+
     return put_cc_response
 
 
@@ -1396,15 +1399,14 @@ def enable_features(cmd, client, resource_group_name, cluster_name, features, ku
         azrbac_skip_authz_check = escape_proxy_settings(azrbac_skip_authz_check)
 
     if enable_cl:
+        if cl_oid is None:
+            raise RequiredArgumentMissingError("Custom locations object id was not passed. Please pass '--custom-locations-oid' to enable custom locations.")
         subscription_id = os.getenv('AZURE_SUBSCRIPTION_ID') if custom_token_passed is True else get_subscription_id(cmd.cli_ctx)
-        enable_cl, custom_locations_oid = check_cl_registration_and_get_oid(cmd, cl_oid, subscription_id)
-        if not enable_cluster_connect and enable_cl:
+        # The following throws an error if custom location requirements are not met since user expects to use custom locations.
+        test_custom_location_requirements(cmd, cl_oid, subscription_id)
+        if not enable_cluster_connect:
             enable_cluster_connect = True
             logger.warning("Enabling 'custom-locations' feature will enable 'cluster-connect' feature too.")
-        if not enable_cl:
-            features.remove("custom-locations")
-            if len(features) == 0:
-                raise ClientRequestError("Failed to enable 'custom-locations' feature.")
 
     # Send cloud information to telemetry
     send_cloud_telemetry(cmd)
@@ -1485,7 +1487,7 @@ def enable_features(cmd, client, resource_group_name, cluster_name, features, ku
         cmd_helm_upgrade.extend(["--set", "systemDefaultValues.clusterconnect-agent.enabled=true"])
     if enable_cl:
         cmd_helm_upgrade.extend(["--set", "systemDefaultValues.customLocations.enabled=true"])
-        cmd_helm_upgrade.extend(["--set", "systemDefaultValues.customLocations.oid={}".format(custom_locations_oid)])
+        cmd_helm_upgrade.extend(["--set", "systemDefaultValues.customLocations.oid={}".format(cl_oid)])
 
     response_helm_upgrade = Popen(cmd_helm_upgrade, stdout=PIPE, stderr=PIPE)
     _, error_helm_upgrade = response_helm_upgrade.communicate()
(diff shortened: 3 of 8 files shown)


yonzhan:

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -385,15 +385,18 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
...
-    # Checking if custom locations rp is registered and fetching oid if it is registered
-    enable_custom_locations, custom_locations_oid = check_cl_registration_and_get_oid(cmd, cl_oid, subscription_id)
+
+    # Checking if custom locations rp is registered and verify passed-in cl_oid.
+    test_custom_location_requirements(cmd, cl_oid, subscription_id)
 
     print("Starting to install Azure arc agents on the Kubernetes cluster.")
+
     # Install azure-arc agents
     utils.helm_install_release(cmd.cli_ctx.cloud.endpoints.resource_manager, chart_path, subscription_id, kubernetes_distro, kubernetes_infra, resource_group_name, cluster_name,
```
keystroke, line 395: this is insane, are these parameters ordered? how could one keep this straight? Didn't you miss passing one bool param here from before?

TheOnlyWei: Can you clarify about the missing Boolean param? Don't remember what that refers to. Yea, this function has a lot of parameters. Might be a way to simplify it, but I will leave that for another PR.

keystroke: Right here in this "utils.helm_install_release" function call, previously the parameters were passed:

    ..., disable_auto_upgrade, enable_custom_locations, custom_locations_oid, helm_client_location, ...

And after the change:

    ..., disable_auto_upgrade, cl_oid, helm_client_location, ...
    
Previously there was another parameter "enable_custom_locations" which I assume is a boolean flag? Hard to follow as all these appear as an array of parameters to this function.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -2141,56 +2143,47 @@ def client_side_proxy(cmd,
...
-
-        if cl_oid is None:
-            logger.warning("Failed to enable Custom Locations feature on the cluster. Unable to fetch Object ID of Azure AD application used by Azure Arc service. Try enabling the feature by passing the --custom-locations-oid parameter directly. Learn more at https://aka.ms/CustomLocationsObjectID")
-            telemetry.set_exception(exception='Unable to fetch oid of custom locations app.', fault_type=consts.Custom_Locations_OID_Fetch_Fault_Type,
-                                    summary='Unable to fetch oid for custom locations app.')
-            return ""
-        else:
-            return cl_oid
+        # Do not use display names for look-up as display names are not unique.
+        result = list(sp_graph_client.list(filter=("objectId eq '{}'".format(cl_oid))))
```
keystroke, line 2178: minor nit, but oid is the key to get a service principal so can query directly for object rather than doing a list query with filter. also, what is the high-level flow here, as it is necessary to get a graph token from user to call graph to lookup this SP just to trace the display name? perhaps this should be skipped entirely?

TheOnlyWei: It uses the Azure CLI context to call the AAD graph to get the SP app information.

src/connectedk8s/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Ignore user-specific test configuration
+azext_connectedk8s/tests/latest/config.json
```
zhoxing-ms, line 2: Why do you need to submit this change?

src/connectedk8s/HISTORY.rst
```diff
@@ -3,13 +3,16 @@
 Release History
 ===============
 
+1.5.0
+++++++
+* Enforce valid custom locations service principal application object id passed in by the user for enabling custom locations feature.
```
zhoxing-ms, line 8: I would like to confirm that this change will not cause a breaking change for the customer, right?

src/connectedk8s/.gitignore
TheOnlyWei: Previously, test code was using hard-coded user-specific configs. Updated to use config.json to be more secure. Basically, a tester fills out the user-specific settings in config.json and then run the test, whereas before it is directly put in the test code.

src/connectedk8s/HISTORY.rst
TheOnlyWei: This is breaking for Azure Cloud users who expect the code to pick the right SP application to use for custom locations feature even though custom locations SP app object id (`--custom-locations-oid`) was not passed, which is the wrong behavior to begin with. This update is a breaking change to enforce correct behavior.

Here is the old code:
https://github.com/Azure/azure-cli-extensions/blob/419c252ee7ff0d85203ef3b6cf19ca71ac30488a/src/connectedk8s/azext_connectedk8s/custom.py#L2174C23-L2174C23

The behavior in the old code checks to see if "Custom Locations RP" SP application exists in AAD graph. It should always exist for Azure Cloud, from my understanding. Therefore it goes inside the `if len(result) != 0:` block. Inside that block, it checks `if cl_oid is not None and cl_oid != result[0].object_id:`. If user didn't pass `cl_oid` (parameter `--custom-locations-oid`), then `return result[0].object_id` is returned anyway even if user has no intention to enable custom locations. This ends up enabling custom locations and is the wrong behaviour since user didn't pass `--custom-locations-oid`. Therefore, currently, custom locations feature is always enabled as long as customer has an SP app named "Custom Locations RP", and this is always the case for Azure Cloud (to my understanding), which is the wrong behaviour.

src/connectedk8s/.gitignore
zhoxing-ms: Got it, thanks~

src/connectedk8s/HISTORY.rst
zhoxing-ms: Have you evaluated the impact of this breaking change on CLI customers?

TheOnlyWei: This will impact Azure Arc users. But there is no loss of functionality, they just need to explicitly pass "--custom-locations-oid" to enable custom locations. Otherwise, it is not enabled by default. Previous code guesses the custom locations object ID via display name of an SP app, which is not unique and may result in buggy behavior.

zhoxing-ms: Yes, I understand. Could we first add the warning logs to guide customers to migrate their usage, and implement the breaking change in the breaking change window of CLI (**Ignite Sprint**)? I believe this will minimize the impact on customers, what do you think of it?

TheOnlyWei: Hi @zhoxing-ms, this warning already exists:
https://github.com/TheOnlyWei/azure-cli-extensions/blob/b315d43fe8c51d47b8cd2f05748e5bed1eec9852/src/connectedk8s/azext_connectedk8s/custom.py#L2154C31-L2154C31

zhoxing-ms: Got it, thanks~

src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py
```diff
@@ -19,10 +19,25 @@
...
+CONFIG = {} # dictionary of configurations
+config_path = os.path.join(os.path.dirname(__file__), "config.json")
+if not os.path.isfile(config_path):
+    CONFIG["customLocationsOid"] = ""
+    CONFIG["rbacAppId"] = "fakeRbacAppId"
+    CONFIG["rbacAppSecret"] = "fakeRbacAppSecret"
+    CONFIG["location"] = "eastus2euap"
+else:
+    with open(config_path, 'r') as f:
+        CONFIG = json.load(f)
```
zhoxing-ms, line 37: >   File "/mnt/vss/_work/1/s/src/connectedk8s/azext_connectedk8s/tests/latest/test_connectedk8s_scenario.py", line 29, in <module>
    with open(configPath, 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/mnt/vss/_work/1/s/src/connectedk8s/azext_connectedk8s/tests/latest/config.json'

`connectedk8s/azext_connectedk8s/tests/latest/config.json` file was used in the testing of CI pipeline, why do you need to remove this file? Could you keep a config file which does not contain sensitive information to make the CI tests pass?

TheOnlyWei: No this is not possible because this is a live test. You need passwords. Luckily, the code doesn't verify that the password actually works, so I can use a fake password for now just for testing.

zhoxing-ms: > Luckily, the code doesn't verify that the password actually works, so I can use a fake password for now just for testing. 

OK, this is exactly the way I want to recommend

zhoxing-ms: > Ideally, Azure CLI CI pipeline should have a way to configure secrets for live tests, but I don't know if Azure CLI pipeline actually runs these tests, or ignore missing configurations files related to manual user configurations.

The CI pipeline will only run tests in playback mode instead of live mode. Regarding the issue of whether the CI pipeline can support configuring secrets files, add our CI expert @wangzelin007  to help take a look

wangzelin007: We cannot store secrets for security reasons. 
If you want run live test successfully, you should create secrets in real time, then complete your test and delete all resources immediately after the test is completed.
This way secrets only exist in memory and will no longer exist after the test is completed, which is the safest.

TheOnlyWei: @wangzelin007 This is caused by manual config file added to the live test to begin with. Azure CLI runs and doesn't see config and throws an error because a user needs to create the file. Ideally either there should be an option to ignore these config files or for Azure CLI to support secret configurations. Configurations and code should always be separate.

The work-around for now is to put in fake values because the code doesn't validate the password, but that is not a long-term solution if you have to write live tests where credentials are validated.

TheOnlyWei: Thanks, since we don't need real password for the tests currently, temporary real password feature can be implemented for another PR when the need arises. 

@zhoxing-ms Are there any other concerns?

zhoxing-ms (approved):

zhoxing-ms:
@keystroke Could you please help review this PR agian?

keystroke (approved):

yanzhudd:
Please resolve the code conflicts.

TheOnlyWei:
Hi @zhoxing-ms can this be merged?

zhoxing-ms (approved):

PR #6689, opened 2023-08-24
{Interactive} Configurations to support error assistance enable/disable

avgale:
**Related commands**
Syntactically incorrect commands

**Description**
Supports configurations for error assistance targeting beginners making syntax error in flags, command structure, or preset parameter values. Early version that supports the aforementioned syntax errors but does not yet recognize error relating to custom parameter values such as a non-existent resource group. The feature provides an explanation of the user's errors and a remedied command suggestion after traditional error messages.

Two separate configuration enable the feature for interactive mode and the default CLI. It is enabled by default in interactive mode and disabled by default outside of interactive mode. These configurations can be set as follows:

az config set core.error_assistance=True/False
az config set interactive.assistance=True/False

The interactive configuration is dependent on the additions in this PR. See the relevant azure-cli PR here:
https://github.com/Azure/azure-cli/pull/27256

Example:
Incorrect Command: 
"""
keyvault create --resource-group 't-interns2' --'testKeyvault' --location 'eastus'
"""

Output:
"""
Issue:
The --name flag is missing before the resource name.

Corrected Command:
az keyvault create --name testKeyvault --resource-group t-interns2 --location eastus
"""

Note: Api key and endpoint have been set as environment variables for security. Edits needs to be made to make it generalizable to all users.

Builds upon copilot for interactive linked below:
https://github.com/Azure/azure-cli-extensions/pull/6600/files


yonzhan:
Interactive

kceiw:
This PR is superseded by a newer one https://github.com/Azure/azure-cli-extensions/pull/6966

PR #6703, opened 2023-08-29
[cosmosdb-preview] Add support for Priority Based Execution, cross account container copy, api-version 2023-09-15-preview

Achint-Agrawal:
This PR contains changes for 20230915-preview RP API release. This includes changes for Priority Based Execution, DTS and CMK.


yonzhan:

src/cosmosdb-preview/azext_cosmosdb_preview/_params.py
```diff
@@ -431,13 +437,35 @@ def load_arguments(self, _):
...
+        c.argument('src_mongo', nargs='+', arg_group='Azure Cosmos DB API for MongoDB collection copy', action=AddMongoCollectionAction, help='Source Mongo collection details')
+        c.argument('src_nosql', nargs='+', arg_group='Azure Cosmos DB API for NoSQL container copy', action=AddSqlContainerAction, help='Source NoSql container details')
+        c.argument('dest_cassandra', nargs='+', arg_group='Azure Cosmos DB API for Apache Cassandra table copy', action=AddCassandraTableAction, help='Destination Cassandra table details')
+        c.argument('dest_mongo', nargs='+', arg_group='Azure Cosmos DB API for MongoDB collection copy', action=AddMongoCollectionAction, help='Destination Mongo collection details')
+        c.argument('dest_nosql', nargs='+', arg_group='Azure Cosmos DB API for NoSQL container copy', action=AddSqlContainerAction, help='Destination NoSql container details')
+        c.argument('host_copy_on_src', arg_type=get_three_state_flag(), help=argparse.SUPPRESS)
+        c.argument('worker_count', type=int, help=argparse.SUPPRESS)
```
evelyn-ys, line 451: May I know why you use `argparse.SUPPRESS` for these two parameters?

src/cosmosdb-preview/linter_exclusions.yml
```diff
@@ -24,6 +24,9 @@ cosmosdb create:
     gremlin_databases_to_restore:
       rule_exclusions:
       - option_length_too_long
+    enable_priority_based_execution:
+      rule_exclusions:
+      - option_length_too_long
```
evelyn-ys, line 29: instead of adding linter exclusion, can you add a short option for this parameter?

evelyn-ys (changes requested):
If you want to release a new version, don't forget to update the version in `setup.py` and `history.rst`

src/cosmosdb-preview/azext_cosmosdb_preview/_params.py
niteshvijay1995: We do not want it to be documented in help section, so that it is not discoverable by default.

src/cosmosdb-preview/linter_exclusions.yml
Achint-Agrawal: The name of the feature is priority based execution. It would be clearer if we used this parameter name. I see there are other parameters like enable_multiple_write_locations, enable_automatic_failover, etc which have added the exclusion as well? Can we go ahead with this name itself for customer clarity?

calvinhzy:
Can you help address these breaking changes as well? 
image

pjohari-ms:
@niteshvijay1995 Can you take a look at the breaking changes mentioned above?

pjohari-ms:
@Achint-Agrawal can you update these files as requested?

> If you want to release a new version, don't forget to update the version in `setup.py` and `history.rst`

Achint-Agrawal:
> If you want to release a new version, don't forget to update the version in `setup.py` and `history.rst`



> @Achint-Agrawal can you update these files as requested?
> 
> > If you want to release a new version, don't forget to update the version in `setup.py` and `history.rst`

Done. Thanks

src/cosmosdb-preview/linter_exclusions.yml
evelyn-ys: I'm not saying replace the original name, but just adding another option. eg, customer can pass `--enable-priority-based-execution` or `--enable-pbe` or some name else

src/cosmosdb-preview/azext_cosmosdb_preview/_params.py
evelyn-ys: Then I guess what you want is `c.ignore()` for defining this argument

niteshvijay1995: ignore will remove the param completely? It is not coming from reflection, rather we are adding it so that only selected customers can use it

pjohari-ms:
@evelyn-ys If this looks good, can we merge it?

evelyn-ys (approved):

PR #6736, opened 2023-09-07
[ContainerApp] az containerapp env: --infrastructure-resource-group

Juancpani:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

Juancpani:
To test locally, run `az extension add -upgrade --source https://infrargparam.blob.core.windows.net/draft/containerapp-0.3.39-py2.py3-none-any.whl`

src/containerapp/azext_containerapp/tests/latest/test_containerapp_env_commands.py
```diff
@@ -460,6 +460,32 @@ def test_containerapp_env_internal_only_e2e(self, resource_group):
         ])
 
 
+    @ResourceGroupPreparer(location="eastus")
+    @live_only()  # passes live but hits CannotOverwriteExistingCassetteException when run from recording
```
smzeng, line 464: nit: do you think we should get this checked out? I know this is true for some test and not all test?

src/containerapp/azext_containerapp/_params.py
```diff
@@ -177,6 +177,7 @@ def load_arguments(self, _):
         c.argument('platform_reserved_cidr', options_list=['--platform-reserved-cidr'], help='IP range in CIDR notation that can be reserved for environment infrastructure IP addresses. It must not overlap with any other Subnet IP ranges')
         c.argument('platform_reserved_dns_ip', options_list=['--platform-reserved-dns-ip'], help='An IP address from the IP range defined by Platform Reserved CIDR that will be reserved for the internal DNS server.')
         c.argument('internal_only', arg_type=get_three_state_flag(), options_list=['--internal-only'], help='Boolean indicating the environment only has an internal load balancer. These environments do not have a public static IP resource, therefore must provide infrastructureSubnetResourceId if enabling this property')
+        c.argument('infrastructure_resource_group', options_list=['--infrastructure-resource-group', '-i'], help='Name for resource group that will contain infrastructure resources. If not provided, the default resource group name will be used.')
```
smzeng, line 180: Does this mean that the rg set using  `az configure --defaults group=<name>` will be used or just a randomally generated rg using our default values be used? Because i tested it and it used a randomly generated name not the default rg set by az configured? Is that expected

src/containerapp/azext_containerapp/tests/latest/test_containerapp_env_commands.py
bgashirabake: this error is bc the request prepared to be sent (but actually won't) doesn't match the one on record from the recording. to debug, check the CannotOverwriteExistingCassetteException for the != line -- this will show you where the mismatch is happening and that way you can tune the test to have a reproducible outcome.

src/containerapp/azext_containerapp/_params.py
Juancpani: This is a different resource group from the one default would set, `az configure` would set the one corresponding to [this parameter](https://learn.microsoft.com/en-us/rest/api/containerapps/preview/managed-environments/create-or-update?tabs=HTTP#:~:text=of%20the%20Environment.-,resourceGroupName,-path), whereas the new field maps onto [this parameter](https://learn.microsoft.com/en-us/rest/api/containerapps/preview/managed-environments/create-or-update?tabs=HTTP#:~:text=of%20Dapr%20component.-,properties.infrastructureResourceGroup,-string).

smzeng: both links go to the same page for me. But I get know that az configure --defaults group=<name> is not setting the same thing.

maybe then say "if not provided a resource group name will be generated?"

src/containerapp/azext_containerapp/tests/latest/test_containerapp_env_commands.py
```diff
@@ -460,6 +460,32 @@ def test_containerapp_env_internal_only_e2e(self, resource_group):
...
+        containerapp_env = self.cmd(f'containerapp env show -g {resource_group} -n {env}').get_output_in_json()
+
+        while containerapp_env["properties"]["provisioningState"].lower() == "waiting":
+            time.sleep(5)
+            containerapp_env = self.cmd(f'containerapp env show -g {resource_group} -n {env}').get_output_in_json()
+
+        self.cmd(f'containerapp env show -n {env} -g {resource_group}', checks=[
+            JMESPathCheck('name', env),
+            JMESPathCheck('properties.infrastructureResourceGroup', infra_rg),
+        ])
```
Greedygre, line 486: please delete the env at the end of test.

src/containerapp/azext_containerapp/_params.py
Greedygre, line 180: We will move all the GA commands to core in this PR: https://github.com/Azure/azure-cli/pull/27078.
For arguments which only existing in extension containerapp,  please add tag `is_preview=True` for it.

src/containerapp/azext_containerapp/containerapp_env_decorator.py
```diff
@@ -111,6 +111,9 @@ def get_argument_min_nodes(self):
     def get_argument_max_nodes(self):
         return self.get_param("max_nodes")
 
+    def get_argument_infrastructure_resource_group(self):
```
Greedygre, line 114: Please move logic for new argument to `ContainerappEnvPreviewCreateDecorator`.
And add logic in `validate_arguments()` and `construct_payload()`.
For example:  `ContainerAppPreviewCreateDecorator`

Here is a sample:

class ContainerappEnvPreviewCreateDecorator(ContainerappEnvCreateDecorator):

    def construct_payload(self):
        super().construct_payload()
        self.set_up_infrastructure_resource_group()

    def validate_arguments(self):
        super().validate_arguments()
        # logic for validate argument infrastructure_resource_group

src/containerapp/azext_containerapp/_params.py
Juancpani: I like that, added.

Juancpani: I'll add the flag. @pagariyaalok do you agree?

src/containerapp/azext_containerapp/containerapp_env_decorator.py
Juancpani: Moved

src/containerapp/azext_containerapp/tests/latest/test_containerapp_env_commands.py
Juancpani: Added.

Juancpani:
@yonzhan I see that this wasn't tagged with containerapps, tagging you to make sure that doesn't cause delays.

yonzhan:
Is this PR ready for review?

Juancpani:
Rebase caused unexpected issues, moving over to: https://github.com/Azure/azure-cli-extensions/pull/6751

PR #6742, opened 2023-09-08
revert wrong CLI whl addition

MilesHolland:
My previous PR was targeting an improperly-named CLI whl. Although the contents of the whl were technically correct, the name/version number was wrong, and thus needs to be removed.


yonzhan:

luigiw (approved):

PR #6745, opened 2023-09-11
[Spring] help customers register AKS RP when creating VNET instance

ShichaoQiu:
help customers register AKS RP when creating VNET instance


yonzhan:
Spring

zhoxing-ms (approved):

PR #6749, opened 2023-09-12
[Spring] Update url logic for log stream in legion arch

felixsong0824:
Update URL logic for log stream in legion arch


felixsong0824:
Cancel this PR since legion arch did not get public released.

PR #6769, opened 2023-09-18
Fleet CLI update 

Ealianis:
---

### Related command
`az fleet`

### General Guidelines

- [x] Have you run `python scripts/ci/test_index.py -q` locally? (`pip


src/fleet/azext_fleet/_help.py
```diff
@@ -23,6 +23,30 @@
...
                          managed cluster and resource group names.
+        - name: --enable-private-cluster
+          type: bool
+          short-summary: Whether to create the Fleet hub as a private cluster or not.
+        - name: --enable-vnet-integration
+          type: bool
+          short-summary: Whether to enable apiserver vnet integration for the Fleet hub or not.
+        - name: --apiserver-subnet-id
+          type: string
+          short-summary: he subnet to be used when apiserver vnet integration is enabled. It is required when creating a new Fleet with BYO vnet.
```
jim-minter, line 34:
```suggestion
          short-summary: the subnet to be used when apiserver vnet integration is enabled. It is required when creating a new Fleet with BYO vnet.
```

src/fleet/azext_fleet/_help.py
```diff
@@ -23,6 +23,30 @@
...
+          short-summary: Type of managed service identity (where both SystemAssigned and UserAssigned types are allowed).
+        - name: --identity-tenant-id
+          type: string
+          short-summary: The tenant ID of the system assigned identity. This property will only be provided for a system assigned identity.
+        - name: --identity-principal-id
+          type: string
+          short-summary: The service principal ID of the system assigned identity. This property will only be provided for a system assigned identity.
+        - name: --identity-client-id
+          type: string
+          short-summary: The client ID of the assigned identity.
```
jim-minter, line 49:
```suggestion
```

```suggestion
```

The service generates these; the user does not provide them.

src/fleet/azext_fleet/_help.py
```diff
@@ -23,6 +23,30 @@
...
+          short-summary: Whether to enable apiserver vnet integration for the Fleet hub or not.
+        - name: --apiserver-subnet-id
+          type: string
+          short-summary: he subnet to be used when apiserver vnet integration is enabled. It is required when creating a new Fleet with BYO vnet.
+        - name: --agent-subnet-id
+          type: string
+          short-summary: The ID of the subnet which the Fleet hub node will join on startup. If this is not specified, a vnet and subnet will be generated and used.
+        - name: --identity-type
+          type: string
+          short-summary: Type of managed service identity (where both SystemAssigned and UserAssigned types are allowed).
```
jim-minter, line 40: If it's a user assigned identity, the user needs to provide the resource ID of the identity somehow.
If it's a system assigned identity, the user just says "give me a system assigned identity".

jim-minter (changes requested):

src/fleet/azext_fleet/_help.py
Ealianis: This is the property to state which identity type is set. As you stated it's either User Assigned or System Assigned.  This summary text is from https://github.com/Azure/azure-rest-api-specs/blob/def187e2e78d7173d8fdd7f77740dd9719e1dfbf/specification/common-types/resource-management/v4/managedidentity.json#L36. 

What I believe you are saying is that we should change the params to take in the RID of the user assigned identity only. The fields TID / Principal ID, Object ID, and type are will be filled by the RP (us) and only readonly from a client side. 

Is that correct?

serbrech: this is completely standard cli functionality.
did you take example on how other cli do this? (param name, types, docs, etc..)

jim-minter: Yes, copy/paste from AKS inasmuch as is possible.  The only valid scenarios for us are:
- no identity
- system identity
- one user identity, user provides its resource ID

Some resources allow system + user at the same time, or multiple user identities at the same time.  This is not allowed by AKS managed clusters or Fleet at this time.

Ealianis: Updated parameters, validators, tests, history. Ready for another review.

Ealianis: Addressing test failures...

kairu-ms (approved):

kairu-ms:
Please fix CI issues

Ealianis:
> Please fix CI issues.

@kairu-ms CI issues resolved. Good to merge.

PR #6772, opened 2023-09-19
Revert "capture az error output and auto-fix the script"

sulixu:
Reverts Azure/azure-cli-extensions#6750


sulixu:
will submit a different PR to revert all changes related to "az aks copilot"

PR #6791, opened 2023-09-21
removed az stack-hci-fm

dbradish-microsoft:
`az stack-hci-vm` is in private preview and will not be published until end of October / early November.  Although this should never have been a top-level node, it was entered this way and therefore causes an empty node as seen here:

image


yonzhan:

kairu-ms (approved):

PR #6802, opened 2023-09-25
Achurchard/regenerate python sdk

Cyclam:
The Hybrid Network API has been updated, so we need to regenerate our Python SDK for use in the CLI extension.

The output of the 'autorest' tool we use to do this has changed significantly between the version used previously, and the latest version. We agreed we should be using the latest version.

Our calling code had to change to match the new API (some models no longer flattened), these changes are included in this PR.

We also found a bug in the generated code: the `HybridNetworkManagementClient` class's signature has reversed the order of the `credential` and `subscription_id` parameters. The calling code is not ours, it is an Azure Core CLI library, and not something we can change. To fix this, we have manually changed the autogenerated code so that the parameter order is the same as before, and matches that expected by the Azure Core CLI code. Specifically, this is in `src/aosm/azext_aosm/vendored_sdks/_client.py`, lines 123-124).

Another change in the autogenerated code required us to add `base_url_bound=False` to our client factory code in `src/aosm/azext_aosm/_client_factory.py`m line 14. I have a question open on Azure SDK's "AutoRest - Engineering" Teams channel asking what the implications are. For now, I'm assuming this is benign as I've not found anything telling me otherwise.

# Testing
1. Mainline testing with a CNF completed
2. UTs run and updated where necessary


yonzhan:

PR #6825, opened 2023-10-03
Get Custom Locations RP app by object id

TheOnlyWei:
ignore error if cannot get custom locations app by object id
update to major version 2.0.0 for enabled-features security update breaking change

---

### Related command
az connectedk8s

### General Guidelines

Can't run it, seems to be a bug with the script as I get the below error:
```
  File "C:\tasks\TheOnlyWei\azure-cli-extensions\scripts\ci\test_index.py", line 23, in <module>
    from wheel.install import WHEEL_INFO_RE
ModuleNotFoundError: No module named 'wheel.install'
```

### About Extension Publish

### DIFF
--- a/src/connectedk8s/setup.py
+++ b/src/connectedk8s/setup.py
@@ -17,7 +17,7 @@ except ImportError:
 # TODO: Confirm this is the right version number you want and it matches your
 # HISTORY.rst entry.
 
-VERSION = '1.5.4'
+VERSION = '2.0.0'
 
 # The full list of classifiers is available at
 # https://pypi.python.org/pypi?%3Aaction=list_classifiers
--- a/src/connectedk8s/HISTORY.rst
+++ b/src/connectedk8s/HISTORY.rst
@@ -3,6 +3,21 @@
 Release History
 ===============
 
+2.0.0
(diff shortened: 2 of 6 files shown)


yonzhan:

src/connectedk8s/setup.py
```diff
@@ -17,7 +17,7 @@
 # TODO: Confirm this is the right version number you want and it matches your
 # HISTORY.rst entry.
 
-VERSION = '1.5.0'
+VERSION = '2.0.0'
```
zhoxing-ms, line 20: We usually only upgrade the major version when there are some breaking changes in PR. Therefore, it is recommend to upgrade the secondary version for this case

zhoxing-ms, line 20:
```suggestion
VERSION = '1.6.0'
```

TheOnlyWei: This is to compensate for the previous breaking change, which still has a breaking change. I initially upgraded only the minor version, due to some assumptions from the code, but those were wrong. So, I updated the major version to make it clear to customers that there is a breaking change.

PR #6857, opened 2023-10-12
[containerapp] Add support to initialize Dapr components

shubham1172:
This PR adds support to initialize dapr-components in a containerapps environment for development purpose, and introduces a new command - `az containerapp env dapr-component init`.

Sample code to try out this feature: https://github.com/shubham1172/aca-dapr-example

Behaviour overview:
1. If user runs `dapr-component init` (DCI) on a fresh ContainerApps environment, it creates the dev service and Dapr component with a binding to the dev service.
1. If user runs DCI when a Dapr component with the same name already exists -
    1.  If the component was created by DCI, overwrite the component.
    1. Otherwise throw an error (users should not lose their component with an overwrite).
1. If the user runs DCI when a dev service already exists, skip its creation.

### Screenshots
Help text:
image

Running the command on a fresh containerapps:
image

Running the command, a second time:
image

Running the command when a Dapr component already exists (user created):
image

Statestore component definition, created by init command:
image

### Related command

`az containerapp env dapr-component`

### General Guidelines

FYI to run `azdev test`, had to do the following:
```bash
export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES
azdev test test_containerapp_env_dapr_components --live
```

For new extensions:

- [N/A] My extension description/summary conforms to the [Extension Summary Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/extensions/extension_summary_guidelines.md).

### About Extension Publish


yonzhan:
containerapp

shubham1172:
Thanks @yonzhan, done!

src/containerapp/azext_containerapp/_dapr_utils.py
```diff
@@ -0,0 +1,340 @@
...
+from ._models import (
+    DaprComponent as DaprComponentModel,
+    DaprMetadata as DaprMetadataModel,
+    DaprServiceComponentBinding as DaprServiceComponentBindingModel,
+)
+
+logger = get_logger(__name__)
+
+SERVICEBIND_METADATA_CREATEDBY_KEY = "SB_CREATED_BY"
+SERVICEBIND_METADATA_CREATEDBY_VALUE = "azext_containerapp_daprutils"
```
anthonychu, line 19: Should this be more generic? Thinking what would happen if we added this easy init logic to the portal too... they would need to use the same value (or if they use a different one, we'd have to check for both)

ahmelsayed: I'm assuming this metadata is non-functional. Something to just track which tooling created the binding? so if we want the same from portal, the UI will have to come up with its own metadata value there. /cc @shubham1172 right?

shubham1172: @ahmelsayed this metadata is used in the `init` logic to figure out if the Dapr Component was created by the `dapr-component init` CLI command or by the users directly (via the existing portal experience or `dapr-component set` CLI command). This determines if a component should be over-written or not (in case it already exists).

@anthonychu when we are using the portal, I am thinking of two things:
1. Users clicking on init in a brand-new managed environment - in which case this metadata is unused.
2. Users clicking on init when some components already exist - on the portal, it should probably show the users outright that this action will overwrite the following components, do you want to proceed **(regardless of the metadata)**.

Now if a component is created by portal's `init` and then a user runs the `init` CLI, then we can do one of the following things:
1. Treat the component created by portal's `init`, the same as the component created by `dapr-component set` command or portal. In this case we continue to throw an error (CURRENT BEHAVIOR)
2. Treat the component created by portal's `init`, the same as the `dapr-component init` command. In this case we can simply overwrite it.

The current experience is in line with the portal's `init` experience (as in only overwrite components created by the tool being used). Do you think we should rather provide an alternate experience as suggested by point 2? We can achieve it by either standardizing the metadata or checking for both as you pointed earlier.

src/containerapp/HISTORY.rst
```diff
@@ -9,6 +9,7 @@ upcoming
 * 'az containerapp compose create': fixed an issue where the environment's resource group was not resolved from --environment when the input value was a resource id.
 * 'az containerapp replica count', returns the replica count of a container app
 * [Breaking Change] 'az containerapp job create': add default values for container app job properties --replica-completion-count, --replica-retry-limit, --replica-timeout, --parallelism, --min-executions, --max-executions, --polling-interval
+* 'az containerapp env dapr-component init': support initializing Dapr components and dev services for an environment
```
Greedygre, line 12: please move the history to the `upcoming`.

shubham1172: Thanks @Greedygre, let me do it once this PR is reviewed. History file creates a lot of merge conflicts.

src/containerapp/azext_containerapp/tests/latest/recordings/test_containerapp_env_dapr_components.yaml
```diff
@@ -859,8 +865,8 @@ interactions:
 - request:
     body: '{"location": "eastus", "tags": null, "properties": {"daprAIInstrumentationKey":
       null, "vnetConfiguration": null, "appLogsConfiguration": {"destination": "log-analytics",
-      "logAnalyticsConfiguration": {"customerId": "0849d8b8-7ccd-4e08-87a3-27e4d75603b4",
-      "sharedKey": "PH5FrLOtTzOnwDq1M/AWYvJm2y2sZ2H2++/ha+WpvC/6RdReJkeutxJW0DytieAF27O4/O7NTJudZu2b5xlXFQ=="}},
+      "logAnalyticsConfiguration": {"customerId": "d0118722-4b37-4428-80e4-b6252f3520fc",
+      "sharedKey": "Yozh2rH7tuhMwRklVyhB/kyuIqgWylr76Fab9s2o11ZfFrdnJDuYWDQtZ1dCoqi8WlYhbec2S62TZGmeVmth8Q=="}},
```
Greedygre, line 869: In extension version 0.3.42, we updated the cluster of environment from V1 to V2.
I notice in this recording file, the environment is still V1 cluster.
Please make sure rerun the test Pass lively with V2 cluster.

src/containerapp/azext_containerapp/tests/latest/recordings/test_containerapp_env_dapr_components.yaml
```diff
@@ -859,8 +865,8 @@ interactions:
...
       "customDomainConfiguration": null, "workloadProfiles": null, "InfrastructureResourceGroup":
       null, "zoneRedundant": false}}'
```
Greedygre, line 871: In extension version 0.3.42, we updated the cluster of environment from V1 to V2.
I notice in this recording file, the environment is still V1 cluster.
Please make sure rerun the test Pass lively with V2 cluster.

src/containerapp/azext_containerapp/_dapr_utils.py
shubham1172: I have updated this to the following for now.

```python
DAPRCOMPONENTINIT_SERVICEBIND_METADATA_CREATEDBY_KEY = "DCI_SB_CREATEDBY"
DAPRCOMPONENTINIT_SERVICEBIND_METADATA_CREATEDBY_VALUE = "azcli_azext_containerapp_daprutils"
```

While checking, I am only checking the presence of the key. The value can be set by CLI or Portal (in future).

src/containerapp/azext_containerapp/_dapr_utils.py
```diff
@@ -0,0 +1,338 @@
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from typing import Dict
+from azure.cli.core.azclierror import ValidationError
+from knack.log import get_logger
+from ._clients import ContainerAppClient, DaprComponentPreviewClient
```
Greedygre, line 9: Please use `ContainerAppPreviewClient`.

src/containerapp/azext_containerapp/_params.py
```diff
@@ -66,6 +66,10 @@ def load_arguments(self, _):
         c.argument('managed_certificates_only', options_list=['--managed-certificates-only', '-m'], help='List managed certificates only.')
         c.argument('private_key_certificates_only', options_list=['--private-key-certificates-only', '-p'], help='List private-key certificates only.')
 
+    with self.argument_context('containerapp env dapr-component init') as c:
```
Greedygre, line 69:
```suggestion
    with self.argument_context('containerapp env dapr-component init', is_preview=True) as c:
```

src/containerapp/azext_containerapp/_params.py
```diff
@@ -66,6 +66,10 @@ def load_arguments(self, _):
...
+        c.argument('statestore', help="The state store component and dev service to create.")
```
Greedygre, line 70: The argument is `statestore`?  Recommend to us: `state-store`.

shubham1172: Actually, this is consistent with Dapr conventions. For example, see component name in https://docs.dapr.io/operations/components/setup-state-store/

src/containerapp/azext_containerapp/_params.py
shubham1172: image
Do we need to mark the args individually as preview? The command itself is in preview.

src/containerapp/azext_containerapp/_dapr_utils.py
```diff
@@ -0,0 +1,338 @@
...
+            raise ValidationError(f"Service type {service_type} is not supported.")
+
+        # Look up the service, if it already exists, return it.
+        logger.debug("Looking up service %s of type %s", service_name, service_type)
+        service_def = None
+        try:
+            service_def = ContainerAppPreviewClient.show(
+                cmd, resource_group_name, service_name
+            )
+        except Exception:  # pylint: disable=broad-except
```
Greedygre, line 249: Please use `handle_non_404_status_code_exception` to avoid ignoring other types of errors.

src/containerapp/azext_containerapp/_dapr_utils.py
```diff
@@ -0,0 +1,338 @@
...
+            )
+
+        # Check if the component already exists.
+        logger.debug("Looking up Dapr component %s", component_name)
+        component_def = None
+        try:
+            component_def = DaprComponentPreviewClient.show(
+                cmd, resource_group_name, environment_name, component_name
+            )
+        except Exception:  # pylint: disable=broad-except
```
Greedygre, line 179: Please use `handle_non_404_status_code_exception` to avoid ignoring other types of errors.

src/containerapp/azext_containerapp/_dapr_utils.py
```diff
@@ -0,0 +1,338 @@
...
+    @staticmethod
+    def _is_component_created_by_daprcomponentinit(component_def) -> bool:
+        """
+        Check if the component was created by dapr-component init.
+
+        :param component_def: component definition to check
+
+        :return: True if the component was created by DaprUtils, False otherwise
+        """
+        from .custom import safe_get
```
Greedygre, line 120:
```suggestion
        from azure.cli.command_modules.containerapp._utils import safe_get
```

Greedygre (approved):
LGTM @Juliehzl Could you help to review this PR?

ahmelsayed (approved):

src/containerapp/HISTORY.rst
```diff
@@ -12,6 +12,7 @@ upcoming
 * 'az containerapp create/update': fix an issue for transforming sensitive values when the scale rules metadata not exists
 * 'az containerapp up': update builder image used when --source is provided with no Dockerfile to support building applications targeting a wider range of platform versions
 * Add Cloud Build support (build without Dockerfile or Docker) in Stage/Canary regions to the 'az containerapp up'/'az containerapp create' and 'az containerapp update' commands
+* 'az containerapp env dapr-component init': support initializing Dapr components and dev services for an environment
```
Greedygre, line 15: Please move this to `upcoming`

Greedygre (approved):

zhoxing-ms:
@shubham1172 Please resolve the conflict

wangzelin007 (approved):

wangzelin007 (approved):

src/containerapp/azext_containerapp/_help.py
```diff
@@ -389,6 +389,18 @@
           az containerapp env certificate delete -g MyResourceGroup --name MyEnvironment --thumbprint MyCertificateThumbprint
 """
 
+helps['containerapp env dapr-component init'] = """
+    type: command
+    short-summary: Initializes Dapr components and dev services for an environment.
```
Juliehzl, line 394: Initializes -> Initialize
to align with other commands

src/containerapp/azext_containerapp/_params.py
Juliehzl, line 70: could you use argument --state-store which will be easy to view for customer like what we do for most arguments? For arguments with multiple words should be separated with `-`

src/containerapp/azext_containerapp/_params.py
```diff
@@ -66,6 +66,10 @@ def load_arguments(self, _):
...
+        c.argument('pubsub', help="The pubsub component and dev service to create.")
```
Juliehzl, line 71: should --pub-sub be easy to understand for customer?

src/containerapp/azext_containerapp/_params.py
shubham1172: Hi, for users coming from the Dapr world, they will be more familiar with `statestore` and `pubsub`. Also please check https://github.com/Azure/azure-cli-extensions/pull/6857#discussion_r1375754681

src/containerapp/azext_containerapp/_params.py
shubham1172: Same as above

PR #6919, opened 2023-10-30
[Container Apps] Enable Cloud Build on ACA CLI

daniv-msft:
---

### Related command
az containerapp up
az containerapp create
az containerapp update

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

daniv-msft:
@greedygre, @howang-ms > I'm still working on this PR (running/adding tests, linter, etc.) but most of the code is there and, as such, I'd love to get some early feedback. Thanks!

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -25,7 +25,7 @@
 
 logger = get_logger(__name__)
 
-PREVIEW_API_VERSION = "2023-05-02-preview"
+PREVIEW_API_VERSION = "2023-08-01-preview"
```
daniv-msft, line 28: @Greedygre, @howang-ms > I assume this API version will be updated for everyone soon. I changed it here to access the new APIs, but I'd be curious to know if you have any visibility on when this will be updated?

Greedygre: Hi @daniv-msft 

For the api-version update in azure-cli-extension, generally we need to wait for api-version public, PR for 2023-08-01-preview is: https://github.com/Azure/azure-rest-api-specs/pull/26352 merged.
After the api-version public, we will upgrade the model class to 2023-08-01-preview if need.

src/containerapp/azext_containerapp/_cloud_build_utils.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # List the builders in the resource group
+        thread = display_spinner("Listing the builders available in the Container Apps environment")
+        builders_list_json_content = BuilderClient.list(cmd, resource_group_name)
+        done_spinner = True
+        thread.join()
+
+        if len(builders_list_json_content["value"]) > 0:
+            builder_name = builders_list_json_content["value"][0]["name"]
+            log_in_file(f"{substatus_indentation}Builder selected: {builder_name}", logs_file)
```
cormacpayne, line 83: Do we not allow more granularity when it comes to selecting the builder? Specifically, if we are given the `environment_name` argument to this `run_cloud_build` function, can we filter the results of `BuilderClient.list` by those associated with the given environment and then select the first one?

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -632,6 +630,29 @@ def run_acr_build(self, dockerfile, source, quiet=False, build_from_source=False
                 dockerfile,
                 quiet,
             )
+            return False
+
+        # Only enable Cloud Build on Stage and Canary while the changes are deployed to all regions.
+        is_cloud_build_enabled = self.env.location and any(self.env.location.lower() == region for region in ["northcentralusstage", "centraluseuap", "eastus2euap"])
```
cormacpayne, line 636: nit: should we extract out this location array into a separate variable that we can easily update as cloud build is enabled for additional locations? Preferably in some constants file, or some file that is related to the cloud build service

src/containerapp/azext_containerapp/custom.py
```diff
@@ -854,12 +854,12 @@ def containerapp_up(cmd,
...
-    app.create_acr_if_needed()
-
     if source:
-        app.run_acr_build(dockerfile, source, quiet=False, build_from_source=not _has_dockerfile(source, dockerfile))
+        used_default_container_registry = app.run_source_to_cloud_flow(source, dockerfile)
+    else:
+        app.create_acr_if_needed()
 
-    app.create(no_registry=bool(repo))
+    app.create(no_registry=bool(repo or used_default_container_registry))
```
snehapar9, line 862: @daniv-msft Not necessarily related to these changes, but @Greedygre curious if you know why it is a no_registry scenario if `repo` was provided? Don't we need to pass the registry details to the GitHub action for non-cloud build scenarios?

src/containerapp/azext_containerapp/_clients.py
daniv-msft: Thank you @Greedygre! OK so once https://github.com/Azure/azure-rest-api-specs/pull/26352 is merged, then we should be good to update to 2023-08-01-preview in the CLI. Depending on when the API specs PR gets merged, I might have to force the use of 2023-08-01-preview for the BuilderClient/BuildClient API calls only as we need these changes available by end of this week. I'll keep you and @howang-ms posted if it's the case.

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -492,3 +492,104 @@ def list(cls, cmd, resource_group_name, env_name, formatter=lambda x: x):
                 env_list.append(formatted)
 
         return env_list
+
+class BuilderClient():
```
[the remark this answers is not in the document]
daniv-msft: I see your point but I guess it's not clear to me that we'll use the whole Builder/Build API available, or when? Ideally, I would rather not add code that we don't need, so if you don't mind I'd prefer to keep this implementation when we have a need for these new methods.

src/containerapp/azext_containerapp/_cloud_build_utils.py
```diff
@@ -0,0 +1,213 @@
...
+        font_bold_green = "\033[32;1m"
+        font_bold_red = "\033[31;1m"
+        font_bold_yellow = "\033[33;1m"
+        font_default = "\033[0m"
+        font_bold = "\033[0;1m"
+        font_blue = "\033[94m"
+        substatus_indentation = "              "
+        def display_spinner(task_title):
+            # Hide the cursor
+            print('\033[?25l', end="")
```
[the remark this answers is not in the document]
daniv-msft: Yes, you're right. Will make that change.

src/containerapp/azext_containerapp/_cloud_build_utils.py
daniv-msft: Yes, you're right as well.

src/containerapp/azext_containerapp/_up_utils.py
daniv-msft: So this is only a thing for this week/next week, as we need to deploy things to get some validation while deploying to more regions. As soon as we have the validation we need, this check will be removed.

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -492,3 +492,104 @@ def list(cls, cmd, resource_group_name, env_name, formatter=lambda x: x):
                 env_list.append(formatted)
 
         return env_list
+
```
Greedygre, line 495: ni: expected 2 blank lines

src/containerapp/azext_containerapp/_cloud_build_utils.py
```diff
@@ -0,0 +1,221 @@
...
+                done_spinner = False
+                thread = display_spinner(f"Buildpack: {log_execution_phase}")
+            log_in_file(log_line, logs_file, no_print=True)
+        done_spinner = True
+        thread.join()
+
+        # TODO: Delete local tar.gz file
+        #os.unlink(tar_file_path)
+
+        final_image = build_json_content["properties"]["destinationContainerRegistry"]["image"]
```
Greedygre, line 200: Local variable 'build_json_content' might be referenced before assignment

src/containerapp/azext_containerapp/custom.py
```diff
@@ -854,12 +854,12 @@ def containerapp_up(cmd,
             _get_registry_from_app(app, source)  # if the app exists, get the registry
         _get_registry_details(cmd, app, source)  # fetch ACR creds from arguments registry arguments
 
-    app.create_acr_if_needed()
-
     if source:
-        app.run_acr_build(dockerfile, source, quiet=False, build_from_source=not _has_dockerfile(source, dockerfile))
+        used_default_container_registry = app.run_source_to_cloud_flow(source, dockerfile)
```
Greedygre, line 858: Local variable 'used_default_container_registry' might be referenced before assignment

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -582,47 +607,20 @@ def build_container_from_source_with_acr_task(self, image_name, source):
         for k, v in old_command_kwargs.items():
             self.cmd.command_kwargs[k] = v
 
-    def run_acr_build(self, dockerfile, source, quiet=False, build_from_source=False):
+
```
Greedygre, line 610: too many blank lines

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -632,6 +630,29 @@ def run_acr_build(self, dockerfile, source, quiet=False, build_from_source=False
                 dockerfile,
                 quiet,
             )
+            return False
+
+        # Only enable Cloud Build on Stage and Canary while the changes are deployed to all regions.
+        is_cloud_build_enabled = self.env.location and any(self.env.location.lower() == region for region in ["northcentralusstage", "centraluseuap", "eastus2euap"])
+        if (self.should_create_acr and is_cloud_build_enabled):
```
Greedygre, line 637: ni: Remove redundant parentheses

src/containerapp/azext_containerapp/_cloud_build_utils.py
```diff
@@ -0,0 +1,221 @@
...
+
+from ._utils import (
+    log_in_file,
+    remove_ansi_characters
+)
+
+class CloudBuildError(Exception):
+    pass
+
+def run_cloud_build(cmd, image_name, source, location, subscription_id, resource_group_name, environment_name, run_full_id, logs_file, logs_file_path):
```
Greedygre, line 29: `image_name` and `subscription_id` not been used. Should we remove them?

src/containerapp/azext_containerapp/_cloud_build_utils.py
```diff
@@ -0,0 +1,221 @@
...
+        font_bold_green = "\033[32;1m"
+        font_bold_red = "\033[31;1m"
+        font_bold_yellow = "\033[33;1m"
+        font_default = "\033[0m"
+        font_bold = "\033[0;1m"
+        font_blue = "\033[94m"
+        hide_cursor = "\033[?25l"
+        display_cursor = "\033[?25h"
+        substatus_indentation = "              "
+        def display_spinner(task_title):
```
Greedygre, line 46: ni: expected 1 blank line before a nested definition

src/containerapp/azext_containerapp/_up_utils.py
daniv-msft: Thanks, removing it

src/containerapp/azext_containerapp/_clients.py
daniv-msft: Updating this, thanks

src/containerapp/azext_containerapp/_cloud_build_utils.py
daniv-msft: This came from a previous version of the code. Thanks for catching!

src/containerapp/azext_containerapp/_cloud_build_utils.py
daniv-msft: Hope I'm not missing something there, but I actually believe we should be good. As we know for sure that the `while (build_provisioning)` loop will always run at least once, then we know that build_json_content will be initialized. Please let me know if I missed something.

src/containerapp/azext_containerapp/_up_utils.py
daniv-msft: Thanks, updating this here and in other places.

src/containerapp/azext_containerapp/custom.py
daniv-msft: I will add a new initialization of this variable to make sure it is always assigned properly.

Greedygre:
Hi @daniv-msft 
Please fix the static analysis:

/mnt/vss/_work/1/s/src/containerapp/azext_containerapp/_up_utils.py:87:1: E302 expected 2 blank lines, found 1
/mnt/vss/_work/1/s/src/containerapp/azext_containerapp/_up_utils.py:644:13: F841 local variable 'build_image_name_with_cache_tag' is assigned to but never used
1     E302 expected 2

Greedygre:
Please add description about the feature in `history.rst`

src/containerapp/azext_containerapp/_cloud_build_utils.py
```diff
@@ -0,0 +1,226 @@
...
+
+from threading import Thread
+import os
+import re
+import requests
+import tempfile
+import time
+
+from ._archive_utils import archive_source_code
+
```
Greedygre, line 15: Please add tests about the feature.

src/containerapp/azext_containerapp/_clients.py
Juliehzl: @daniv-msft we should wait for 2023-08-01-preview api version manifest rollout to worldwide. The manifest rollout needs to take at least 5 days. So it is hard to support it by end of this week.

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -492,3 +492,106 @@ def list(cls, cmd, resource_group_name, env_name, formatter=lambda x: x):
...
+
+class BuilderClient():
```
Greedygre, line 497: If only `BuilderClient` need api-version 2023-08-01-preview in this feature, I recommend you only update the version for BuilderClient.
For example: 
`api_version = "2023-08-01-preview"`

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -492,3 +492,106 @@ def list(cls, cmd, resource_group_name, env_name, formatter=lambda x: x):
...
+        request_url = url_fmt.format(
+            management_hostname.strip('/'),
+            sub_id,
+            resource_group_name,
+            builder_name,
+            cls.api_version)
+        body_data = {
+            "location": location,
+            "properties": {
+                "environmentId": f"/subscriptions/{sub_id}/resourcegroups/{resource_group_name}/providers/microsoft.app/managedenvironments/{environment_name}"
```
Juliehzl, line 527: could you use same pattern with other resource id?
resourcegroup -> resourceGroup
microsoft.app -> Microsoft.App 
...

src/containerapp/azext_containerapp/_clients.py
daniv-msft: Thanks @Juliehzl. @Greedygre and I were talking about this separately.
In this case what I might do is just using the new API version in the BuildClient/BuilderClient I'm introducing, and asking the partner team who needs to test these changes to register for PrereleaseApiVersionAllowed.
That way, it will unblock the partners without making this API version being rushed.

src/containerapp/azext_containerapp/_clients.py
Greedygre: For example:
`BUILDER_CLIENT_API_VERSION = "2023-08-01-preview"`

src/containerapp/azext_containerapp/_cloud_build_utils.py
```diff
@@ -0,0 +1,226 @@
...
+                loop_counter = 0
+                start_time = time.time()
+                time_elapsed = 0
+                while done_spinner is False and fail_spinner is False and cancel_spinner is False:
+                    loop_counter = (loop_counter + 1) % 17
+                    loading_bar_left_spaces_count = loop_counter - 9 if loop_counter > 9 else 0
+                    loading_bar_right_spaces_count = 6 - loop_counter if loop_counter < 7 else 0
+                    spinner = f"[{' ' * loading_bar_left_spaces_count}{'=' * (7 - loading_bar_left_spaces_count - loading_bar_right_spaces_count)}{' ' * loading_bar_right_spaces_count}]"
+                    time_elapsed = time.time() - start_time
+                    print(f"\r    {spinner} {task_title} ({time_elapsed:.1f}s)", end="", flush=True)
```
Juliehzl, line 63: avoid use print(), see https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#general-patterns.

is it possible to use progress bar in https://github.com/Azure/azure-cli/blob/main/src/azure-cli-core/azure/cli/core/commands/progress.py?

src/containerapp/azext_containerapp/_cloud_build_utils.py
```diff
@@ -0,0 +1,226 @@
...
+                done_spinner = False
+                thread = display_spinner(f"Buildpack: {log_execution_phase}")
+            log_in_file(log_line, logs_file, no_print=True)
+        done_spinner = True
+        thread.join()
+
+        # TODO: Delete local tar.gz file
+        # os.unlink(tar_file_path)
+
+        final_image = build_json_content["properties"]["destinationContainerRegistry"]["image"]
```
Juliehzl, line 205: will ["properties"]["destinationContainerRegistry"]["image"] also exist in build json? do we need null check?

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -492,3 +492,106 @@ def list(cls, cmd, resource_group_name, env_name, formatter=lambda x: x):
...
+            r = send_raw_request(cmd.cli_ctx, "GET", request_url)
+
+        return r.json()
+
+
+class BuildClient():
+    api_version = PREVIEW_API_VERSION
+
+    @classmethod
+    def create(cls, cmd, builder_name, build_name, resource_group_name, location):
```
Juliehzl, line 547: is it an async operation?

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -492,3 +492,106 @@ def list(cls, cmd, resource_group_name, env_name, formatter=lambda x: x):
                 env_list.append(formatted)
 
         return env_list
+
+
+class BuilderClient():
+    api_version = PREVIEW_API_VERSION
+
+    @classmethod
+    def list(cls, cmd, resource_group_name):
```
Juliehzl, line 501: one more question: as builder/build are azure resources, `up` command will help create the resource implicitly when --source is specified, but when container app is deleted, the builder resource will still be there, right?

so how can customer delete it as I do not see delete operation for your client?

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -582,56 +604,51 @@ def build_container_from_source_with_acr_task(self, image_name, source):
...
                 image_name_with_tag,
                 source,
                 dockerfile,
-                quiet,
+                False
             )
+            return False
+
+        # Only enable Cloud Build on Stage and Canary while the changes are deployed to all regions.
+        is_cloud_build_enabled = self.env.location and any(self.env.location.lower() == region for region in ["northcentralusstage", "centraluseuap", "eastus2euap"])
```
Juliehzl, line 632: make sure region is normalized because there may be spaces in friendly location, e.g. "North Central US (stage)"

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -582,56 +604,51 @@ def build_container_from_source_with_acr_task(self, image_name, source):
...
+        if self.should_create_acr and is_cloud_build_enabled:
```
Juliehzl, line 633: what is should_create_acr is true but cloud build is not enabled?

src/containerapp/azext_containerapp/_clients.py
daniv-msft: Thanks, updating this accordingly so that only BuildClient and BuilderClient get the new API version

src/containerapp/azext_containerapp/_cloud_build_utils.py
daniv-msft: Yes, this value will always be available (and if the API didn't return it, it would be a bug worth getting an exception here), even before the build completes as we can get all these values beforehand.

src/containerapp/azext_containerapp/_up_utils.py
daniv-msft: Yes, good catch. Once we remove the `is_cloud_build_enabled` (== next week once the deployment is complete), then we wouldn't fall in this case anymore, but as of today I agree it's possible to be in a state where we reach the next steps without a container registry. I'm adding a `create_acr_if_needed` call to make sure this case won't be a problem.

src/containerapp/azext_containerapp/_cloud_build_utils.py
daniv-msft: This one is tricky. We've been asked to match (or as close as possible) to the Azure Dev CLI style, and I needed more freedom than what the typical Logger or Progress Bar gave me.
I agree that the style is different from what is usually done with the Az CLI, but in the end I think it is better for the customers.

src/containerapp/azext_containerapp/_up_utils.py
daniv-msft: Please find it line 640. I'll remove it next week with the PR removing the is_cloud_build_enabled check

src/containerapp/azext_containerapp/_clients.py
daniv-msft: As of today, there is no way in the CLI to delete the builder. This is because the builder is, as of today, a different concept from the container apps environment but that in the coming months, it will be a part of it and as such its lifecycle will be handled with the container apps environment. Please note that the builder does not occur costs for the customer.

Greedygre:
Please merge main to fix the conflict.

showpune (commented):
@daniv-msft , should we update the builder for local builder?

showpune (commented):
@daniv-msft , if the source location is a file, the tar zip file is bad as arcname is empty

Juliehzl (approved):
:shipit:

yonzhan:
Please fix conflicting files.

src/containerapp/azext_containerapp/_clients.py
daniv-msft: Yes it is! Thanks for catching, I updated the code to use the usual polling method

src/containerapp/azext_containerapp/_up_utils.py
daniv-msft: Please note that this is only a temporary solution to allow our partner team to test end to end in a couple of regions. This whole condition will be removed by mid/late next week. The current regions seem to work OK from my testing.

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -582,56 +603,53 @@ def build_container_from_source_with_acr_task(self, image_name, source):
...
                 dockerfile,
-                quiet,
+                False
             )
+            return False
+
+        # Only enable Cloud Build on Stage and Canary while the changes are deployed to all regions.
+        location = "eastus"
+        if self.env.location:
+            location = self.env.location
```
daniv-msft, line 633: @Greedygre, would you know if there is a better way to get the location to use? I've reused the same logic as what is done for the ACR creation (try to read `self.env.location` and if not available default to "eastus") but I'm not super satisfied with it.
Specifically, something I don't get is why we cannot rely more on `self.env.location` in all cases? From my testing this `self.env.location` is only set if I'm using `--location northcentralusstage`. However don't all environments always have a location? Why would we get `None` for an environment?

src/containerapp/azext_containerapp/_clients.py
daniv-msft: Updated! Thanks

yanzhudd:
please run related tests in live mode and upload the recording files.

yanzhudd (approved):

PR #6984, opened 2023-11-11
{containerapp} Re-record failing source-to-cloud patch tests

cormacpayne:
The current set of patching tests that we have available for the source-to-cloud flow are currently failing as they default into our new cloud build service to build the provided application, when, for these patching tests, they should use the local builder flow. To enable this, we create an ACR instance before the `containerapp up` command is called, which signals to the flow that the cloud build service should be skipped (as it currently only runs when no container registry exists in the provided resource group). This then allows an application image to be built, pushed to the given ACR instance and able to be tested with for the existing `containerapp patch` commands.

In addition, a default location was being set before the `containerapp up` call was being called, which was affecting flow where ACA environments were previously created with one location, and the `up` flow expects it have been created with the default location that was set. This line was removed to avoid ambiguity with regards to the location that was observed as the default throughout the lifecycle of the test.

---

### Related command

az containerapp patch

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
containerapp

Greedygre (approved):

Greedygre:
Recommend Title:
{containerapp} Re-record failing source-to-cloud patch tests

Because this change will not impact customer, it should uses {}.

PR title guidance: https://github.com/Azure/azure-cli/blob/f9a4685d6bfee8bd5b8eff867f40a7d674515dae/doc/authoring_command_modules/README.md#format-pr-title

cormacpayne:
@Greedygre thanks for the tip -- I just updated the title

Greedygre (approved):

Greedygre:
Hi @zhoxing-ms @yanzhudd 

Could you please help to review and merge this PR? Thanks!

yanzhudd (approved):

PR #7011, opened 2023-11-20
{AKS} `az aks update`: force `--network-policy=cilium` whenever `--network-dataplane=cilium`

robogatikov:
Recent versions of the aks-preview Azure CLI extension use an AKS preview API (2023-08-02preview) that introduced networkPolicy=none. This causes the command to upgrade to Cilium dataplane to fail with a validation error when --network-policy isn't set explicitly to "cilium".

1. Azure CLI retrieves the current cluster, which defaults to networkPolicy=none.
2. Azure CLI modifies the cluster object, leaving networkPolicy=none if not explicitly overridden by a CLI flag.
3. Azure CLI sends the modified cluster object to AKS, which fails on validation that networkPolicy=cilium when networkDataplane=cilium.

Update Azure Preview CLI to force `--network-policy=cilium` when it is not specified and when `--network-dataplane=cilium` allows the validation to succeed. 

---

### Related command
`az aks update --network-dataplane=cilium`

### General Guidelines

### DIFF
--- a/src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
+++ b/src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
@@ -68,6 +68,7 @@ from azext_aks_preview._consts import (
     CONST_NETWORK_PLUGIN_AZURE,
     CONST_NETWORK_PLUGIN_MODE_OVERLAY,
     CONST_NETWORK_DATAPLANE_CILIUM,
+    CONST_NETWORK_POLICY_CILIUM,
     CONST_PRIVATE_DNS_ZONE_NONE,
     CONST_PRIVATE_DNS_ZONE_SYSTEM,
     CONST_AZURE_KEYVAULT_SECRETS_PROVIDER_ADDON_NAME,
@@ -3436,6 +3437,9 @@ class AKSPreviewManagedClusterUpdateDecorator(AKSManagedClusterUpdateDecorator):
         network_policy = self.context.get_network_policy()
         if network_policy:
             mc.network_profile.network_policy = network_policy
+
+        if network_dataplane == CONST_NETWORK_DATAPLANE_CILIUM:
+            mc.network_profile.network_policy = CONST_NETWORK_POLICY_CILIUM
 
         return mc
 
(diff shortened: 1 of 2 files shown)


yonzhan:
AKS

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -3437,6 +3438,9 @@ def update_network_plugin_settings(self, mc: ManagedCluster) -> ManagedCluster:
         if network_policy:
             mc.network_profile.network_policy = network_policy
 
+        if network_dataplane == CONST_NETWORK_DATAPLANE_CILIUM:
```
wedaly, line 3441: suggest setting this only when `network_policy` isn't explicitly set.

 ```python
network_policy = self.context.get_network_policy()
if network_policy:
    mc.network_profile.network_policy = network_policy
elif network_dataplane == CONST_NETWORK_DATAPLANE_CILIUM:
    mc.network_profile.network_policy = CONST_NETWORK_POLICY_CILIUM
 ```
 
 preferrably with a comment explaining why this is necessary.

robogatikov: Thank you, updated accordingly

robogatikov:
> Hi @robogatikov, Please write the description of changes which can be perceived by customers into HISTORY.rst. If you want to release a new extension version, please update the version in setup.py as well.

I don't think there is anything in this pr that needs to be announced to customers

FumingZhang (approved):
LGTM

yanzhudd:
> This causes the command to upgrade to Cilium dataplane to fail with a validation error when --network-policy isn't set explicitly to "cilium".

if convenient, it is suggested to add test for this scenario and run it to verify this fix.

robogatikov:
> > This causes the command to upgrade to Cilium dataplane to fail with a validation error when --network-policy isn't set explicitly to "cilium".
> 
> if convenient, it is suggested to add test for this scenario and run it to verify this fix.

@yanzhudd, yes, sure. The test that I modified covers this scenario: we don't set `--network-policy` when upgrading (it is an empty string), and then we verify that network policy gets set to "cilium" since network dataplane is "cilium"

yanzhudd (approved):

PR #7085, opened 2023-12-14
[containerapp] Fix telemetry collection when using local builder

cormacpayne:
Currently, when `containerapp create/update/up` are called and the provided source is built locally using the builder, the previously set value for `collect_telemetry` within the Azure CLI is not respected when running the builder. This PR will set the `ORYX_DISABLE_TELEMETRY` environment variable on the builder if it's found that the user previously disabled telemetry collection.

---

### Related command

az containerapp create
az containerapp update
az containerapp up

### General Guidelines

For new extensions:

~- [ ] My extension description/summary conforms to the [Extension Summary Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/extensions/extension_summary_guidelines.md).~

### About Extension Publish

### DIFF
--- a/src/containerapp/azext_containerapp/_up_utils.py
+++ b/src/containerapp/azext_containerapp/_up_utils.py
@@ -44,6 +44,7 @@ from azure.cli.command_modules.containerapp._utils import (
     register_provider_if_needed,
     format_location
 )
+from azure.cli.core import telemetry as telemetry_core
 from azure.core.exceptions import HttpResponseError
 from azure.mgmt.containerregistry import ContainerRegistryManagementClient
 from knack.log import get_logger
@@ -560,16 +561,20 @@ class ContainerApp(Resource):  # pylint: disable=too-many-instance-attributes
         except Exception as ex:
             raise ValidationError(f"Unable to run 'pack config' command to set default builder: {ex}") from ex
 
-        # If the user specifies a target port, pass it to the buildpack
-        if self.target_port:
-            command.extend(['--env', f"ORYX_RUNTIME_PORT={self.target_port}"])
-
         logger.warning("Selecting a compatible builder for the provided application source...")
         could_build_image = False
         for builder_image in builder_image_list:
             # Run 'pack build' to produce a runnable application image for the Container App
             # Specify the image as the 'build-cache' image to ensure that the local cache is used for build layers
             command = [pack_exec_path, 'build', cache_image_name, '--builder', builder_image, '--path', source, '--tag', image_name]
+
+            # If the user specifies a target port, pass it to the buildpack
+            if self.target_port:
+                command.extend(['--env', f"ORYX_RUNTIME_PORT={self.target_port}"])
+
+            # If the user has disabled telemetry collection for the CLI, propagate that to the builder
+            if not telemetry_core.is_telemetry_enabled():
+                command.extend(['--env', 'ORYX_DISABLE_TELEMETRY=true'])
 
             logger.debug(f"Calling '{' '.join(command)}'")
             try:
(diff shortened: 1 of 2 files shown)


src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -560,17 +561,21 @@ def build_container_from_source_with_buildpack(self, image_name, source, cache_i
...
         logger.warning("Selecting a compatible builder for the provided application source...")
         could_build_image = False
         for builder_image in builder_image_list:
             # Run 'pack build' to produce a runnable application image for the Container App
             # Specify the image as the 'build-cache' image to ensure that the local cache is used for build layers
             command = [pack_exec_path, 'build', cache_image_name, '--builder', builder_image, '--path', source, '--tag', image_name]
 
+            # If the user specifies a target port, pass it to the buildpack
+            if self.target_port:
+                command.extend(['--env', f"ORYX_RUNTIME_PORT={self.target_port}"])
```
daniv-msft, line 573: I'm curious about that part of the change. Does that mean that we were not setting this ORYX_RUNTIME_PORT anymore as we were reassigning command?

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -560,17 +561,21 @@ def build_container_from_source_with_buildpack(self, image_name, source, cache_i
...
+
+            # If the user has disabled telemetry collection for the CLI, propagate that to the builder
+            if not telemetry_core.is_telemetry_enabled():
+                command.extend(['--env', 'ORYX_DISABLE_TELEMETRY=true'])
```
daniv-msft, line 577: I assume that you tried this but just to be sure: could you please confirm that we can pass multiple `--env` values without problem?

cormacpayne: That's correct, you can provide multiple `--env` values to `pack build` -- we do this in a couple of our client tools and per their help documentation:

_This flag may be specified multiple times_

src/containerapp/azext_containerapp/_up_utils.py
cormacpayne: It looks like the `command` variable was being overridden _after_ the target port was being set, so the target port was never being propagated to the builder. This change ensures that both the target port and telemetry environment variables are appended to the `command` variable we construct just above.

daniv-msft: Yes exactly. Basically, I'm surprised that we didn't hear anyone reporting that issue yet. Thanks for catching it!

src/containerapp/azext_containerapp/_up_utils.py
daniv-msft: Thanks for confirming!

daniv-msft (approved):

cormacpayne:
For the reviewer:

It looks like the CI is failing for a couple of reasons:
- The `azdev style containerapp` is failing in the pipeline, but when I run it locally against `main` and my branch for this PR, both report the same score, so it doesn't appear any style changes have been added/removed as a part of this PR, so I'm not sure why that part of CI is failing
- The Python integration tests are failing for a test that's independent from the changes I made, so I'm not sure if it was a flaky test and should be re-ran

yanzhudd:
> For the reviewer:
> 
> It looks like the CI is failing for a couple of reasons:
> 
> * The `azdev style containerapp` is failing in the pipeline, but when I run it locally against `main` and my branch for this PR, both report the same score, so it doesn't appear any style changes have been added/removed as a part of this PR, so I'm not sure why that part of CI is failing
> * The Python integration tests are failing for a test that's independent from the changes I made, so I'm not sure if it was a flaky test and should be re-ran

please try to rerun the failed test in live mode and upload the re-recording file.

PR #7087, opened 2023-12-14
upgrade connectedk8s to use Microsoft Graph

TheOnlyWei:
---

### Related command
az connectedk8s

### General Guidelines

- [X] Have you run `python scripts/ci/test_index.py -q` locally? (`pip install wheel==0.30.0` required)

### DIFF
--- a/src/connectedk8s/HISTORY.rst
+++ b/src/connectedk8s/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+1.6.4
+++++++
+* replace AAD Graph with Microsoft Graph for getting Custom Locations RP service principal application object ID.
+
 1.6.3
 ++++++
 * Skip checking operation version support for non-stable release trains.
@@ -18,7 +22,6 @@ Release History
 
 1.6.0
 ++++++
-
 * Updated Helm from 3.6.3 to 3.12.2
 
 1.5.6
--- a/src/connectedk8s/azext_connectedk8s/custom.py
+++ b/src/connectedk8s/azext_connectedk8s/custom.py
@@ -29,6 +29,7 @@ from knack.util import CLIError
 from knack.log import get_logger
 from knack.prompting import prompt_y_n
 from knack.prompting import NoTTYException
+from azure.cli.command_modules.role import graph_client_factory
 from azure.cli.core.commands.client_factory import get_subscription_id
 from azure.cli.core.commands import LongRunningOperation
 from azure.cli.core._profile import Profile
@@ -40,10 +41,8 @@ from kubernetes import client as kube_client, config
 from Crypto.IO import PEM
 from Crypto.PublicKey import RSA
 from Crypto.Util import asn1
-from azext_connectedk8s._client_factory import _graph_client_factory
 from azext_connectedk8s._client_factory import cf_resource_groups
 from azext_connectedk8s._client_factory import resource_providers_client
-from azext_connectedk8s._client_factory import get_graph_client_service_principals
 from azext_connectedk8s._client_factory import cf_connected_cluster_prev_2022_10_01, cf_connected_cluster_prev_2023_11_01
 from azext_connectedk8s._client_factory import cf_connectedmachine
 import azext_connectedk8s._constants as consts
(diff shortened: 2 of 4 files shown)


yonzhan:

src/connectedk8s/HISTORY.rst
```diff
@@ -2,9 +2,13 @@
 
 Release History
 ===============
-1.6.0
```
NarayanThiru, line 5: @TheOnlyWei , can you share why you are making this change?  Can you update the description with this explanation and link to the ADO Task tracking this work?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -29,6 +29,7 @@
 from knack.log import get_logger
 from knack.prompting import prompt_y_n
 from knack.prompting import NoTTYException
+from azure.cli.command_modules.role import graph_client_factory
```
NarayanThiru, line 32: With this change to use MS Graph, does the customer/client need different permissions?  Specifically, is it possible for this to fail, for someone for whom it used to work before?  This will become a regression - hence asking.

NarayanThiru (changes requested):

src/connectedk8s/HISTORY.rst
TheOnlyWei: Replied in Teams chat with details.

src/connectedk8s/azext_connectedk8s/custom.py
TheOnlyWei: The customer will need MS Graph permission for the extension to get the object ID for them if they don't want to pass it in. There is no way around this in an automated way without using MS Graph, since AAD Graph is deprecated. The behavior will be the same, in that if they don't have permission, it will try to use the one passed in, and if the customer didn't pass anything in, then Custom Locations RP won't be enabled. The AAD Graph delegated user permissions should be automatically converted to the equivalent MS Graph user permissions. This call specifically uses the list API, and the permissions are specified below:

https://learn.microsoft.com/en-us/graph/api/serviceprincipal-list?view=graph-rest-1.0&tabs=http#permissions

src/connectedk8s/HISTORY.rst
```diff
@@ -2,9 +2,13 @@
...
+
+1.6.1
```
bganapa, line 6: Version needs update

src/connectedk8s/setup.py
```diff
@@ -17,7 +17,7 @@
 # TODO: Confirm this is the right version number you want and it matches your
 # HISTORY.rst entry.
 
-VERSION = '1.6.0'
+VERSION = '1.6.1'
```
bganapa, line 20: version update

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -2199,14 +2198,13 @@ def check_cl_registration_and_get_oid(cmd, cl_oid, subscription_id):
...
     try:
-        sp_graph_client = get_graph_client_service_principals(cmd.cli_ctx)
-        sub_filters = []
-        sub_filters.append("appId eq '{}'".format("bc313c14-388c-4e7d-a58e-70017303ee3b"))  # Lookup using well-known first-party application id
-        result = list(sp_graph_client.list(filter=(' and '.join(sub_filters))))
-        if len(result) != 0:
-            if cl_oid is not None and cl_oid != result[0].object_id:
-                logger.debug("The 'Custom-locations' OID passed is different from the actual OID({}) of the Custom Locations RP app. Proceeding with the correct one...".format(result[0].object_id))
-            return result[0].object_id  # Using the fetched OID
+        graph_client = graph_client_factory(cmd.cli_ctx)
```
bganapa, line 2201: CLI is already making use of MS graph in other places, you are just reusing the same here . Is that correct? @TheOnlyWei

src/connectedk8s/HISTORY.rst
bganapa: Lets use the PR for discussion @TheOnlyWei for why the change is needed

bganapa:
The PR needs @julienstroheker's approval. Please dont merge it without his approval

bganapa:
@TheOnlyWei Can you please list what manual validations are done against azure to cover the changes?

src/connectedk8s/HISTORY.rst
TheOnlyWei: @bganapa I gave Narayan internal links in Teams Chat, so didn't make sense to link them in a public discussion.

But the reason I am making this change, is because AAD graph is deprecated.

src/connectedk8s/azext_connectedk8s/custom.py
TheOnlyWei: I am using the Azure CLI MS Graph client that they use for their RBAC module.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -2257,14 +2256,13 @@ def check_cl_registration_and_get_oid(cmd, cl_oid, subscription_id):
...
+        app_id = "bc313c14-388c-4e7d-a58e-70017303ee3b"
+        app_object = graph_client.service_principal_list(filter="appId eq '{}'".format(app_id))
```
julienstroheker, line 2261: Should we handle the case where caller does not have `Application.Read.All` permission on graph ?

IMO this should be wrap into a `try` `catch` block and returns an actionable error message.

julienstroheker: just checked the warning message the next line, we might just need to update the doc and add a prerequisites about this.

Remind me the current flow, if we do return `""` is the whole operation is failing ?

TheOnlyWei: Custom Locations is enabled by default. If user doesn't have permission for retrieving the service principal via MS Graph, then it won't be enabled, and a warning message will be printed, which is the current behavior.

julienstroheker: Make senses. Is just that would be great to provide more insights about the real failure here, if this is a graph permission issue or just that the call did returns no `oid`.
Can we ask PM to edit the doc : https://aka.ms/CustomLocationsObjectID to makes sure we ask customer to make sure they are having the `Application.Read.All` permission.

julienstroheker: Actually I just checked and looks like `Application.Read.All` is granted by default on any default AAD User

TheOnlyWei: For service principals, they need to be granted the permission explicitly by an admin.

bganapa (approved):

yanzhudd:
please rerun related tests in live mode to verify this change.

TheOnlyWei:
Hi @yanzhudd, we have already run the live tests.

yanzhudd (approved):

PR #7103, opened 2023-12-19
Properly handle k8s clusters of kind provisioned cluster for connectedk8s

sidiesen:
The new kind ProvisionedCluster is being operated via a separate az CLI extension (az arck8s).
Therefore we need to update the connectedk8s extension to disable several subcommands that are not supported by the connectedk8s extension for provisioned clusters.

In order to achieve this, we need to:

update some of the service calls to a newer API version that supports the "kind" property
display an appropriate error message is a disallowed command is being executed against a provisioned cluster

---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -315,14 +315,23 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
         configmap_rg_name = configmap.data["AZURE_RESOURCE_GROUP"]
         configmap_cluster_name = configmap.data["AZURE_RESOURCE_NAME"]
         if connected_cluster_exists(client, configmap_rg_name, configmap_cluster_name):
+            preview_cluster_resource = None
+            public_key = None
```
NarayanThiru, line 319: Why are you setting these, if you are anyway overwriting them later, in lines 322 and 324?

Looks like earlier there was some reason to reset these.  Do you know what was it?  Can you confirm that is not broken with your change?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -315,14 +315,23 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
...
+
+            try:
+                preview_cluster_resource = get_connectedk8s_2023_11_01(cmd, configmap_rg_name,
```
NarayanThiru, line 322: >get_connectedk8s_2023_11_01 [](http://example.com/codeflow?start=43&length=27)

If the cluster doesn't exist, this will return null, right?  So, check the value of preview_cluster_resource, before using its properties.  Do the same in all the places where you are calling get_connectedk8s_2023_11_01.  There are 4 or 5 more instances

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -418,6 +430,25 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
                                custom_locations_oid, helm_client_location, enable_private_link, arm_metadata, onboarding_timeout, container_log_path)
     return put_cc_response
 
+def validate_existing_provisioned_cluster_for_reput(cluster_resource, kubernetes_distro, kubernetes_infra, enable_private_link, private_link_scope_resource_id, distribution_version, azure_hybrid_benefit, location):
+    if cluster_resource.kind == "ProvisionedCluster":
```
NarayanThiru, line 434: >cluster_resource.kind [](http://example.com/codeflow?start=7&length=21)

To be safe, convert the value to lowercase and compare against 'provisionedcluster', instead.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -418,6 +430,25 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
...
+        if azure_hybrid_benefit is not None:
+            raise InvalidArgumentValueError("Updating the 'azure hybrid benefit' property of a Provisioned Cluster is not supported from the Connected Cluster CLI. Please use the 'az aksarc update' CLI command.\nhttps://learn.microsoft.com/en-us/cli/azure/aksarc?view=azure-cli-latest#az-aksarc-update")
```
NarayanThiru, line 436: >Please use the 'az aksarc update' CLI command [](http://example.com/codeflow?start=164&length=45)

Why not leave this out as well?  The link can explain this too.  And, any time there is an update, it will be only in the documentation - the code need not change.  Discuss with the PM too and see if this makes sense.

NarayanThiru, line 436: >azure hybrid benefit [](http://example.com/codeflow?start=59&length=20)

Any reason for explicit validation and error message for this property only?  Especially since you are validating this as part of the validation_values array as well?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -315,14 +315,23 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
...
                     configmap_cluster_name.lower() == cluster_name.lower()):
                 # Re-put connected cluster
-                try:
-                    public_key = client.get(configmap_rg_name,
-                                            configmap_cluster_name).agent_public_key_certificate
-                except Exception as e:  # pylint: disable=broad-except
-                    utils.arm_exception_handler(e, consts.Get_ConnectedCluster_Fault_Type, 'Failed to check if connected cluster resource already exists.')
+                
+                # If cluster is of kind provisioned cluster, there are several properties that cannot be updated
+                validate_existing_provisioned_cluster_for_reput(preview_cluster_resource, kubernetes_distro, kubernetes_infra, enable_private_link, private_link_scope_resource_id, distribution_version, azure_hybrid_benefit, location)
```
NarayanThiru, line 333: Why not move this validation outside the if block, since you are doing this in both if and else cases?

src/connectedk8s/azext_connectedk8s/custom.py
sidiesen: On the one hand this is purely for code readability - the values of these variables are set in a try block, where any error happening might cause these variables not to be set to the expected value - this way it would be obvious on first glance what default value is assigned to the variable in such error cases.
Additionally, while I know it is not not necessary for Python, it establishes the "variable scope" outside of the try block, which in my opinion also improves on the readability

src/connectedk8s/azext_connectedk8s/custom.py
sidiesen: If the cluster doesn't exist, this will raise an error with the failing HTTTP status, not just return null.
It looks like this:
image

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -331,6 +340,9 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
                     disable_cluster_connect(cmd, client, resource_group_name, cluster_name, kube_config, kube_context, values_file, release_namespace, helm_client_location)
                 return cc_response
             else:
+                # If cluster is of kind provisioned cluster, there are several properties that cannot be updated
+                validate_existing_provisioned_cluster_for_reput(preview_cluster_resource, kubernetes_distro, kubernetes_infra, enable_private_link, private_link_scope_resource_id, distribution_version, azure_hybrid_benefit, location)    
+
```
NarayanThiru, line 345: Actually, why is this validation needed, since this seems to be the error case.  Moreover, you can probably do this error check first in an if block and move the current if block out, below the new if block - as the default.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -859,6 +896,11 @@ def delete_connectedk8s(cmd, client, resource_group_name, cluster_name,
 
     logger.warning("This operation might take a while ...\n")
 
+    # Check if the cluster is of supported type for deletion
+    preview_cluster_resource = get_connectedk8s_2023_11_01(cmd, resource_group_name, cluster_name)
+    if preview_cluster_resource.kind == "ProvisionedCluster":
```
NarayanThiru, line 901: >preview_cluster_resource.kind [](http://example.com/codeflow?start=7&length=29)

Check preview_cluter_resource is not null before using and convert kind to lowercase and compare against 'provisionedcluster'.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1006,6 +1048,12 @@ def update_connected_cluster(cmd, client, resource_group_name, cluster_name, htt
 
     proxy_cert = proxy_cert.replace('\\', r'\\\\')
 
+    # Fetch Connected Cluster for agent version
+    connected_cluster = get_connectedk8s_2023_11_01(cmd, resource_group_name, cluster_name)
+
+    if connected_cluster.kind == "ProvisionedCluster":
```
NarayanThiru, line 1054: >connected_cluster.kind [](http://example.com/codeflow?start=7&length=22)

Same as earlier, null check and lowercase comparison

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1149,6 +1197,12 @@ def update_connected_cluster(cmd, client, resource_group_name, cluster_name, htt
 
 
 def upgrade_agents(cmd, client, resource_group_name, cluster_name, kube_config=None, kube_context=None, arc_agent_version=None, upgrade_timeout="600"):
+    # Check if cluster supports upgrading
+    connected_cluster = get_connectedk8s_2023_11_01(cmd, resource_group_name, cluster_name)
+
+    if connected_cluster.kind == "ProvisionedCluster":
```
NarayanThiru, line 1203: >connected_cluster.kind == "ProvisionedCluster" [](http://example.com/codeflow?start=7&length=46)

One more, null check and lowercase comparison

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1395,7 +1449,11 @@ def enable_features(cmd, client, resource_group_name, cluster_name, features, ku
     enable_cluster_connect, enable_azure_rbac, enable_cl = utils.check_features_to_update(features)
 
     # Check if cluster is private link enabled
-    connected_cluster = get_connectedk8s(cmd, client, resource_group_name, cluster_name)
+    connected_cluster = get_connectedk8s_2023_11_01(cmd, resource_group_name, cluster_name)
+
+    if connected_cluster.kind == "ProvisionedCluster":
```
NarayanThiru, line 1454: >connected_cluster.kind == "ProvisionedCluster [](http://example.com/codeflow?start=7&length=45)

Another one of the same

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -1520,6 +1578,12 @@ def disable_features(cmd, client, resource_group_name, cluster_name, features, k
     confirmation_message = "Disabling few of the features may adversely impact dependent resources. Learn more about this at https://aka.ms/ArcK8sDependentResources. \n" + "Are you sure you want to disable these features: {}".format(features)
     utils.user_confirmation(confirmation_message, yes)
 
+    # Fetch Connected Cluster for agent version
+    connected_cluster = get_connectedk8s_2023_11_01(cmd, resource_group_name, cluster_name)
+
+    if connected_cluster.kind == "ProvisionedCluster":
```
NarayanThiru, line 1584: >connected_cluster.kind == "ProvisionedCluster [](http://example.com/codeflow?start=7&length=45)

Another one

src/connectedk8s/azext_connectedk8s/_client_factory.py
```diff
@@ -46,6 +46,19 @@ def cf_connected_cluster_prev_2022_10_01(cli_ctx, _):
     return cf_connectedk8s_prev_2022_10_01(cli_ctx).connected_cluster
 
 
+def cf_connectedk8s_prev_2023_11_01(cli_ctx, *_):
+    from azext_connectedk8s.vendored_sdks.preview_2023_11_01 import ConnectedKubernetesClient
+    if os.getenv('AZURE_ACCESS_TOKEN'):
```
NarayanThiru, line 51: >AZURE_ACCESS_TOKEN [](http://example.com/codeflow?start=18&length=18)

This literal is used in so many places.  Move this to a constant?

NarayanThiru:
I assume you did not make any changes to the files under vendored_sdks?  Since these are generated files, I did not review them.  If you have made any changes, let me know and I'll review the changes

NarayanThiru:
Are there tests that need to be updated?

src/connectedk8s/azext_connectedk8s/custom.py
sidiesen: It was the guidance from a PM to have this error message separately listed - I shared a design document with you in a separate thread for your reference.

src/connectedk8s/azext_connectedk8s/custom.py
NarayanThiru: Have you confirmed that it errors out when invoked from within the CLI too?  If so, yes, you don't need to check for null.  In that case, is that error message thrown sufficient for the user?

sidiesen: Yes, the screenshot I shared above was me invoking that code path for the connectedk8s show subcommand with a private build of the connectedk8s extension containing my changes.
I think the error message is sufficient enough to convey to the user that the referenced cluster resource doesn't exist.

src/connectedk8s/azext_connectedk8s/custom.py
sidiesen: you are right, that validation was not put there on purpose - there is already an appropriate error message thrown

src/connectedk8s/azext_connectedk8s/custom.py
sidiesen: see comment above

src/connectedk8s/azext_connectedk8s/custom.py
sidiesen: Yes, that screenshot I shared above was from when I invoked that code path via CLI locally - this was done with a private build of the connectedk8s extension containing my changes.
I think the message thrown is sufficient, it is similar behavior as what it was previously.

NarayanThiru (approved):
:shipit:

yonzhan:
Please fix CI issues.

yonzhan:
Add @wangzelin007 to help with

wangzelin007:
You can fix these issues later, I've added a new warning message.
I'm sorry for the confusion.
image

src/connectedk8s/azext_connectedk8s/_client_factory.py
```diff
@@ -22,9 +22,9 @@
 
 def cf_connectedk8s(cli_ctx, *_):
     from azext_connectedk8s.vendored_sdks import ConnectedKubernetesClient
-    if os.getenv('AZURE_ACCESS_TOKEN'):
+    if os.getenv(consts.Azure_Access_Token_Variable):
```
wangzelin007, line 25: Will changing from reading from environment variables to reading from configuration files have an impact on the user experience?
Do users have to manually fill in `Azure_Access_Token_Variable` in` _constants.py`?

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -418,6 +427,25 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
...
+            kubernetes_infra,
+            converted_priv_link_value,
+            private_link_scope_resource_id,
+            distribution_version,
+            azure_hybrid_benefit,
+            location,
+        ]
+
+        for value in validation_values:
+            if not value is None:
```
wangzelin007, line 446: If `value is not None` is better and more readable.

src/connectedk8s/azext_connectedk8s/custom.py
```diff
@@ -418,6 +427,25 @@ def create_connectedk8s(cmd, client, resource_group_name, cluster_name, correlat
                                custom_locations_oid, helm_client_location, enable_private_link, arm_metadata, onboarding_timeout, container_log_path)
     return put_cc_response
 
+def validate_existing_provisioned_cluster_for_reput(cluster_resource, kubernetes_distro, kubernetes_infra, enable_private_link, private_link_scope_resource_id, distribution_version, azure_hybrid_benefit, location):
+    if (cluster_resource != None) and (cluster_resource.kind.lower() == "provisionedcluster"):
```
wangzelin007, line 431: It is recommended to set `provisionedcluster` to a constant, which will facilitate unified modification in the future.

src/connectedk8s/azext_connectedk8s/_client_factory.py
sidiesen: the value is still being read from environment variables - it is only the (hardcoded) environment variable name 'AZURE_ACCESS_TOKEN' that moved to constants as per Narayan's previous comments on this PR.
There is no change for users.

src/connectedk8s/azext_connectedk8s/custom.py
sidiesen: good catch, changed it

src/connectedk8s/azext_connectedk8s/custom.py
sidiesen: updated the PR and made those changes

wangzelin007 (approved):

wangzelin007 (approved):

PR #7131, opened 2023-12-28
{CI} Add build wheel check and extension name check

wangzelin007:
---

### Related command

The name in setup.py should be the same as the extension dir name.
Otherwise the test case test_index will report an error:
image

At the same time, we should restrict setup.py to not allow the use of underscores, otherwise the version information of the build package cannot be obtained after user install the extension.
image

Only if the name contains an underscore, the version information cannot be displayed. This problem only appeared recently, and only packages built on the ci environment have problems, which cannot be reproduced in the local environment. The root cause has not yet been found.

Add build wheel check and extension name check to avoid the same error:
image

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
CI

AllyW (approved):

jsntcy (approved):

PR #7293, opened 2024-02-16
[ContainerApp] Bugfix - Remove pre-existing containers for source to cloud flow 

snehapar9:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

src/containerapp/azext_containerapp/containerapp_decorator.py
```diff
@@ -1012,9 +1013,19 @@ def construct_payload(self):
         self.set_up_service_bindings()
         self.set_up_unbind_service_bindings()
         self.set_up_source()
+        self.set_up_containers_for_private_registry_image()
         if self.get_argument_max_inactive_revisions() is not None:
             safe_set(self.new_containerapp, "properties", "configuration", "maxInactiveRevisions", value=self.get_argument_max_inactive_revisions())
 
+    def set_up_containers_for_private_registry_image(self):
```
daniv-msft, line 1020: Nit: please consider renaming "private registry" to "default registry" as it's the wording we're pushing for now (and it is what is used in the UI)

src/containerapp/azext_containerapp/containerapp_decorator.py
```diff
@@ -1012,9 +1013,19 @@ def construct_payload(self):
...
+        containers = safe_get(self.new_containerapp, "properties", "template", "containers", default=[])
+        # Check for private registry image and if found remove the containers from the container app and ensure there is only one container with the build image.
+        for container in containers:
```
daniv-msft, line 1023: The tricky part is that in that case, we actually want to remove all previous containers, not only the non-Cloud Build ones.
We also can't rely on the use of the default container registry or not, because soon we will have support for BYO ACR and they can be stored anywhere.

I remember that you mentioned that there was an update/append option in the CLI. Could you please share where this is? This would seem to me like the right approach.
The problem is that here we're already at the containerapp_decorator level, so we lost all information about whether we're in a Cloud Build context or not.

snehapar9: It may be better to remove all previous containers in `containerapp_up` [here](https://github.com/Azure/azure-cli-extensions/blob/main/src/containerapp/azext_containerapp/custom.py#L1193) since we determine if cloud build scenario or not here. 

[This](https://github.com/Azure/azure-cli-extensions/blob/main/src/containerapp/azext_containerapp/containerapp_decorator.py#L305) is the code flow which determines if new container must be appended or replaced, however this flow is also in the containerapp_decorator class so we would not have any information about Cloud Build or not. So it may be better to remove all containers even before updating the container app.

src/containerapp/azext_containerapp/containerapp_decorator.py
```diff
@@ -1084,7 +1086,7 @@ def _update_container_app_source(self, cmd, source, build_env_vars, registry_ser
         app.run_source_to_cloud_flow(source, dockerfile, build_env_vars, can_create_acr_if_needed=False, registry_server=registry_server)
 
         # Validate an image associated with the container app exists
-        containers = safe_get(self.containerapp_def, "properties", "template", "containers", default=[])
+        containers = safe_get(self.new_containerapp, "properties", "template", "containers", default=[])
```
daniv-msft, line 1089: Could you please clarify where is this new_containerapp defined? Do we even need this with your other change above?

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -521,7 +527,7 @@ def build_container_from_source_with_cloud_build_service(self, source, build_env
 
         run_full_id = uuid.uuid4().hex
         logs_file_path = os.path.join(tempfile.gettempdir(), f"{'build{}'.format(run_full_id)[:12]}.txt")
-        logs_file = open(logs_file_path, "w")
+        logs_file = open(logs_file_path, "w", encoding="utf-8")
```
daniv-msft, line 530: Is the `encoding` part coming from this other PR? https://github.com/Azure/azure-cli-extensions/pull/7281/files
If yes, should we remove it from here (or take the rest of the changes so that we can close the other PR)?

src/containerapp/azext_containerapp/containerapp_decorator.py
```diff
@@ -3,6 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 # pylint: disable=line-too-long, consider-using-f-string, no-else-return, duplicate-string-formatting-argument, expression-not-assigned, too-many-locals, logging-fstring-interpolation, broad-except, pointless-statement, bare-except
+from re import match
```
daniv-msft, line 6: Could you please clarify where is this one used?

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -410,6 +410,7 @@ def __init__(
         env_vars=None,
         workload_profile_name=None,
         ingress=None,
+        use_default_container_registry=None
```
daniv-msft, line 413: Please consider renaming to `force_single_container_updates` or something of that kind.
Here `use_default_container_registry` won't work once we have BYO ACR. Also, it would make more sense in the containerapp_decorator.py to have something like `force_single_container_updates` (the default container registry is not a concern of the decorator)

src/containerapp/azext_containerapp/_up_utils.py
Greedygre: I left a comment for this PR https://github.com/Azure/azure-cli-extensions/pull/7281/files early this Week, about if there is any test cover this issue.
Do we have test case to cover this issue when code relies on UTF-8 logs?

src/containerapp/azext_containerapp/containerapp_decorator.py
snehapar9: Thanks Dan! This change is not valid anymore in my latest commits.

src/containerapp/azext_containerapp/containerapp_decorator.py
snehapar9: This change is not related to the bugfix in this PR but I think it makes sense to change this to `self.new_containerapp` instead because this payload was already populated before it reached this flow (since this is the preview update flow) so keeping it as `self.containerapp_def` might not have the changes from `self.new_containerapp` in some cases.

This is not a regression so I can remove this and create a separate PR if it makes more sense.

src/containerapp/azext_containerapp/_up_utils.py
snehapar9: Merged these [changes](https://github.com/Azure/azure-cli-extensions/pull/7281/files) into this PR.

src/containerapp/azext_containerapp/containerapp_decorator.py
snehapar9: This is not used anywhere with the latest changes so removed it.

src/containerapp/azext_containerapp/_up_utils.py
snehapar9: Updated.

snehapar9:
Fixed linting errors for all the files that are a part of this PR but there are some linting errors in some un-related files that has not been fixed in this PR.

daniv-msft (approved):

src/containerapp/azext_containerapp/_up_utils.py
```diff
@@ -410,6 +410,7 @@ def __init__(
         env_vars=None,
         workload_profile_name=None,
         ingress=None,
+        force_single_container_updates=None
```
Greedygre, line 413: You add a new argument `force_single_container_updates`?
Need to update `_params.py` and add description about it.

Greedygre: Also, please add an example for the new argument in `_help.py`.

src/containerapp/azext_containerapp/containerapp_decorator.py
```diff
@@ -1084,7 +1096,7 @@ def _update_container_app_source(self, cmd, source, build_env_vars, registry_ser
         app.run_source_to_cloud_flow(source, dockerfile, build_env_vars, can_create_acr_if_needed=False, registry_server=registry_server)
 
         # Validate an image associated with the container app exists
-        containers = safe_get(self.containerapp_def, "properties", "template", "containers", default=[])
+        containers = safe_get(self.new_containerapp, "properties", "template", "containers", default=[])
```
Greedygre, line 1099: This is **Validate an image associated with the container app exists**, so it get existing containers from `self.containerapp_def`, which is the existing containerapp info from `client.show()`.

If there some logic can make sure the `new.containerapp` contains all existing containers?

src/containerapp/azext_containerapp/containerapp_decorator.py
```diff
@@ -222,8 +225,13 @@ def construct_payload(self):
...
             # Check if updating existing container
             updating_existing_container = False
+            if self.get_argument_force_single_container_updates():
+                # Remove n-1 containers where n is the number of containers in the containerapp and replace remaining container with the new container
+                # Fails if all containers are removed
+                while (len(self.new_containerapp["properties"]["template"]["containers"]) > 1):
+                    self.new_containerapp["properties"]["template"]["containers"].pop()
             for c in self.new_containerapp["properties"]["template"]["containers"]:
-                if c["name"].lower() == self.get_argument_container_name().lower():
+                if c["name"].lower() == self.get_argument_container_name().lower() or self.get_argument_force_single_container_updates():
```
Greedygre, line 234: Can you add some description about this change in here about this argument?

src/containerapp/azext_containerapp/containerapp_decorator.py
```diff
@@ -222,8 +225,13 @@ def construct_payload(self):
 
             # Check if updating existing container
             updating_existing_container = False
+            if self.get_argument_force_single_container_updates():
```
Greedygre, line 228: This logic is for preview feature and not related to any GA feature, should not be added in `ContainerAppUpdateDecorator`.
If you need to reuse this part of logic for update_containers, please split it as follow example then in `ContainerAppPreviewUpdateDecorator` over write it.:
```

def should_update_existing_container(self, curent_container_name=None):
  return curent_container_name.lower() == self.get_argument_container_name().lower() or self.get_argument_force_single_container_updates()

def set_up_updating_existing_container():
    # using should_update_existing_container
    if self.should_update_existing_container():


```

src/containerapp/azext_containerapp/containerapp_decorator.py
```diff
@@ -222,8 +225,13 @@ def construct_payload(self):
...
+                # Remove n-1 containers where n is the number of containers in the containerapp and replace remaining container with the new container
+                # Fails if all containers are removed
+                while (len(self.new_containerapp["properties"]["template"]["containers"]) > 1):
+                    self.new_containerapp["properties"]["template"]["containers"].pop()
```
Greedygre, line 232: When will the value of `force_single_container_updates` be `True`?

I check the previous logic, currently only when `No container registry provided. Let's use the default container registry through Cloud Build.`, the value can be `True`.
Why you pop all other containers? Will it cause existing containers delete that customer not aware to delete? @snehapar9  @daniv-msft 
```
        if self.should_create_acr:
            # No container registry provided. Let's use the default container registry through Cloud Build.
            self.image = self.build_container_from_source_with_cloud_build_service(source, build_env_vars, location)
            return True
```

src/containerapp/azext_containerapp/containerapp_decorator.py
snehapar9: @Greedygre, `self.containerapp_def` is populated from `client.show()` and `new.containerapp` is populated from `self.containerapp_def` even before it reaches this code flow [here](https://github.com/Azure/azure-cli-extensions/blob/main/src/containerapp/azext_containerapp/containerapp_decorator.py#L214). So, `new_containerapp` should have all the containers? Please let me know if this is not the case.

src/containerapp/azext_containerapp/containerapp_decorator.py
snehapar9: Added a comment about this argument in my latest commit.

src/containerapp/azext_containerapp/containerapp_decorator.py
snehapar9: @Greedygre could you please elaborate on how splitting this would ensure the change is only introduced in the `ContainerAppPreviewUpdateDecorator`? 

Since the payload for `self.new_containerapp`  is populated in `ContainerAppUpdateDecorator` , there would be no way to properly determine cloud build scenario to remove all the existing containers in `ContainerAppPreviewUpdateDecorator`

src/containerapp/azext_containerapp/containerapp_decorator.py
snehapar9: @Greedygre the goal here is to delete all existing containers for cloud build scenarios to ensure the containerapp only has one container with the newly built image. 
The `force_single_container_updates` argument is set only for cloud build scenarios.

src/containerapp/azext_containerapp/containerapp_decorator.py
Greedygre: So it should use `self.containerapp_def` because no one will change it, but `new_containerapp` may be changed somehow, right?
What is the benefit to use `new_containerapp`?

src/containerapp/azext_containerapp/containerapp_decorator.py
Greedygre: @snehapar9 

`set only for cloud build scenario` is here?

```
    if self.should_create_acr:
            # No container registry provided. Let's use the default container registry through Cloud Build.
            self.image = self.build_container_from_source_with_cloud_build_service(source, build_env_vars, location)
            return True
```

Is there any warning or info in CLI for customer or ant Doc that about `delete all existing containers for cloud build scenarios to ensure the containerapp only has one container with the newly built image.`?

snehapar9: Customers would not have control over it because they are using the cloud build scenario, so we want to make sure there is only the container with the image built by the source to cloud builder.
Not sure if it necessary for the customer to be warned? @daniv-msft any thoughts on this?

src/containerapp/azext_containerapp/containerapp_decorator.py
Greedygre: Here is an example. You can organize the code to make it more elegant.
In ContainerAppUpdateDecorator:
image

In ContainerAppPreviewUpdateDecorator, overwrite key logic that needs to be modified in the subclass

```
    def _need_update_container(self):
        return self.get_argument_image() or self.get_argument_force_single_container_updates() or self.get_argument_container_name() or self.get_argument_set_env_vars() is not None or self.get_argument_remove_env_vars() is not None or self.get_argument_replace_env_vars() is not None or self.get_argument_remove_all_env_vars() or self.get_argument_cpu() or self.get_argument_memory() or self.get_argument_startup_command() is not None or self.get_argument_args() is not None or self.get_argument_secret_volume_mount() is not None

    def set_up_containers(self):
        if self.get_argument_force_single_container_updates():
            # Remove n-1 containers where n is the number of containers in the containerapp and replace remaining container with the new container
            # Fails if all containers are removed
            while (len(self.new_containerapp["properties"]["template"]["containers"]) > 1):
                self.new_containerapp["properties"]["template"]["containers"].pop()
        super().set_up_containers()

    def should_update_existing_container(self, c) -> bool:
        return c["name"].lower() == self.get_argument_container_name().lower() or self.get_argument_force_single_container_updates()
```

src/containerapp/azext_containerapp/containerapp_decorator.py
daniv-msft: Here the ARM update to the container app will create a new revision, so if the customers encounter any issues they can always go back to their previous version. That said here in this command the customers are asking us to redeploy on their Container App with the source code provided. I don't think that many of them would be surprised to lose their current containers (and if they did they could always roll back to the previous revision).

src/containerapp/azext_containerapp/custom.py
```diff
@@ -1202,13 +1203,14 @@ def containerapp_up(cmd,
             _get_registry_from_app(app, source)  # if the app exists, get the registry
         _get_registry_details(cmd, app, source)  # fetch ACR creds from arguments registry arguments
 
-    used_default_container_registry = False
+    force_single_container_updates = False
     if source:
-        used_default_container_registry = app.run_source_to_cloud_flow(source, dockerfile, build_env_vars, can_create_acr_if_needed=True, registry_server=registry_server)
+        force_single_container_updates = app.run_source_to_cloud_flow(source, dockerfile, build_env_vars, can_create_acr_if_needed=True, registry_server=registry_server)
+        app._set_force_single_container_updates(force_single_container_updates)
```
Greedygre, line 1209: Warning: 
Access to a protected member _set_force_single_container_updates of a class 

Please change to `set_force_single_container_updates`

snehapar9: Updated.

src/containerapp/azext_containerapp/containerapp_decorator.py
snehapar9: @Greedygre Is there a way to make sure the preview code is executed first? The above proposed changes fails the test because the code in `ContainerAppPreviewUpdateDecorator` did not execute first before the code in `ContainerAppUpdateDecorator`.

daniv-msft (approved):

yonzhan:
Please fix CI issues.

src/containerapp/azext_containerapp/containerapp_decorator.py
```diff
@@ -585,6 +505,93 @@ def set_up_update_containerapp_yaml(self, name, file_name):
...
+                            # check that the only volume is of type secret
+                            volume_name = c["volumeMounts"][0]["volumeName"]
+                            for v in self.new_containerapp["properties"]["template"]["volumes"]:
+                                if v["name"].lower() == volume_name.lower():
+                                    if v["storageType"] != "Secret":
+                                        raise ValidationError(
+                                            "Usage error: --secret-volume-mount can only be used to update volume mounts with volumes of type secret. To update other types of volumes please use --yaml")
+                                    break
+                            c["volumeMounts"][0]["mountPath"] = self.get_argument_secret_volume_mount()
+            return updating_existing_container
```
Greedygre, line 590: should always `return updating_existing_container`, currently you only return updating_existing_container in the `For`

Greedygre, line 590:
```suggestion
        return updating_existing_container
```

snehapar9: Fixed in my latest commit

Greedygre:
LGTM, please rerun tests relate to `az containerapp up/update` with `--live`.

Greedygre (approved):

Greedygre:
Hi @zhoxing-ms  @yanzhudd 

Can you help to review and merge this PR when you have chance? Thanks.

Greedygre (approved):

Greedygre:
The credential Scan is Pass before.

image

zhoxing-ms (approved):

zhoxing-ms:
> The credential Scan is Pass before.

Known issue, @wangzelin007 has sent an email to `Microsoft Security Code Analysis Support` for help.

Greedygre:
Hi @wangzelin007 @zhoxing-ms 

Can you help to merge this PR? Thanks. Hope the Credential Scan not block this.

Greedygre:
Hi @wangzelin007 

https://github.com/Azure/azure-cli-extensions/commit/d6a8f0cb598bfa2e7801485a1c866d53709629a5

Can you help to review and merge this PR? I have remove all secrets in the yaml file, but the GitGuardian Security Checks is still failed. These secrets are always existing before, and they are deleted and out of date.

Thanks for your help.

wangzelin007:
> Hi @wangzelin007
> 
> [d6a8f0c](https://github.com/Azure/azure-cli-extensions/commit/d6a8f0cb598bfa2e7801485a1c866d53709629a5)
> 
> Can you help to review and merge this PR? I have remove all secrets in the yaml file, but the GitGuardian Security Checks is still failed. These secrets are always existing before, and they are deleted and out of date.
> 
> Thanks for your help.

Please fix the Integration Tests first, I can help bypass other security checks.

Greedygre:
> > Hi @wangzelin007
> > [d6a8f0c](https://github.com/Azure/azure-cli-extensions/commit/d6a8f0cb598bfa2e7801485a1c866d53709629a5)
> > Can you help to review and merge this PR? I have remove all secrets in the yaml file, but the GitGuardian Security Checks is still failed. These secrets are always existing before, and they are deleted and out of date.
> > Thanks for your help.
> 
> Please fix the Integration Tests first, I can help bypass other security checks.

Thank you, the integration Tests has pass, please help to merge it.

wangzelin007 (approved):

PR #7216, opened 2024-01-26
[Containerapp] az containerapp env java-component: Support Java component

ShichaoQiu:
Support Java component in Azure Container Apps

PR #7335, opened 2024-03-01
{containerapp} remove additional conversion of binding name of Java component

ShichaoQiu:
remove additional conversion of binding name of Java component


yonzhan:
Containerapp

Greedygre (approved):

Greedygre:
This PR is for fix a logic which has not been released for customer in PR https://github.com/Azure/azure-cli-extensions/pull/7216, so no need to update `history`

src/containerapp/azext_containerapp/java_component_decorator.py
```diff
@@ -57,6 +57,10 @@ def create(self):
                 environment_name=self.get_argument_environment_name(), name=self.get_argument_java_component_name(),
                 java_component_envelope=self.java_component_def, no_wait=self.get_argument_no_wait())
         except Exception as e:
+            stringErr = str(e)
+            if "JavaComponentsNotAllowedForSubscription" in stringErr:
```
Juliehzl, line 61: Is it possible to get ErrorCode?

Juliehzl (commented):
add test case for '-' scenario

src/containerapp/azext_containerapp/java_component_decorator.py
ShichaoQiu: The error code is 400. You mean that we need to check both of the error code and the error message, right?

ShichaoQiu: As we discussed offline, will close this

Greedygre (approved):

zhoxing-ms (approved):

PR #7398, opened 2024-03-19
Use common.utils to check helm/docker installed

Cyclam:
At some point the check_tool_installed method was moved to common.utils as a function, but it looks like the existing code wasn't deleted from `artifact.py`.

This PR fixes that. The code diff should be fairly self-explanatory.

The exception type raised changes from CLIError to MissingDependency, but as none of the code as `try`/`except` for this function, I think that's fine.


yonzhan:

PR #7512, opened 2024-04-16
[AKS]: Fix Scenario test failure for Azure Container Storage

mukhoakash:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
AKS

mukhoakash:
Live test for the failing test case: test_aks_update_with_azurecontainerstorage https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=91616856&view=results

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -4934,79 +4934,6 @@ def postprocessing_after_mc_created(self, cluster: ManagedCluster) -> None:
         :return: None
         """
         super().postprocessing_after_mc_created(cluster)
-        enable_azure_container_storage = self.context.get_intermediate("enable_azure_container_storage")
```
FumingZhang, line 4937: Glad to see the live test passed, but I have some concerns about this change.

Users may not necessarily use the version of azure-cli (2.59.0+) that contains the changes in [#28251](https://github.com/Azure/azure-cli/pull/28251). For example, if the user installs azure-cli 2.58.0 and the latest version of aks-preview that contains the current changes, then these enable/disable operations will not be performed, which is equivalent to the container storage related options being ignored, and this is unexpected.

File [azext_metadata.json](https://github.com/Azure/azure-cli-extensions/blob/main/src/aks-preview/azext_aks_preview/azext_metadata.json#L2) is used to restrict aks-preview's dependence on the version of azure-cli, which is currently 2.56.0. And we don’t want to bump it yet, because this will make the restrictions too strict.

The most ideal solution is that the operation can be implemented as idempotent. If it is difficult to achieve the former, is it possible to do some checks before starting these operations in aks-preview? If these operations have already been done, skip them, otherwise execute them.

mukhoakash: Ensured that there is no depedency on az cli version in the latest set of changes.

mukhoakash:
[Live test](https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=92551354&view=results) with the latest changes.
- test_aks_update_with_azurecontainerstorage
- test_aks_create_with_azurecontainerstorage
- test_aks_create_with_azurecontainerstorage_with_nodepool_name

FumingZhang (approved):
LGTM

zhoxing-ms (approved):

Moazzem-Hossain-pixel (commented):
Need support

PR #7600, opened 2024-05-09
{AKS} Support Mixed SKU VMs agentpool

HuShaoRu:
---

### Related command

az aks create --vm-sizes
az aks nodepool add --vm-sizes
az aks nodepool manual-scale add
az aks nodepool manual-scale update
az aks nodepool manual-scale delete

### General Guidelines

For new extensions:

### About Extension Publish


src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1012,7 +1014,13 @@ def aks_scale(cmd,  # pylint: disable=unused-argument
                 raise CLIError(
                     "Cannot scale cluster autoscaler enabled node pool.")
 
-            agent_profile.count = int(node_count)
+            if agent_profile.type == CONST_VIRTUAL_MACHINES:
+                if len(agent_profile.virtual_machines_profile.scale.manual) == 1:
+                    agent_profile.virtual_machines_profile.scale.manual[0].count = int(node_count)
+                else:
+                    raise CLIError("Cannot scale virtual machines node pool with more than one size.")
```
FumingZhang, line 1021: Better use a more specific error type listed in `azure.cli.core.azclierror` ,e.g., `InvalidArgumentValueError`

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1397,7 +1407,13 @@ def aks_agentpool_scale(cmd,    # pylint: disable=unused-argument
     if new_node_count == instance.count:
         raise CLIError(
             "The new node count is the same as the current node count.")
-    instance.count = new_node_count  # pylint: disable=no-member
+    if instance.type_properties_type == CONST_VIRTUAL_MACHINES:
+        if len(instance.virtual_machines_profile.scale.manual) == 1:
+            instance.virtual_machines_profile.scale.manual[0].count = new_node_count
+        else:
+            raise CLIError("Cannot scale virtual machines node pool with more than one size.")
```
FumingZhang, line 1414: better use a more specific error type

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1715,6 +1731,105 @@ def aks_agentpool_delete_machines(cmd,   # pylint: disable=unused-argument
...
+                                   client,
+                                   resource_group_name,
+                                   cluster_name,
+                                   nodepool_name,
+                                   vm_sizes,
+                                   node_count,
+                                   no_wait=False):
+    instance = client.get(resource_group_name, cluster_name, nodepool_name)
+    if instance.type_properties_type != CONST_VIRTUAL_MACHINES:
+        raise CLIError("Cannot add manual to a non-virtualmachines node pool.")
```
FumingZhang, line 1744: better use a more specific error type

src/aks-preview/azext_aks_preview/_params.py
```diff
@@ -929,6 +929,8 @@ def load_arguments(self, _):
             is_preview=True,
             arg_type=get_enum_type(health_probe_modes),
         )
+        # virtual machines
+        c.argument("vm_sizes", is_preview=True)
```
FumingZhang, line 933: If PM agrees to change the behavior of this option to use spaces to separate multiple values, may set `nargs="*"` and python/cli argument parse would help parse user input to a list of strings, which would save your work to split user input.

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1715,6 +1731,105 @@ def aks_agentpool_delete_machines(cmd,   # pylint: disable=unused-argument
...
+                                      current_vm_sizes,
+                                      vm_sizes=None,
+                                      node_count=None,
+                                      no_wait=False):
+    if vm_sizes is None and node_count is None:
+        raise RequiredArgumentMissingError("specify --vm-sizes or --node-count or both.")
+
+    instance = client.get(resource_group_name, cluster_name, nodepool_name)
+    if instance.type_properties_type != CONST_VIRTUAL_MACHINES:
+        raise CLIError("Cannot update manual in a non-virtualmachines node pool.")
```
FumingZhang, line 1777: better use a more specific error type

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -1715,6 +1731,105 @@ def aks_agentpool_delete_machines(cmd,   # pylint: disable=unused-argument
...
+                                   no_wait=False):
+    instance = client.get(resource_group_name, cluster_name, nodepool_name)
+    if instance.type_properties_type != CONST_VIRTUAL_MACHINES:
+        raise CLIError("Cannot add manual to a non-virtualmachines node pool.")
+    ManualScaleProfile = cmd.get_models(
+        "ManualScaleProfile",
+        resource_type=CUSTOM_MGMT_AKS_PREVIEW,
+        operation_group="managed_clusters",
+    )
+    new_manual_scale_profile = ManualScaleProfile(sizes=vm_sizes.split(","), count=int(node_count))
```
FumingZhang, line 1750: `vm_sizes.split(",")` seems not safe enough, user may provide value with extra space, if server cannot handle this well, may results in unexpected behavior

src/aks-preview/azext_aks_preview/agentpool_decorator.py
```diff
@@ -618,6 +618,17 @@ def get_gateway_prefix_size(self) -> Union[int, None]:
...
 
+    def get_vm_sizes(self) -> List[str]:
+        """Obtain the value of vm_sizes.
+        :return: list of strings
+        """
+        raw_value = self.raw_param.get("vm_sizes")
+        if raw_value is not None:
+            vm_sizes = raw_value.split(",")
+        else:
+            vm_sizes = [self.get_node_vm_size()]
```
FumingZhang, line 629: What is the expected behavior if user specifies vm_sizes and node_vm_size at the same time? Would sever refuse request or take some preference on these two properties?

BTW, these 2 options (`--vm-sizes`, `--node-vm-size`) are doing similar things but named differently? Better add the `node` prefix for the newly added option, could confirm with PM.

FumingZhang (commented):
Queued [live test](https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=93316408&view=results) for updated test case
- test_vms_agentpool_type

src/aks-preview/azext_aks_preview/custom.py
HuShaoRu: change to ClientRequestError now

HuShaoRu: change to ClientRequestError now

src/aks-preview/azext_aks_preview/custom.py
HuShaoRu: change to ClientRequestError now

src/aks-preview/azext_aks_preview/custom.py
HuShaoRu: change to ClientRequestError now

src/aks-preview/azext_aks_preview/custom.py
HuShaoRu: change to ClientRequestError now

src/aks-preview/azext_aks_preview/custom.py
HuShaoRu: if user provides spaces in `--node-vm-size`, server will also return error. I think it's expected.

src/aks-preview/azext_aks_preview/agentpool_decorator.py
HuShaoRu: as shown in codes, we use value from `--vm-sizes` first, if it's not specified, we use the value from `--node-vm-size` (or the default value of it).

HuShaoRu: the 2 options are different. `--node-vm-size` only accepts one value. VirtualMachines agentpool can have one or more than one vmsize so we add new parameter `--vm-sizes` for it. We just let the old parameter also work for our scenario.

Moazzem-Hossain-pixel: If you can do the necessary changes that will be appreciated.

HuShaoRu: If there's no strong reason we prefer not to add node- prefix, mostly because
1.it matches the field name in api and no confusion on what vm size is for.
2.less likely to be confused with node-vm-size which only takes 1 size.

Moazzem-Hossain-pixel: Pls bug fixes

HuShaoRu:
> Queued [live test](https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=93316408&view=results) for updated test case
> 
> * test_vms_agentpool_type

src/aks-preview/azext_aks_preview/_params.py
HuShaoRu: don't know how others think about it, but I think the difference is not significant. If we don't accept spaces users don't need to add extra quotes around it. We may discuss it more when testing it in preview stage.

HuShaoRu:
We may not pass live test now. Some related features are only supported in some subscriptions (will open to other subscriptions soon). I chose the right subscription locally so I could get the test_.yaml.

FumingZhang:
> We may not pass live test now. Some related features are only supported in some subscriptions (will open to other subscriptions soon). I chose the right subscription locally so I could get the test_.yaml.

If the test case could not be executed in a bare sub, please exclude test case from the default test matrix (`src/aks-preview/azcli_aks_live_test/configs/ext_matrix_default.json`). And add it back later when the feature is ready later.

Moazzem-Hossain-pixel (approved):
Thanks

FumingZhang (approved):
LGTM

zhoxing-ms (approved):

PR #7766, opened 2024-07-01
chore: do not use network_plugin default from sdk

tyler-lloyd:
---

### Related command

the cli should not set network_plugin defaults. it should let the AKS RP handle defaulting.

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
+++ b/src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
@@ -2968,6 +2968,16 @@ class AKSPreviewManagedClusterCreateDecorator(AKSManagedClusterCreateDecorator):
         mc = super().set_up_network_profile(mc)
         network_profile = mc.network_profile
 
+        # network_plugin is typically defaulted to kubenet. AKS-RP is moving
+        # away from specifying this default in the API and making it based
+        # on the k8s version being used. The CLI should not be responsible
+        # for setting default values and should pass properties as empty
+        # unless specified by the user.
+        if (network_profile.network_plugin is not None and
+            self.context.raw_param.get("network_plugin") is None and
+            self.context.decorator_mode == DecoratorMode.CREATE):
+            self.mc.network_profile.network_plugin = ""
+
         # set up pod_cidrs, service_cidrs and ip_families
         (
             pod_cidrs,
--- a/src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
+++ b/src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
@@ -2977,7 +2977,7 @@ class AKSPreviewManagedClusterCreateDecorator(AKSManagedClusterCreateDecorator):
             network_profile.network_plugin is not None and
             self.context.raw_param.get("network_plugin") is None
         ):
-            self.mc.network_profile.network_plugin = ""
+            network_profile.network_plugin = ""
 
         # set up pod_cidrs, service_cidrs and ip_families
         (
--- a/src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
+++ b/src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
@@ -2975,8 +2975,7 @@ class AKSPreviewManagedClusterCreateDecorator(AKSManagedClusterCreateDecorator):
         # unless specified by the user.
         if (
             network_profile.network_plugin is not None and
-            self.context.raw_param.get("network_plugin") is None and
-            self.context.decorator_mode == DecoratorMode.CREATE
+            self.context.raw_param.get("network_plugin") is None
         ):
             self.mc.network_profile.network_plugin = ""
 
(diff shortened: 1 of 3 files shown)


yonzhan:

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2968,6 +2968,18 @@ def set_up_network_profile(self, mc: ManagedCluster) -> ManagedCluster:
...
 
+        # network_plugin is typically defaulted to kubenet. AKS-RP is moving
+        # away from specifying this default in the API and making it based
+        # on the k8s version being used. The CLI should not be responsible
+        # for setting default values and should pass properties as empty
+        # unless specified by the user.
+        if (
+            network_profile.network_plugin is not None and
+            self.context.raw_param.get("network_plugin") is None and
+            self.context.decorator_mode == DecoratorMode.CREATE
```
FumingZhang, line 2979: may omit `self.context.decorator_mode == DecoratorMode.CREATE` as `set_up_network_profile` is only called in the code path of `aks create` (get_xxx could be used by both create and update and DecoratorMode is used there)

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
```diff
@@ -2968,6 +2968,18 @@ def set_up_network_profile(self, mc: ManagedCluster) -> ManagedCluster:
         mc = super().set_up_network_profile(mc)
         network_profile = mc.network_profile
 
+        # network_plugin is typically defaulted to kubenet. AKS-RP is moving
```
FumingZhang, line 2971: not only set by SDK, we also set it to kubenet when the lb sku is standard
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acs/managed_cluster_decorator.py#L5887-L5890

tyler-lloyd: we should remove that then. but this PR would still work right since setup_network_profile in the cli is called before setup_network_profile from the extension, correct? so if we always clear the value when a user does not specify anything then it takes care of that defaulting you linked.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -6097,6 +6097,7 @@ def test_aks_create_with_pod_identity_enabled(
         # create
         create_cmd = (
             "aks create --resource-group={resource_group} --name={name} --location={location} "
+            "--network-plugin kubenet "
```
paulgmiller, line 6100: Should we leave this blank and let defaulting do its thing?

paulgmiller: Are we asserting the result of this create has kubenet?

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
paulgmiller: Confused why that code exists at all. Should't the api be doing defaulting there not cli? Might be worth stating a conversation with Pengfei's team about. 

    # Attention: RP would return UnexpectedLoadBalancerSkuForCurrentOutboundConfiguration internal server error
            # if load_balancer_sku is set to basic and load_balancer_profile is assigned.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
tyler-lloyd: the reason I added kubenet explicity is because we're asserting this `self.check("podIdentityProfile.allowNetworkPluginKubenet", True)` so it seems like we have some property around pod identity specific to kubenet. I am not sure if that property is set if you had overlay (my guess is no)

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
tyler-lloyd: fixed

src/aks-preview/azext_aks_preview/managed_cluster_decorator.py
FumingZhang: yeah, the function in core cli is called earlier. Also agree with Paul, defaulting should be done by the sever. This piece of code has been around for a long time (4 years+), and perhaps the server did not have a suitable default logic at the time.

FumingZhang:
Failures in the CI check are not related to the change, will fix in another PR #7775

FumingZhang:
#7775 merged, please rebase from main

tyler-lloyd:
@FumingZhang - looks like CI is passing now. thanks for the fix.

FumingZhang (approved):

tyler-lloyd:
@zhoxing-ms PTAL

yanzhudd (approved):

PR #7802, opened 2024-07-16
[AMG] Migrate

leozhang-msft:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:
AMG

yonzhan:
Please fix CI issues

leozhang-msft:
Hello, sorry this is still a draft, the feature isn't complete yet, will fix CI issues once feature is done

src/amg/azext_amg/custom.py
```diff
@@ -297,6 +297,36 @@ def restore_grafana(cmd, grafana_name, archive_file, components=None, remap_data
             destination_datasources=data_sources)
 
 
+def migrate_grafana(cmd, grafana_name, source_instance_ip, source_instance_token, dry_run=False, overwrite=False, folders_to_include=None,
```
qingquanzhao, line 300: source_endpoint?

src/amg/azext_amg/_help.py
```diff
@@ -55,6 +55,15 @@
            az grafana restore -g MyResourceGroup -n MyGrafana --archive-file backup\\dashboards\\ServiceHealth-202307051036.tar.gz --components dashboards folders --remap-data-sources
 """
 
+helps['grafana migrate'] = """
+    type: command
+    short-summary: Migrates an existing Grafana instance to an Azure Managed Grafana instance.
```
Jeralan, line 60:
```suggestion
    short-summary: Migrate an existing Grafana instance to an Azure Managed Grafana instance.
```
Matches the imperative grammar used in the other short summaries.

src/amg/azext_amg/_params.py
```diff
@@ -66,6 +66,12 @@ def load_arguments(self, _):
         c.argument("archive_file", options_list=["-a", "--archive-file"], help="archive to restore Grafana artifacts from")
         c.argument("remap_data_sources", options_list=["-r", "--remap-data-sources"], arg_type=get_three_state_flag(),
                    help="during restoration, update dashboards to reference data sources defined at the destination workspace through name matching")
+        
+    with self.argument_context("grafana migrate") as c:
+        c.argument("source_instance_ip", options_list=["-s", "--src-ip"], help="Grafana instance IP to migrate from")
+        c.argument("source_instance_token", options_list=["-t", "--src-token"], help="Grafana instance token to migrate from")
+        c.argument("dry_run", options_list=["-d", "--dry-run"], arg_type=get_three_state_flag(), help="Preview changes without committing. Takes priority over --overwrite.")
+        c.argument("overwrite", options_list=["--overwrite"], arg_type=get_three_state_flag(), help="Overwrite previous dashboards, library panels, and folders with the same uid with new ones (version replacement)")
```
Jeralan, line 74: I feel like the overwrite help line can be cleaned up, there's some redundant language

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,248 @@
...
+    output.append((Style.SUCCESS, f"\n\nSnapshots {dry_run_status}updated:"))
+    for folder, snapshots in snapshots_synced_summary.items():
+        output.append((Style.PRIMARY, f"\n    {folder}/\n        "))
+        output.append((Style.SECONDARY, "\n        ".join(snapshots)))
+
+    # TODO: annotations don't auto update, it creates a new one.
+    output.append((Style.SUCCESS, f"\n\nAnnotations {dry_run_status}updated (by id):"))
+    output.append((Style.PRIMARY, "\n    " + "\n    ".join(annotations_synced_summary)))
+
+    print_styled_text(output)
```
Jeralan, line 83: Change "overwrote" to "overwritten" in these outputs

src/amg/azext_amg/restore.py
```diff
@@ -65,54 +65,76 @@ def _restore_components(grafana_url, restore_functions, tmpdir, components, http
...
     dashboard_title = content['dashboard'].get('title', '')
-    print_styled_text([
+
+    to_print = [
         (Style.WARNING, f'Create dashboard {dashboard_title}: '),
         (Style.SUCCESS, 'SUCCESS') if result[0] == 200 else (Style.ERROR, 'FAILURE')
-    ])
+    ]
+    if result[0] == 412:
+        to_print.append((Style.ERROR, ' (version mismatch, please enable --overwrite if you want to overwrite it)'))
```
Jeralan, line 113: overwrite what?

src/amg/azext_amg/restore.py
```diff
@@ -162,27 +210,63 @@ def _create_snapshot(grafana_url, file_path, http_headers):
...
+            (Style.WARNING, f'Overwrite folder {folder_name}: '),
+            (Style.SUCCESS, 'SUCCESS') if result[0] in [200] else (Style.ERROR, 'FAILURE')
+        ])
+    else:
+        to_print = [
+            (Style.WARNING, f'Create folder {folder_name}: '),
+            (Style.SUCCESS, 'SUCCESS') if result[0] in [200] else (Style.ERROR, 'FAILURE')
+        ]
+
+        if result[0] == 412:
```
Jeralan, line 241: can this be simplified to only check result[0] once?

src/amg/azext_amg/restore.py
```diff
@@ -65,54 +65,76 @@ def _restore_components(grafana_url, restore_functions, tmpdir, components, http
...
-        uid_mapping = create_datasource_mapping(source_datasources, destination_datasources)
+        set_uid_mapping(source_datasources, destination_datasources)
 
     if "dashboard" in exts:  # dashboard restoration can't work if linked library panels don't exist
         exts.insert(0, "library_panel")
 
     if "folder" in exts:  # make "folder" be the first to restore, so dashboards can be positioned under a right folder
         exts.insert(0, exts.pop(exts.index("folder")))
 
+    # print(exts)
```
qingquanzhao, line 76: Remove these?

src/amg/azext_amg/_params.py
```diff
@@ -66,6 +66,12 @@ def load_arguments(self, _):
         c.argument("archive_file", options_list=["-a", "--archive-file"], help="archive to restore Grafana artifacts from")
         c.argument("remap_data_sources", options_list=["-r", "--remap-data-sources"], arg_type=get_three_state_flag(),
                    help="during restoration, update dashboards to reference data sources defined at the destination workspace through name matching")
+        
+    with self.argument_context("grafana migrate") as c:
+        c.argument("source_instance_ip", options_list=["-s", "--src-ip"], help="Grafana instance IP to migrate from")
```
qingquanzhao, line 71: endpoint?

src/amg/azext_amg/_params.py
```diff
@@ -66,6 +66,12 @@ def load_arguments(self, _):
...
+        c.argument("source_instance_token", options_list=["-t", "--src-token"], help="Grafana instance token to migrate from")
```
qingquanzhao, line 72: Probably more specific terms like "Service Account or API Key"

src/amg/azext_amg/_params.py
leozhang-msft: How about "Replace existing dashboards, panels, and folders with new ones using the same UID", basically, this flag will allow customers to overwrite previous dashboards with same uid. There is an edge case where there is a title conflict but different uid, it will keep the old uid but then overwrite everything else. How do you think I phrase that weird case?

src/amg/azext_amg/custom.py
leozhang-msft: Yeah, that sounds good, but I want to take advantage of -n flag to be consistent with other commands. How would I implement that?

src/amg/azext_amg/_params.py
Jeralan: Are you saying if the new dashboards match existing dashboards in title, and you use the --overwrite flag, it will overwrite the dashboard, but keep the existing UID?

Can you clarify for me how conflicts are identified? And then how they are resolved when this --overwrite flag is used and when it is not used?

src/amg/azext_amg/_help.py
```diff
@@ -55,6 +55,15 @@
            az grafana restore -g MyResourceGroup -n MyGrafana --archive-file backup\\dashboards\\ServiceHealth-202307051036.tar.gz --components dashboards folders --remap-data-sources
 """
 
+helps['grafana migrate'] = """
+    type: command
+    short-summary: Migrate an existing Grafana instance to an Azure Managed Grafana instance.
+    examples:
+       - name: Simple local to AMG example. Migrate dashboards and folders from a local Grafana instance to an Azure Managed Grafana instance.
```
Jeralan, line 62:
```suggestion
       - name: Migrate dashboards and folders from a local Grafana instance to an Azure Managed Grafana instance.
```

src/amg/azext_amg/backup_core.py
```diff
@@ -0,0 +1,255 @@
...
+            folders_to_include = [f.lower() for f in folders_to_include]
+            dashboards = [d for d in dashboards if (d.get('folderTitle', '').lower() in folders_to_include or
+                                                    not d.get('folderTitle', '') and 'general' in folders_to_include)]
+        if folders_to_exclude:
+            folders_to_exclude = [f.lower() for f in folders_to_exclude]
+            dashboards = [d for d in dashboards if ((d.get('folderTitle', '')
+                                                    and d.get('folderTitle', '').lower() not in folders_to_exclude)
+                                                    or
+                                                    (not d.get('folderTitle', '')
+                                                    and 'general' not in folders_to_exclude))]
```
Jeralan, line 52: might be best to use "filter" here. Can reduce the number of .get calls too.

src/amg/azext_amg/backup_core.py
```diff
@@ -0,0 +1,255 @@
...
+
+        _print_an_empty_line()
+        if len(dashboards) == 0:
+            break
+        current_page += 1
+        current_run_dashboards = _get_individual_dashboard_setting(dashboards, grafana_url, http_headers)
+        # add the previous list to the list where we added everything.
+        all_dashboards = all_dashboards + current_run_dashboards
```
Jeralan, line 60: can we use += for all_dashboards?

src/amg/azext_amg/backup.py
```diff
@@ -1,31 +1,19 @@
...
 import os
-import random
 import shutil
-import string
+import json
 import re
-import tarfile
-import time
+import random
+import string
```
Jeralan, line 9: is there a reason the import order is changed? Keep the header too.

src/amg/azext_amg/backup.py
```diff
@@ -189,32 +118,17 @@ def _save_library_panel_setting(panel_name, file_name, library_panel_settings, f
...
-
 # Save snapshots
 def _save_snapshots(grafana_url, backup_dir, timestamp, http_headers, **kwargs):  # pylint: disable=unused-argument
     folder_path = f'{backup_dir}/snapshots/{timestamp}'
 
     if not os.path.exists(folder_path):
         os.makedirs(folder_path)
 
-    _get_all_snapshots_and_save(folder_path, grafana_url, http_get_headers=http_headers)
-    _print_an_empty_line()
```
Jeralan, line 217: is there a reason the empty line print was removed?

src/amg/azext_amg/backup_core.py
```diff
@@ -0,0 +1,255 @@
...
+        for board in dashboards:
+            logger.info('name: %s', board['title'])
+        return dashboards
+    logger.warning("Get dashboards FAILED, status: %s, msg: %s", status, content)
+    return []
+
+
+def _get_individual_dashboard_setting(dashboards, grafana_url, http_headers):
+    all_individual_dashboards = []
+    if dashboards:
```
Jeralan, line 83: can we change this dashboards check to reduce the indentation level?

src/amg/azext_amg/backup_core.py
```diff
@@ -0,0 +1,255 @@
...
+        panels = _get_all_library_panels_in_grafana(current_page, grafana_url, http_headers)
+
+        _print_an_empty_line()
+        if len(panels) == 0:
+            break
+        current_page += 1
+
+        # don't log here, log later depending on backup versus migrate.
+
+        # Since we are not excluding anything. We can just add the panels to the list since this is all the data we need.
```
Jeralan, line 112: clean up the new lines around these comments

src/amg/azext_amg/backup_core.py
```diff
@@ -0,0 +1,255 @@
...
+        for panel in library_panels:
+            logger.info('name: %s', panel['name'])
+        return library_panels
+    logger.warning("Get library panel FAILED, status: %s, msg: %s", status, content)
+    return []
+
+
+def get_all_snapshots(grafana_url, http_headers):
+    (status, content) = search_snapshot(grafana_url, http_headers)
+
```
Jeralan, line 133:
```suggestion
if status != 200:
    logger.warning("...")
    return []
```

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,259 @@
...
+    snapshots_synced_summary = _migrate_snapshots(all_snapshots, restore_url, restore_headers, dry_run)
+   
+    # get all the annotations
+    all_annotations = get_all_annotations(backup_url, backup_headers)
+    annotations_synced_summary = _migrate_annotations(all_annotations, restore_url, restore_headers, dry_run)
+
+    dry_run_status = "to be " if dry_run else ""
+    output = [
+        (Style.IMPORTANT, f"\n\nDry run: {dry_run if dry_run else False}\n"),
+        (Style.IMPORTANT, f"Overwrite dashboards, folders, and library panels: {overwrite if overwrite else False}\n"),
```
Jeralan, line 44: dry_run and overwrite are booleans right? Can these both just be simplified to {dry_run} and {overwrite} ?

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,259 @@
...
+    output.append((Style.SUCCESS, f"\n\nDashboards {dry_run_status}overwritten:"))
+    for folder, dashboards in dashboards_overwrote_summary.items():
+        output.append((Style.PRIMARY, f"\n    {folder}/\n        "))
+        output.append((Style.SECONDARY, "\n        ".join(dashboards)))
+
+    # TODO: snapshots don't auto update, it creates a new one.
+    output.append((Style.SUCCESS, f"\n\nSnapshots {dry_run_status}updated:"))
+    for folder, snapshots in snapshots_synced_summary.items():
+        output.append((Style.PRIMARY, f"\n    {folder}/\n        "))
+        output.append((Style.SECONDARY, "\n        ".join(snapshots)))
```
Jeralan, line 87: these printouts are pretty repetitive, can we use a helper function here?

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,259 @@
...
+        if is_successful:
+            panel_folder_name = library_panel['meta']['folderName']
+            if exists_before:
+                if panel_folder_name not in library_panels_overwrote_summary:
+                    library_panels_overwrote_summary[panel_folder_name] = set()
+                library_panels_overwrote_summary[panel_folder_name].add(library_panel['name'])
+            else:
+                if panel_folder_name not in library_panels_created_summary:
+                    library_panels_created_summary[panel_folder_name] = set()
+                library_panels_created_summary[panel_folder_name].add(library_panel['name'])
```
Jeralan, line 177: is there a reason we use sets here, but use lists for the folders summary?

src/amg/azext_amg/_params.py
leozhang-msft: Yup, so there a couple cases (due to quirks of the Grafana HTTP API).

For dashboard specifically, when we send a request to create the dashboard, it will either create it successfully or not. If it doesn't it will return with a status and error code. Depending on the error code, we will pass the message to the user. So Grafana API will check that the Dashboard Title or UID is taken. If it is taken, then it will return if an error.

Now the overwrite flag comes into play when we receive the flag, we can pass it into the request and when we do, Grafana API will automatically overwrite. However, through testing I noticed that:
- If UIDs match, everything will be overwritten (including title, if title is different)
- If Titles match, everything will be overwritten EXCEPT it will keep the old (existing) dashboard UID
- If both UID and title match, then UID and title will be same and the dashboard data will just be overwritten (so new panels, etc).

src/amg/azext_amg/backup.py
leozhang-msft: I moved them to backup_core.py

src/amg/azext_amg/_params.py
Jeralan: I see. If it overwrites based on UID or Title maybe "Overwrite previous dashboards, library panels, and folders with the same uid or title" would be good? I think we should include it will also overwrite based on title.

src/amg/azext_amg/backup.py
Jeralan: I am not seeing a line print in get_all_snapshots

src/amg/azext_amg/backup.py
```diff
@@ -254,35 +153,21 @@ def _save_folders(grafana_url, backup_dir, timestamp, http_headers, **kwargs):
...
-            logger.info("name: %s", folder['title'])
-        return folders
-    logger.warning("Get folders FAILED, status: %s, msg: %s", status, content)
-    return []
+    folders = get_all_folders(grafana_url, http_headers, **kwargs)
+
+    log_file_path = folder_path + '/' + log_file
+    with open(log_file_path, 'w+', encoding="utf8") as f:
+        for folder_set in folders:
+            (folder_settings, _) = folder_set
```
Jeralan, line 161: are parentheses needed here?

src/amg/azext_amg/_params.py
leozhang-msft: Okay that sounds good!

leozhang-msft: I also added support for Snapshots & Annotations, but snapshot can have same title and annotations will only overwrite if certain conditions are true.

src/amg/azext_amg/backup.py
leozhang-msft: Oh, that's true, thanks for catching that!

src/amg/azext_amg/_params.py
Jeralan: Ok, so my understanding here then is dashboards and folders use the Grafana HTTP API overwrite flags, but library panels and snapshots and annotations use our own defined behavior. 

I don't think it will be feasible to explain the nuance of all of these cases in the help line. We want to most avoid cases where the user doesn't expect something to be overwritten when it is, so I think mentioning both uid and title is a necessity. The inverse case where something is not overwritten that was expected to be is less harmful.

The only other necessary change I think is removing "with new ones." I think this is redundant as it is implied by overwriting,

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,302 @@
...
+    (folders_created_summary, folders_overwrote_summary) = _migrate_folders(
+        all_folders, all_restore_folders, restore_url, restore_headers, dry_run, overwrite)
+
+    valid_folder_uids = set(folder[0]['uid'] for folder in all_folders + all_restore_folders)
+    all_dashboards = get_all_dashboards(backup_url,
+                                        backup_headers,
+                                        folders_to_include=folders_to_include,
+                                        folders_to_exclude=folders_to_exclude)
+    all_library_panels = get_all_library_panels(backup_url, backup_headers)
+    if folders_to_exclude is None or 'General' not in folders_to_exclude:
```
ABZhang0, line 43: is 'General' case sensitive?

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,302 @@
...
+        # edge case for general folder. The uid is empty string.
+        valid_folder_uids.add('')
+
+    all_library_panels_filtered = [panel for panel in all_library_panels if panel['folderUid'] in valid_folder_uids]
```
ABZhang0, line 47: I remember when I was implementing the 'sync' command I discovered when you fetched the library panels using the corresponding api endpoint it would show panels in the root directory as folderUid = 'general'. When creating the same panels in the destination workspace, I needed to map the panels with folderUid == 'general' to '' (empty string) for it to create properly in the root directory. Did you do the same here?

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,302 @@
...
+        exists_before = check_library_panel_exists(restore_url, library_panel, restore_headers)
+
+        library_panel['id'] = None
+        is_successful = True
+        if not dry_run:
+            # in create_library_panel, it does the patch if the library panel already exists.
+            is_successful = create_library_panel(restore_url, library_panel, restore_headers, overwrite)
+
+        if not is_successful:
+            continue
```
ABZhang0, line 190: does 'migrate' have partial success behavior? Seems like if it fails to migrate over a library panel it will just skip on ahead instead of terminating

src/amg/azext_amg/backup.py
```diff
@@ -130,57 +95,27 @@ def _save_dashboard_setting(dashboard_name, file_name, dashboard_settings, folde
...
     log_file = f'library_panels_{timestamp}.txt'
+    log_file_path = folder_path + '/' + log_file
 
     if not os.path.exists(folder_path):
         os.makedirs(folder_path)
 
-    current_page = 1
-    while True:
-        panels = _get_all_library_panels_in_grafana(current_page, grafana_url, http_headers)
+    all_panels = get_all_library_panels(grafana_url, http_headers)
```
ABZhang0, line 107: do we include folders_to_include & folders_to_exclude arguments for the library panel save?

ABZhang0 (changes requested):

src/amg/azext_amg/tests/latest/test_amg_migrate_scenario.py
```diff
@@ -0,0 +1,335 @@
...
+
+    def __init__(self, method_name):
+        super().__init__(method_name, recording_processors=[
+            ApiKeyServiceAccountTokenReplacer()
+        ])
+
+    def _setup_migrate_instances(self):
+        # migrate from amg1 to amg2
+        amg1 = self.cmd('grafana create -g {rg} -n {name} -l {location}').get_output_in_json()
+        amg2 = self.cmd('grafana create -g {rg} -n {name2} -l {location}').get_output_in_json()
```
ABZhang0, line 31: migrating between two AMG instances for the test?

src/amg/azext_amg/restore.py
```diff
@@ -81,38 +81,66 @@ def _restore_components(grafana_url, restore_functions, tmpdir, components, http
...
+def create_dashboard(grafana_url, content, http_headers, overwrite):
     payload = {
         'dashboard': content['dashboard'],
         'folderId': get_folder_id(content, grafana_url, http_post_headers=http_headers),
-        'overwrite': True
+        'overwrite': overwrite
     }
 
     datasources_missed = set()
     remap_datasource_uids(payload, uid_mapping, datasources_missed)
```
ABZhang0, line 102: I remember you said the change I made to the remap function was causing issues. Are you still experiencing the problem here?

src/amg/azext_amg/backup_core.py
```diff
@@ -0,0 +1,255 @@
...
+    all_dashboards = []
+    
+    # Go through all the pages, we are unsure how many pages there are
+    while True:
+        dashboards = _get_all_dashboards_in_grafana(current_page, limit, grafana_url, http_headers)
+
+        # only include what users want
+        folders_to_include = kwargs.get('folders_to_include')
+        folders_to_exclude = kwargs.get('folders_to_exclude')
+        if folders_to_include:
```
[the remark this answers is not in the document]
leozhang-msft: The folders to exclude gets priority so it won't be included. Do you think we should send a error saying that you cannot do such a thing?

src/amg/azext_amg/migrate.py
leozhang-msft: Yep, should I support lowercase as well?

src/amg/azext_amg/migrate.py
leozhang-msft: Huh, interesting, when I was testing the folderuid=(empty string) that's why I had to add the valid_folder_uids of the empty string since it never had general.

src/amg/azext_amg/migrate.py
leozhang-msft: Yeah, if it fails, it will skip over that library panel. If there happens to be an issue where the dashboard doesn't have the library panel, it will fail to create that dashboard to so it will also skip over it.

src/amg/azext_amg/backup.py
leozhang-msft: Nope, backup never had that feature for library panels so I didn't add it. But in migrate, it is handled by using set()

src/amg/azext_amg/tests/latest/test_amg_migrate_scenario.py
leozhang-msft: Yep, it is easier to create two AMG instances. One of the instances is basically treated as a local so we get the service account token and that flow. If I were to try to do a local to AMG test, I will need to mock a lot more functions for basically the same effect.

src/amg/azext_amg/restore.py
leozhang-msft: Yeah, both in migrate & restore it is broken, just re-tested again (Make AMG instance, create a datasource, then migrate, then the dashboard datasource uid isn't overwritten). I am not sure why my test didn't catch it though, going to double check.

leozhang-msft: Okay fixed the test, now it is failing as expected:

https://github.com/Azure/azure-cli-extensions/pull/7802/commits/6efcf36e699007a365b5897f6f831c2b2c37935f

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,302 @@
...
+
+def migrate(backup_url, backup_headers, restore_url, restore_headers, dry_run,
+            overwrite, folders_to_include=None, folders_to_exclude=None):
+    # pylint: disable=too-many-locals
+
+    # get all datasources to be backedup
+    all_datasources = get_all_datasources(backup_url, backup_headers)
+    all_restore_datasources = get_all_datasources(restore_url, restore_headers)
+    if (all_restore_datasources is None) or (all_datasources is None):
+        logger.error("ABORTING!! Datasources are not found. Please check the URLs, headers, or api key/service token.")
```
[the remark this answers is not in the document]
leozhang-msft: Nope, I guess the error message wasn't good enough. The issue is mainly when the source / restore url doesn't seem to work, so that the response isn't 200 and thus the function will return None.

src/amg/azext_amg/backup_core.py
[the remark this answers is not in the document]
ABZhang0: probably should have an error message for the user, yes. We can have a preliminary check for this kind of user error

src/amg/azext_amg/migrate.py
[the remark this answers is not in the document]
ABZhang0: do you mean if the get_all_datasources request fails?

src/amg/azext_amg/migrate.py
ABZhang0: yes, definitely. I think you can refer to sync.py. Everything is casted to lowercase first before any string comparison

src/amg/azext_amg/migrate.py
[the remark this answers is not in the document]
leozhang-msft: get_all_datasources can fail since we don't do a health check on the restore url (only on the backup/destination url), I probably shouldn't have the logic here but in the main migrate function.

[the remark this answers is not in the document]
leozhang-msft: Here is a better fix: https://github.com/Azure/azure-cli-extensions/pull/7802/commits/f2e2f43ce7538c3f01a415a541a980de93a00b9f

src/amg/azext_amg/migrate.py
leozhang-msft: Fixed :) https://github.com/Azure/azure-cli-extensions/pull/7802/commits/4d06109d54d45d38f74a1889eb5e71a582f4ec50

src/amg/azext_amg/backup_core.py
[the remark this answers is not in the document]
leozhang-msft: Fixed https://github.com/Azure/azure-cli-extensions/pull/7802/commits/4d06109d54d45d38f74a1889eb5e71a582f4ec50

src/amg/azext_amg/restore.py
leozhang-msft: Was weird edge case with test datasource, put in a read UID so now does remapping: [e23ecf8](https://github.com/Azure/azure-cli-extensions/pull/7802/commits/e23ecf80394275203417b81a8b1fd3600386d56c)

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,331 @@
...
+    if folders_to_include_set.intersection(folders_to_exclude_set):
+        raise ArgumentUsageError("Folders to include and exclude cannot have the same folder")
+
+    # get all datasources to be backed up
+    all_source_datasources = get_all_datasources(backup_url, backup_headers)
+    all_destination_datasources = get_all_datasources(restore_url, restore_headers)
+    (datasources_created_summary, datasources_remapped_summary) = _migrate_datasources(
+        all_source_datasources, all_destination_datasources, restore_url, restore_headers, dry_run)
+
+    all_folders = get_all_folders(backup_url,
```
ABZhang0, line 38: nit: clearer naming for variables since there are a lot of variables/operations. Either 'source' & 'destination' (e.g. all_source_folders, all_source_dashboards, all_source_snapshots) or 'backup' & 'restore'. Otherwise, it's ambiguous

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,331 @@
...
+        output.append((Style.SECONDARY, "\n        ".join(items)))
+
+
+def append_annotations_summary(output, status, summary_type, annotations_summary, max_text_length=50):
+    output.append((Style.SUCCESS, f"\n\n{summary_type} {status}:"))
+    if len(annotations_summary) > 0:
+        postprocessed_summary = [(a, b[:max_text_length] + "...") if len(b) >
+                                 max_text_length else (a, b) for a, b in annotations_summary]
+        output.append((Style.PRIMARY, "\n    " +
+                       "\n    ".join([f"id: {a}, text: {b}" for a, b in postprocessed_summary])))
```
ABZhang0, line 132: what's the distinction between these 3 different summary helpers? Is it possible to simplify it into 1?

src/amg/azext_amg/migrate.py
```diff
@@ -0,0 +1,331 @@
...
+                                        folders_to_exclude=folders_to_exclude)
+    all_library_panels = get_all_library_panels(backup_url, backup_headers)
+
+    valid_folder_uids.add('')
+    if folders_to_include:
+        if 'general' not in folders_to_include_set:
+            valid_folder_uids.discard('')
+    if folders_to_exclude:
+        if 'general' in folders_to_exclude_set:
+            valid_folder_uids.discard('')
```
ABZhang0, line 62: simplify:

`if 'general' in folders_to_include or 'general' not in folders_to_exclude:
   valid_folder_uids.add('')`

src/amg/azext_amg/restore.py
```diff
@@ -121,83 +144,211 @@ def _create_library_panel(grafana_url, file_path, http_headers):
...
+
+def create_snapshot(grafana_url, snapshot, http_headers, overwrite):
     try:
         snapshot['name'] = snapshot['dashboard']['title']
     except KeyError:
         snapshot['name'] = "Untitled Snapshot"
+    snapshot_name = snapshot['name']
+
+    exists_before = False
+    # TODO: in backup we don't save the key.
```
ABZhang0, line 212: so we're missing this functionality at the moment?

src/amg/azext_amg/migrate.py
leozhang-msft: There might be a way to simplify into one, but I think it will make the code messier. They are used to show:
image

So append_summary would take in a simple list, while append_nested_summary takes in a dictionary, the reason why is because we need things to go under folders so we also want to show those. The simple list are for components that don't need folders. 

append_annotations_summary takes in a list, but each lsit has id & text, for annotation to show more information about the annotation

src/amg/azext_amg/restore.py
leozhang-msft: Yeah, I did not change the functionality in the backup functionality (we didn't store the key in the backup) but we needed to save key for migrate so we can use it to help check if the snapshot exists before (since snapshots don't have uids, keys are kinda used for that here).

ABZhang0 (approved):

qingquanzhao (approved):

kairu-ms (approved):

PR #7857, opened 2024-08-05
Move Grafana CRUD commands to AAZ

ABZhang0:
### Related command
`az grafana create`, `az grafana update`, `az grafana list`, `az grafana show`, `az grafana delete`

### General Guidelines


yonzhan:

kairu-ms:
Please check whether those breaking change are expexted 
image

src/amg/azext_amg/custom.py
```diff
@@ -25,55 +28,108 @@
...
+                 'Once provided, CLI won\'t make the current logged-in user as Grafana Admin',
+        )
+        args_schema['principal_ids'].Element = AAZStrArg()
+        args_schema['identity']._registered = False  # pylint: disable=protected-access
+
+        return args_schema
+
+    def pre_operations(self):
+        args = self.ctx.args
+        if args.skip_role_assignments and args.principal_ids:
```
kairu-ms, line 57:
```suggestion
        if has_value(args.skip_role_assignments) and has_value(args.principal_ids):
```

src/amg/azext_amg/custom.py
```diff
@@ -25,55 +28,108 @@
...
+            raise ArgumentUsageError("--skip-role-assignments | --assignee-object-ids")
+
+        if not args.skip_system_assigned_identity:
```
kairu-ms, line 60:
```suggestion
        if not has_value(args.skip_system_assigned_identity):
```

src/amg/azext_amg/custom.py
```diff
@@ -25,55 +28,108 @@
...
+            args.identity = {"type": "SystemAssigned"}
+
+        process_grafana_create_namespace(self.ctx, self.ctx.args)
+
+    # override the output method to create role assignments after instance creation
+    def _output(self, *args, **kwargs):
```
kairu-ms, line 66: It should be in `post_operations` function instead of `_output`

ABZhang0: Unfortunately `post_operations` does not have ``self.ctx.vars.instance`` available to reference, and `post_instance_create`/`post_instance_update`, which does have that, is only available to override in subcommands. I found this way to achieve what was needed while still using the recommended inheritance method of customization. Is this acceptable?

kairu-ms: OK, that's fine. It's because it's a LRO and the instance will created only when LRO pulling finished.

kairu-ms (approved):

PR #8242, opened 2024-11-06
[Draft - Do not merge] [Quantum] Use Post Storage URI and SAS token like Quantum Python SDK

warren-jones:
### NOTE: Restore old version number before clicking "Ready for review"

The Azure Quantum CLI should use a Post Storage URI and a SAS token like Python SDK, azure-quantum-python.

Our CLI extension has a potential issue with directly using a user's credentials to access storage account to upload input or download output files, when our other SDK uses the PostStorageUri endpoint in our service. If we had the CLI extension use the same method, then we won't run into user permission issues with using CLI (if they have access to workspace, but not to storage account).

---

### Related commands
`az quantum job submit`
`az quantum execute`
`az run`

---

### DIFF
--- a/src/quantum/azext_quantum/__init__.py
+++ b/src/quantum/azext_quantum/__init__.py
@@ -13,7 +13,7 @@ import azext_quantum._help  # pylint: disable=unused-import
 # This is the version reported by the CLI to the service when submitting requests.
 # This should be in sync with the extension version in 'setup.py', unless we need to
 # submit using a different version.
-CLI_REPORTED_VERSION = "1.0.0b4"
+CLI_REPORTED_VERSION = "1.0.0b4.dev3"
 
 
 class QuantumCommandsLoader(AzCommandsLoader):
--- a/src/quantum/setup.py
+++ b/src/quantum/setup.py
@@ -17,7 +17,7 @@ except ImportError:
 # This version should match the latest entry in HISTORY.rst
 # Also, when updating this, please review the version used by the extension to
 # submit requests, which can be found at './azext_quantum/__init__.py'
-VERSION = '1.0.0b4'
+VERSION = '1.0.0b4.dev3'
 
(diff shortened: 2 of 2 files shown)


yonzhan:

04diiguyi (approved):

yonzhan:
Still in draft status

src/quantum/azext_quantum/operations/job.py
```diff
@@ -234,15 +235,19 @@ def submit(cmd, resource_group_name, workspace_name, location, target_id, job_in
...
-    connection_string_dict = show_storage_account_connection_string(cmd, resource_group_name, storage)
-    connection_string = connection_string_dict["connectionString"]
-    container_client = create_container(connection_string, container_name)
     blob_name = "inputData"
 
+    resource_id = "/subscriptions/" + ws_info.subscription + "/resourceGroups/" + ws_info.resource_group + "/providers/Microsoft.Quantum/Workspaces/" + ws_info.name
+    workspace = Workspace(resource_id=resource_id, location=location)
+
+    knack_logger.warning("Getting Azure credential token...")
+    container_uri = workspace.get_container_uri(job_id=job_id)
```
ashwinmayya, line 244: Does anything here create an actual container from the containerUri? Previously, it looks like create_container would do this. 
We need to create a container if it doesn't already exists. The service just returns a URI, it doesn't actually create it today., although we are planning to modify it so that it creates it. The cli should handle both cases.

ashwinmayya: It looks like upload_blob will create a container if it doesn't already exist, so I think we are good.

ashwinmayya (approved):

kikomiss (approved):

yonzhan:
This PR is in draft

PR #8324, opened 2024-11-26
[containerapp] Add new options for labels mode

Tratcher:
This adds support for the new containerapp activeRevisionMode 'labels' mode and the associated target-label option. These are used when creating or updating applications. There are also new commands to show the history of label revision assignments.

All commands except label-history already exist and need to be overridden by the extension to support these new features.

Needs tests.

---

### Related command

* 'az containerapp create': New target-label option for Labels revision mode.
* 'az containerapp update': New revisions-mode and target-label options for Labels revision mode.
* 'az containerapp up': New revisions-mode and target-label options for Labels revision mode.
* 'az containerapp label-history': New list and show commands to see label revision assignment history.
* 'az containerapp revision set-mode': New target-label option for Labels revision mode.

### General Guidelines


src/containerapp/azext_containerapp/custom.py
```diff
@@ -1256,9 +1266,9 @@ def containerapp_up(cmd,
         target_port = 80 if not target_port else target_port
 
     if image:
-        if ingress and not target_port:
-            target_port = 80
-            logger.warning("No ingress provided, defaulting to port 80. Try `az containerapp up --ingress %s --target-port <port>` to set a custom port.", ingress)
+        if ingress and not target_port and target_port != 0:
+            target_port = 0
+            logger.warning("No target-port provided, defaulting to auto-detect. Try `az containerapp up --ingress %s --target-port <port>` to set a custom port.", ingress)
```
Tratcher, line 1271: A small bugfix for my prior feature 😁.

yanzhudd:
Hi @Greedygre 
please review this PR as well, thanks

src/containerapp/HISTORY.rst
```diff
@@ -5,6 +5,11 @@ Release History
 upcoming
 ++++++
 * 'az containerapp debug': Open an SSH-like interactive shell within a container app debug console.
+* 'az containerapp create': New target-label option for Labels revision mode.
+* 'az containerapp update': New revisions-mode and target-label options for Labels revision mode.
+* 'az containerapp up': New revisions-mode and target-label options for Labels revision mode.
```
Greedygre, line 10: Should the customer use sub command group for this property?

src/containerapp/azext_containerapp/custom.py
```diff
@@ -445,7 +447,8 @@ def create_containerapp(cmd,
                         transport="auto",
                         ingress=None,
                         allow_insecure=False,
-                        revisions_mode="single",
+                        revisions_mode=None,
```
Greedygre, line 450: Why change the default value of `revision-mode`? It is a breaking change.

image

src/containerapp/azext_containerapp/containerapp_decorator.py
```diff
@@ -797,6 +813,7 @@ def parent_construct_payload(self):
         config_def = deepcopy(ConfigurationModel)
         config_def["secrets"] = secrets_def
         config_def["activeRevisionsMode"] = self.get_argument_revisions_mode()
+        config_def["targetLabel"] = self.get_argument_target_label()
```
Greedygre, line 816: The targetLabel should only be set when the revision mode is `Labels`, right?
We can add argument validation for this case.

Greedygre:
please make sure all checks in CI passed.

If the PR is to add new feature support, please get approval from your team who could be responsible for your business logic.

src/containerapp/azext_containerapp/commands.py
```diff
@@ -270,3 +270,14 @@ def load_command_table(self, args):
         g.custom_command('update', 'update_maintenance_config')
         g.custom_command('remove', 'remove_maintenance_config', confirmation=True)
         g.custom_show_command('list', 'list_maintenance_config')
+
+    with self.command_group('containerapp label-history', is_preview=True) as g:
+        g.custom_show_command('show', 'show_labelhistory')
+        g.custom_command('list', 'list_labelhistory')
+
+    with self.command_group('containerapp revision') as g:
```
Juliehzl, line 278: add is_preview=True

src/containerapp/azext_containerapp/commands.py
```diff
@@ -270,3 +270,14 @@ def load_command_table(self, args):
...
+        g.custom_command('set-mode', 'set_revision_mode', exception_handler=ex_handler_factory())
```
Juliehzl, line 279: add is_preview=True

src/containerapp/azext_containerapp/custom.py
```diff
@@ -445,7 +446,8 @@ def create_containerapp(cmd,
...
```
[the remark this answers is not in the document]
Juliehzl: Why do you want to fix the warning issue in YAML scenario?

src/containerapp/azext_containerapp/custom.py
Tratcher: The default is not actually changing, 'single' is still applied as the default in the CP. See https://github.com/Azure/azure-cli-extensions/pull/8324#discussion_r1859074278

src/containerapp/HISTORY.rst
Tratcher: ?

src/containerapp/azext_containerapp/custom.py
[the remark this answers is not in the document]
Tratcher: Why not? I was testing labels mode in the YAML scenario. As is, this warning is incorrect, it warns even when no extra parameters are given.

If we want to keep the default as is then the warning should be removed.

src/containerapp/azext_containerapp/containerapp_decorator.py
Tratcher: I will add this for create, but it's harder for up and update. revisions-mode and target-label are both saved in the app model and can be updated independently. target-label is also not cleared from the model if you leave labels mode, so I don't want to generate errors based on its presence.

The RP will enforce that target-label is required in labels mode. It will not complain if target-label is present outside of labels mode.

src/containerapp/azext_containerapp/custom.py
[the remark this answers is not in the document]
Greedygre: Hi @Tratcher 
We should not change the default value of `revisons_mode` in CLI side.
If we want to fix the warning, we can remove `or self.get_argument_revisions_mode()` from the yaml warning check [here](https://github.com/Azure/azure-cli-extensions/blob/cef77fd5e8c3fb684ffab04bca16bb320e4d2dc3/src/containerapp/azext_containerapp/containerapp_decorator.py#L1120) .

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3359,3 +3370,176 @@ def containerapp_debug(cmd, resource_group_name, name, container=None, revision=
...
+            "label": target_label
+        }])
+
+    containerapp_def["properties"]["configuration"]["activeRevisionsMode"] = mode
+    containerapp_def["properties"]["configuration"]["targetLabel"] = target_label
+
+    _get_existing_secrets(cmd, resource_group_name, name, containerapp_def)
+
+    try:
+        r = ContainerAppPreviewClient.create_or_update(
```
Greedygre, line 3426: We should use `Patch` method for the set-mode, because we only want to update the `mode` parameter, shouldn't we?

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3359,3 +3370,176 @@ def containerapp_debug(cmd, resource_group_name, name, container=None, revision=
...
+        safe_set(containerapp_def, "properties", "configuration", "ingress", "traffic", value=[{
+            "revisionName": containerapp_def["properties"]["latestRevisionName"],
+            "weight": 100,
+            "label": target_label
+        }])
+
+    containerapp_def["properties"]["configuration"]["activeRevisionsMode"] = mode
+    containerapp_def["properties"]["configuration"]["targetLabel"] = target_label
+
+    _get_existing_secrets(cmd, resource_group_name, name, containerapp_def)
```
Greedygre, line 3423: If we change to use the `Patch`, we don't need to `_get_existing_secrets`.

src/containerapp/HISTORY.rst
Tratcher: These are covered by the new tests.

src/containerapp/HISTORY.rst
```diff
@@ -5,6 +5,11 @@ Release History
 upcoming
 ++++++
 * 'az containerapp debug': Open an SSH-like interactive shell within a container app debug console.
+* 'az containerapp create': New target-label option for Labels revision mode.
+* 'az containerapp update': New revisions-mode and target-label options for Labels revision mode.
```
ShichaoQiu, line 9:
```suggestion
* 'az containerapp up/update': New revisions-mode and target-label options for Labels revision mode.
```

src/containerapp/azext_containerapp/_help.py
```diff
@@ -2198,3 +2198,30 @@
       text: |
           az containerapp debug -n MyContainerapp -g MyResourceGroup --revision MyRevision --replica MyReplica --container MyContainer
 """
+
+helps['containerapp label-history'] = """
+    type: group
+    short-summary: Show the history for one more labels on the Container App.
```
ShichaoQiu, line 2204:
```suggestion
    short-summary: Show the history for one or more labels on the Container App.
```

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3359,3 +3370,174 @@ def containerapp_debug(cmd, resource_group_name, name, container=None, revision=
             if conn.is_connected:
                 logger.info("Caught KeyboardInterrupt. Sending ctrl+c to server")
                 conn.send(SSH_CTRL_C_MSG)
+
+
+def list_labelhistory(cmd, resource_group_name, name):
```
Greedygre, line 3375: ni:
Typo: In word 'labelhistory'

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3359,3 +3370,174 @@ def containerapp_debug(cmd, resource_group_name, name, container=None, revision=
...
+    _validate_subscription_registered(cmd, CONTAINER_APPS_RP)
+    try:
+        return LabelHistoryPreviewClient.list(cmd, resource_group_name, name)
+    except Exception as e:
+        handle_raw_exception(e)
+
+
+def show_labelhistory(cmd, resource_group_name, name, label):
```
Greedygre, line 3383: ni:
same as above

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3359,3 +3370,174 @@ def containerapp_debug(cmd, resource_group_name, name, container=None, revision=
...
+
+
+def set_revision_mode(cmd, resource_group_name, name, mode, target_label=None, no_wait=False):
+    _validate_subscription_registered(cmd, CONTAINER_APPS_RP)
+
+    containerapp_def = None
+    try:
+        containerapp_def = ContainerAppPreviewClient.show(cmd=cmd, resource_group_name=resource_group_name, name=name)
+    except:
+        pass
```
Greedygre, line 3398:
```suggestion
    except Exception as e:
        handle_raw_exception(e)
```

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3359,3 +3370,174 @@ def containerapp_debug(cmd, resource_group_name, name, container=None, revision=
...
+    _validate_subscription_registered(cmd, CONTAINER_APPS_RP)
+
+    if not name:
+        name = _get_app_from_revision(revision)
+
+    containerapp_def = None
+    try:
+        containerapp_def = ContainerAppPreviewClient.show(cmd=cmd, resource_group_name=resource_group_name, name=name)
+    except:
+        pass
```
Greedygre, line 3441:
```suggestion
    except Exception as e:
        handle_raw_exception(e)
```

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3359,3 +3370,174 @@ def containerapp_debug(cmd, resource_group_name, name, container=None, revision=
...
+            "revisionName": containerapp_def["properties"]["latestRevisionName"],
+            "weight": 100,
+            "label": target_label
+        }])
+
+    containerapp_def["properties"]["configuration"]["activeRevisionsMode"] = mode
+    containerapp_def["properties"]["configuration"]["targetLabel"] = target_label
+
+    try:
+        r = ContainerAppPreviewClient.update(
```
Greedygre, line 3424: We are using `update` (which is `PATCH`), the payload should only relate to the params that are needed.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3359,3 +3370,174 @@ def containerapp_debug(cmd, resource_group_name, name, container=None, revision=
...
+
+
+def remove_revision_label(cmd, resource_group_name, name, label, no_wait=False):
+    _validate_subscription_registered(cmd, CONTAINER_APPS_RP)
+
+    containerapp_def = None
+    try:
+        containerapp_def = ContainerAppPreviewClient.show(cmd=cmd, resource_group_name=resource_group_name, name=name)
+    except:
+        pass
```
Greedygre, line 3508:
```suggestion
    except Exception as e:
        handle_raw_exception(e)
```

Greedygre:
Hi @Tratcher 
This PR LGTM expect above ni comments, can you take a look and address them? 
Also, please sync the main branch to fix the integration test.
Thanks.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3415,3 +3426,174 @@ def delete_http_route_config(cmd, resource_group_name, name, http_route_config_n
...
+        raise ValidationError("Ingress is requried for labels mode.")
+
+    containerapp_patch_def = {}
+    safe_set(containerapp_patch_def, "properties", "configuration", "activeRevisionsMode", value=mode)
+    safe_set(containerapp_patch_def, "properties", "configuration", "targetLabel", value=target_label)
+
+    # If we're going into labels mode, replace the default traffic config with the target label and latest revision.
+    # Otherwise all revisions will be deactivated.
+    traffic = safe_get(containerapp_def, "properties", "configuration", "ingress", "traffic")
+    if mode.lower() == "labels" and len(traffic) == 1 and safe_get(traffic[0], "latestRevision") is True:
```
Greedygre, line 3473: If the traffic length > 1, we don't need to update traffic lable?

Greedygre (approved):
Please make sure

src/containerapp/azext_containerapp/custom.py
Tratcher: If it's more than 1 then they're using a manual configuration and we shouldn't change it. If it's just 1 entry then it's likely the default config.

src/containerapp/azext_containerapp/custom.py
```diff
@@ -1257,9 +1268,9 @@ def containerapp_up(cmd,
         target_port = 80 if not target_port else target_port
 
     if image:
-        if ingress and not target_port:
-            target_port = 80
-            logger.warning("No ingress provided, defaulting to port 80. Try `az containerapp up --ingress %s --target-port <port>` to set a custom port.", ingress)
+        if ingress and not target_port and target_port != 0:
+            target_port = 0
+            logger.warning("No target-port provided, defaulting to auto-detect. Try `az containerapp up --ingress %s --target-port <port>` to set a custom port.", ingress)
```
zhoxing-ms, line 1273: Will this change lead to a breaking change?

linter_exclusions.yml
```diff
@@ -401,6 +401,12 @@ containerapp github-action add:
     service_principal_tenant_id:
       rule_exclusions:
       - option_length_too_long
+containerapp label-history show:
+  rule_exclusions:
+  -  missing_command_test_coverage
+containerapp label-history list:
+  rule_exclusions:
+  -  missing_command_test_coverage
```
zhoxing-ms, line 409: Why not add some tests for these commands?

Tratcher: The commands won't work properly until our next control plane deployment goes out, there was a change in the object structure. I can add tests after that.

src/containerapp/azext_containerapp/custom.py
Tratcher: No, the auto-detection will still cover the old scenario where the app was listening on port 80.

linter_exclusions.yml
zhoxing-ms: Got it, thanks~

src/containerapp/azext_containerapp/custom.py
zhoxing-ms: Sounds good

zhoxing-ms (approved):

Greedygre:
Hi @Tratcher 

I got error when I tried following command:
` az containerapp revision set-mode --mode labels -n xinyutest2 -g client.env_rg_eastus`

cli.azure.cli.core.azclierror: az containerapp revision set-mode: 'labels' is not a valid value for '--mode'. Allowed values: single, multiple.

This is because you didn't update the --mode setting:
image

```
    with self.argument_context('containerapp revision set-mode') as c:
        c.argument('mode', arg_type=get_enum_type(['single', 'multiple']), help="The active revisions mode for the container app.")
```

But I notice you have test for ` az containerapp revision set-mode --mode labels`, do you only update it from the `azure-cli` locally?

My suggestion is please update `azure-cli-extensions` - `_params.py` and adding following code:
```
    with self.argument_context('containerapp revision set-mode') as c:
        c.argument('mode', arg_type=get_enum_type(['single', 'multiple',  'labels']), help="The active revisions mode for the container app.")
```

PR #8349, opened 2024-11-28
[Spring] Detect failed Update App operation during polling

domainname:
Update App operation may fail due to quota usage. So I'm adding a few lines to detect such a failure during polling.

### DIFF
--- a/src/spring/azext_spring/app.py
+++ b/src/spring/azext_spring/app.py
@@ -311,7 +311,7 @@ def app_update(cmd, client, resource_group, service, name,
     wait_till_end(cmd, *pollers)
 
     for p in pollers:
-        if p.status() == 'Failed':
+        if p.status() == 'Failed' and isinstance(p.result(), Exception):
             raise p.result()
 
     return app_get(cmd, client, resource_group, service, name)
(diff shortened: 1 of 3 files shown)


yonzhan:

src/spring/azext_spring/app.py
```diff
@@ -309,6 +309,11 @@ def app_update(cmd, client, resource_group, service, name,
     if no_wait:
         return
     wait_till_end(cmd, *pollers)
+
+    for p in pollers:
+        if p.status() == 'Failed' and isinstance(p.result(), Exception):
+            raise p.result()
```
yuwzho, line 315:
```suggestion
    if any(p.status() == 'Failed' and isinstance(p.result(), Exception) for p in pollers):
        raise next(p.result() for p in pollers if p.status() == 'Failed' and isinstance(p.result(), Exception))
```

nit: could be written like this in python

src/spring/setup.py
```diff
@@ -16,7 +16,7 @@
 
 # TODO: Confirm this is the right version number you want and it matches your
 # HISTORY.rst entry.
-VERSION = '1.26.0'
+VERSION = '1.26.1'
```
zhoxing-ms, line 19: Due to the impact of CCOA, if you want to merge this PR in advance, please do not upgrade the extension version for now

zhoxing-ms (approved):

PR #8379, opened 2024-12-19
[devcenter] Update data plane API to v2024-10-01-preview

am-lim:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

src/devcenter/azext_devcenter/utils.py
```diff
@@ -88,4 +95,15 @@ def get_delayed_time(delay_time, action_time):
...
+    delayed_time = action_time + timedelta(hours=hours, minutes=minutes)
+    return delayed_time.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+
+
+def get_dataplane_endpoint(cli_ctx, endpoint=None, dev_center=None, project_name=None):
+    validate_endpoint(endpoint, dev_center)
+    if endpoint is None and dev_center is not None:
+        project = get_project_data(cli_ctx, dev_center, project_name)
+        endpoint = project["devCenterUri"]
+    endpoint = endpoint.split('//', 1)[-1]
```
kairu-ms, line 107: It seems the user can defined data-plane endpoints by three ways:

- directly provide the endpoint
- provide the `dev_center` and using the first devCenterUri of the project
- provide both `dev_center` and `project_name` and using the devCenterUri

My question is why need the third way if providing the `dev_center` only can get the devCenterUri in response.

And for the second way actually you can config the data plane client like this which will retrieve the `properties.devCenterUri` as endpoint from devcenter resource from the control plane api.
image

kairu-ms:
The git hooks are available for __azure-cli__ and __azure-cli-extensions__ repos. They could help you run required checks before creating the PR. Please sync the latest code with latest `dev` branch (for __azure-cli__) or `main` branch (for __azure-cli-extensions__). After that please run the following commands to enable git hooks:
 
```
pip install azdev --upgrade
azdev setup -c <your azure-cli repo path> -r <your azure-cli-extensions repo path>
```

src/devcenter/azext_devcenter/utils.py
am-lim: @kairu-ms For the question about why the third way is included, for commands that have a project name, we fail earlier and let the user know that the reason for the failure is that the project does not exist within the dev center. Otherwise, for example the user may receive a 401 not found for their get command or an empty list for their list command - but the user would be unaware that this is caused because the project does not exist within the chosen dev center. 

I don't know if we can config the data plane client "By resource property" in the screenshot. This is because users using the data plane are only assumed to have read access to the project, not the dev center. Also, some commands do not have project name as a parameter (For example, in some list commands the user can list all the resources available in a dev center).  This is why we need to use the azure resource graph query on all projects.  Given this context, not sure if we can set up the devCenterUri this way?

kairu-ms (approved):

PR #8439, opened 2025-01-26
Frank/analysis follow up

frankqianms:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

src/apic-extension/azext_apic_extension/command_patches.py
```diff
@@ -519,3 +543,199 @@ def pre_operations(self):
             "region_name": args.region_name,
             "msi_resource_id": args.msi_resource_id
         }
+
+
+@register_command(
+    "apic import apim",
+    is_preview=True,
```
blackchoey, line 550: I'm not sure whether we expect to set this as preview, since the original command is in GA status. Let's confirm with PM.

src/apic-extension/azext_apic_extension/command_patches.py
```diff
@@ -519,3 +543,199 @@ def pre_operations(self):
...
+    def _build_arguments_schema(cls, *args, **kwargs):
+        # pylint: disable=protected-access
+        args_schema = super()._build_arguments_schema(*args, **kwargs)
+        args_schema.filter._registered = False
+        return args_schema
+
+    def pre_operations(self):
+        super().pre_operations()
+        args = self.ctx.args
+        args.analyzer_type = "spectral"
```
blackchoey, line 625: Are you tring to set a default vaule for --analyzer-type, but still exposes this parameter, or you're trying to hide this parameters to users?

src/apic-extension/azext_apic_extension/command_patches.py
```diff
@@ -519,3 +543,199 @@ def pre_operations(self):
...
+
+
+class DeleteApiAnalysisConfig(DefaultWorkspaceParameter, DeleteApiAnalysis):
+    pass
+
+
+class ImportApiAnalysisRuleset(DefaultWorkspaceParameter, ImportRuleset):
+    # pylint: disable=C0301
```
blackchoey, line 633: which line is too long? are we able to optimize the code instead of disable this lint rule?

src/apic-extension/azext_apic_extension/command_patches.py
```diff
@@ -519,3 +543,199 @@ def pre_operations(self):
...
+    # Zip and encode the ruleset folder to base64
+    def zip_folder_to_buffer(self, folder_path):
+        # pylint: disable=unused-variable
```
blackchoey, line 636: You could try  to add `_` in front of the unused variables to avoid the lint warning, which seems widely used in many projects when some function parameters or returned value from external library is indeed not used.

src/apic-extension/azext_apic_extension/command_patches.py
```diff
@@ -519,3 +543,199 @@ def pre_operations(self):
...
+    def pre_operations(self):
+        super().pre_operations()
+        args = self.ctx.args
+
+        args.format = 'inline-zip'
+        args.value = self.zip_folder_to_buffer(str(args.ruleset_folder_path))
+
+
+class ExportApiAnalysisRuleset(DefaultWorkspaceParameter, ExportRuleset):
+    # pylint: disable=C0301
```
blackchoey, line 668: Also avoid disable this rule if we're able to optimize the code to avoid the lint warning

src/apic-extension/azext_apic_extension/command_patches.py
```diff
@@ -519,3 +543,199 @@ def pre_operations(self):
...
+    def _build_arguments_schema(cls, *args, **kwargs):
+        # pylint: disable=protected-access
+        args_schema = super()._build_arguments_schema(*args, **kwargs)
+
+        args_schema.ruleset_folder_path = AAZStrArg(
+            options=["--path"],
+            help="The folder path to extract the ruleset files.",
+            required=True,
+        )
+        # args_schema.result = None
```
blackchoey, line 685: Remove this comment?

src/apic-extension/azext_apic_extension/command_patches.py
```diff
@@ -519,3 +543,199 @@ def pre_operations(self):
...
+
+            if response_format == 'link':
+                logger.info('Fetching specification from: %s', exportedResults)
+                getReponse = requests.get(exportedResults, timeout=10)
+                if getReponse.status_code == 200:
+                    exportedResults = getReponse.content.decode()
+                else:
+                    error_message = f'Error while fetching the results from the link. Status code: {getReponse.status_code}'
+                    logger.error(error_message)
+                    raise getReponse.raise_for_status()
```
blackchoey, line 704: Can you share what kind of CLI error is raised with this line? Want to ensure we can get meaningful error type report from telemetry.

src/apic-extension/azext_apic_extension/command_patches.py
```diff
@@ -519,3 +543,199 @@ def pre_operations(self):
...
+        result = self.deserialize_output(self.ctx.vars.instance, client_flatten=True)
+        arguments = self.ctx.args
+
+        if result:
+            response_format = result['format']
+            exportedResults = result['value']
+
+            if response_format == 'link':
+                logger.info('Fetching specification from: %s', exportedResults)
+                getReponse = requests.get(exportedResults, timeout=10)
```
blackchoey, line 698: How do we decide the timeout value for this request?

src/apic-extension/azext_apic_extension/command_patches.py
```diff
@@ -243,6 +262,11 @@ def _build_arguments_schema(cls, *args, **kwargs):
 
 
 # `az apic service commands`
+@register_command(
+    "apic import-from-apim",
+    hide=True,
```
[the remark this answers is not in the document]
frankqianms: do you mean that we'd better to set it in aaz?

src/apic-extension/azext_apic_extension/command_patches.py
frankqianms: actually, this value is only allowed to be set as "spectral". not sure if it still needs to be exposed to users, will talk with PM on this.

src/apic-extension/azext_apic_extension/command_patches.py
frankqianms: i think HTTPError, RequestException or ConnectionError will be raised. For example, if the link is invalid like 'https://example.com/nonexistent', then an error with status 404 will be raised

src/apic-extension/azext_apic_extension/command_patches.py
frankqianms: I refered to `ExportMetadataSchemaExtension` and `ExportSpecificationExtension` as both commands set the same timeout.

src/apic-extension/azext_apic_extension/command_patches.py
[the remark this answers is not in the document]
blackchoey: Yes. I'm thinking maybe we want to have a practice on what kind of changes should be made in aaz and what kind of changes should made with customized python code. So we and other can better maintain the CLI. Feel free to share your thoughts.

src/apic-extension/azext_apic_extension/command_patches.py
blackchoey: Can you investigate which Azure CLI error type is used in this case? https://github.com/Azure/azure-cli/blob/dev/doc/error_handling_guidelines.md does not mention these errors, I'm not sure whether these will be mapped to CLIInternalError, which should not be the case.

src/apic-extension/azext_apic_extension/command_patches.py
blackchoey: Thanks. Can we investigate more on the timeout? I personally would suggest use the default timeout value (if the method have) in all these commands if we don't have special reason to change it.

src/apic-extension/azext_apic_extension/command_patches.py
```diff
@@ -345,7 +345,6 @@ class ShowIntegrationExtension(DefaultWorkspaceParameter, ShowIntegration):
 
 @register_command(
     "apic integration create apim",
-    is_preview=True,
```
blackchoey, line 324: the integration feature should still in preview,  so we need to keep this flag. The import apim functionality should be in GA status per my understanding. Let's double confirm with PM.

src/apic-extension/linter_exclusions.yml
```diff
@@ -0,0 +1,96 @@
+---
```
frankqianms: This file is added to pass the linter for "disallowed_html_tag_from_parameter". As  "generic_update_add", "generic_update_remove", "generic_update_set" are auto generated for update az commands.

Is it okay to use the exclusion?

blackchoey: I remember we didn't meet such lint issues previously. Is this because some new linters are introduced recently?

src/apic-extension/azext_apic_extension/tests/latest/test_register_command.py
```diff
@@ -47,7 +47,7 @@ def test_register_with_yml_spec(self):
             self.check('description', 'API Description'), # default value when spec does not have description
             self.check('name', 'openapi'), # hard coded when spec is swagger or openapi
             self.check('specification.name', 'openapi'),
-            self.check('specification.version', '3-0-0'),
+            self.check('specification.version', '3.0.0'),
```
blackchoey, line 50: What causes the behavior change?

frankqianms: Due to api center server side format changes, we removed replace logic here https://github.com/Azure/azure-cli-extensions/pull/8439/files#diff-4e0b385dc3ef344952de17f1d2095f3c5a7a428e4c4bb18e4954f29fefd46698:~:text=extracted_definition_version%20%3D%20first_value.lower(). And this line of test change is syncronized.

src/apic-extension/linter_exclusions.yml
frankqianms: Checking with CLI team. Waiting for their ideas.

frankqianms: From azdev v0.1.92 release note: azdev linter: Set disallowed_html_tags and broken_site_link detection in high severity to block future format issues
https://pypi.org/project/azdev/0.2.0/#:~:text=0.1.92,future%20format%20issues

src/apic-extension/azext_apic_extension/tests/latest/test_register_command.py
blackchoey: I would suggest keep previous behavior if no explicit ask to change it. Not sure whether users rely on existing naming convention to do something.

frankqianms: It is from partners ask.  And keeping the previous behavior will cause the display format error in the Azure portal.

blackchoey: Thanks. Looks good to me as it just changes the display name.

blackchoey:
LGTM

PR #8492, opened 2025-02-19
[load] Adding CLI support for engine reference identity feature

Himanshu49:
Added two new parameter --engine-ref-ids and --engine-ref-id-type to 'az load test create' and 'az load test update' commands.
Added their corresponding validators, logic and tests.

---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/load/HISTORY.rst
+++ b/src/load/HISTORY.rst
@@ -4,7 +4,7 @@ Release History
 ===============
 
 
-1.5.1
+1.6.0
 ++++++
 * Add support for engine reference identity using CLI. Engine reference identity can be set using `--engine-ref-id-type` and `--engine-ref-ids` argument in 'az load test create' and 'az load test update' commands. Engine reference identity set in YAML config file under key `referenceIdentities` with `kind` as `Engine` will also be honoured.
 
--- a/src/load/setup.py
+++ b/src/load/setup.py
@@ -10,7 +10,7 @@ from setuptools import setup, find_packages
 
 
 # HISTORY.rst entry.
-VERSION = '1.5.1'
+VERSION = '1.6.0'
 
 # The full list of classifiers is available at
 # https://pypi.python.org/pypi?%3Aaction=list_classifiers


src/load/azext_load/data_plane/utils/argtypes.py
```diff
@@ -392,6 +392,23 @@
...
+    type=str,
+    completer=get_generic_completion_list(
+        utils.get_enum_values(models.EngineIdentityType)
+    ),
+    choices=utils.get_enum_values(models.EngineIdentityType),
+    help="Type of identity to be used for the engine.",
+)
+
+engine_reference_identities = CLIArgumentType(
+    options_list=["--engine-reference-identities"],
```
mbhardwaj-msft, line 406: I believe the number of characters in the option have to be <=22

src/load/azext_load/data_plane/utils/utils.py
```diff
@@ -337,6 +337,14 @@ def convert_yaml_to_test(cmd, data):
         new_body["passFailCriteria"] = utils_yaml_config.yaml_parse_failure_criteria(data=data)
     if data.get(LoadTestConfigKeys.AUTOSTOP) is not None:
         new_body["autoStopCriteria"] = utils_yaml_config.yaml_parse_autostop_criteria(data=data)
+    
+    if data.get(LoadTestConfigKeys.REFERENCE_IDENTITIES):
+        for identity in data[LoadTestConfigKeys.REFERENCE_IDENTITIES]:
+            if identity.get(LoadTestConfigKeys.KIND) == LoadTestConfigKeys.ENGINE:
```
mitsha-microsoft, line 343: Should we handle if the get fails?

mitsha-microsoft: Does it throw or return `None`?

src/load/azext_load/data_plane/utils/validators.py
```diff
@@ -522,3 +523,27 @@ def validate_regionwise_engines(cmd, namespace):
             )
         regionwise_engines.append({"region": key.strip().lower(), "engineInstances": value})
     namespace.regionwise_engines = regionwise_engines
+
+def validate_engine_reference_identities(namespace):
+    """Extracts multiple space-separated identities"""
+    if isinstance(namespace.engine_reference_identities, list):
+        engine_reference_identities_list = []
+        for item in namespace.engine_reference_identities:
+            if not is_valid_resource_id(item):
```
mitsha-microsoft, line 532: Can we not use a common validator for all reference identities?

src/load/azext_load/data_plane/utils/utils.py
mbhardwaj-msft, line 343: do we need null check on `identity`?

src/load/azext_load/data_plane/utils/utils.py
```diff
@@ -514,6 +524,17 @@ def create_or_update_test_with_config(
...
+    # if argument is provided prefer that over yaml values
+    if engine_reference_identity_type:
+        validators.validate_engine_identities_id_and_type(engine_reference_identity_type, engine_reference_identities)
+        if engine_reference_identity_type:
+            new_body["engineBuiltinIdentityType"] = engine_reference_identity_type
+        if engine_reference_identities:
+            new_body["engineBuiltinIdentityIds"] = engine_reference_identities
+    elif yaml_test_body.get("engineBuiltinIdentityType"):
+        new_body["engineBuiltinIdentityType"] = yaml_test_body.get("engineBuiltinIdentityType")
+        new_body["engineBuiltinIdentityIds"] = yaml_test_body.get("engineBuiltinIdentityIds")
```
mbhardwaj-msft, line 536: do we need to explicitly handle the patch using backfill from existing test payload, or is API able to handle that?

src/load/azext_load/data_plane/utils/utils.py
```diff
@@ -745,9 +775,10 @@ def upload_zipped_artifacts_helper(
                 file_to_upload=zip_artifact, file_type=AllowedFileTypes.ZIPPED_ARTIFACTS,
                 wait=wait
             )
-            if wait and file_response.get("validationStatus") != "VALIDATION_SUCCESS":
+            if wait and file_response.get("validationStatus") not in ("VALIDATION_SUCCESS", "NOT_VALIDATED"):
```
mbhardwaj-msft, line 778: wouldn't the `NOT_VALIDATED` state conflict with fragments feature, since delayed validation would be possible going forward?

src/load/azext_load/data_plane/utils/argtypes.py
```diff
@@ -392,6 +392,23 @@
...
 )
 
+engine_reference_identity_type = CLIArgumentType(
+    options_list=["--engine-reference-identity-type"],
+    type=str,
+    completer=get_generic_completion_list(
+        utils.get_enum_values(models.EngineIdentityType)
+    ),
+    choices=utils.get_enum_values(models.EngineIdentityType),
+    help="Type of identity to be used for the engine.",
```
mbhardwaj-msft, line 402: do we not want to include allowed values in help text?

src/load/azext_load/data_plane/utils/utils_yaml_config.py
```diff
@@ -137,3 +139,27 @@ def yaml_parse_loadtest_configuration(cmd, data):
     if data.get(LoadTestConfigKeys.SPLIT_CSV) is not None:
         load_test_configuration["splitAllCSVs"] = _yaml_parse_splitcsv(data=data)
     return load_test_configuration
+
+def yaml_parse_engine_identities(data):
+    engine_identities = []
+    reference_type = None
+    reference_identities = data.get(LoadTestConfigKeys.REFERENCE_IDENTITIES)
+    for identity in reference_identities:
```
mbhardwaj-msft, line 147: can we check if this flow is working for when user sets type to None after it's been set some other type?

src/load/azext_load/data_plane/utils/argtypes.py
```diff
@@ -392,6 +392,23 @@
     help="Specify the engine count for each region in the format: region1=engineCount1 region2=engineCount2 .... Use region names in the format accepted by Azure Resource Manager (ARM). Ensure the regions are supported by Azure Load Testing. Multi-region load tests can only target public endpoints.",
 )
 
+engine_reference_identity_type = CLIArgumentType(
+    options_list=["--engine-reference-identity-type"],
+    type=str,
+    completer=get_generic_completion_list(
```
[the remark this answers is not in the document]
Himanshu49: It works in bash when activate interactive mode. Will check in powershell once.

src/load/azext_load/data_plane/utils/argtypes.py
Himanshu49: 'choices' parameter of CLIArgumentType takes care of that. help will have all the allowed values

src/load/azext_load/data_plane/utils/utils.py
Himanshu49: If the key is not present the get method returns None. It should be fine in my opinion. Have tested locally with empty and other variation.

Himanshu49: yes, this can happen if null is explicitly provided in yaml.

src/load/azext_load/data_plane/utils/utils_yaml_config.py
Himanshu49: Tried it, working as expected

src/load/azext_load/data_plane/utils/validators.py
Himanshu49: We dont have validator for other reference identities as of now, making it generic reference id validator.

src/load/azext_load/data_plane/utils/argtypes.py
Himanshu49: After discussion with Nag, renamed to --engine-ref-ids and --engine-ref-id-type

src/load/azext_load/data_plane/utils/utils.py
Himanshu49: Updated

kairu-ms (approved):

src/load/azext_load/data_plane/utils/argtypes.py
```diff
@@ -392,6 +392,23 @@
...
+        utils.get_enum_values(models.EngineIdentityType)
+    ),
+    choices=utils.get_enum_values(models.EngineIdentityType),
+    help="Type of identity to be configured for the engine.",
+)
+
+engine_ref_ids = CLIArgumentType(
+    options_list=["--engine-ref-ids"],
+    nargs="+",
```
mitsha-microsoft, line 407: For my understanding, what is `nargs="+"` ?

src/load/azext_load/data_plane/utils/models.py
```diff
@@ -11,6 +11,12 @@ class IdentityType(str, Enum):
     UserAssigned = "UserAssigned"
 
 
+class EngineIdentityType(str, Enum):
```
mitsha-microsoft, line 14: I dont think we need this, this should be already present in the SDK models now?

src/load/azext_load/data_plane/utils/utils.py
Himanshu49: Going forward NOT_VALIDATED is a valid file status for files other than test plan. What conflict can this cause?

src/load/azext_load/data_plane/utils/models.py
mitsha-microsoft: Can retain it for now though

src/load/azext_load/data_plane/utils/validators.py
```diff
@@ -522,3 +523,28 @@ def validate_regionwise_engines(cmd, namespace):
...
+    engine_ref_id_type = incoming_engine_ref_id_type or exisiting_engine_ref_id_type
+    if engine_ref_id_type != EngineIdentityType.UserAssigned and engine_ref_ids:
+        raise InvalidArgumentValueError(
+            "engine-ref-ids should not be provided when engine-ref-id-type is None or SystemAssigned"
+        )
+
+    # If engine_ref_id_type is UserAssigned, then engine_ref_ids is expected.
+    if incoming_engine_ref_id_type == EngineIdentityType.UserAssigned and engine_ref_ids is None:
+        raise InvalidArgumentValueError(
+            "Atleast one engine-ref-ids should be provided when engine-ref-id-type is UserAssigned"
```
mitsha-microsoft, line 549: Lets expand and not use the acronym in the error messages?

"Atleast one engine-reference-identity should be provided..."

Same for the error message in L543

src/load/azext_load/data_plane/utils/argtypes.py
Himanshu49: expects at least one value if option is used. Empty is not valid input.

src/load/azext_load/data_plane/utils/validators.py
Himanshu49: but the option they would be using --engine-ref-ids. IMO either we can go with engine reference identities or keep it same way.

mitsha-microsoft (approved):

mbhardwaj-msft (approved):

src/load/azext_load/data_plane/utils/argtypes.py
```diff
@@ -392,6 +392,23 @@
...
+    validator=validators.validate_engine_ref_ids,
+    help="Space separated list of fully qualified resource IDs of the managed identities to be configured on the engine. Required only for user assigned identities. ",
```
mbhardwaj-msft, line 409: Have we updated the `help.py` as well?

src/load/azext_load/data_plane/utils/utils_yaml_config.py
```diff
@@ -137,3 +139,31 @@ def yaml_parse_loadtest_configuration(cmd, data):
...
+
+# pylint: disable=line-too-long
+# Disabling this because dictionary key are too long
+def yaml_parse_engine_identities(data):
+    engine_identities = []
+    reference_type = None
+    reference_identities = data.get(LoadTestConfigKeys.REFERENCE_IDENTITIES)
+    for identity in reference_identities:
+        if identity and identity.get(LoadTestConfigKeys.KIND) == LoadTestConfigKeys.ENGINE:
+            if reference_type and identity.get(LoadTestConfigKeys.TYPE) != reference_type:
```
mbhardwaj-msft, line 152: nit: instead of doing `.get` multiple times, we can store them in variables and use those

src/load/azext_load/data_plane/utils/utils.py
```diff
@@ -337,8 +339,17 @@ def convert_yaml_to_test(cmd, data):
         new_body["passFailCriteria"] = utils_yaml_config.yaml_parse_failure_criteria(data=data)
     if data.get(LoadTestConfigKeys.AUTOSTOP) is not None:
         new_body["autoStopCriteria"] = utils_yaml_config.yaml_parse_autostop_criteria(data=data)
+
+    if data.get(LoadTestConfigKeys.REFERENCE_IDENTITIES):
+        for identity in data[LoadTestConfigKeys.REFERENCE_IDENTITIES]:
+            if identity and identity.get(LoadTestConfigKeys.KIND) == LoadTestConfigKeys.ENGINE:
+                new_body["engineBuiltinIdentityType"], new_body["engineBuiltinIdentityIds"] = utils_yaml_config.yaml_parse_engine_identities(data=data)
+                if new_body["engineBuiltinIdentityType"] in [EngineIdentityType.NoneValue, EngineIdentityType.SystemAssigned]:
+                    new_body.pop("engineBuiltinIdentityIds")
```
mbhardwaj-msft, line 348: why are we needing to pop? isn't `yaml_parse_engine_identities` only appending when type is `UserAssigned`?

src/load/azext_load/data_plane/utils/utils.py
```diff
@@ -337,8 +339,17 @@ def convert_yaml_to_test(cmd, data):
         new_body["passFailCriteria"] = utils_yaml_config.yaml_parse_failure_criteria(data=data)
     if data.get(LoadTestConfigKeys.AUTOSTOP) is not None:
         new_body["autoStopCriteria"] = utils_yaml_config.yaml_parse_autostop_criteria(data=data)
+
+    if data.get(LoadTestConfigKeys.REFERENCE_IDENTITIES):
```
mbhardwaj-msft, line 343: I would suggest putting the logic inside this if block in a separate method. Having all the logic in this method keeps increasing complexity of this method and reduces cognition

src/load/HISTORY.rst
```diff
@@ -2,6 +2,10 @@
 
 Release History
 ===============
+1.5.0
+++++++
+* Add support for engine reference identity using CLI. Engine reference identity can be set using `--engine-reference-identity-type` and `--engine-reference-identities` argument in 'az load test create' and 'az load test update' commands. Engine reference identity set in YAML config file under key `referenceIdentities` with `kind` as `Engine` will also be honoured.
```
mbhardwaj-msft, line 7: since we've updated the params, need to update here too

src/load/azext_load/data_plane/utils/argtypes.py
```diff
@@ -392,6 +392,23 @@
     help="Specify the engine count for each region in the format: region1=engineCount1 region2=engineCount2 .... Use region names in the format accepted by Azure Resource Manager (ARM). Ensure the regions are supported by Azure Load Testing. Multi-region load tests can only target public endpoints.",
 )
 
+engine_ref_id_type = CLIArgumentType(
```
mbhardwaj-msft, line 395: for my info, can the params type and ids occur independently? or do we expect both arguments to be present in the cmd always?

src/load/azext_load/data_plane/utils/utils.py
mbhardwaj-msft: my bad, it's a not in, lgtm

mbhardwaj-msft (approved):

src/load/azext_load/data_plane/utils/argtypes.py
mbhardwaj-msft: just curious if we could have some of the validations which are later in the code at this level itself, related to type and the ids

Himanshu49: If existing type is UA and they give new list of ids then it is possible. For that scenario we need to validate combination of existing type and incoming ids.

src/load/HISTORY.rst
Himanshu49: Updated

src/load/azext_load/data_plane/utils/utils.py
Himanshu49: To avoid pop, we can add different logic of checking before adding to dictionary. But maybe this is also fine

src/load/azext_load/data_plane/utils/argtypes.py
Himanshu49: Have updated two examples now in help.py

src/load/azext_load/data_plane/utils/utils.py
Himanshu49: Updated. Moved it altogether to utils_yaml_config file

src/load/azext_load/data_plane/utils/utils_yaml_config.py
Himanshu49: Updated

src/load/HISTORY.rst
```diff
@@ -4,6 +4,11 @@ Release History
 ===============
 
 
+1.5.1
+++++++
+* Add support for engine reference identity using CLI. Engine reference identity can be set using `--engine-ref-id-type` and `--engine-ref-ids` argument in 'az load test create' and 'az load test update' commands. Engine reference identity set in YAML config file under key `referenceIdentities` with `kind` as `Engine` will also be honoured.
+
```
AllyW, line 10: Adding params is taken as feature for cli extension modules and asks a minor version upgrade, therefore the new version should be `1.6.0` instead of `1.5.1`.

For more info about cli extension versioning schema, please refer to: https://github.com/Azure/azure-cli/blob/dev/doc/extensions/versioning_guidelines.md

src/load/azext_load/data_plane/utils/utils.py
mbhardwaj-msft: this is fine if check is needed anyways, just wanted to confirm the flow

mbhardwaj-msft (approved):

kairu-ms (approved):

PR #8530, opened 2025-03-06
[ContainerRegistry] Add 'acrcssc' extension for public preview

cegraybl:
---

### Related command
CSSC (Container Secure Supply Chain) is an extension that configures your registry for continuous scanning and patching of container images. With the Continuous Patching feature in Azure Container Registry, you can automatically scan and patch designated artifacts for OS-level vulnerabilities. The workflow allows you to schedule recurring ACR tasks that scan your list of configured images for vulnerabilities (CVEs) using Trivy and then patch them using Copa.

### General Guidelines

For new extensions:

### About Extension Publish

### Pending items before the publishing PR
- [x] Fix any issues on the description of this PR
- [x] Fill `HISTORY.rst` and `README.rst` with required documentation
- [x] Finist test cases PR: https://github.com/AzureCR/azure-cli-extensions/pull/25
- [x] Perform internal team (ACR) bugbash
- [x] Close all pending issues: tbd


yonzhan:

src/acrcssc/setup.py
```diff
@@ -0,0 +1,61 @@
...
+from codecs import open
+from setuptools import setup, find_packages
+try:
+    from azure_bdist_wheel import cmdclass
+except ImportError:
+    from distutils import log as logger
+    logger.warn("Wheel is not available, disabling bdist_wheel hook")
+
+VERSION = '1.1.1'
+
```
AllyW, line 17: Please adjust the initial version for `acrcssc` to be `1.0.0b1`, instead of `1.1.1`

For more info about the extension versioning schema, please refer to doc here:https://github.com/Azure/azure-cli/blob/dev/doc/extensions/versioning_guidelines.md#initialization

src/acrcssc/azext_acrcssc/_validators.py
```diff
@@ -0,0 +1,183 @@
...
+    _validate_continuouspatch_config(config)
+
+
+def _validate_continuouspatch_file(config_path):
+    if not os.path.exists(config_path):
+        raise InvalidArgumentValueError(f"Config path file: {config_path} does not exist in the path specified")
+    if not os.path.isfile(config_path):
+        raise InvalidArgumentValueError(f"Config path file: {config_path} is not a valid file")
+    if os.path.getsize(config_path) > CONTINUOUSPATCH_CONFIG_SCHEMA_SIZE_LIMIT:
+        raise InvalidArgumentValueError(f"Config path file: {config_path} is too large. Max size limit is 10 MB")
```
huanwu, line 46: Instead of using hardcode 10MB, can we define a constant value.
e.g. raise InvalidArgumentValueError(
    f"Config file '{config_path}' exceeds the size limit of {CONTINUOUSPATCH_CONFIG_SCHEMA_SIZE_LIMIT / (1024 * 1024)} MB."
)

src/acrcssc/azext_acrcssc/_validators.py
```diff
@@ -0,0 +1,183 @@
...
+    try:
+        with open(config_path, 'r') as f:
+            config = json.load(f)
+            validate(config, CONTINUOUSPATCH_CONFIG_SCHEMA_V1)
+            return config
+    except Exception as e:
+        logger.error(f"Error validating the continuous patch config file: {e}")
+        raise InvalidArgumentValueError("File used for --config is not a valid config JSON file. Use --help to see the schema of the config file.")
+    finally:
+        f.close()
```
huanwu, line 64: Is it required as the file is opened using with open(...) and it is auto closed when exiting the with block.

src/acrcssc/azext_acrcssc/_validators.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def _validate_continuouspatch_config(config):
+    if not isinstance(config, dict):
+        raise InvalidArgumentValueError("Config file is not a valid JSON file. Use --help to see the schema of the config file.")
+    for repository in config.get("repositories", []):
```
huanwu, line 70: Question, shall we throw the warning if 'repositories' not exist in the config or repositories list is empty?

src/acrcssc/azext_acrcssc/_validators.py
```diff
@@ -0,0 +1,183 @@
...
+    if schedule is None:
+        return
+    # Extract the numeric value and unit from the timespan expression
+    match = re.match(r'(\d+)(d)$', schedule)
+    if not match:
+        raise InvalidArgumentValueError(error_msg=ERROR_MESSAGE_INVALID_TIMESPAN_FORMAT, recommendation=RECOMMENDATION_SCHEDULE)
+
+    value = int(match.group(1))
+    unit = match.group(2)
+    if unit == 'd' and (value < 1 or value > 30):  # day of the month
```
huanwu, line 133: Suggestion: use constant instead of hardcode.

src/acrcssc/azext_acrcssc/_validators.py
```diff
@@ -0,0 +1,183 @@
...
+    
+
+def validate_run_type(dryrun, run_immediately):
+    if dryrun and run_immediately:
+        raise InvalidArgumentValueError(error_msg="The --dryrun and --run-immediately options cannot be used together. Use one or the other.")
+
+
+def validate_task_type(task_type):
+    if (task_type not in [item.value for item in CSSCTaskTypes]):
+        raise InvalidArgumentValueError(error_msg=ERROR_MESSAGE_INVALID_TASK)
```
huanwu, line 151: Suggestion: Improve the error message so the customer knows the current type and the expected valid types.

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,568 @@
...
+
+    resource_group = parse_resource_id(registry.id)[RESOURCE_GROUP]
+    schedule_cron_expression = None
+    cssc_tasks_exists, task_list = check_continuous_task_exists(cmd, registry)
+
+    if schedule is not None:
+        schedule_cron_expression = convert_timespan_to_cron(schedule)
+        logger.debug(f"converted schedule to cron expression: {schedule_cron_expression}")
+
+    cssc_tasks_exists, task_list = check_continuous_task_exists(cmd, registry)
```
huanwu, line 70: This is duplicate with line 64

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,568 @@
...
+        # Seen Managed Identity taking time, see if there can be an alternative (one alternative is to schedule the cron expression with delay)
+        # NEED TO SKIP THE TIME.SLEEP IN UNIT TEST CASE OR FIND AN ALTERNATIVE SOLUITION TO MI COMPLETE
+        time.sleep(30)
+        _trigger_task_run(cmd, registry, resource_group, CONTINUOUSPATCH_TASK_SCANREGISTRY_NAME)
+
+
+def delete_continuous_patch_v1(cmd, registry, dryrun):
+    logger.debug("Entering delete_continuous_patch_v1")
+    cssc_tasks_exists, _ = check_continuous_task_exists(cmd, registry)
+    cssc_config_exists = check_continuous_task_config_exists(cmd, registry)
```
huanwu, line 151: Question, if the customer accidentally deletes the task config, how can they recover it?

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,568 @@
...
+    if not dryrun and (cssc_tasks_exists or cssc_config_exists):
```
huanwu, line 152: Suggestion: Check tasks and config separately to avoid redundant deletions, such as when the config doesn't exist but tasks do
e.g. 
if not dryrun and cssc_tasks_exists
    delete tasks ...
if not dryrun and cssc_config_exists
        delete_oci_artifact_continuous_patch(cmd, registry, dryrun)

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,568 @@
...
+def list_continuous_patch_v1(cmd, registry):
+    logger.debug("Entering list_continuous_patch_v1")
+    cssc_tasks_exists, _ = check_continuous_task_exists(cmd, registry)
+    if not cssc_tasks_exists:
+        logger.warning(f"{ERROR_MESSAGE_WORKFLOW_TASKS_DOES_NOT_EXIST}")
+        return
+
+    acr_task_client = cf_acr_tasks(cmd.cli_ctx)
+    resource_group_name = parse_resource_id(registry.id)[RESOURCE_GROUP]
+    tasks_list = acr_task_client.list(resource_group_name, registry.name)
```
huanwu, line 174: Do you fetch all tasks from the server and then filter out the CSSC tasks? This could lead to performance issues if there are many tasks in the registry. Could you fetch only the CSSC-specific tasks to improve performance?

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,568 @@
...
+        if not cssc_tasks_exists:
+            raise InvalidArgumentValueError("--config parameter is needed to perform dry-run check.")
+
+        # attempt to get the config file from the registry, since the configuration should exist
+        logger.debug("Retrieving the configuration file from the registry.")
+        _, config_file_path = get_oci_artifact_continuous_patch(cmd, registry)
+        if config_file_path is None:
+            raise AzCLIError("Failed to retrieve the configuration file from the registry.")
+
+    if is_create and cssc_tasks_exists:
```
huanwu, line 193: Suggestion: Move this before the config_file_path check to avoid redundant operations when config_file_path is None, but both is_create and cssc_tasks_exists are True as the get_oci_artifact_continuous_patch is expensive.

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,568 @@
...
+        raise AzCLIError(f"{ERROR_MESSAGE_WORKFLOW_TASKS_ALREADY_EXISTS}")
+
+    file_name = None
+    tmp_folder = None
+    acr_archive_utils_logger_level = acr_archive_utils_logger.getEffectiveLevel()
+    try:
+        file_name = os.path.basename(config_file_path)
+        tmp_folder = os.path.join(os.getcwd(), tempfile.mkdtemp(prefix="cli_temp_cssc"))
```
huanwu, line 201: Suggestion: consider using with tempfile.TemporaryDirectory(prefix="cli_temp_cssc") as tmp_folder to generate the temp folder

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,568 @@
...
+
+    except Exception as exception:
+        raise AzCLIError(f"Failed to delete task {task_name} from registry {registry.name} : {exception}")
+
+    logger.debug(f"Task {task_name} deleted successfully")
+
+
+def _delete_task_role_assignment(cli_ctx, acrtask_client, registry, resource_group, task_name, dryrun):
+    role_client = cf_authorization(cli_ctx)
+    acrtask_client = cf_acr_tasks(cli_ctx)
```
huanwu, line 439: Is this necessary as acrtask_client is already passed as a parameter?

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,568 @@
...
+    if identity:
+        assigned_roles = role_client.role_assignments.list_for_scope(
+            registry.id,
+            filter=f"principalId eq '{identity.principal_id}'"
+        )
+
+        for role in assigned_roles:
+            if dryrun:
+                logger.debug(f"Dry run, skipping deletion of role assignments, task: {task_name}, role name: {role.name}")
+                return None
```
huanwu, line 459: Do you want to return early "return None" on the first role found during a dryrun, meaning it won't log or process additional role assignments.

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,568 @@
...
+            logger.debug(f"Deleting role assignments of task {task_name} from the registry")
+            role_client.role_assignments.delete(
```
huanwu, line 461: Suggestion: add a try-except block around role deletion to log failures but continue processing other roles.

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,568 @@
...
+        source_registry_credentials = acr_tasks_client.models.SourceRegistryCredentials(
+            login_mode=auth_mode)
+
+    custom_registries = None
+    if login_server:
+        # if null username and password (or identity), then remove the credential
+        custom_reg_credential = None
+
+        is_identity_credential = False
+        if not username and not password:
```
huanwu, line 521: Question: what happen if username is none, but password is not none?

src/acrcssc/azext_acrcssc/helper/_deployment.py
```diff
@@ -0,0 +1,150 @@
...
+        parameters=deployment)
+    logger.debug(poller)
+
+    # Wait for the deployment to complete and get the outputs
+    deployment: DeploymentExtended = LongRunningOperation(
+        cmd_ctx
+    )(poller)
+    logger.debug("Finished deploying")
+
+    if deployment.properties is not None:
```
huanwu, line 134: Suggestion: simply the condition check by removing the if-else block and directly throw the error if deployment.properties is None.
e.g.
if deployment.properties is None:
    raise RuntimeError("The deployment has no properties.\nAborting")

depl_props = deployment.properties

src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py
```diff
@@ -0,0 +1,233 @@
...
+
+        # the ORAS client can only work with files under the current directory
+        temp_artifact = tempfile.NamedTemporaryFile(
+            prefix="cssc_config_tmp_",
+            mode="w+b",
+            dir=os.getcwd(),
+            delete=False)
+
+        temp_artifact_name = temp_artifact.name
+        user_artifact = open(cssc_config_file, "rb")
```
huanwu, line 56: use with statement to ensure the file is closed properly.
with open(cssc_config_file, "rb") as user_artifact:
                shutil.copyfileobj(user_artifact, temp_artifact)

src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py
```diff
@@ -0,0 +1,233 @@
...
+
+    oras_client = None
+    temp_artifact_name = None
+    try:
+        oras_client = _oras_client(registry)
+        # we might have to handle the tag lock/unlock for the cssc config file,
+        # to make it harder for the user to change it by mistake
+
+        # the ORAS client can only work with files under the current directory
+        temp_artifact = tempfile.NamedTemporaryFile(
```
huanwu, line 49: Suggestion: use with statement

src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py
```diff
@@ -0,0 +1,233 @@
...
+            prefix="cssc_config_tmp_",
+            mode="w+b",
+            dir=os.getcwd(),
+            delete=False)
+
+        temp_artifact_name = temp_artifact.name
+        user_artifact = open(cssc_config_file, "rb")
+        shutil.copyfileobj(user_artifact, temp_artifact)
+        temp_artifact.close()
```
huanwu, line 58: Suggestion, call temp_artifact.flush() to ensure all data is written before use

src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py
```diff
@@ -0,0 +1,233 @@
...
+
+        if dryrun:
```
huanwu, line 60: Question: we push the config file even in dry run?

src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py
```diff
@@ -0,0 +1,233 @@
...
+            oci_target_name = f"{CSSC_WORKFLOW_POLICY_REPOSITORY}/{CONTINUOUSPATCH_OCI_ARTIFACT_CONFIG}:{CONTINUOUSPATCH_OCI_ARTIFACT_CONFIG_TAG_V1}"
+
+        oras_client.push(
+            target=oci_target_name,
+            files=[temp_artifact.name])
+    except Exception as exception:
+        raise AzCLIError(f"Failed to push OCI artifact to ACR: {exception}")
+    finally:
+        if oras_client:
+            oras_client.logout(hostname=str.lower(registry.login_server))
```
huanwu, line 72: Suggestion: add try except

src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py
```diff
@@ -0,0 +1,233 @@
...
+def get_oci_artifact_continuous_patch(cmd, registry):
+    logger.debug("Entering get_oci_artifact_continuous_patch with parameter: %s", registry.login_server)
+    config = None
+    file_name = None
+    try:
+        oras_client = _oras_client(registry)
+
+        oci_target_name = f"{CSSC_WORKFLOW_POLICY_REPOSITORY}/{CONTINUOUSPATCH_OCI_ARTIFACT_CONFIG}:{CONTINUOUSPATCH_OCI_ARTIFACT_CONFIG_TAG_V1}"
+
+        oci_artifacts = oras_client.pull(
```
huanwu, line 86: question: is it possible oci_artifacts is none?

src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py
```diff
@@ -0,0 +1,233 @@
...
+            target=oci_target_name,
+            stream=True)
+        trigger_task = get_task(cmd, registry, CONTINUOUSPATCH_TASK_SCANREGISTRY_NAME)
```
huanwu, line 89: Suggestion: add logs for the oci_artifacts, trigger_task to assist with troubleshooting if an exception is thrown in the try block.

src/acrcssc/azext_acrcssc/helper/_utility.py
```diff
@@ -0,0 +1,97 @@
...
+
+    if date_time is None:
+        date_time = datetime.now(timezone.utc)
+
+    cron_hour = date_time.hour
+    cron_minute = date_time.minute
+    cron_expression = ""
+
+    if unit == 'd':  # day of the month
+        if value < 1 or value > 30:
```
huanwu, line 37: Define a constant, as I noticed that 1 and 30 are used in multiple places.

src/acrcssc/azext_acrcssc/helper/_utility.py
```diff
@@ -0,0 +1,97 @@
...
+                return match.group(1)
+            return match.group(1) + 'd'
+
+        return None
+    except IndexError:
+        return None
+
+
+def create_temporary_dry_run_file(file_location, tmp_folder):
+    templates_path = os.path.dirname(
```
huanwu, line 63: Suggestion, avoid using os.path.dirname twice e.g.
base_path = os.path.dirname(os.path.abspath(__file__))
templates_path = os.path.join(base_path, "..", "templates")

src/acrcssc/azext_acrcssc/helper/_utility.py
```diff
@@ -0,0 +1,97 @@
...
+        os.path.join(
+            os.path.dirname(
+                os.path.abspath(__file__)),
+            "../templates/"))
+    logger.debug(f"templates_path:  {templates_path}")
+
+    os.makedirs(tmp_folder, exist_ok=True)
+    file_template_copy = templates_path + "/" + TMP_DRY_RUN_FILE_NAME
+
+    shutil.copy2(file_template_copy, tmp_folder)
```
huanwu, line 73: add error handling for the copy method

src/acrcssc/azext_acrcssc/helper/_utility.py
```diff
@@ -0,0 +1,97 @@
...
+def create_temporary_dry_run_file(file_location, tmp_folder):
+    templates_path = os.path.dirname(
+        os.path.join(
+            os.path.dirname(
+                os.path.abspath(__file__)),
+            "../templates/"))
+    logger.debug(f"templates_path:  {templates_path}")
+
+    os.makedirs(tmp_folder, exist_ok=True)
+    file_template_copy = templates_path + "/" + TMP_DRY_RUN_FILE_NAME
```
huanwu, line 71: use join e.g. file_template_copy = os.path.join(templates_path, TMP_DRY_RUN_FILE_NAME)

src/acrcssc/azext_acrcssc/helper/_utility.py
```diff
@@ -0,0 +1,97 @@
...
+
```
huanwu, line 72: Add a check to ensure the file_template_copy exist before attempting to copy

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,557 @@
...
+                output += "\n" + line
+
+        return output
+
+    @staticmethod
+    def _get_missing_taskrun(taskrun_client, registry, run_id):
+        try:
+            resourceid = parse_resource_id(registry.id)
+            resource_group = resourceid[RESOURCE_GROUP]
+            runs = WorkflowTaskStatus.get_taskruns_with_filter(taskrun_client,
```
huanwu, line 475: Is it possible runs is none or empty?

src/acrcssc/azext_acrcssc/_validators.py
cegraybl: not necessary as the `with open()` will handle the file life cycle, will remove

src/acrcssc/azext_acrcssc/_validators.py
cegraybl: that is handled on the schema validation, it allows us to check that the arrays contain at least one element
https://github.com/AzureCR/azure-cli-extensions/blob/fc0beb69cbe54574219e1ac38f494257d0c04f21/src/acrcssc/azext_acrcssc/helper/_constants.py#L133

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
cegraybl: they would have to push the file as an OCI artifact (ORAS) or recreate the workflow.
We could create a lock on the config itself, but if that is the case, please open an item for the feature change

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
cegraybl: `dryrun` has been removed from the delete path, current code is 
```
if cssc_tasks_exists:
    ...
if cssc_config_exists:
   ....

File: src/acrcssc/azext_acrcssc/helper/_taskoperations.py (hunk 13 above)
Comment (cegraybl) [reply src/acrcssc/azext_acrcssc/helper/_taskoperations.py:458]: no, removed

File: src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py (hunk 21 above)
Comment (cegraybl) [reply src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py:60]: on dryrun we push  the artifact under a different tag (*:dryrun instead of *:v1) so the task can read the config file from it

File: src/acrcssc/azext_acrcssc/helper/_workflow_status.py (hunk 30 above)
Comment (cegraybl) [reply src/acrcssc/azext_acrcssc/helper/_workflow_status.py:514]: yes, it was detected during testing.
Since the task runs are retrieved first and the log later, there is a window where a task run id is referenced in the logs but does not exist in the list of task runs we originally retrieved.
This function is to get the missing task runs

File: src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py (hunk 23 above)
Comment (cegraybl) [reply src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py]: does not seem like it, since the pull will attempt to get the manifest before the artifact, if it doesn't exist it will throw
https://github.com/oras-project/oras-py/blob/71599b903a05ac09e921af1762dfea9fa6fa8c7c/oras/provider.py#L932C9-L933C43

File: src/acrcssc/azext_acrcssc/helper/_taskoperations.py (hunk 16 above)
Comment (cegraybl) [reply src/acrcssc/azext_acrcssc/helper/_taskoperations.py:537]: I think this is picked up from the ACR module in the base CLI, and not all parameters are being used
https://github.com/Azure/azure-cli/blob/6f3fa67f6544566b564bef5d61fcc0a7ddc24550/src/azure-cli/azure/cli/command_modules/acr/_utils.py#L281

will remove parameters and code not being used to generate `CustomRegistryCredentials`

File: src/acrcssc/azext_acrcssc/helper/_taskoperations.py (hunk 10 above)
Comment (cegraybl) [reply src/acrcssc/azext_acrcssc/helper/_taskoperations.py:189]: changed call from 'list()' to 'get()' using the CSSC task names

**Comment by cegraybl** [comment]
@AllyW, @zhoxing-ms we have published this PR. the versioning is fixed under [this PR](https://github.com/AzureCR/azure-cli-extensions/pull/43) that addresses @huanwu comments.
The rest of the changes are ready for review: tests, linter and style are successful.
Please comment if there are any concerns.

**Comment by cegraybl** [comment]
@yonzhan @zhoxing-ms a question on release. When is the cutoff for an extension to be released? are the extensions release monthly, or do they go live on a different schedule?

File: src/acrcssc/azext_acrcssc/helper/_taskoperations.py [hunk 31]
```diff
@@ -0,0 +1,545 @@
...
+            raise AzCLIError(f"{ERROR_MESSAGE_WORKFLOW_TASKS_ALREADY_EXISTS}")
+        _create_cssc_workflow(cmd, registry, schedule_cron_expression, resource_group, dryrun)
+    else:
+        if not cssc_tasks_exists:
+            raise AzCLIError(f"{ERROR_MESSAGE_WORKFLOW_TASKS_DOES_NOT_EXIST}")
+
+        _update_cssc_workflow(cmd, registry, schedule_cron_expression, resource_group, dryrun, task_list)
+
+    if cssc_config_file is not None:
+        create_oci_artifact_continuous_patch(registry, cssc_config_file, dryrun)
```
huanwu, line 81: Per Sajay’s comments, the OCI artifact should be pushed before the task is created. If this step fails, the process should simply exit.

zhoxing-ms:
> When is the cutoff for an extension to be released? are the extensions release monthly, or do they go live on a different schedule?

The release time of extensions is flexible and can be released according to requirements

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
cegraybl: fixed on https://github.com/AzureCR/azure-cli-extensions/pull/45

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+        if scan_error_reason != "":
+            result["scan_error_reason"] = scan_error_reason
+
+        if patch_error_reason != "":
+            result["patch_error_reason"] = patch_error_reason
+
+        result["patch_date"] = patch_date
+        result["patch_task_ID"] = patch_task_id
+        result["last_patched_image"] = patched_image
+        result["workflow_type"] = workflow_type
```
stevenhvtran, line 396: nit: these can be moved to where result is instantiated

src/acrcssc/azext_acrcssc/templates/acrcli/artifact.json
stevenhvtran, line 1: how is this file used?

src/acrcssc/azext_acrcssc/_validators.py
```diff
@@ -0,0 +1,183 @@
...
+    if schedule is None:
+        return
+    # Extract the numeric value and unit from the timespan expression
+    match = re.match(r'(\d+)(d)$', schedule)
+    if not match:
+        raise InvalidArgumentValueError(error_msg=ERROR_MESSAGE_INVALID_TIMESPAN_FORMAT, recommendation=RECOMMENDATION_SCHEDULE)
+
+    value = int(match.group(1))
+    unit = match.group(2)
+    if unit == 'd' and (value < CONTINUOUSPATCH_SCHEDULE_MIN_DAYS or value > CONTINUOUSPATCH_SCHEDULE_MAX_DAYS):  # day of the month
```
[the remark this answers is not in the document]
cegraybl: already discussed from task scheduling, this is a limitation on cron expressions

src/acrcssc/azext_acrcssc/cssc.py
```diff
@@ -0,0 +1,152 @@
...
+    dryrun_output = acr_cssc_dry_run(cmd,
+                                     registry=registry,
+                                     config_file_path=config,
+                                     is_create=is_create,
+                                     remove_internal_statements=not dryrun)
+    if dryrun:
+        print(dryrun_output)
+    else:
+        validate_continuous_patch_v1_image_limit(dryrun_output)
+        create_update_continuous_patch_v1(cmd, registry, config, schedule, dryrun, run_immediately, is_create)
```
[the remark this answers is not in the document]
cegraybl: change on the feature, dryrun eventually meant only running the quick task to test the configuration and filter the images. Since the change affects the signature of multiple functions, not sure if it's worth it to fully remove now, specially so close to public preview.
@stevenhvtran could you open an item on this to remove for next version?

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,550 @@
...
+            registry.id,
+            filter=f"principalId eq '{identity.principal_id}'"
+        )
+
+        for role in assigned_roles:
+            try:
+                logger.debug(f"Deleting role assignments of task {task_name} from the registry")
+                role_client.role_assignments.delete(
+                    scope=registry.id,
+                    role_assignment_name=role.name
```
huanwu, line 490: Question, is it possible this role is used by other non cssc tasks?

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+                logger.debug("Scan Taskrun: %s has no logs, silent failure", scan.run_id)
+                continue
+
+            image = WorkflowTaskStatus._get_image_from_tasklog(scan.task_log_result)
+
+            if not image:
+                continue
+
+            # need to check if we have the latest scan task
+            task = scan
```
huanwu, line 289: Are these two lines required? They are covered in line 293

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+                                                             scan, scan.task_log_result)
+            else:
+                all_status[image] = WorkflowTaskStatus(image)
+            all_status[image].scan_task = task
+            all_status[image].scan_logs = logs
+            patch_task_id = all_status[image].get_patch_task_from_scan_tasklog()
+            # missing the patch task id means that the scan either failed, or succeeded and patching is not needed.
+            # this is important, because patching status depends on both the patching task status (if it exists)
+            # and the scan task status
+            if patch_task_id is not None:
```
huanwu, line 304: nit, if patch_task_id

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+                # it is possible for the patch task to be mentioned in the logs, but the API has not returned the
+                # taskrun for it yet, attempt to retrieve it from client
+                try:
+                    patch_task = next((task for task in patch_taskruns if task.run_id == patch_task_id))
```
huanwu, line 308: Considering using a default return to avoid the StopIteration e.g.
patch_task = next((task for task in patch_taskruns if task.run_id == patch_task_id), None)
if not patch_task:
    patch_task = WorkflowTaskStatus._get_missing_taskrun(...)

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+                except StopIteration:
+                    patch_task = WorkflowTaskStatus._get_missing_taskrun(taskrun_client, registry, patch_task_id)
+
+                all_status[image].patch_task = patch_task
+                if WorkflowTaskStatus._task_status_to_workflow_status(patch_task) == WorkflowTaskState.FAILED.value:
+                    failed_patch_tasklog_retrieval.append(all_status[image])
+
+        if len(failed_patch_tasklog_retrieval) > 0:
```
huanwu, line 316: nit, if failed_patch_tasklog_retrieval:

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+                registry_name=registry_name,
+                run_id=run_id)
+            log_file_sas = response.log_link
+        except (AttributeError, CloudError) as e:
+            logger.debug("%s Exception: %s", error_msg, e)
+            raise AzCLIError(error_msg)
+        except ResourceNotFoundError as e:
+            logger.debug("log file not found for run_id: %s, registry: %s, "
+                         "resource_group: %s -- exception: %s",
+                         run_id, registry_name, resource_group_name, e)
```
huanwu, line 448: Do you need return empty string for ResouceNotFoundError?

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+                    progress_bar=IndeterminateProgressBar(cmd.cli_ctx, message=await_task_message),
+                    poller_done_interval_ms=1500  # every poller call will do a call to the API
+                )(polling_method)
+                logger.debug("Task result: %s", result)
+            except TimeoutError:
+                logger.debug("Timeout waiting for task run to complete, workflow task run ID: %s", run_id)
+                logger.debug("An attempt to retrieve the logs will be done, if there are any")
+
+        blobClient = get_sdk(cmd.cli_ctx, ResourceType.DATA_STORAGE_BLOB, '_blob_client#BlobClient')
+        return WorkflowTaskStatus._download_logs(blobClient.from_blob_url(log_file_sas))
```
huanwu, line 467: Do you need add try block here to catch the exception thrown during download logs?

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+    def _download_logs(blob_service):
+        blob = blob_service.download_blob()
+        blob_text = blob.readall().decode('utf-8')
+
+        return blob_text
+
+    @staticmethod
+    def remove_internal_acr_statements(blob_content):
+        logger.debug("Removing internal ACR statements from logs, blob content size: %s", len(blob_content))
+        lines = blob_content.split("\n")
```
huanwu, line 491: nit, lines = blob_content.splitlines()

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+        starting_identifier = "DRY RUN mode enabled"
+        terminating_identifier = "Total matches found"
+        print_line = False
+        output = ""
+
+        for line in lines:
+            if line.startswith(starting_identifier):
```
huanwu, line 498: line.strip().startswith(starting_identifier)

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+                print_line = True
+            elif line.startswith(terminating_identifier):
```
huanwu, line 500: line.strip().startswith(terminating_identifier)

src/acrcssc/azext_acrcssc/templates/task/cssc_scan_image.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+  - cmd: cssc jq 'if .Results == null or (.Results | length) == 0 then 0 else [.Results[] | select(.Vulnerabilities != null) | .Vulnerabilities | length] | add end' /workspace/data/vulnerability-report_trivy_$DATE.json > /workspace/data/vulCount.txt
+  - cmd: cssc jq 'if .Metadata.OS.EOSL == null then false else .Metadata.OS.EOSL end' /workspace/data/vulnerability-report_trivy_$DATE.json > /workspace/data/eosl.txt
+  - cmd: |        
+        bash -c 'echo "Vulnerabilities found for image {{.Values.SOURCE_REPOSITORY}}:{{.Values.SOURCE_IMAGE_TAG}} -> $(cat /workspace/data/vulCount.txt)"'
+        bash -c 'echo "EOSL for the image {{.Values.SOURCE_REPOSITORY}}:{{.Values.SOURCE_IMAGE_TAG}} -> $(cat /workspace/data/eosl.txt)"'
+
+  - id: trigger-patch-task
+    retries: 3
+    retryDelay: 10
```
huanwu, line 38: Why choose 10? This value is different from the scan task.

src/acrcssc/azext_acrcssc/cssc.py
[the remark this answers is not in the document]
stevenhvtran: Okay, happy for this to be done after the initial release. I've created an item to track now

src/acrcssc/azext_acrcssc/azext_metadata.json
```diff
@@ -0,0 +1,4 @@
+{
+    "azext.isPreview": true,
+    "azext.minCliCoreVersion": "2.15.0"
```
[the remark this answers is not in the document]
cegraybl: I think this is the default when running the azdev tools to create the extension, will set to [v2.60](https://github.com/Azure/azure-cli/releases/tag/azure-cli-2.60.0) which is the version when the extension development started. We haven't noted any breaking changes since development started with the newer cli versions.

src/acrcssc/azext_acrcssc/_validators.py
```diff
@@ -0,0 +1,183 @@
...
+        acrtask_client = cf_acr_tasks(cmd.cli_ctx)
+        for task_name in CONTINUOUSPATCH_ALL_TASK_NAMES:
+            task = get_task(cmd, registry, task_name, acrtask_client)
+            if task is None:
+                missing_tasks.append(task_name)
+            else:
+                task_list.append(task)
+
+        if len(missing_tasks) > 0:
+            logger.debug(f"Failed to find tasks {', '.join(missing_tasks)} from registry {registry.name}")
```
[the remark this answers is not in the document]
cegraybl: Since this function is used to validate also that tasks do not exist (in the case of create, for example), this function logs the error but does not report on it, the caller handles depending on the situation, i.e. [during creation](https://github.com/AzureCR/azure-cli-extensions/blob/91df80422b0b6aeb3a9e8c92978f4d0258cf3e35/src/acrcssc/azext_acrcssc/helper/_taskoperations.py#L71C9-L72C79) or [during delete](https://github.com/AzureCR/azure-cli-extensions/blob/91df80422b0b6aeb3a9e8c92978f4d0258cf3e35/src/acrcssc/azext_acrcssc/helper/_taskoperations.py#L177C9-L177C73)

src/acrcssc/azext_acrcssc/helper/_deployment.py
```diff
@@ -0,0 +1,150 @@
...
+
+def validate_template(cmd_ctx, resource_group, deployment_name, template):
+    # Validation is automatically re-attempted in live runs, but not in test
+    # playback, causing them to fail. This explicitly re-attempts validation to
+    # ensure the tests pass
+    api_clients = cf_resources(cmd_ctx)
+    validation_res = None
+    deployment = Deployment(
+        properties=template,
+        # tags = { "test": CSSC_TAGS }, #we need to know if tagging
```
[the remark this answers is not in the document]
cegraybl: removed

src/acrcssc/azext_acrcssc/templates/task/cssc_scan_image.yaml
cegraybl: @Ruchii-27 can you provide context for the retryDelay?

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
cegraybl: added to `_download_logs()` so it returns an empty string in case of exception and log the issue

src/acrcssc/azext_acrcssc/templates/task/cssc_scan_image.yaml
Ruchii-27: To wait a little longer before retrying as intermittent auth issues were observed from ARM during az acr task run

src/acrcssc/azext_acrcssc/helper/_deployment.py
```diff
@@ -0,0 +1,150 @@
...
+        mode=DeploymentMode.incremental)
+    try:
+        validate_template(cmd_ctx, resource_group, deployment_name, template)
+        if (dryrun):
+            logger.debug("Dry run, skipping deployment")
+            return None
+
+        return deploy_template(cmd_ctx, resource_group, deployment_name, template)
+    except Exception as exception:
+        logger.debug(f'Failed to validate and deploy template: {exception}')
```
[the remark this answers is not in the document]
cegraybl: I think it's better to remove the logger.debug() since the error is also raised as an exception, that will be an error visible to the customer and duplicated if we keep the debug message

src/acrcssc/azext_acrcssc/cssc.py
```diff
@@ -0,0 +1,152 @@
...
+    # every time we perform a create or update operation, we need to validate for the number of images selected on the
+    # configuration file. The way to do this is by silently running the dryrun operation. If the limit is exceeded, we
+    # will not proceed with the operation.
+    dryrun_output = acr_cssc_dry_run(cmd,
+                                     registry=registry,
+                                     config_file_path=config,
+                                     is_create=is_create,
+                                     remove_internal_statements=not dryrun)
+    if dryrun:
+        print(dryrun_output)
```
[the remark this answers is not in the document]
cegraybl: The task gives the error message in dryrun and the CLI displays it
https://github.com/AzureCR/azure-cli-extensions/blob/91df80422b0b6aeb3a9e8c92978f4d0258cf3e35/src/acrcssc/azext_acrcssc/templates/tmp_dry_run_template.yaml#L15

src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py
```diff
@@ -0,0 +1,226 @@
...
+            validate(json_config, CONTINUOUSPATCH_CONFIG_SCHEMA_V1)
+        except ValidationError as e:
+            logger.error("Error validating the continuous patch config file: %s", e)
+            return None
+
+        self.version = json_config.get("version", "")
+        repositories = json_config.get("repositories", [])
+        for repo in repositories:
+            enabled = repo.get("enabled", True)  # optional field, default to True
+            repository = Repository(repo["repository"], repo["tags"], enabled)
```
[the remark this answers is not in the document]
cegraybl: yes, it's part of the schema validation
https://github.com/AzureCR/azure-cli-extensions/blob/91df80422b0b6aeb3a9e8c92978f4d0258cf3e35/src/acrcssc/azext_acrcssc/helper/_constants.py#L133C17-L133C51

src/acrcssc/azext_acrcssc/helper/_taskoperations.py
```diff
@@ -0,0 +1,550 @@
...
+
+    if schedule is not None:
+        schedule_cron_expression = convert_timespan_to_cron(schedule)
+        logger.debug(f"converted schedule to cron expression: {schedule_cron_expression}")
+
+    if is_create_workflow:
+        if cssc_tasks_exists:
+            raise AzCLIError(f"{ERROR_MESSAGE_WORKFLOW_TASKS_ALREADY_EXISTS}")
+
+        if cssc_config_file is not None:
```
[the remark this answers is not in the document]
cegraybl: during create config cannot be null, it's a combination of the parameter restriction for the command and validation catching if the config is none/empty/does not parse/etc
https://github.com/AzureCR/azure-cli-extensions/blob/91df80422b0b6aeb3a9e8c92978f4d0258cf3e35/src/acrcssc/azext_acrcssc/_validators.py#L140
Will remove the conditional from create

during update, since the parameter is optional, the condition below is used to determine if the configuration should be updated or not

src/acrcssc/azext_acrcssc/helper/_deployment.py
```diff
@@ -0,0 +1,150 @@
...
+        # tags = { "test": CSSC_TAGS },
+        # we need to know if tagging is something that will help us,
+        # tasks are proxy resources, so not sure how that would work
+    )
+
+    poller = api_client.deployments.begin_create_or_update(
+        resource_group_name=resource_group,
+        deployment_name=deployment_name,
+        parameters=deployment)
+    logger.debug(poller)
```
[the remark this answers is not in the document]
cegraybl: removed, the execution and result of the deployment is logged under the `cli.azure.cli.core.sdk.policies` in the LongRunningOperation execution

src/acrcssc/azext_acrcssc/helper/_workflow_status.py
```diff
@@ -0,0 +1,600 @@
...
+                                                           taskrun.run_id,
+                                                           registry.name,
+                                                           resource_group,
+                                                           await_task_run=False)
+                if tasklog == "":
+                    logger.debug("Taskrun: %s has no logs, silent failure", taskrun.run_id)
+                taskrun.task_log_result = tasklog
+            except ResourceNotFoundError as e:
+                logger.debug("Failed to get logs for taskrun: %s with exception: %s", taskrun.run_id, e)
+
```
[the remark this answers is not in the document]
cegraybl: "process_taskrun()" handles the retrieval of a single task, will modify so the wrapping function "_retrieve_all_tasklogs()" will update the progress_indicator while the thread finish executing with an interval

src/acrcssc/azext_acrcssc/helper/_utility.py
```diff
@@ -0,0 +1,105 @@
...
+    ERROR_MESSAGE_INVALID_TIMESPAN_VALUE,
+    RESOURCE_GROUP,
+    TMP_DRY_RUN_FILE_NAME)
+from .._client_factory import cf_acr_tasks
+
+logger = get_logger(__name__)
+# pylint: disable=logging-fstring-interpolation
+
+
+def convert_timespan_to_cron(schedule, date_time=None):
```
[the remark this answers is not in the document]
cegraybl: will call "convert_timespan_to_cron()" from "_validate_schedule()"

src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py
```diff
@@ -0,0 +1,226 @@
...
+    def from_file(self, file_path, trigger_task=None):
+        with open(file_path, "r") as file:
+            return self.from_json(file.read(), trigger_task)
+
+    def from_json(self, json_str, trigger_task=None):
+        import json
+        from jsonschema import validate
+
+        try:
+            json_config = json.loads(json_str)
```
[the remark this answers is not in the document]
cegraybl: to avoid a circular import, will call "ContinuousPatchConfig.from_file()" from "validate_continuouspatch_json()"

src/acrcssc/azext_acrcssc/helper/_ociartifactoperations.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+        token = result.stdout.strip()
+        if not token or token == "":
+            logger.debug("Failed to retrieve ACR token: Token is empty.")
+            raise AzCLIError("Failed to retrieve ACR token. The token is empty.")
+        return token
+    except subprocess.CalledProcessError as error:
+        logger.debug(f"Error while retrieving ACR token: {error.stderr.strip()}")
+
+        unauthorized = (error.stderr
```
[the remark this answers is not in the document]
cegraybl: removed stderr check, since we call 'subprocess.run' with 'check=True' stderr should always be initialized, it could still be empty if the process only logs to stdout

huanwu (approved):

stevenhvtran (approved):

Ruchii-27 (approved):

src/acrcssc/azext_acrcssc/_params.py
```diff
@@ -0,0 +1,38 @@
...
+        c.argument("schedule", options_list=["--schedule"], help=f"schedule to run the scan and patching task. E.g. `<n>d` where n is the number of days between each run. Max value is {CONTINUOUSPATCH_SCHEDULE_MAX_DAYS}d.", required=False)
+        c.argument("run_immediately", options_list=["--run-immediately"], help="Set this flag to trigger the immediate run of the selected workflow task. Default value: false.", arg_type=get_three_state_flag(), required=False)
+        c.argument("dryrun", options_list=["--dry-run"], help="Use this flag to see the qualifying repositories and tags that would be affected by the workflow. Default value: false. 'config' parameter is mandatory to provide with dry-run", arg_type=get_three_state_flag(), required=False)
```
zhoxing-ms, line 32: Same as above
```suggestion
    with self.argument_context("acr supply-chain workflow update") as c:
        c.argument("config", help="Configuration file path containing the json schema for the list of repositories and tags to filter within the registry. Schema example:{\"repositories\":[{\"repository\":\"alpine\",\"tags\":[\"tag1\",\"tag2\"],\"enabled\":true},{\"repository\":\"python\",\"tags\":[\"*\"],\"enabled\":false}], \"version\": \"v1\", \"tag-convention\": \"floating\"}}. \"tag-convention\" is an optional property, values can be \"incremental\" (the default behavior, will increase the patch version of the tag, for example \"{repository}:{original-tag}-1\", \"{repository}:{original-tag}-2\", etc), or \"floating\" (will reuse the tag \"{repository}:{original-tag}-patched\" for patching)")
        c.argument("schedule", help=f"schedule to run the scan and patching task. E.g. `<n>d` where n is the number of days between each run. Max value is {CONTINUOUSPATCH_SCHEDULE_MAX_DAYS}d.")
        c.argument("run_immediately", help="Set this flag to trigger the immediate run of the selected workflow task. Default value: false.", arg_type=get_three_state_flag())
        c.argument("dryrun", help="Use this flag to see the qualifying repositories and tags that would be affected by the workflow. Default value: false. 'config' parameter is mandatory to provide with dry-run", arg_type=get_three_state_flag())
```

src/acrcssc/azext_acrcssc/_params.py
```diff
@@ -0,0 +1,38 @@
...
+
+    with self.argument_context("acr supply-chain workflow list") as c:
+        c.argument("status", arg_type=get_enum_type(WorkflowTaskState), options_list=["--run-status"], help="Status to filter the supply-chain workflow image status.", required=False)
+
+    with self.argument_context("acr supply-chain workflow delete") as c:
+        c.argument("yes", options_list=["--yes", "-y"], help="Proceed with the deletion without user confirmation", required=False)
```
zhoxing-ms, line 38:
```suggestion
    with self.argument_context("acr supply-chain workflow list") as c:
        c.argument("status", arg_type=get_enum_type(WorkflowTaskState), help="Status to filter the supply-chain workflow image status.")

    with self.argument_context("acr supply-chain workflow delete") as c:
        c.argument("yes", options_list=["--yes", "-y"], help="Proceed with the deletion without user confirmation")
```

src/acrcssc/azext_acrcssc/commands.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def load_command_table(self, _):
+    # required to mark the command as 'preview', can be expanded if additional commands are added
+    self.command_group("acr supply-chain", client_factory=cf_acr, is_preview=True)
+
+    with self.command_group("acr supply-chain workflow", client_factory=cf_acr, is_preview=True) as g:
+        g.custom_command("create", "create_acrcssc")
+        g.custom_command("update", "update_acrcssc")
+        g.custom_command("delete", "delete_acrcssc")
```
zhoxing-ms, line 17: Do we need to consider adding `confirmation=True` for the delete command?

src/acrcssc/azext_acrcssc/tests/latest/test_cssc_scenario.py
```diff
@@ -0,0 +1,39 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import os
+from azure.cli.testsdk import ScenarioTest, ResourceGroupPreparer, live_only
+
+
+class AcrcsscScenarioTest(ScenarioTest):
+    @live_only()
```
zhoxing-ms, line 11: May I ask why do you need to mark this test as ` @live_only()`?

src/acrcssc/azext_acrcssc/commands.py
cegraybl: we are doing a direct `user_confirmation()` on the command, did not find any documentation when adding it, but what is the preferred method?
https://github.com/AzureCR/azure-cli-extensions/blob/f48d94961ebad8403ebfe10687ee0d9187d6591e/src/acrcssc/azext_acrcssc/cssc.py#L114C5-L114C22

src/acrcssc/azext_acrcssc/_params.py
```diff
@@ -0,0 +1,38 @@
...
+        c.argument("schedule", options_list=["--schedule"], help=f"schedule to run the scan and patching task. E.g. `<n>d` where <n> is the number of days between each run. Max value is {CONTINUOUSPATCH_SCHEDULE_MAX_DAYS}d.", required=True)
+        c.argument("run_immediately", options_list=["--run-immediately"], help="Set this flag to trigger the immediate run of the selected workflow task. Default value: false.", arg_type=get_three_state_flag(), required=False)
+        c.argument("dryrun", options_list=["--dry-run"], help="Use this flag to see the qualifying repositories and tags that would be affected by the workflow. Default value: false. 'config' parameter is mandatory to provide with dry-run", arg_type=get_three_state_flag(), required=False)
```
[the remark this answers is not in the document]
cegraybl: github gives me errors when applying your commit, will PR and merge a fix with the parameter changes

src/acrcssc/azext_acrcssc/tests/latest/test_cssc_scenario.py
cegraybl: the end-2-end scenario needs to do authenticate and do a 'docker push' an artifact through a module (oras-py) to the registry being tested, test keeps failing when not in live mode. Could not get it to work without live_only(), but would appreciate if you have any recommendation or doc I can follow.
Added unit tests for all sections of the extension to properly test where I can mock the required calls (everything that is not extension code) if it makes any difference.

src/acrcssc/azext_acrcssc/_params.py
cegraybl: not sure if a bug on cli core, but removing the `required=False` on `--run-status` marks it as required. The other ones are fine
image

cegraybl: similar to `--config` on `update` command

cegraybl: also didn't remove the `options_list` on the parameters that change the signature (i.e dryrun -> dry-run)

zhoxing-ms (dismissed):

src/acrcssc/azext_acrcssc/tests/latest/test_cssc_scenario.py
zhoxing-ms: OK, got it, thanks~

src/acrcssc/azext_acrcssc/_params.py
```diff
@@ -0,0 +1,38 @@
...
+        c.argument("schedule", help=f"schedule to run the scan and patching task. E.g. `<n>d` where n is the number of days between each run. Max value is {CONTINUOUSPATCH_SCHEDULE_MAX_DAYS}d.")
+        c.argument("run_immediately", help="Set this flag to trigger the immediate run of the selected workflow task. Default value: false.", arg_type=get_three_state_flag())
+        c.argument("dryrun", options_list=["--dry-run"], help="Use this flag to see the qualifying repositories and tags that would be affected by the workflow. Default value: false. 'config' parameter is mandatory to provide with dry-run", arg_type=get_three_state_flag())
+
+    with self.argument_context("acr supply-chain workflow list") as c:
+        c.argument("status", arg_type=get_enum_type(WorkflowTaskState), options_list=["--run-status"], help="Status to filter the supply-chain workflow image status.", required=False)
```
zhoxing-ms, line 35:
```suggestion
        c.argument("status", arg_type=get_enum_type(WorkflowTaskState), options_list=["--run-status"], help="Status to filter the supply-chain workflow image status.")
```

src/acrcssc/azext_acrcssc/commands.py
zhoxing-ms: If you want to customize the confirmation information, then using `user_confirmation()` is fine

src/acrcssc/azext_acrcssc/_params.py
cegraybl: mentioned in a previous comment, only happening for the `--config` (under update) and the `--run-status` (under list) parameters, when the `required=False` is removed the parameter becomes required, changing the behavior of the command

image

zhoxing-ms (approved):

src/acrcssc/setup.py
```diff
@@ -0,0 +1,61 @@
...
+from codecs import open
+from setuptools import setup, find_packages
+try:
+    from azure_bdist_wheel import cmdclass
+except ImportError:
+    from distutils import log as logger
+    logger.warn("Wheel is not available, disabling bdist_wheel hook")
+
+VERSION = '1.0.0b1'
```
zhoxing-ms, line 16: Why the version in history note is `1.1.1` but the version here is `1.0.0b1`?

src/acrcssc/azext_acrcssc/_params.py
```diff
@@ -0,0 +1,38 @@
...
+from azure.cli.core.commands.parameters import (get_resource_name_completion_list, get_three_state_flag, get_enum_type)
+from .helper._constants import CONTINUOUSPATCH_SCHEDULE_MAX_DAYS
+
+
+def load_arguments(self: AzCommandsLoader, _):
+    from .helper._constants import CSSCTaskTypes
+    from .helper._workflow_status import WorkflowTaskState
+
+    with self.argument_context("acr supply-chain workflow") as c:
+        c.argument('resource_group', options_list=['--resource-group', '-g'], help='Name of resource group. You can configure the default group using `az configure --defaults group=<name>`', completer=get_resource_name_completion_list(REGISTRY_RESOURCE_TYPE))
```
zhoxing-ms, line 18:
```suggestion
        c.argument('resource_group', resource_group_name_type, completer=get_resource_name_completion_list(REGISTRY_RESOURCE_TYPE))
```

src/acrcssc/setup.py
cegraybl: we changed to `1.0.0b1` to follow CLI versioning, but history didn't get updated. Will keep version `1.0.0b1` for this one and modify the HISTORY file

zhoxing-ms (approved):

PR #8636, opened 2025-04-03
Update CLI to create role assignment

MuhammadAliFleet:
---

This PR updates the create_fleet process to create a role assignment for the Fleet RP on the subnet when creating a private fleet.

### Related command
az fleet create -n <FLEET> -l <LOCATION> --enable-hub --enable-private-cluster
az fleet create -n <FLEET> -l <LOCATION> --enable-hub --enable-private-cluster --enable-vnet-integration

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

src/fleet/azext_fleet/custom.py
```diff
@@ -109,6 +111,21 @@ def create_fleet(cmd,
...
 
+    if enable_private_cluster:
+        if not add_role_assignment(cmd, 'Network Contributor', FLEET_1P_APP_ID, scope=agent_subnet_id):
+            raise CLIError(
+                "failed to create role assignment for Fleet RP.\nDo you have owner permissions on the vnet?\n"
+                "Is the Microsoft.Containerservice namespace registered?\n"
+                "Please try registering Fleet namespace before retrying: az provider register --namespace 'Microsoft.Containerservice'"  # pylint: disable=line-too-long
+            )
+    if enable_vnet_integration:
+        if not add_role_assignment(cmd, 'Network Contributor', FLEET_1P_APP_ID, scope=apiserver_subnet_id):
```
MuhammadAliFleet, line 122: FLEET_1P_APP_ID automatically gets resolved to object ID when creating role assignment

https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acs/_graph.py#L30C5-L31C100

src/fleet/azext_fleet/custom.py
```diff
@@ -109,6 +111,21 @@ def create_fleet(cmd,
         identity=managed_service_identity
     )
 
+    if enable_private_cluster:
+        if not add_role_assignment(cmd, 'Network Contributor', FLEET_1P_APP_ID, scope=agent_subnet_id):
+            raise CLIError(
+                "failed to create role assignment for Fleet RP.\nDo you have owner permissions on the vnet?\n"
+                "Is the Microsoft.Containerservice namespace registered?\n"
+                "Please try registering Fleet namespace before retrying: az provider register --namespace 'Microsoft.Containerservice'"  # pylint: disable=line-too-long
```
jim-minter, line 119:
```suggestion
                "Is the Microsoft.Containerservice namespace registered?\n"
                "Please try registering Fleet namespace before retrying: az provider register --namespace 'Microsoft.Containerservice'"  # pylint: disable=line-too-long
```

1. Case on Microsoft.ContainerService (capital S)
2. Consistently call it the Microsoft.ContainerService namespace, not the Fleet namespace
3. This error message doesn't feel super friendly - if we think this is in any way a likely case, we should check and register ourselves if necessary.

jim-minter: 4. Please indicate the vnet in question to the user?

src/fleet/azext_fleet/custom.py
```diff
@@ -109,6 +111,21 @@ def create_fleet(cmd,
...
+            )
+    if enable_vnet_integration:
```
jim-minter, line 121: Please use a separate function if possible to avoid duplication.

src/fleet/azext_fleet/custom.py
```diff
@@ -109,6 +111,21 @@ def create_fleet(cmd,
         identity=managed_service_identity
     )
 
+    if enable_private_cluster:
+        if not add_role_assignment(cmd, 'Network Contributor', FLEET_1P_APP_ID, scope=agent_subnet_id):
```
jim-minter, line 115: Have you tested to see what happens if a role assignment already exists but with a different uuid?

src/fleet/azext_fleet/custom.py
jim-minter: For #3, I would be fine with removing the comments about namespace registration for now if you think that's appropriate.

src/fleet/setup.py
```diff
@@ -16,7 +16,7 @@
 
 # TODO: Confirm this is the right version number you want and it matches your
 # HISTORY.rst entry.
-VERSION = '1.4.0'
+VERSION = '1.4.1'
```
frantran, line 19: see https://github.com/Azure/azure-cli-extensions/pull/8636#discussion_r2029225465

src/fleet/azext_fleet/custom.py
MuhammadAliFleet: The add_role_assignment function returns `True` and does not create a new role assignment if a role assignment already exists but with a different uuid.

We have a new uuid everytime we call [add_role_assignment](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli/azure/cli/command_modules/acs/_roleassignments.py#L85C1-L86C1) function, and the check ensures if the resource already exists, we return `True`

src/fleet/azext_fleet/custom.py
MuhammadAliFleet: Azure automatically registers the namespace when creating a resource, but since we are adding a step before the creation of fleet resource, If an end user is looking to create fleet for the first time ever that happens to be private fleet, this would fail the role assignment step since the Microsoft.ContainerService is not registered and the SP is not provisioned in the customers tenant.

Should we register namespace ourselves?

MuhammadAliFleet: Separated provider registration handling from role assignment.

src/fleet/azext_fleet/_helpers.py
```diff
@@ -148,3 +152,15 @@ def _load_kubernetes_configuration(filename):
         raise
     except (yaml.parser.ParserError, UnicodeDecodeError) as ex:
         raise CLIError(f'Error parsing {filename} ({str(ex)})') from ex
+
+
+def validate_subnet(cmd, subnet_id):
```
jim-minter, line 157: consider putting this in _validators.py ?

src/fleet/azext_fleet/custom.py
```diff
@@ -109,6 +109,9 @@ def create_fleet(cmd,
         identity=managed_service_identity
     )
 
+    if enable_private_cluster or enable_vnet_integration:
```
jim-minter, line 112: does `enable_vnet_integration` imply `enable_private_cluster` already?

src/fleet/azext_fleet/_helpers.py
jim-minter: maybe add a comment about why we check the provider -- it's to catch the obvious other case that would cause the SP not to exist and `add_role_assignment` to hence fail

jim-minter (approved):
LGTM

kairu-ms (approved):

MuhammadAliFleet:
@microsoft-github-policy-service agree

PR #8686, opened 2025-04-24
fix(cssc): 31250186 remove the explicit 'required=False' under _params.py

cegraybl:
- for update command, for parameter --config
- for list command, for parameter --run-status


yonzhan:

PR #8713, opened 2025-05-02
networkfabric cli - changes from swagger update for run-ro command - fix for bug in command output

idanielsteven:
Due to a mismatch of the underlying API and the swagger specification - the "networkfabric device run-ro" (RunReadOnlyCommand) would not generate any output after successful completion. To fix this, the swagger was updated to match what the API was sending, this allowed the CLI to being outputting the expected results.

---

### Related command

- networkfabric device run-ro


yonzhan:

kairu-ms (approved):

PR #9139, opened 2025-09-08
[Azure Container Apps] Support session pool health probe

ShichaoQiu:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

bowen5 (approved):

src/containerapp/azext_containerapp/_params.py
```diff
@@ -469,6 +469,7 @@ def load_arguments(self, _):
         c.argument('startup_command', nargs='*', options_list=['--command'], help="A list of supported commands on the container that will executed during startup. Space-separated values e.g. \"/bin/queue\" \"mycommand\". Empty string to clear existing values")
         c.argument('args', nargs='*', help="A list of container startup command argument(s). Space-separated values e.g. \"-c\" \"mycommand\". Empty string to clear existing values")
         c.argument('target_port', type=int, validator=validate_target_port_range, help="The session port used for ingress traffic.")
+        c.argument('probe_yaml', type=file_type, help='Path to a .yaml file with the configuration of the container probes.', is_preview=True)
```
bowen5: Shall we mention here that probe is only supported in custom container.

Also, if customer provide a probe setting here for interpreter pool, will our CP return error or just ignore the config? From e2e perspective we'd better to fail somewhere (either CLI or CP, CP would be better) instead of silently ignored.

bowen5 (approved):

src/containerapp/azext_containerapp/_params.py
ShichaoQiu: for issue 1, Add `Only applicable for Custom Container Session Pool` in help info
for issue 2, 
- Add parameter validation in CLI side, for example:
image

- I also check if we pass the value into CP, CP will return 400, for example
image

src/containerapp/HISTORY.rst
```diff
@@ -6,7 +6,7 @@ upcoming
 ++++++
 * 'az containerapp update/up': Disallow changing `--revisions-mode` to Labels.
 * 'az containerapp session code-interpreter': Fix `--path` in examples
-* 'az containerapp sessionpool create/update': Support `--lifecycle-type` and `--max-alive-period`
+* 'az containerapp sessionpool create/update': Support `--lifecycle-type`, `--max-alive-period` and `--probe-yaml`
```
Greedygre: They are two separate PR, the History description should separate too.

ShichaoQiu: Thanks a lot, fixed

src/containerapp/azext_containerapp/containerapp_sessionpool_decorator.py
```diff
@@ -475,6 +511,10 @@ def validate_arguments(self):
                 current_lifecycle_type.lower() != LifecycleType.Timed.name.lower():
             raise ValidationError(f"--cooldown-period is not supported for the current --lifecycle-type '{current_lifecycle_type}'.")
 
+        # Validate container related arguments
+        if safe_get(self.existing_pool_def, "properties", "containerType").lower() != ContainerType.CustomContainer.name.lower():
+            self.check_container_related_arguments()
```
Greedygre, line 516: @ShichaoQiu 
What is the original behavior if the containerType != `CustomContainer` but input container_related_args?
With this change it will throw error `ValidationError(f"'{arg_name}' can not be set when container type is not 'CustomContainer'.")`.
Will it cause breaking change?

src/containerapp/azext_containerapp/tests/latest/test_containerapp_sessionpool.py
```diff
@@ -486,3 +486,146 @@ def test_containerapp_sessionpool_oncontainerexit(self, resource_group, subnet_i
...
+                JMESPathCheck('properties.customContainerTemplate.containers[0].probes[0].httpGet.httpHeaders[0].name', "Custom-Header"),
+                JMESPathCheck('properties.customContainerTemplate.containers[0].probes[0].httpGet.httpHeaders[0].value', "liveness probe"),
+                JMESPathCheck('properties.customContainerTemplate.containers[0].probes[0].initialDelaySeconds', 7),
+                JMESPathCheck('properties.customContainerTemplate.containers[0].probes[0].periodSeconds', 3),
+            ])
+
+        health_probe_yaml_text = """
+                                probes:
+                                """
+        write_test_file(health_probe_file_name, health_probe_yaml_text)
```
Greedygre, line 617: Please clean the test file at the end with `clean_up_test_file`

ShichaoQiu: fixed

src/containerapp/azext_containerapp/containerapp_sessionpool_decorator.py
ShichaoQiu: Good question. The original behavior is that the CLI command will break, for example,
image

New behavior is as below:
image

I think it is not a breaking change, it is a fix? Do you have any suggestion?

src/containerapp/azext_containerapp/custom.py
```diff
@@ -3116,7 +3116,8 @@ def create_session_pool(cmd,
                         registry_user=None,
                         mi_user_assigned=None,
                         registry_identity=None,
-                        mi_system_assigned=False):
+                        mi_system_assigned=False,
+                        probe_yaml=None):
```
Juliehzl, line 3120: Add example for update command

Juliehzl (approved):

src/containerapp/azext_containerapp/custom.py
ShichaoQiu: Good catch, have added the example for update command

PR #9140, opened 2025-09-08
[Azure Container Apps] Support stop session for custom container session pool

ShichaoQiu:
---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

src/containerapp/HISTORY.rst
```diff
@@ -8,6 +8,7 @@ upcoming
 * 'az containerapp session code-interpreter': Fix `--path` in examples
 * 'az containerapp sessionpool create/update': Support `--lifecycle-type` and `--max-alive-period`
 * 'az containerapp env premium-ingress': Deprecate `--min-replicas` and `--max-replicas` parameters, use workload profile scale instead.
+* 'az containerapp session custom-container': Support stop-session for custom container sessions
```
bowen5: whether modeling this as `az containerapp session stop-session` (apply to both Interpreter and CustomContainer) could be a feasible option? I'm thinking that later we might also have stop/delete session for code interpreter (https://learn.microsoft.com/en-us/rest/api/data-plane/containerapps/session-management/delete?view=rest-data-plane-containerapps-2025-02-02-preview&tabs=HTTP), and also list session related commands.

Or sth like `az containerapp session management stop-session`? How would you suggest

ShichaoQiu: I think `az containerapp session stop` will be better, have updated the logic

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -1513,6 +1513,23 @@ def extract_path_from_filename(path, filename):
             return path.rstrip("/") + "/" + path_in_filename.lstrip('/'), filename
 
 
+class SessionCustomContainerPreviewClient():
+    # pylint: disable=too-few-public-methods
+    api_version = PREVIEW_API_VERSION
```
yitaopan: Session api version may not exactly following control plane api version releases, do you think we need to add a dedicated property for this so that in the future when updating this PREVIEW_API_VERSION won't break this function?

But In this case I guess we need to update this api version regularly by ourselves

yitaopan (approved):

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -1513,6 +1513,23 @@ def extract_path_from_filename(path, filename):
...
+
+    @classmethod
+    def stop_session(cls, cmd, identifier, session_pool_endpoint):
+        url_fmt = "{}/.management/stopSession?identifier={}&api-version={}"
+        request_url = url_fmt.format(
+            session_pool_endpoint,
+            identifier,
+            cls.session_dp_api_version)
+
+        r = send_raw_request(cmd.cli_ctx, "POST", request_url, resource=SESSION_RESOURCE)
```
bowen5, line 1528: can we get the pool from cp first and check whether it is a custom container pool, and fail on client side when pool type is not custom container. Otherwise I suppose it would return 404 for python pool, which is not good experience.

src/containerapp/azext_containerapp/_clients.py
ShichaoQiu: good catch, have added a sperate api version for session data plane.

src/containerapp/azext_containerapp/_clients.py
ShichaoQiu: good catch, have added the validation logic in CLI side, the experience is as below:
image

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -1513,6 +1513,23 @@ def extract_path_from_filename(path, filename):
             return path.rstrip("/") + "/" + path_in_filename.lstrip('/'), filename
 
 
+class SessionCustomContainerPreviewClient():
+    # pylint: disable=too-few-public-methods
+    session_dp_api_version = "2025-02-02-preview"  # may be different from ACA CP
```
Juliehzl, line 1518: is it possible that the api version will be used by other clients?

src/containerapp/azext_containerapp/_clients.py
```diff
@@ -1513,6 +1513,23 @@ def extract_path_from_filename(path, filename):
...
+
+    @classmethod
+    def stop_session(cls, cmd, identifier, session_pool_endpoint):
```
Juliehzl, line 1521: is it a long running opreation?

src/containerapp/azext_containerapp/_clients.py
Greedygre, line 1518: Currently other client is using:
```
PREVIEW_API_VERSION = "2025-02-02-preview"
```

Do you want the api-version for SessionCustomContainerPreviewClient controlled by itself and should not be impacted by api-version upgrade?

ShichaoQiu: yeah, we want the api-version for SessionCustomContainerPreviewClient controlled by itself and should not be impacted by api-version upgrade

ShichaoQiu: no, it is only used by session custom container data plane.

src/containerapp/azext_containerapp/_clients.py
ShichaoQiu: No, it is not a long running operation.

Greedygre (approved):

PR #9163, opened 2025-09-13
[Containerapp] `az containerapp function`: Add function key management commands

khushishah513:
---

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/containerapp/HISTORY.rst
+++ b/src/containerapp/HISTORY.rst
@@ -4,6 +4,10 @@ Release History
 ===============
 upcoming
 ++++++
+* 'az containerapp function list-keys': List function keys for a specific function in a container app
+* 'az containerapp function update-keys': Update specific function key for a specific function in a container app
+* 'az containerapp function list-hostkeys': List host keys for a container app
+* 'az containerapp function update-hostkeys': Update specific host key for a container app
 * 'az containerapp update/up': Disallow changing `--revisions-mode` to Labels.
 * 'az containerapp session code-interpreter': Fix `--path` in examples
 * 'az containerapp sessionpool create/update': Support `--lifecycle-type` and `--max-alive-period`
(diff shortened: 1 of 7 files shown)


yonzhan:

src/containerapp/HISTORY.rst
```diff
@@ -4,6 +4,10 @@ Release History
 ===============
 upcoming
 ++++++
+* 'az containerapp function list-keys': List function keys for a specific function in a container app
+* 'az containerapp function update-keys': Update specific function key for a specific function in a container app
+* 'az containerapp function list-hostkeys': List host keys for a container app
+* 'az containerapp function update-hostkeys': Update specific host key for a container app
```
krishnajaju, line 10: please review these commands from Nitesh/Deep as well.

krishnajaju, line 10: we also have master key and system keys. Instead of adding separate commands for each type of keys, can we follow the same approach as that of [`az functionapp keys`](https://learn.microsoft.com/en-us/cli/azure/functionapp/keys?view=azure-cli-latest#az-functionapp-keys-list) and [`az functionapp function keys`](https://learn.microsoft.com/en-us/cli/azure/functionapp/function/keys?view=azure-cli-latest)

My suggestion is to have only one command that handles all kinds of function keys, following the az functionapp conventions.

>az containerapp function keys list/delete/set --key-type {functionKeys, hostKey, masterKey, systemKeys} [--function-name] {} [--key-name] {} [--key-value] {}

- if keyType is `functionKeys`, --function-name becomes mandatory parameter, else it is optional.
- keyType is mandatory parameter
- list --> to view all keys with given keyType
- delete --> delete a key with given keyType and keyName (keyName required only when keyType is functionKeys. Is it possible to delete master key, host key in case of other Functions SKUs using `az functionapp keys` command?
- set --> add new key or update existing key. (Only update operation supported in host and master key. Add or update supported in functionKeys)
- function-name, key-name and key-value parameters will be applicable only in certain combination of commands.

krishnajaju: Reference - https://learn.microsoft.com/en-us/azure/azure-functions/function-keys-how-to?tabs=azure-portal#understand-keys

PR #9265, opened 2025-10-02
Remove GA ingress, routing

Tratcher:
These features are moving to the core CLI and no longer need to be overridden in the extension.
https://github.com/Greedygre/azure-cli/pull/4

Also fixed a typo in the name of `create_vnet_subnet` so it is the same in both repos.


yonzhan:

src/containerapp/azext_containerapp/commands.py
```diff
@@ -39,13 +39,6 @@ def load_command_table(self, args):
         g.custom_command('delete', 'delete_managed_environment', supports_no_wait=True, confirmation=True, exception_handler=ex_handler_factory())
         g.custom_command('update', 'update_managed_environment', supports_no_wait=True, exception_handler=ex_handler_factory())
 
-    with self.command_group('containerapp env http-route-config', is_preview=True) as g:
```
Greedygre, line 42: To avoid breaking change, we can remove the `is_preview=True` in extension first.
Then after azure-cli release, we update `"azext.minCliCoreVersion": <azure-cli-version>` in [azext_metadata.json](https://github.com/Azure/azure-cli-extensions/blob/main/src/containerapp/azext_containerapp/azext_metadata.json) first, then remove these GA commands in extension.

Tratcher: what if we just wait to merge this until after the azure-cli release?

Greedygre: @Tratcher It depends on if you mind customer will see the `Preview` tag if they already install latest containerapp extension and azure-cli version.
If you care the `Preview` Tag, it should separate to 2 PR.

image

Greedygre: But yes we can just wait to merge until after the azure-cli release.
Besides I will discuss the azure-cli-extension containerapp module release plan for ignite with co-workers after holiday to double check if we have plan to release a new version to separate these.

Tratcher:
CLI 2.79.0 has shipped, this can merge now.

src/containerapp/azext_containerapp/_help.py
```diff
@@ -2337,102 +2337,3 @@
       text: |
           az containerapp label-history show -n my-containerapp -g MyResourceGroup --label LabelName
 """
```
Greedygre, line 2343: 1. Please update `\src\containerapp\azext_containerapp\azext_metadata.json`, set `"azext.minCliCoreVersion": "2.79.0"`

2. Update `HISTORY.rst`:
```
* Update azure cli dependency version >= "2.79.0"
* `az containerapp env http-route-config/premium-ingress`: Remove GA commands which exist in azure-cli version 2.79.0
```

Greedygre: I updated the branch

PR #9268, opened 2025-10-03
Feature/eas llm setup

alex3267006:
---

### Related command

This pull request introduces major improvements to the AKS Agent extension, focusing on simplifying LLM model setup and management for users. The most significant change is the addition of the new `az aks agent-init` command, which enables interactive configuration of LLM providers and models, removing the need for manual environment variable setup. Documentation and help text have been updated to reflect these changes, and new backend modules have been added to support provider selection and configuration management. There are also code changes to ensure configuration consistency and prevent conflicts.

**User Experience and CLI Improvements**

* Added the new `az aks agent-init` command, allowing users to interactively add and configure LLM models for AKS troubleshooting. This command saves model configurations locally, supports multiple providers, and guides users through the setup process.
* Updated documentation in `README.rst` and CLI help to highlight the new workflow: users no longer need to set environment variables manually, and can easily select or switch models using `--model` or `--config-file`. 

**Configuration Management and Validation**

* Introduced `LLMConfigManager` (`llm_config_manager.py`) for robust loading, saving, and retrieval of LLM model configurations, ensuring the last used or specifically selected model is used by default.
* Added provider registry and interactive selection logic in `llm_providers/__init__.py`, making it easy for users to choose among Azure, OpenAI, Anthropic, Gemini, and other compatible providers.

**Provider Support and Validation**

* Added new provider classes for Azure, OpenAI, OpenAI-Compatible, Gemini and Anthropic, each with schema-based parameter validation and live connection testing to ensure correct configuration before use. 

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

src/aks-agent/azext_aks_agent/custom.py
```diff
@@ -14,6 +19,31 @@
 
 
 # pylint: disable=unused-argument
+def aks_agent_init(cmd):
+    """Initialize AKS agent llm configuration."""
+    print("Welcome to AKS Agent LLM configuration setup.")
```
mainred: Let's use log for output.

src/aks-agent/azext_aks_agent/custom.py
```diff
@@ -14,6 +19,31 @@
...
+    params = provider.prompt_params()
+
+    llm_config_manager = LLMConfigManager()
+    # If the connection to the model endpoint is valid, save the configuration
+    is_valid, message, action = provider.validate_connection(params)
+    if is_valid and action == "save":
+        logger.info("%s", message)
+        llm_config_manager.save(provider.name, params)
+        logger.info("Configuration saved successfully.")
+        print("LLM configuration setup successfully.")
```
mainred: print seems to be used by intention, what's the difference this CLI?

src/aks-agent/azext_aks_agent/custom.py
```diff
@@ -51,6 +150,26 @@ def aks_agent(
...
+    llm_config: Dict,
+    llm_config_manager: LLMConfigManager
+) -> bool:
+    # Check if provider name is not empty
+    if not provider_name:
+        raise ValueError("No provider name.")
+    # Check if provider is supported
+    if provider_name not in PROVIDER_REGISTRY:
+        raise ValueError(
+            f"Unsupported provider {provider_name}.")
```
mainred: If the provider is not in PROVIDER_REGISTRY, the user can still use that llm in aks agent, right?

mainred: But they need to provide their own setting through the configuration file or environment variables.
`Unsupported provider {provider_name} for LLM initialization, and supported llm providers are {PROVIDER_REGISTRY} . please refer to doc to setup the manually`

src/aks-agent/azext_aks_agent/agent/llm_providers/azure_provider.py
```diff
@@ -0,0 +1,73 @@
...
+        api_key = params.get("AZURE_API_KEY")
+        api_base = params.get("AZURE_API_BASE")
+        api_version = params.get("AZURE_API_VERSION")
+        model_name = params.get("MODEL_NAME")
+
+        if not all([api_key, api_base, api_version, model_name]):
+            return False, "Missing required Azure parameters.", "retry_input"
+
+        # REST API reference: https://learn.microsoft.com/en-us/azure/ai-foundry/openai/api-version-lifecycle?tabs=rest
+        api_base = api_base.rstrip('/') + '/'
```
mainred: api_base should be correctly passed down here? but it's ok to recheck the format.

src/aks-agent/azext_aks_agent/agent/llm_providers/azure_provider.py
```diff
@@ -0,0 +1,73 @@
...
+            },
+            "AZURE_API_VERSION": {
+                "secret": False,
+                "default": "2025-04-01-preview",
+                "hint": None,
+                "validator": non_empty
+            },
+            "MODEL_NAME": {
+                "secret": False,
+                "default": "gpt-4.1",
```
mainred: I am not sure if it's a good idea to default the model, instead, we should raise when the mode is empty.

src/aks-agent/azext_aks_agent/agent/llm_providers/base.py
```diff
@@ -0,0 +1,106 @@
...
+from typing import Dict, Callable, Tuple, Any
+import re
+
+
+def non_empty(v: str) -> bool:
+    return bool(v and v.strip())
+
+
+def is_url(v: str) -> bool:
+    regex = re.compile(
```
mainred: is there any special requirement for the url format? How about using python library

```
from urllib.parse import urlparse

def is_valid_url_format_urllib(url_string):
    try:
        parsed_url = urlparse(url_string)
        # A minimal valid URL typically requires a scheme and a network location
        return all([parsed_url.scheme, parsed_url.netloc])
    except Exception:
        return False
```

src/aks-agent/azext_aks_agent/agent/llm_providers/base.py
```diff
@@ -0,0 +1,106 @@
...
+                "secret": True/False,
+                "default": "default value or None",
+                "hint": "Additional hint to show user",
+                "validator": Callable[[str], bool]  # function to validate input
+            }
+        }
+        """
+        raise NotImplementedError()
+
+    def prompt_params(self):
```
mainred, line 52: Holmes is using rich.Console to interactively get the user answers https://rich.readthedocs.io/en/latest/console.html. How about we try this? But remember using the same library version as holmes's

mainred (commented):
Thanks for your changes, the code change, docs and UTs are great.
Left the first-round review.

src/aks-agent/azext_aks_agent/custom.py
alex3267006: Fixed.

src/aks-agent/azext_aks_agent/custom.py
alex3267006: Use console.print() instead.

src/aks-agent/azext_aks_agent/custom.py
alex3267006: Currently we only support these providers, in which the openai_compatible_provider may have more flexibility.
Have added the prompt.

src/aks-agent/azext_aks_agent/agent/llm_providers/azure_provider.py
alex3267006: Simplified.

src/aks-agent/azext_aks_agent/agent/llm_providers/azure_provider.py
alex3267006: Removed it to Hint. We will check non-empty

src/aks-agent/azext_aks_agent/agent/llm_providers/base.py
alex3267006: Simplified.

src/aks-agent/azext_aks_agent/agent/llm_providers/base.py
alex3267006: Using console.print() instead.

src/aks-agent/azext_aks_agent/custom.py
```diff
@@ -14,6 +19,34 @@
 
 
 # pylint: disable=unused-argument
+def aks_agent_init(cmd):
+    """Initialize AKS agent llm configuration."""
+    from rich.console import Console
+    console = Console()
+    logger.info("Welcome to AKS Agent LLM configuration setup.")
+    console.print("Welcome to AKS Agent LLM configuration setup.")
```
mainred: For user prompt, we use console to print colored text with the same experience as agent. and for the others, we use log.
In this case, I think it's more like a prompt, can we use console only? and the color used in holmes is [here](https://github.com/robusta-dev/holmesgpt/blob/master/holmes/interactive.py#L43). In this case, we use HELP_COLOR

src/aks-agent/azext_aks_agent/custom.py
```diff
@@ -14,6 +19,34 @@
...
+    llm_config_manager = LLMConfigManager()
+    # If the connection to the model endpoint is valid, save the configuration
+    is_valid, message, action = provider.validate_connection(params)
+    if is_valid and action == "save":
+        logger.info("%s", message)
+        llm_config_manager.save(provider.name, params)
+        logger.info("Configuration saved successfully.")
+        console.print("LLM configuration setup successfully.")
+    elif not is_valid and action == "retry_input":
+        logger.warning("%s Please re-run `aks agent init` to correct the input parameters.", message)
```
mainred:
```suggestion
        logger.warning("%s Please re-run `az aks agent-init` to correct the input parameters.", message)
```

src/aks-agent/azext_aks_agent/commands.py
```diff
@@ -14,3 +14,4 @@ def load_command_table(self, _):
         "aks",
     ) as g:
         g.custom_command("agent", "aks_agent")
+        g.custom_command("agent-init", "aks_agent_init")
```
mainred, line 17: Instead of a introducing another subcommand which is not under the agent scope, how about we adding a flag `--init` like az aks agent --init to initialize the configuration and llm setting? cc @aritraghosh 
This just comes to me, and we can discuss.

src/aks-agent/azext_aks_agent/agent/llm_providers/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    choices = _provider_choices_numbered()
+    if not choices:
+        raise ValueError("No providers are registered.")
+    while True:
+        for idx, name in choices:
+            console.print(f" {idx}. {name}")
+        sel_idx = input("Enter the number of your choice: ").strip().lower()
+
+        if sel_idx == "/exit":
```
mainred, line 72: How does the user know /exit will exit the program?

mainred: holmesgpt can provide some reference to this interactive mode.

https://github.com/robusta-dev/holmesgpt/blob/master/holmes/interactive.py#L55

src/aks-agent/azext_aks_agent/agent/llm_providers/anthropic_provider.py
```diff
@@ -0,0 +1,60 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+
+import requests
+from .base import LLMProvider, non_empty
+
+
+class AnthropicProvider(LLMProvider):
```
mainred, line 11: Did we have a chance to test the code against the the supported llms?

src/aks-agent/azext_aks_agent/custom.py
alex3267006: Fixed.

src/aks-agent/azext_aks_agent/custom.py
alex3267006: Fixed.

src/aks-agent/azext_aks_agent/commands.py
alex3267006: Last week I inquired this issue with Julie Zhu. The current 'az aks agent' is already defined as a command for Q&A ('az aks' is a command group). So we need to use some other key "verb" as a new command (not a subcommand of az aks agent).

src/aks-agent/azext_aks_agent/agent/llm_providers/__init__.py
alex3267006: We have some hints in retry prompt. Updated using holmes.interative.SlashCommands.

src/aks-agent/azext_aks_agent/agent/llm_providers/anthropic_provider.py
alex3267006: Have tested azure, openai, anthropic and gemini, tried ollama with some small models but not supported by litellm. Will try others later.

mainred:
@alex3267006  let's 
1. update the setup.py to include the dependencies like rich
2. since it's a big change, let's bump the aks agent version as well.
3. follow the PR title pattern to use `{AKS}: introduce agent-init subcommand for llm setup`

The CI failed for the missing dependencies.

mainred (dismissed):

alex3267006:
@microsoft-github-policy-service agree company="Microsoft"

src/aks-agent/azext_aks_agent/commands.py
yanzhudd: To clarify, what we discussed last week was that you cannot define "az aks agent" as both a command and a command group at the same time. This means that if "az aks agent" has already been defined as a command, you can no longer define it as a command group.

However, you can still add new parameters to the existing command by modifying the _params.py file.

yanzhudd: BTW, please refer to this doc regarding the naming specification for Azure CLI commands: https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md#command-naming-and-behavior

mainred: We discuss whether or not to make agent as a command group, under which we can add `agent init` or `agent ask`, but finally we draw a conclusion to use what the PR presents.
`az aks agent` is the main entrypoint of the agent, and the other tools are not frequently used, so we want to make `az aks agent` simply without an extra subcommand.

mainred: So init-agent is the preferred command in this case?

mainred: > Avoid hyphenated command names when moving the commands into a subgroup would eliminate the need (e.g. database show and database get instead of show-database and get-database).

Or you suggest we move the init under agent command group to be `az aks agent init`? But considering the frequency of the agent command, and the announcement of the usage, we want to keep agent as a command and add another command to init the agent if possible

yanzhudd: What I just mentioned above is simply a clarification — you can still add parameters to an existing command in the future, but you cannot define "az aks agent" as both a command and a command group at the same time, in case you weren’t aware of that.

mainred: Thanks for the clarification. we'll keep `az aks agent` as a command not a command group.

mainred:
@yanzhudd please help merge this PR, thanks.

yanzhudd:
please resolve the code conflict

yanzhudd (approved):

PR #9380, opened 2025-11-03
Fix live test for quantum extension

xinyi-joffre:
Update live test to point to new H2 target, since previous H1 target was brought offline. This is test change only, no changes to the actual extension.


yonzhan:

xinyi-joffre:
@yonzhan, would we be able to get help to merge this tiny change? This PR doesn't change actual test contents or the azure cli extension, but is needed only because a upstream target name changed from our partner, since they brought down the target and replaced it with another one.

PR #9409, opened 2025-11-10
[confcom] Add a `--with-containers` flag to policy and fragment gen

DomAyre:
### Why

We currently have a custom input json format to describe containers to be permitted in a policy. This can be used by acipolicygen and acifragmentgen.

This json is basically the same as the final container definition in the policy except it's arbitrarily different, e.g. "env_rules" vs "environmentVariables".

Therefore it would make more sense to just be able to supply the container definition directly. In future PRs we will provide tools to generate these from inputs such as arm_templates.

### How

- [x] Add new flag `--with-containers` to acipolicygen and acifragmentgen
- [x] Define a schema for the container definition in the final policy
- [x] When flag is set, put the container definition given into the final serialised policy (filling out default values where missing)

---

### General Guidelines

### DIFF
--- a/src/confcom/azext_confcom/custom.py
+++ b/src/confcom/azext_confcom/custom.py
@@ -241,7 +241,7 @@ def acifragmentgen_confcom(
     key: str,
     chain: str,
     minimum_svn: str,
-    container_definitions: list,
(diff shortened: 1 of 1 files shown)


yonzhan:

PR #9414, opened 2025-11-11
aks bastion: allow using public FQDN

stevekuznetsov:
When users have public FQDN on their private cluster and want to use normal DNS resolution for bastion, we should let them.


yonzhan:

PR #9420, opened 2025-11-12
Test built extension wheels in CI

DomAyre:
### Why

This is a tiny step in addressing 
- #9425 

As an extension developer, I authored a PR which broke my extension by introducing code which was missing an `__init__.py` file which stopped it being included in the built extension package
- #9409 

This wasn't caught by CI because it doesn't run any testing on the built package wheel, only that a wheel can be built without errors. 

Ideally all the tests we define as extension authors should run on the thing that will ultimately be released, but the way we have to run our testing through azdev makes this non trivial as azdev doesn't support installing extensions from source wheel.

So for now, we can do better than nothing by running the extension with `--help` which should at least catch broken imports like my case.

### How

- [x] Update `ci/test_source.py::test_source_wheels()` to optionally output the wheels to a provided dir
- [x] Update `ci/test_source.py::test_extension()` to optionally install built wheels and attempt to run that extension
- [x] When running the file directly, run `test_source_wheels()` before `test_extension()` with a tempdir containing the wheels

---

### General Guidelines

### DIFF
--- a/scripts/ci/test_source.py
+++ b/scripts/ci/test_source.py
@@ -89,21 +89,25 @@ def test_extension(whl_dir: Optional[Path] = None):
         cmd = ['azdev', 'extension', 'remove', ext_name]
         run_command(cmd, check_return_code=True)
 
+        # This catches code missing in the final wheel package, but until the full
+        # set of extension tests can be run against the wheel, it will not catch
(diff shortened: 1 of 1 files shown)


yonzhan:

PR #9429, opened 2025-11-14
[confcom] Run testing against built extension

DomAyre:
### Why

This repos CI uses azdev to run each extensions tests, however this means it's testing the local modules which might have breaking differences compared to what we ship to customers.

### How

- [x] Add a new autouse pytest fixture (currently scoped to the confcom extension but designed to apply to all extensions) which builds wheels for each extension being tested, and swaps out the non-test modules

The modules actually running the tests are the only part not swapped for the built version to handle tests which rely on unshipped code/data

---

### General Guidelines

Note: This is strictly a testing only change, so no version bump or history update is required


yonzhan:

kairu-ms (approved):

PR #9633, opened 2026-02-25
[confcom] fragment attach updates

micromaomao:
Fragments should be attached to the manifest corresponding to the correct platform, but for a multiarch image, `oras attach` without any additional parameter, which is what the tooling currently does, attaches it to the index. This causes fragment to not be loaded in ACI for such images (which are now generated by Docker by default, due to having an extra provenance manifest with unknown/unknown platform).

This PR fixes this by adding `--platform` flag to `oras attach`, and corrects the behaviour of both `az confcom fragment attach` and `az confcom acifragmentgen --upload-fragment`.

Also fix unrelated bug in `az confcom fragment attach`.

Tested both the old and new command with both --provenance=true and --provenance=false images.

### DIFF
--- a/src/confcom/azext_confcom/tests/latest/test_confcom_acifragmentgen.py
+++ b/src/confcom/azext_confcom/tests/latest/test_confcom_acifragmentgen.py
@@ -25,18 +25,20 @@ def docker_image():
         ["docker", "run", "-d", "-p", "0:5000", "registry:2"],
         stdout=subprocess.PIPE,
         text=True,
-    ).stdout
+        check=True,
+    ).stdout.strip()
 
     registry_port = subprocess.run(
         ["docker", "port", registry_id],
         stdout=subprocess.PIPE,
         text=True,
+        check=True,
     ).stdout.split(":")[-1].strip()
 
     test_container_ref = f"localhost:{registry_port}/hello-world:latest"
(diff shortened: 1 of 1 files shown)


yonzhan:

micromaomao:
Update HISTORY.rst
Add Signed-off-by

micromaomao:
Changed version increment from 1.7.2 to 1.8.0

micromaomao:
log warning on oras manifest fetch[-config] failures
format

micromaomao:
Fix lint, remove unused `import platform`

micromaomao:
Use prepend_docker_registry in get_image_platforms
Minor grammar fix

Always use linux/amd64 for automatic fragment discovery:

Originally the change made this command fail if an image with multiple platforms
was used, but this breaks policy generation whenever an actual multiarch image
(i.e. amd64+arm64) was used in the input, regardless of whether we intend to
fetch fragments from it.

Therefore, let's make it always use linux/amd64 for fragment discovery.  Since
this command is considered "legacy", it should not be used for arm64 policies
anyway (when that comes).

micromaomao:
This PR is ready to merge as agreed with @DomAyre 
@necusjz can you do an azp run?

necusjz (approved):

PR #9638, opened 2026-02-26
Add ARM64 support for Helm installation in connectedk8s and k8s-extension

ashnanze:
## Summary
- Adds ARM64-aware Helm client installation logic in both `connectedk8s` and `k8s-extension`.
- Detects host architecture and uses the ARM64 Helm artifact path when running on ARM64 machines.
- Uses the official Helm distribution source for ARM64 binaries.
- Ensures the versioned local download directory exists before saving the Helm archive.

## Why
- Enables onboarding and extension workflows to run on ARM64 environments.
- Prevents ARM64 Helm bootstrap failures caused by missing local download paths.

## Validation
- Built and installed the updated extension `.whl`.
- Validated behavior on an ARM64 VM.
- Validated behavior on an AMD64 VM.
- Confirmed the updated Helm installation flow works across both architectures.


yonzhan:

PR #9697, opened 2026-03-18
Add fileshares extension for 2025-09-01-preview api version

ankushbindlish2:
- Add Microsoft.FileShares CLI extension (2025-09-01-preview API)
- 21 commands across 4 command groups (all Preview stage)
- Command groups: file-shares, file-share, file-share-snapshot, private-endpoint-connection, private-link-resource
- Generated from AAZ command models

---

### General Guidelines

For new extensions:

### About Extension Publish


yonzhan:

VincentLiu777:
Thanks for getting both the AAZ and CLI extension PRs up so quickly. The overall structure looks clean. Just a few adjustments to bring the CLI UX in line with Azure conventions before we merge.
 
Private Endpoint Connection Commands
 
The current PR uses create and update for private-endpoint-connection, but the established Azure CLI pattern uses approve and reject (see az storage account private-endpoint-connection for reference):
 
   - Remove create — private endpoint connections originate from the consumer side; the resource owner doesn't create them directly.
   - Add approve and reject — these are the standard verbs for accepting or denying pending connections.
   - show, list, update, and delete are good as-is.
 
az fileshare check-name-availability
 
    --name should be required — the command doesn't work without it.
    --type should be hidden and auto-populated with Microsoft.FileShares/fileShares, since the extension only targets one resource type. No reason to make the user type it every time.
 
Parameter Naming for az fileshare commands [create, update, and get-provisioning-recommendation]
 
    --provisioned-io-per-sec → --provisioned-iops (shorter, widely understood abbreviation)
    --provisioned-throughput-mi-b-per-sec → --provisioned-throughput-MiB (cleaner, preserves the unit acronym)
    --provisioned-storage-gi-b → --provisioned-storage-GiB (standard GiB casing)
 
These renames also apply to get-provisioning-recommendation, which currently uses --storage-gib.
 
Flatten Nested Parameters az fileshare commands [create, update] 
 
  Two parameters are currently nested inside object args, which adds friction for users:
 
   --root-squash {NoRootSquash, RootSquash, AllSquash} — currently requires --nfs-props root-squash=RootSquash. Since NFS is the only protocol today, promoting this to a top-level flag is much more intuitive.
   --allowed-subnets <subnet-id> ... — currently requires --public-access allowed-subnets=<id>. Same idea — a flat top-level flag is simpler.
 
  Both can be done via custom parameter overrides in _params.py with mapping logic in custom.py.
 
Thanks.

necusjz (approved):

PR #9719, opened 2026-03-25
[aks-agent] Add Microsoft Entra ID authentication support for Azure OpenAI

mainred:
## Summary
- Add support for keyless authentication using Microsoft Entra ID for Azure OpenAI
- Allow users to skip providing an API key during initialization to enable workload identity-based authentication
- Configure aks-agent pod to use the same service account as aks-mcp pod with workload identity enabled

## Changes
- Bump version to 1.0.0b22 and aks-agent to v0.7.1
- Allow empty API key during Azure OpenAI configuration with hint message
- Configure helm values for aks-agent pod: `workloadIdentity.enabled=true`, `serviceAccount.create=false`, and use the same service account name as aks-mcp
- Skip creating secrets and environment variables when API key is empty
- Enable `azureADTokenAuth` flag in helm when API key is not provided

## Test plan
- [ ] Test initialization with empty API key for Azure OpenAI
- [ ] Verify helm values include workloadIdentity and serviceAccount configuration
- [ ] Verify aks-agent pod uses the same service account as aks-mcp
- [ ] Test Azure OpenAI authentication with workload identity

### DIFF
--- a/src/aks-agent/azext_aks_agent/agent/llm_providers/__init__.py
+++ b/src/aks-agent/azext_aks_agent/agent/llm_providers/__init__.py
@@ -5,7 +5,7 @@
 
 from typing import List, Tuple
 
-from azext_aks_agent.agent.console import ERROR_COLOR, HELP_COLOR, INFO_COLOR
+from azext_aks_agent.agent.console import ERROR_COLOR, HELP_COLOR
 from rich.console import Console
 
 from .anthropic_provider import AnthropicProvider
--- a/src/aks-agent/azext_aks_agent/custom.py
+++ b/src/aks-agent/azext_aks_agent/custom.py
@@ -19,7 +19,6 @@ from azext_aks_agent.agent.console import (
 from azext_aks_agent.agent.k8s import AKSAgentManager, AKSAgentManagerClient
 from azext_aks_agent.agent.k8s.aks_agent_manager import AKSAgentManagerLLMConfigBase
 from azext_aks_agent.agent.llm_providers import prompt_provider_choice
-from azext_aks_agent.agent.llm_providers.azure_entraid_provider import AzureEntraIDProvider
 from azext_aks_agent.agent.telemetry import CLITelemetryClient
 from azure.cli.core.azclierror import AzCLIError
 from azure.cli.core.commands.client_factory import get_subscription_id
@@ -179,7 +178,7 @@ def _setup_helm_deployment(console, aks_agent_manager: AKSAgentManager):
         console.print(
             f"\n👤 Current service account in namespace '{aks_agent_manager.namespace}': {service_account_name}",
             style="cyan")
-        
+
         # Check if using Azure Entra ID provider and show role assignment reminder
         model_list = aks_agent_manager.get_llm_config()
         if model_list and any("azure/" in model_name and not model_config.get("api_key") for model_name, model_config in model_list.items()):
@@ -210,7 +209,7 @@ def _setup_helm_deployment(console, aks_agent_manager: AKSAgentManager):
             "To have access to Azure resources, the service account should be annotated with "
             "'azure.workload.identity/client-id: <managed-identity-client-id>'.",
             style=WARNING_COLOR)
-        
+
         # Check if using Azure Entra ID provider and show role assignment note
         model_list = aks_agent_manager.get_llm_config()
         if model_list and any("azure/" in model_name and not model_config.get("api_key") for model_name, model_config in model_list.items()):


yonzhan:

mainred:
cc @yanzhudd

src/aks-agent/azext_aks_agent/agent/k8s/aks_agent_manager.py
```diff
@@ -927,6 +927,22 @@ def _create_helm_values(self):
...
+            "enabled": True,
+        }
+        helm_values["serviceAccount"] = {
+            "create": False,
+            "name": self.aks_mcp_service_account_name,
+        }
+
+        has_empty_api_key = any(
+            not model_config.get("api_key") or not model_config.get("api_key").strip()
+            for model_config in self.llm_config_manager.model_list.values()
```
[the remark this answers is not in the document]
mainred: we validate the emptiness when the user provide the api key.

src/aks-agent/azext_aks_agent/agent/llm_providers/__init__.py
```diff
@@ -19,11 +20,11 @@
 
 _PROVIDER_CLASSES: List[LLMProvider] = [
```
gossion, line 21: **Potential bug: `PROVIDER_REGISTRY` key collision**

Both `AzureProvider` and `AzureEntraIDProvider` have `model_route = "azure"`, which means their `name` property (derived from `model_route`) is both `"azure"`. If `PROVIDER_REGISTRY` is keyed by `name`, the second registration (`AzureEntraIDProvider`) will silently overwrite the first (`AzureProvider`).

This could cause issues when looking up a provider by name. Consider giving `AzureEntraIDProvider` a distinct `name` (e.g., by overriding the `name` property to return `"azure_entraid"`).

src/aks-agent/azext_aks_agent/agent/k8s/aks_agent_manager.py
```diff
@@ -927,6 +927,22 @@ def _create_helm_values(self):
             "create": False,
         }
 
+        # Configure aks-agent pod to use the same service account as aks-mcp for workload identity
+        helm_values["workloadIdentity"] = {
+            "enabled": True,
+        }
+        helm_values["serviceAccount"] = {
+            "create": False,
```
gossion, line 935: **workload identity is enabled unconditionally for all deployments**

These Helm values (`workloadIdentity.enabled=true`, `serviceAccount.create=false`) are set for every deployment, even when the user is purely using API key authentication and does not have workload identity configured on their cluster.

This could cause deployment failures or unexpected behavior on clusters without workload identity. Consider gating this behind the `has_empty_api_key` check below:

```python
has_empty_api_key = any(
    not (model_config.get("api_key") or "").strip()
    for model_config in self.llm_config_manager.model_list.values()
)
if has_empty_api_key:
    helm_values["workloadIdentity"] = {"enabled": True}
    helm_values["serviceAccount"] = {
        "create": False,
        "name": self.aks_mcp_service_account_name,
    }
    helm_values["azureADTokenAuth"] = True
```

src/aks-agent/azext_aks_agent/agent/llm_providers/azure_entraid_provider.py
```diff
@@ -0,0 +1,58 @@
...
+            "api_version": {
+                "secret": False,
+                "default": "2025-04-01-preview",
+                "hint": None,
+                "validator": non_empty
+            }
+        }
+
+    def validate_connection(self, params: dict) -> Tuple[str, str]:
+        api_base = params.get("api_base")
```
gossion, line 51: Unlike `AzureProvider.validate_connection`, this method only checks parameter presence and does not actually validate the connection to the Azure OpenAI endpoint. This is understandable since the Entra ID token is not available at CLI configuration time (it will be obtained by the pod via workload identity).

However, it would be helpful to add a comment here explaining *why* connection validation is skipped, so future maintainers don't mistake this for an incomplete implementation.

```python
def validate_connection(self, params: dict) -> Tuple[str, str]:
    # Connection validation is skipped for Entra ID auth because the token
    # is obtained at runtime via workload identity, not available at CLI config time.
    ...
```

src/aks-agent/azext_aks_agent/agent/k8s/aks_agent_manager.py
gossion, line 941: Nit: the empty api_key condition can be simplified. The `or` short-circuit already handles `None`/empty string, so the second `.strip()` check only adds value for whitespace-only strings. A simpler equivalent:

```python
has_empty_api_key = any(
    not (model_config.get("api_key") or "").strip()
    for model_config in self.llm_config_manager.model_list.values()
)
```

src/aks-agent/azext_aks_agent/custom.py
```diff
@@ -179,6 +179,19 @@ def _setup_helm_deployment(console, aks_agent_manager: AKSAgentManager):
             f"\n👤 Current service account in namespace '{aks_agent_manager.namespace}': {service_account_name}",
             style="cyan")
 
+        # Check if using Azure Entra ID provider and show role assignment reminder
+        model_list = aks_agent_manager.get_llm_config()
+        if model_list and any("azure/" in model_name and not model_config.get("api_key") for model_name, model_config in model_list.items()):
+            console.print(
+                f"\n⚠️  IMPORTANT: If using keyless authentication with Azure OpenAI, ensure the 'Cognitive Services OpenAI User' or 'Azure AI Developer' role "
+                f"is assigned to the workload identity (service account: {service_account_name}).",
+                style=f"bold {INFO_COLOR}"
```
gossion, line 188: This Entra ID detection + warning block is duplicated almost identically in the `helm_status == "not_found"` branch below. Consider extracting it into a helper function like `_warn_entraid_role_assignment(console, model_list, service_account_name)` to reduce duplication.

src/aks-agent/HISTORY.rst
```diff
@@ -12,6 +12,13 @@ To release a new version, please select a new version number (usually plus 1 to
 Pending
 +++++++
 
+1.0.0b22
```
gossion, line 15: Minor: HISTORY.rst says "Bump aks-agent to v0.7.1" but the PR description says "v0.7.0". The code in `_consts.py` matches the changelog (v0.7.1). Please update the PR description to be consistent.

src/aks-agent/azext_aks_agent/agent/llm_providers/__init__.py
mainred: It's by design. We want two kinds of azure openai.

src/aks-agent/azext_aks_agent/agent/k8s/aks_agent_manager.py
mainred: It's also by design. It's ok to not have workload identity.

src/aks-agent/azext_aks_agent/agent/llm_providers/azure_entraid_provider.py
mainred: We cannot check validation for keyless auzre openai because it relies on azure role of the current user.

src/aks-agent/azext_aks_agent/custom.py
mainred: This minor change can be included in a following PR.

src/aks-agent/azext_aks_agent/agent/k8s/aks_agent_manager.py
mainred: each llm provider with check the emptiness of the apikey

src/aks-agent/HISTORY.rst
mainred: Done

gossion (approved):
lgtm

PR #9815, opened 2026-04-22
Add ARM64 support for Helm installation in connectedk8s

ashnanze:
## Summary
- Bumps Helm version from `v3.12.2` to `v3.20.1`.
- Detects host architecture via `platform.machine()` (maps `aarch64`/`arm64` to `arm64`, defaults to `amd64`).
- Uses the detected architecture in all Helm artifact tags, download file names, and install paths (previously hardcoded to `amd64`).
- Adds `_resolve_helm_pull_target()` to resolve the correct OCI pull target: tries arch-specific tag first (`helm-v3.20.1-linux-arm64`), falls back to manifest list tag (`helm-v3.20.1`) with annotation-based matching.
- Ensures the versioned local download directory exists before saving the Helm archive.

## Why
- Enables `az connectedk8s connect` to work on ARM64 machines by downloading the correct Helm binary.
- Previously, ARM64 hosts would download the AMD64 binary, which would fail to execute. 

## Validation
- Built and installed the updated extension `.whl`.
- Validated behavior on an ARM64 VM.
- Validated behavior on an AMD64 VM.
- Both VMs successfully completed `az connectedk8s connect`.
- Confirmed the updated Helm installation flow works across both architectures.

## Directory handling note
- This fix also addresses a latent path issue in Helm download setup.
- The download target file path is `~/.azure/helm/<version>/<archive-file>`.
- Previous logic only created the parent directory `~/.azure/helm`, not the required version subdirectory `~/.azure/helm/<version>`.
- Failures occurred when `~/.azure/helm/<version>` was missing, even if `~/.azure/helm` already existed.
- It could appear to work when that specific `<version>` directory had already been created by a prior run.
- The fix now creates `download_location` (`~/.azure/helm/<version>`) directly before download.


yonzhan:

PR #9903, opened 2026-05-27
remove depericated --location

s-ddavydenko:
## Summary

Adds `--workspace-kind` parameter to `az quantum workspace create`, allowing users to specify whether to create a V1 or V2 Azure Quantum workspace.

## Motivation

V2 workspaces have a different architecture and must be explicitly requested at creation time. The service already supports this via the `workspaceKind` field in the REST API, but there was no way to set it from the CLI — users had to use the portal or call the API directly.

## Changes

- Added `--workspace-kind` (`V1` or `V2`) argument to `az quantum workspace create` in `_params.py`.
- Updated `operations/workspace.py` to pass `workspace_kind` through both code paths: the direct API path (`--skip-role-assignment`) sets it on `WorkspaceResourceProperties`, and the ARM template path includes it as a `workspaceKind` parameter.
- Updated the ARM template `create-workspace-and-assign-role.json` to accept and forward `workspaceKind` to the workspace resource properties
- Added V2 example to `az quantum workspace create` help text.
- Bumped extension version

## Testing

- All offline unit tests pass locally 
- New Unit test added that verifies previous behavior for no breaking changes and checks new workspaces are created with the right workspaceKind field by serializing the JSON.
- Manually verified end-to-end via `test_v2_workspace_temp.py`: created a V2 workspace and confirmed workspaceKind is V2 on azure

### DIFF
--- a/src/quantum/HISTORY.rst
+++ b/src/quantum/HISTORY.rst
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+1.0.0b15
++++++++++++++++
+* Removed deprecated ``--location``/``-l`` parameter from ``quantum execute``, ``quantum run``, ``quantum job submit``, ``quantum job cancel``, ``quantum job list``, ``quantum job output``, ``quantum job show``, ``quantum job wait``, ``quantum target list``, ``quantum workspace set``, and ``quantum workspace quotas`` commands.
+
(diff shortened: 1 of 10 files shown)


yonzhan:
quantum

Issue #7740, opened 2024-06-25
{Package} Upgrade setuptools and wheel to the latest version.


wangzelin007:
### Describe the bug

Now we pin the setuptools version to 70.0.0
and the wheel version to 0.30.0
Otherwise the metadata.json file will not be generated in the dist-info directory of the wheel package.
We should modify the logic of reading metadata and upgrade setuptools and wheel to the latest versions.
Otherwise it will be difficult to maintain the related code.
https://github.com/Azure/azure-cli-extensions/blob/main/azure-pipelines.yml#L79
https://github.com/Azure/azure-cli-extensions/blob/main/azure-pipelines.yml#L104-L105
https://github.com/Azure/azure-cli-extensions/blob/main/scripts/ci/util.py#L57
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/extension/tests/latest/test_dev_type_extension.py#L42
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/extension/tests/latest/test_wheel_type_extension.py#L46
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/extension/tests/latest/test_wheel_type_extension.py#L77
https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/extension/tests/latest/test_wheel_type_extension.py#L77
https://github.com/Azure/azure-cli-dev-tools/blob/dev/azdev/operations/extensions/util.py#L48

### Related command

None

### Errors

```
metadata in the wheel file is: {
  "azext.minCliCoreVersion": "2.56.0",
  "azext.isPreview": true
}
Traceback (most recent call last):
  File "/mnt/vss/_work/1/s/scripts/ci/azdev_linter_style.py", line 290, in <module>
    main()
  File "/mnt/vss/_work/1/s/scripts/ci/azdev_linter_style.py", line 281, in main
    azdev_on_internal_extension(modified_files, azdev_type)
  File "/mnt/vss/_work/1/s/scripts/ci/azdev_linter_style.py", line 233, in azdev_on_internal_extension
    azdev_extension.check_extension_name()
  File "/mnt/vss/_work/1/s/scripts/ci/azdev_linter_style.py", line 116, in check_extension_name
    if metadata['name'] != extension_name:
       ~~~~~~~~^^^^^^^^
KeyError: 'name'
```

### Issue script & Debug output

None

[...]

PR #9922, opened 2026-06-08
{CI} Upgrade azdev to decouple from wheel 0.30.0

aryasadeghi1:
Part of #7740. Needs azdev 0.2.11b1 (azure-cli-dev-tools#549).

We've been pinned to wheel==0.30.0 because the old metadata.json it wrote is what azdev read to build index.json entries. Newer wheel stopped writing that file, so this switches the CI over to azdev's pkginfo-based metadata and drops
the wheel==0.30.0 / setuptools==70.0.0 pins.

Changes:
- azure-pipelines.yml: remove the wheel==0.30.0 installs. IndexVerify now installs azdev, otherwise test_index.py can't import util anymore.
- azdev_setup.yml: drop setuptools==70.0.0. azdev is installed from source for now since 0.2.11b1 isn't on PyPI yet (there's a TODO to switch back).
- scripts/ci/util.py: get_ext_metadata reads metadata via azdev's pkginfo_to_dict. The import is lazy so scripts that only pull SRC_PATH etc. from util don't suddenly need azdev installed.
- scripts/ci/test_index.py: diff the generated metadata against index.json by meaning instead of byte-for-byte, ignoring the fields wheel used to add (generator, metadata_version, test_requires).
- tests/e2e/packaging: added e2e tests covering az/azdev add/show/list/remove, azdev build, the containerapp build-isolation case, and metadata parity.

`pytest tests/e2e/packaging` passes locally (13 tests).

Still TODO after this: publish azdev 0.2.11b1 and flip the source installs back
to azdev==0.2.11b1, and eventually the pyproject.toml migration.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -83,7 +83,7 @@ jobs:
       # pkginfo-based metadata module (decoupled from wheel==0.30.0), so azdev must be
       # installed here. Install from the merged azure-cli-dev-tools source until azdev
       # 0.2.11b1 is published to PyPI, then switch this to: pip install --upgrade "azdev==0.2.11b1"
-      pip install --upgrade "git+https://github.com/Azure/azure-cli-dev-tools.git@dev#egg=azdev"
+      pip install --upgrade "git+https://github.com/Azure/azure-cli-dev-tools.git@49390f30c6a822620d2fbdf88889c4445d22666f#egg=azdev"
       pip install requests packaging setuptools
       export CI="ADO"
       python ./scripts/ci/test_index.py -v
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -79,7 +79,12 @@ jobs:
   - bash: |
       #!/usr/bin/env bash
       set -ev
-      pip install wheel==0.30.0 requests packaging setuptools
+      # test_index.py imports scripts/ci/util.py, which reads wheel metadata via azdev's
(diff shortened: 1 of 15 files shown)


yonzhan:
CI

azure-pipelines.yml
```diff
@@ -79,7 +79,12 @@ jobs:
   - bash: |
       #!/usr/bin/env bash
       set -ev
-      pip install wheel==0.30.0 requests packaging setuptools
+      # test_index.py imports scripts/ci/util.py, which reads wheel metadata via azdev's
+      # pkginfo-based metadata module (decoupled from wheel==0.30.0), so azdev must be
+      # installed here. Install from the merged azure-cli-dev-tools source until azdev
+      # 0.2.11b1 is published to PyPI, then switch this to: pip install --upgrade "azdev==0.2.11b1"
+      pip install --upgrade "git+https://github.com/Azure/azure-cli-dev-tools.git@dev#egg=azdev"
```
wangzelin007: Out of curiosity, is the @dev (vs the pinned SHA in azdev_setup.yml) intentional?

aryasadeghi1: It's not intentional, I will push a fix for this!

aryasadeghi1: Whilst its inconsistency across the .yml files it should not influence the behaviour as @dev without SHA includes the changes applied.

aryasadeghi1: This is now addressed!

wangzelin007 (approved):

PR #9927, opened 2026-06-10
{AKS} az aks maintenancewindow: Add CRUD commands for peer MaintenanceWindow resources

ibaboulfetouh:
### Related command

`az aks maintenancewindow`

### Summary

Wires up `az aks maintenancewindow {create,show,list,update,delete,wait}` against the existing `MaintenanceWindowsOperations` exposed by the vendored `azure_mgmt_preview_aks` SDK (API version `2026-04-02-preview`, already in `21.0.0b4`).

`MaintenanceWindow` is a resource-group-scoped ARM peer resource that defines a reusable maintenance schedule. When the `maintenanceWindowId` link field becomes available on `maintenanceConfigurations` (May 2026 preview API), customers will be able to link multiple managedCluster maintenance configurations to one shared `MaintenanceWindow` so the schedule lives in a single place.

### Approach

Hand-written commands modeled after the existing `aks maintenanceconfiguration` block. The peer-resource shape is simpler than MTC's because there's no cluster context, no legacy time-in-week schedule shape, and no per-config-name validation gymnastics — just one `MaintenanceWindowResource` (Location + Tags + Properties) wrapping the same `Schedule` enum/types that already power `aksManagedAutoUpgradeSchedule`.

**Reuse, not duplication:** `constructSchedule`, `constructDailySchedule`, `constructWeeklySchedule`, `constructAbsoluteMonthlySchedule`, and `constructRelativeMonthlySchedule` in `azext_aks_preview/maintenanceconfiguration.py` are imported as-is — they validate cross-arg exclusivity for all four schedule types and build the `Schedule` model. The new `maintenancewindow.py` only adds the outer `MaintenanceWindowResource` wrapper. No copy-paste; one source of truth for schedule construction.

**LRO support:** `create`/`delete`/`update` accept `--no-wait` (the SDK exposes `begin_create_or_update` + `begin_delete` as LRO). `update` branches at runtime — tags-only goes through the synchronous `update_tags()` PATCH; any schedule arg present routes through the full `begin_create_or_update()` PUT.

**Feature gate:** Requires `Microsoft.ContainerService/AKSSharedMaintenanceWindowPreview` to be registered on the subscription (auto-approval). Following the established convention (e.g. `EnableFIPSPreview` in `_help.py`), this is documented in the command help, not enforced client-side — the RP returns 403 if missing.

### Changes

| File | Change |
|---|---|
| `_client_factory.py` | + `cf_maintenance_windows` |
| `commands.py` | + `maintenance_window_sdk` CliCommandType + `aks maintenancewindow` command group |
| `maintenancewindow.py` | **NEW** — `constructMaintenanceWindowResource`, `hasAnyScheduleArg` |
| `custom.py` | + `aks_maintenancewindow_{list,show,create,update,delete}` (5 handlers) |
| `_params.py` | + `argument_context` blocks (location auto-bound via CLI core, matching nodepool-snapshot pattern) |
| `_help.py` | + group + 5 per-command help entries with examples |
| `tests/latest/test_maintenancewindow.py` | **NEW** — 11 unit tests |
| `HISTORY.rst` | Pending entry |

### Test plan

- [x] `python -m unittest azext_aks_preview.tests.latest.test_maintenancewindow` — **11/11 pass** (4 schedule happy-paths + 4 validation cases + 3 `hasAnyScheduleArg`)
- [x] `python -m unittest azext_aks_preview.tests.latest.test_maintenanceconfiguration` — **10/10 still pass** (no regressions from the new imports in `custom.py`)
- [x] `az aks maintenancewindow --help` — group lists all 6 commands with the AFEC-flag note
- [x] `az aks maintenancewindow create --help` — all 7 schedule args present, `--location -l` auto-bound, `--no-wait` auto-added, examples render
- [x] `az aks maintenancewindow update --help` — documents the tags-only-vs-PUT contract
- [ ] Live end-to-end against a standalone (once a standalone with the May preview is available)

### General Guidelines

- [x] Have you run `azdev style <YOUR_EXT>` locally? *(local `azdev` not installed; `py_compile` clean on all 7 touched files)*
- [ ] Have you run `python scripts/ci/test_index.py -q` locally? *(no setup.py version bump in this PR; the next aks-preview release will roll all pending changes together)*
- [x] My extension version conforms to the [Extension version schema](https://github.com/Azure/azure-cli/blob/release/doc/extensions/versioning_guidelines.md) *(no version change in this PR)*


src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -935,6 +939,162 @@ def aks_maintenanceconfiguration_update(
...
+# ---------------------------------------------------------------------------
+
+def aks_maintenancewindow_list(
+    cmd,  # pylint: disable=unused-argument
+    client,
+    resource_group_name=None,
+):
+    if resource_group_name is None or resource_group_name == "":
+        return client.list_by_subscription()
+    return client.list(resource_group_name)
```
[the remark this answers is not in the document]
ibaboulfetouh: The vendored SDK `MaintenanceWindowsOperations` exposes `list(resource_group_name)` (returns `ItemPaged[MaintenanceWindowResource]`) and `list_by_subscription()` — confirmed in `azext_aks_preview/vendored_sdks/azure_mgmt_preview_aks/operations/_operations.py` line 9674. There is no `list_by_resource_group` method on this operations group.

Empirically confirmed by running `az aks maintenancewindow list -g <rg>` locally: the call resolved through to ARM and returned `InvalidResourceType` (RP-side; the test sub doesn't have the `AKSSharedMaintenanceWindowPreview` AFEC flag registered) rather than a Python `AttributeError`, which proves the SDK method exists and was invoked.

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -935,6 +939,162 @@ def aks_maintenanceconfiguration_update(
...
+    # Full PUT path: location is required by the MaintenanceWindowResource
+    # model (TrackedResource), so fall back to the existing resource's
+    # location if the caller didn't specify one.
+    if location is None:
+        existing = client.get(resource_group_name, maintenance_window_name)
+        raw_parameters["location"] = existing.location
+        if tags is None:
+            # Preserve existing tags on a schedule-only update so we don't
+            # accidentally clear them — PUT replaces the resource.
+            raw_parameters["tags"] = existing.tags
```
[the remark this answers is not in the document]
ibaboulfetouh: Good catch — the refetch-and-merge code that motivated this comment shouldn't have been there in the first place; it conflated tags-only PATCH with full PUT. Removed in 00d5c5f. The contract is now strict and documented in the help:

- tags-only → PATCH (sync, `--no-wait` ignored)
- any schedule arg → full PUT; caller owns the entire resource body

PUT replaces the resource; tags must be re-passed if the caller wants to keep them. Updated `_help.py` `long-summary` on `aks maintenancewindow update` to state this explicitly. `--location` still defaults to the RG location when omitted (same fallback as `create`).

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -935,6 +939,162 @@ def aks_maintenanceconfiguration_update(
...
+    # tags-only API is synchronous, so honor the flag by warning the user
+    # we ignored it rather than silently dropping it.
+    if not schedule_args_present:
+        if tags is None:
+            raise RequiredArgumentMissingError(
+                "Nothing to update. Provide --tags for a tags-only PATCH, or any of "
+                "--schedule-type / --interval-* / --day-of-week / --day-of-month / "
+                "--week-index / --duration / --utc-offset / --start-date / --start-time "
+                "for a full PUT."
+            )
```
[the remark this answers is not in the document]
ibaboulfetouh: Addressed in 00d5c5f via option (b) — the help long-summary now states the minimum required set explicitly: `--schedule-type` + `--start-time` + `--duration` + per-type fields (`--interval-weeks` / `--day-of-week` / etc.).

Not pursuing option (a) (refetch + merge missing schedule fields) — that would re-introduce the same "implicit hidden state" pattern that comment #2 was flagging on tags, and would conflict with the PATCH-or-full-PUT contract the command intentionally documents.

yonzhan:
AKS

src/aks-preview/azext_aks_preview/_help.py
```diff
@@ -1970,6 +1970,129 @@
                 }
 """
 
+helps['aks maintenancewindow'] = """
+    type: group
+    short-summary: Commands to manage MaintenanceWindow peer ARM resources.
+    long-summary: |
+        MaintenanceWindow is a resource-group-scoped ARM resource that defines a
+        reusable maintenance schedule. Once the `maintenanceWindowId` field is
```
anushkasingh16: nit: When we talk about it being available - do we think that's a detail customers need here?

We could just say "Users can link this resource to a pre-existing maintenanceconfiguration via the maintenanceWindowId. Multiple managedCluster maintenance configurations to a single MaintenanceWindow so the schedule lives in one place"

src/aks-preview/azext_aks_preview/commands.py
```diff
@@ -219,6 +226,22 @@ def load_command_table(self, _):
...
+    # surfaces 403 if it isn't.
+    with self.command_group(
+        "aks maintenancewindow",
+        maintenance_window_sdk,
+        client_factory=cf_maintenance_windows,
+    ) as g:
+        g.custom_command("list", "aks_maintenancewindow_list")
+        g.custom_show_command("show", "aks_maintenancewindow_show")
+        g.custom_command("create", "aks_maintenancewindow_create", supports_no_wait=True)
+        g.custom_command("update", "aks_maintenancewindow_update", supports_no_wait=True)
```
anushkasingh16, line 248: Double confirmation on delete 

The command is registered with confirmation=True and the handler manually calls prompt_y_n. Users will be prompted twice. We should pick one way to confirm

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -935,6 +939,157 @@ def aks_maintenanceconfiguration_update(
...
+        return client.update_tags(
+            resource_group_name,
+            maintenance_window_name,
+            TagsObject(tags=tags),
+        )
+
+    # Full PUT path: the caller owns the resource body. No refetch / merge —
+    # this is a true PUT, mirroring `aks_maintenancewindow_create`. Default
+    # --location to the RG location when omitted (same fallback as create).
+    if location is None:
```
anushkasingh16: If the window was created in a different location, the PUT may attempt an invalid location change. Maybe we should fetch the existing resource's location?

src/aks-preview/azext_aks_preview/maintenancewindow.py
```diff
@@ -0,0 +1,94 @@
...
+        resource_type=CUSTOM_MGMT_AKS_PREVIEW,
+        operation_group="maintenance_windows",
+    )
+    MaintenanceWindowResourceProperties = cmd.get_models(
+        "MaintenanceWindowResourceProperties",
+        resource_type=CUSTOM_MGMT_AKS_PREVIEW,
+        operation_group="maintenance_windows",
+    )
+
+    properties = MaintenanceWindowResourceProperties(
```
anushkasingh16, line 63: The constructed resource body never includes not_allowed_dates, existing blackout dates would get erased on schedule updates

src/aks-preview/azext_aks_preview/tests/latest/test_maintenancewindow.py
```diff
@@ -0,0 +1,178 @@
...
+        "week_index": None,
+        "duration_hours": None,
+        "utc_offset": None,
+        "start_date": None,
+        "start_time": None,
+        "no_wait": False,
+    }
+
+
+class TestConstructMaintenanceWindowResource(unittest.TestCase):
```
FumingZhang, line 52: glad to see these unit test cases. In addition, could we include an end-to-end scenario test as well?

src/aks-preview/azext_aks_preview/maintenancewindow.py
yewmsft: +1 — and the same applies to *any* future schedule fields beyond what `constructMaintenanceWindowResource` enumerates. The structural fix is either: (a) accept a `--not-allowed-dates` arg and add it to construct, OR (b) for an `update` semantics, refetch the existing MW, deep-merge user-supplied fields onto it, then PUT — which is the conventional CLI `update` shape (matches `az aks update`, `az aks nodepool update`, etc.). Option (b) also resolves @anushkasingh16's location-mismatch concern below in one go.

The current shape is functionally "replace" with a fast-path tags-only PATCH, named `update`. That's surprising — users expect `update` to be read-modify-write.

src/aks-preview/azext_aks_preview/custom.py
yewmsft: confirmed: ARM rejects PUTs that change the `location` of a TrackedResource (it's immutable post-create). Defaulting to RG location on update will hard-fail any MW that was created in a non-default region.

Two-line fix without going full read-modify-write: in `aks_maintenancewindow_update`, when `location is None`, fetch `client.get(rg, name).location` instead of `get_rg_location`. The fetch is one round-trip; cheap. Better still: combine with the not_allowed_dates fix and refactor `update` to read-modify-write end-to-end.

src/aks-preview/azext_aks_preview/commands.py
yewmsft: +1. `confirmation=True` in commands.py registration gives users a uniform `--yes` flag handled by the CLI core (which honors `AZURE_CORE_DISABLE_CONFIRM_PROMPT` and `--yes` consistently with every other AKS delete), while the handler-side `prompt_y_n` is hand-rolled. Drop the handler-side prompt and rely on `confirmation=True` — the `yes` parameter on the handler can also go away then.

yewmsft:
First-pass review on commit `00d5c5f1`.

Replied inline on 3 of @anushkasingh16's threads (not_allowed_dates erasure, location-immutability on PUT, double-confirmation on delete) — all real bugs, all share a root cause: the `update` handler is structurally a full PUT named `update`, where every CLI user (and our own `az aks update` / `az aks nodepool update` siblings) expects read-modify-write.

**Recommend:** refactor `aks_maintenancewindow_update` to `client.get()` → merge user-supplied fields onto the fetched body → `client.begin_create_or_update()`. This single change fixes all three concerns at once, matches the canonical CLI `update` semantics, and avoids any per-field deny-list maintenance (today not_allowed_dates; tomorrow whatever new MW field the RP adds).

One additional concern of my own:

- `_params.py:2712` — `--duration` help says "Range 4-24" but no validator enforces it. Server will reject with an opaque ARM 400; UX would be much better to fail at `argparse` time with a clear local error. Pattern: add a `validate_duration_hours` to `_validators.py` mirroring `validate_utc_offset` / `validate_start_date`. Same applies to any future bounded-int args (`interval_days`, etc. — check what the RP enforces and mirror locally where cheap).

Otherwise the CRUD shape, the `hasAnyScheduleArg` tags-vs-PUT routing, the `prompt_y_n` confirmation default-n, and the test coverage (7 happy-path cases via mocked sdk) are all the right shapes for an initial preview-only CLI command. No blocking concerns once the three-bug-one-fix refactor lands.

src/aks-preview/azext_aks_preview/_help.py
ibaboulfetouh: Done in fc075fb6. Rewrote the long-summary per your suggestion (drop the "Once available" framing in favor of a customer-focused description). New text:

> "MaintenanceWindow is a resource-group-scoped ARM resource that defines a reusable maintenance schedule. Users can link this resource to a pre-existing maintenanceConfiguration via `maintenanceWindowId`, so multiple managedCluster maintenance configurations share a single MaintenanceWindow and the schedule lives in one place."

src/aks-preview/azext_aks_preview/commands.py
ibaboulfetouh: Done in fc075fb6. Kept `confirmation=True` on the command registration (auto-injects `--yes/-y` via CLI core, respects `AZURE_CORE_DISABLE_CONFIRM_PROMPT`), dropped the handler-side `prompt_y_n` and the `yes` kwarg. Matches `aks delete` / `aks nodepool delete`. Verified with `az aks maintenancewindow delete --help` — single `--yes -y` arg now.

src/aks-preview/azext_aks_preview/custom.py
ibaboulfetouh: Done in fc075fb6 — subsumed by the RMW refactor (see thread 3392454224 for the framing). `aks_maintenancewindow_update` now does `client.get()` first and uses `existing.location` for the PUT body. If `--location` is supplied with a different value than the existing one, we warn and keep the existing location rather than letting ARM 400 the PUT.

Test: `test_location_mismatch_is_warned_not_erroring` pins this — supplying `--location eastus` on a `westus2` window emits a warning and the PUT body keeps `westus2`.

src/aks-preview/azext_aks_preview/maintenancewindow.py
ibaboulfetouh: Both addressed in fc075fb6. Two-part fix:

1. **`update` is now read-modify-write** — Ye's suggestion (thread 3398625216). The whole shape now matches `az aks update` / `az aks nodepool update` / `az containerapp update` (all do fetch → merge → PUT). After surveying the CLI, `update_tags()` is only used by `aks nodepool snapshot update` because snapshots have no other mutable fields; there's no precedent for the "PATCH if tags-only, PUT otherwise" branching shape we had. Refetch picks up `existing.properties.not_allowed_dates` for free, so any subset-of-fields update preserves blackout dates without per-field merge code. Also future-proofs against new MW properties the RP adds.

2. **`--config-file` support added** (MTC parity) — for the case where users do need to *change* `notAllowedDates`. Mirrors `aks maintenanceconfiguration add --config-file` exactly: JSON-supplied `properties` block replaces the existing `properties` wholesale on update. Help carries an example JSON with `notAllowedDates`.

Tests: `test_tags_only_update_preserves_schedule_and_dates` asserts the blackout dates are preserved on a tags-only update; `test_config_file_replaces_properties_keeps_tags_location` asserts the JSON path works end-to-end.

ibaboulfetouh:
Thanks for the structural framing — the RMW refactor lands all three of the cross-thread bugs (A2 double-confirm aside) with one change, and aligns the command shape with every other multi-field-mutable resource in the CLI. Posting the same response shape inline on each of the three threads you replied to, plus the new `--config-file` complement for the case where users do need to mutate `notAllowedDates`.

All five new tests (`TestAksMaintenancewindowUpdateRMW` × 9 + `TestAksMaintenancewindowCreateConfigFile` × 2 + `TestValidateDurationHours` × 8) plus the existing 8 construction tests pass locally (28 total). Sibling MTC scenario tests still pass with no regressions. Help rendering verified.

Y1 (`--duration` 4-24 range validator) landed too — added `validate_duration_hours` in `_validators.py` and wired into the MW arg block.

anushkasingh16:
LGTM!

src/aks-preview/azext_aks_preview/tests/latest/test_maintenancewindow.py
ibaboulfetouh: Added in the latest revision — `AksMaintenanceWindowScenarioTest::test_aks_maintenancewindow_crud` in this file. It's a recorded `ScenarioTest` that exercises the full lifecycle end-to-end against ARM: create (weekly schedule + tags) → show (asserts the schedule/tags round-trip) → update (`--duration` change, asserts the existing schedule/tags are preserved) → list → delete → show (expects 404). The cassette is committed alongside so it replays offline in CI.

ibaboulfetouh:
@yewmsft thanks — pushed a new revision that lands your recommendation plus the rest of the outstanding feedback. Summary of what changed since `00d5c5f1`:

**Your points**
- **RMW refactor (the three-bug-one-fix):** `aks_maintenancewindow_update` is now read-modify-write — `client.get()` → merge user-supplied fields onto the fetched body → `begin_create_or_update()`. This resolves all three threads at once (not_allowed_dates erasure, location-immutability on PUT, and the tags-vs-PUT routing) and matches the canonical `az aks update` / `az aks nodepool update` semantics, so there's no per-field deny-list to maintain.
- **`--duration` range:** added `validate_duration_hours` to `_validators.py` (mirroring `validate_utc_offset` / `validate_start_date`) and wired it on the param, so `4–24` now fails at argparse time with a clear local error instead of an opaque ARM 400.

**Anushka's points** (also addressed in the same pass)
- Dropped the double delete-confirmation (kept CLI-core `confirmation=True`, removed the handler-side `prompt_y_n`).
- `--not-allowed-dates` blackout dates are preserved through update (the RMW merge keeps existing `properties` unless a config file replaces them wholesale).

**E2E test (Fuming's ask)**
- Added `AksMaintenanceWindowScenarioTest::test_aks_maintenancewindow_crud` — a recorded scenario exercising create → show → update → list → delete → show(404) end-to-end against ARM, with the cassette committed for offline CI playback.

**Housekeeping**
- Rebased onto latest `main`.
- Fixed two `azdev linter` findings on the new command: the `--utc-offset` create examples now use the `=value` form (the space form parsed the leading `-` as a flag), and added the missing help entry for `aks maintenancewindow wait`.
- Moved the changelog entry into the `Pending` section of HISTORY.rst.

FumingZhang (dismissed):

FumingZhang:
Please fix failed CI checks, the case `test_aks_maintenancewindow` is still present in test_aks_commands.py, so if it can be replaced by the newly added test case `test_aks_maintenancewindow_crud`,remove it then. @ibaboulfetouh 

https://dev.azure.com/azclitools/public/_build/results?buildId=323409&view=logs&j=85692d73-7fd1-54cf-e8a2-b651dc664eb4&t=791c622c-dd9a-5515-3d48-9f66a91ff10c&l=5145

ibaboulfetouh:
Thanks @FumingZhang — root-caused it. The failure wasn't from a duplicate test; it was a regression introduced in this PR: an earlier revision deleted `recordings/test_aks_maintenancewindow.yaml` thinking it was orphaned, but that cassette actually belongs to the **pre-existing** `test_aks_maintenancewindow` ScenarioTest in `test_aks_commands.py` (a `maintenanceConfiguration` test whose name predates this new `maintenanceWindows` peer resource). With its recording gone, that test had nothing to play back and failed in CI.

I've **restored its cassette from `main`** so it plays back green again (verified locally). That keeps the existing `maintenanceConfiguration` coverage intact and is the minimal change to unblock CI.

The genuinely new test this PR adds is `test_aks_maintenancewindow_crud` in `test_maintenancewindow.py`, which has its own cassette and is unaffected. Both pass in playback now. Pushed in `4b772f53e`. PTAL 🙏

FumingZhang (approved):

yanzhudd (approved):

PR #10039, opened 2026-06-24
{CI} Fix update_index wheel filename + bump pinned azdev to latest

aryasadeghi1:
Two follow-ups to #9922 so the decoupled (pkginfo) azdev path works end to end:

- scripts/ci/update_index.py saved the downloaded wheel as just the extension name, but pkginfo.Wheel needs a .whl-named file, so the publish index update failed with "Not a known wheel archive format". Use the real filename from the URL (as the entry already does).

- Bump the pinned azdev (azdev_setup.yml + azure-pipelines.yml) 49390f3 -> 644d296, the current dev commit that includes azdev's matching `extension update-index` fix. Stays pinned until azdev 0.2.11 is published to PyPI.

### DIFF
--- a/.azure-pipelines/templates/azdev_setup.yml
+++ b/.azure-pipelines/templates/azdev_setup.yml
@@ -11,14 +11,9 @@ steps:
       python -m venv env
       chmod +x env/bin/activate
       source ./env/bin/activate
-      # clone azure-cli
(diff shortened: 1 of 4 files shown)


yonzhan:
CI

necusjz (dismissed):

necusjz (approved):

PR #10056, opened 2026-06-30
[Migrate] protected item id to target object id (CUSTOMER REPORTED BUG)

saifaldin14:
For the migrate command we need to rename the protected item id parameter to target object id


yonzhan:
Migrate

necusjz (approved):

PR #10059, opened 2026-07-01
[documentdb] Add DocumentDB (Mongo vCore) CLI extension

amatarritamicrosoft:
## Summary

Adds a new Azure CLI **extension** `documentdb` for **Azure Cosmos DB for MongoDB (vCore)** — management-plane commands for `Microsoft.DocumentDB/mongoClusters` (api-version `2026-06-01`).

Generated with the AAZ codegen tool. Companion command-model PR in Azure/aaz: **Azure/aaz#1040**.

## Command tree

- `az documentdb mongocluster` create / update / show / delete / list / list-connection-strings / check-name-availability / wait
- `az documentdb mongocluster firewall-rule` create / update / show / delete / list / wait
- `az documentdb mongocluster identity` assign / remove / show / wait (UserAssigned only)
- `az documentdb mongocluster replica` list / promote
- `az documentdb mongocluster user` create / update / show / delete / list / wait

## Design notes

- **Flattened scalar arguments** on create/update (no JSON blobs): `--admin-user`/`-u`, `--admin-password`/`-p`, `--tier`, `--storage-size`, `--storage-type`, `--shard-count`, `--high-availability`, `--data-api-mode`, `--auth-allowed-modes`. Core provisioning args required on create, optional on update.
- `user create`/`update` accept a repeatable `--role` (singular option) and a custom `--type`/`-t` flag mapping to the `identityProvider` discriminated union (custom.py).
- Private-endpoint / private-link commands intentionally omitted (covered by `az network` per current guidance).
- New `documentdb` entry added to `src/service_name.json`.

## Dependency (why this is a draft)

`az documentdb` is currently reserved/blocked in `azure-cli` core. This extension cannot be exercised until the core unblock ships in a released `azure-cli`:
- Core unblock PR: **Azure/azure-cli#33686**

Marking ready for review once that lands.

## Testing

- Live-tested end-to-end against Azure (create/update/delete of real mongoClusters, flattened flags, `--role`, `--type`).
- Command models pass `aaz-dev command-model verify`.

---

- [x] The PR title and description conform to the repository guidelines.
- [x] I have read the [Authoring Command Modules](https://github.com/Azure/azure-cli/tree/dev/doc/authoring_command_modules) doc.
- [x] The code generated follows the [Command Guidelines](https://github.com/Azure/azure-cli/blob/dev/doc/command_guidelines.md).


yonzhan:
documentdb

src/documentdb/azext_documentdb/tests/latest/test_documentdb.py
```diff
@@ -5,6 +5,9 @@
...
 # --------------------------------------------------------------------------------------------
 
-# pylint: skip-file
-# flake8: noqa
+from azure.cli.testsdk import *
 
+
+class DocumentdbScenario(ScenarioTest):
+    # TODO: add tests here
+    pass
```
necusjz: would mind add some real test cases?

src/documentdb/README.md
```diff
@@ -0,0 +1,5 @@
+# Azure CLI Documentdb Extension #
+This is an extension to Azure CLI to manage Documentdb resources.
+
+## How to use ##
+Please add commands usage here.
```
necusjz: better to enrich that part.

necusjz (approved):

src/documentdb/azext_documentdb/custom.py
```diff
@@ -0,0 +1,277 @@
...
+    # point the CLI parser uses); ``_args_schema`` is the internal build cache.
+    _arguments_schema = None
+    _args_schema = None
+
+    @classmethod
+    def _build_arguments_schema(cls, *args, **kwargs):
+        args_schema = super()._build_arguments_schema(*args, **kwargs)
+        _keep_only_args(args_schema, {"cluster_name", "resource_group", "admin_password"})
+        password = args_schema.admin_password
+        password._options = ["--password", "-p"]
```
olivertowers, line 117: Argument name consistency between this command (--password, -p) and what the cluster create/update accepts( --admin-password), let's add the same set of aliases on both sides so that it is consistent.

src/documentdb/azext_documentdb/commands.py
```diff
@@ -12,4 +12,10 @@
...
 
 def load_command_table(self, _):  # pylint: disable=unused-argument
-    pass
+    from azext_documentdb.custom import (
+        UserCreate, UserUpdate, ResetPassword, ReplicaCreate, Restore)
+    self.command_table['documentdb mongocluster user create'] = UserCreate(loader=self)
+    self.command_table['documentdb mongocluster user update'] = UserUpdate(loader=self)
+    self.command_table['documentdb mongocluster reset-password'] = ResetPassword(loader=self)
+    self.command_table['documentdb mongocluster replica create'] = ReplicaCreate(loader=self)
+    self.command_table['documentdb mongocluster restore'] = Restore(loader=self)
```
olivertowers, line 21: When I looked into this - the promote action (POST API call) handling that AAZ generated had incorrect deserialization mapping where it would hit a None type error in python that required a workaround (python sdk did the right thing, but AAZ made an assumption that was not correct based on Typespec).

I just want to confirm that this this didn't repro in the newer AAZ code gen or if there was something else you did to fix this.

src/documentdb/azext_documentdb/tests/latest/test_documentdb.py
```diff
@@ -0,0 +1,123 @@
...
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+import unittest
+
+from azure.cli.testsdk import ScenarioTest, ResourceGroupPreparer
+from azure.cli.testsdk.scenario_tests.decorators import AllowLargeResponse
+
+
+class DocumentdbScenario(ScenarioTest):
```
olivertowers, line 12: I don't see tests for restore, replica, CMK, users etc.

src/documentdb/azext_documentdb/commands.py
```diff
@@ -12,4 +12,10 @@
 
 
 def load_command_table(self, _):  # pylint: disable=unused-argument
-    pass
+    from azext_documentdb.custom import (
+        UserCreate, UserUpdate, ResetPassword, ReplicaCreate, Restore)
+    self.command_table['documentdb mongocluster user create'] = UserCreate(loader=self)
+    self.command_table['documentdb mongocluster user update'] = UserUpdate(loader=self)
+    self.command_table['documentdb mongocluster reset-password'] = ResetPassword(loader=self)
+    self.command_table['documentdb mongocluster replica create'] = ReplicaCreate(loader=self)
```
olivertowers, line 20: Should we add a `wait` commands for these sub-command groups?
 
don't know if the standard pattern is that for every sub-command that supports a LRO command, there should be a corresponding 'wait' command eg.

```bash
az documendb mongocluster replica create .... -nowait
....<do some other stuff>
az documentdb mongocluster replica wait
````

Comment (olivertowers) [reply src/documentdb/azext_documentdb/commands.py:20]: Perhaps this is already generated somewhere I just couldn't see it - I did think the liniter/style checks would have failed on this.

**Review by olivertowers — COMMENTED** [review]
Some things we need to fix up 

Key gap - we need to add testing for replica/restore/CMK (user managed identity) to validate the command set fully.

File: src/documentdb/azext_documentdb/aaz/latest/documentdb/mongocluster/firewall_rule/_create.py [hunk 7]
```diff
@@ -9,22 +9,22 @@
 # flake8: noqa
 
 from azure.cli.core.aaz import *
-from azext_workload_orchestration_preview.aaz.latest.workload_orchestration._resource_validator import ValidateResourceExists
 
 
 @register_command(
-    "workload-orchestration context site-reference create",
+    "documentdb mongocluster firewall-rule create",
```
olivertowers, line 15: Did we reach agreement that we would use `firewall-rule` instead of the AZ cli standard `network rule`?

amatarritamicrosoft: Abinav Rameesh review the tree command and there was not issue from his side, I will share the doc why is better to use firewall-rule instead of network rule

src/documentdb/azext_documentdb/commands.py
amatarritamicrosoft: Confirmed, it does not reproduce. In the current codegen, promote is modeled as a location based LRO with only a 202 response and no success body schema. AAZ therefore generates no response deserialization (no  on_200  /  _schema_on_200 ; the LRO result callback is  None ), so the previous None type path no longer exists. It is fully generated code with no manual workaround.

src/documentdb/azext_documentdb/commands.py
amatarritamicrosoft: Wait commands are generated for most of the sub groups:  mongocluster ,  mongocluster user ,  mongocluster firewall-rule  and  mongocluster identity  all expose a  wait . The one group without its own  wait  is  replica , and that is expected. A replica is itself a  Microsoft.DocumentDB/mongoClusters  resource, the same resource type as the primary, so it is already covered by the top level  az documentdb mongocluster wait -n <replica> -g <rg> . I verified that  mongocluster wait --exists  returns 0 against a cluster, and that both  replica create  and  replica promote  expose  --no-wait . A dedicated  replica wait  would just duplicate the mongocluster one. On the linter: the CLI linter and style checks do not require a  wait  command for every LRO command, so no failure is expected here.

yonzhan:
documentdb

PR #10152, opened 2026-07-28
[appnet-preview] Auto-detect --member-location from member cluster on member join

jack4it:

| Breaking Changes |
| :--: |
| ️✔️ None |

### Summary
Make `--member-location` optional on `az appnet member join`. When omitted, it now defaults to the **location of the member cluster** referenced by `--member-resource-id`, removing a redundant region input that had to be kept in sync with the cluster by hand. An explicit `--member-location` is still honored; a warning is logged if it differs from the cluster's region.

### Changes
- `custom.py` – `MemberJoin` overrides `_build_arguments_schema` to make `--member-location` optional and strip the resource-group location auto-fill (so an omitted value reaches the detection logic instead of silently defaulting to the RG location). `pre_operations` resolves the location from the cluster via a generic ARM GET on `--member-resource-id`.
- Existing upgrade-mode validation in `pre_operations` is preserved.
- New unit tests (`tests/latest/test_member_location.py`) covering default / explicit-mismatch-warning / missing-both-raises / invalid-resource-id paths. No live ARM calls.
- Version bump `1.0.0b2` -> `1.0.0b3` and `HISTORY.rst` entry.

### Notes
- The AAZ-generated `_join.py` is intentionally **not** edited; the behavior change is applied through the `MemberJoin` command subclass in `custom.py`.
- `--member-resource-id` remains optional at the CLI layer, but if both `--member-location` and a resolvable `--member-resource-id` are absent, the command now fails fast with `RequiredArgumentMissingError` rather than silently using the resource group's location.

Drafted with GitHub Copilot CLI.

### DIFF
--- a/src/appnet-preview/azext_appnet_preview/custom.py
+++ b/src/appnet-preview/azext_appnet_preview/custom.py
@@ -16,6 +16,7 @@ from azure.cli.core.azclierror import ArgumentUsageError, RequiredArgumentMissin
 from azure.cli.core.commands.client_factory import get_mgmt_service_client
 from azure.cli.core.profiles import ResourceType
 from azure.mgmt.core.tools import is_valid_resource_id
+from azure.core.exceptions import HttpResponseError
 
 logger = get_logger(__name__)
 
@@ -85,5 +86,13 @@ class MemberJoin(Join):
         if not is_valid_resource_id(resource_id):
             return None
         client = get_mgmt_service_client(self.cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES)
-        cluster = client.resources.get_by_id(resource_id, MANAGED_CLUSTER_API_VERSION)
+        try:
+            cluster = client.resources.get_by_id(resource_id, MANAGED_CLUSTER_API_VERSION)
+        except HttpResponseError as ex:
+            # Reading the cluster is best-effort: if it can't be reached (404/403/etc.),
+            # treat the location as unavailable so an explicitly provided --member-location
+            # still works, and the auto-detect path surfaces the actionable
+            # RequiredArgumentMissingError instead of a raw ARM error.
+            logger.debug("Could not read location from member cluster '%s': %s", resource_id, ex)
+            return None
         return cluster.location
--- a/src/appnet-preview/azext_appnet_preview/custom.py
+++ b/src/appnet-preview/azext_appnet_preview/custom.py
@@ -7,16 +7,41 @@
 
 # pylint: disable=too-many-lines
 # pylint: disable=too-many-statements
+# pylint: disable=protected-access
 
 from knack.log import get_logger
 from .aaz.latest.appnet.member._join import Join
-from azure.cli.core.aaz import has_value
-from azure.cli.core.azclierror import ArgumentUsageError
+from azure.cli.core.aaz import has_value, AAZResourceLocationArgFormat
+from azure.cli.core.azclierror import ArgumentUsageError, RequiredArgumentMissingError
+from azure.cli.core.commands.client_factory import get_mgmt_service_client
+from azure.cli.core.profiles import ResourceType
+from azure.mgmt.core.tools import is_valid_resource_id
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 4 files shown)


german1608 (approved):

yonzhan:
appnet-preview

src/appnet-preview/azext_appnet_preview/custom.py
```diff
@@ -28,3 +53,37 @@ def pre_operations(self):
...
+
+    def _get_member_cluster_location(self, args):
+        if not has_value(args.member_resource_id):
+            return None
+        resource_id = args.member_resource_id.to_serialized_data()
+        if not is_valid_resource_id(resource_id):
+            return None
+        client = get_mgmt_service_client(self.cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES)
+        cluster = client.resources.get_by_id(resource_id, MANAGED_CLUSTER_API_VERSION)
+        return cluster.location
```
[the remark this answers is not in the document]
jack4it: Good catch — fixed in 433fe76a. `_get_member_cluster_location` now catches `HttpResponseError` and returns `None`, so an explicit `--member-location` is honored when the cluster is unreadable and the auto-detect path still raises the intended `RequiredArgumentMissingError`. Added two covering tests.

src/appnet-preview/azext_appnet_preview/custom.py
```diff
@@ -28,3 +54,45 @@ def pre_operations(self):
...
+    def _get_member_cluster_location(self, args):
+        if not has_value(args.member_resource_id):
+            return None
+        resource_id = args.member_resource_id.to_serialized_data()
+        if not is_valid_resource_id(resource_id):
+            return None
+        client = get_mgmt_service_client(self.cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES)
+        try:
+            cluster = client.resources.get_by_id(resource_id, MANAGED_CLUSTER_API_VERSION)
+        except HttpResponseError as ex:
```
deveshdama: `HttpResponseError` catches 404/403, but `ServiceRequestError`/`ServiceResponseError` subclass `AzureError` directly, so connection failures and timeouts still escape and an explicit `--member-location` aborts anyway. `except AzureError` covers it.

src/appnet-preview/azext_appnet_preview/custom.py
```diff
@@ -7,16 +7,42 @@
...
 class MemberJoin(Join):
+
+    @classmethod
+    def _build_arguments_schema(cls, *args, **kwargs):
+        args_schema = super()._build_arguments_schema(*args, **kwargs)
+        # --member-location is derived from the member cluster (--member-resource-id) when
+        # not supplied, so it is no longer required on the command line. Replace the format
+        # so AAZ does NOT auto-fill it from the resource group's location before
+        # pre_operations runs; an empty value must reach our detection logic below.
+        args_schema.member_location._required = False
```
deveshdama, line 38: `AAZResourceLocationArg.__init__` also sets `configured_default='location'`, which neither of these overrides clears. Anyone with `az configure --defaults location=...` set gets that value instead, so the auto-detect silently never runs for them. Needs `_configured_default = None` too.

src/appnet-preview/azext_appnet_preview/custom.py
jack4it: Fixed in 75b6593b — broadened to `except AzureError` so `ServiceRequestError`/`ServiceResponseError` (connection/timeout) are also treated as location-unavailable, not just 404/403. Added a transport-failure test (`test_arm_transport_failure_returns_none`).

src/appnet-preview/azext_appnet_preview/custom.py
jack4it: Fixed in 75b6593b — added `member_location._configured_default = None` in `_build_arguments_schema` so `az configure --defaults location=...` no longer preempts auto-detect. Added a schema assertion covering it.

deveshdama (approved):
LGTM!

necusjz (approved):

PR #10161, opened 2026-07-30
Quantum: Add new update job operation

v-elegacheva:

| Breaking Changes |
| :--: |
| ⚠️ None |

<details>

><details>
>
>|---|---|---|---|
>|⚠️ [1001 - CmdAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1001.md)|quantum job update|cmd `quantum job update` added||
>
></details>
</details>

## Summary

Adds a new `az quantum job update` command that lets users update a submitted job's name, priority, and/or tags after submission. This implements the CLI side of the merged Data Plane API Spec. Only the fields that user provides are sent in the patch, `--job-priority` is validated against the `Priority` enum, at least one updatable field is required, and the refreshed job is returned. Note that `--job-tags` replaces the job's existing tags rather than appending.

## What Changed

- Added the `az quantum job update` command, wired up with the workspace validator and job table transformer.
- Implemented the update operation using the data-plane jobs client's merge-patch API, which builds the patch from only the provided fields, validated priority, requires at least one updatable field, and returns the refreshed job.
- Added the `--job-priority` and `--job-tags` parameters
- Added command help with examples for renaming, changing priority, and replacing tags.
- Bumped the extension version and updated the changelog.
- Added unit and live scenario test coverage for the new command and its parameters.

### DIFF
--- a/src/quantum/azext_quantum/_params.py
+++ b/src/quantum/azext_quantum/_params.py
@@ -39,8 +39,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
     job_name_type = CLIArgumentType(help='A friendly name to give to this run of the program.')
     job_name_filter_type = CLIArgumentType(help='Job name to be listed (search by prefix), example "My Job".')
     job_id_type = CLIArgumentType(options_list=['--job-id', '-j'], help='Job unique identifier in GUID format.')
-    job_priority_type = CLIArgumentType(options_list=['--job-priority'], help='The updated priority of the job. Allowed values: Standard, High.')
-    job_tags_type = CLIArgumentType(options_list=['--job-tags'], help='Space-separated list of tags to associate with the job. Replaces any existing tags.', nargs='+')
+    job_name_update_type = CLIArgumentType(options_list=['--job-name'], help='The updated friendly name of the job.')
+    job_priority_type = CLIArgumentType(options_list=['--job-priority'], help='The updated priority of the job.', arg_type=get_enum_type(['Standard', 'High']))
+    job_tags_type = CLIArgumentType(options_list=['--job-tags'], help='Space-separated list of tags to associate with the job. Replaces any existing tags. Pass "" to clear all tags.', nargs='+')
     job_params_type = CLIArgumentType(options_list=['--job-params'], help='Job parameters passed to the target as a list of key=value pairs, json string, or `@{file}` with json content.', action=JobParamsAction, nargs='+')
     target_capability_type = CLIArgumentType(options_list=['--target-capability'], help='Target-capability parameter passed to the compiler.')
     shots_type = CLIArgumentType(help='The number of times to run the program on the given target.')
@@ -133,7 +134,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
         c.argument('entry_point', entry_point_type)
 
     with self.argument_context('quantum job update') as c:
-        c.argument('job_name', options_list=['--job-name'], help='The updated friendly name of the job.')
+        c.argument('job_name', job_name_update_type)
         c.argument('job_priority', job_priority_type)
         c.argument('job_tags', job_tags_type)
 
--- a/src/quantum/azext_quantum/_params.py
+++ b/src/quantum/azext_quantum/_params.py
@@ -39,6 +39,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
     job_name_type = CLIArgumentType(help='A friendly name to give to this run of the program.')
     job_name_filter_type = CLIArgumentType(help='Job name to be listed (search by prefix), example "My Job".')
     job_id_type = CLIArgumentType(options_list=['--job-id', '-j'], help='Job unique identifier in GUID format.')
+    job_priority_type = CLIArgumentType(options_list=['--job-priority'], help='The updated priority of the job. Allowed values: Standard, High.')
+    job_tags_type = CLIArgumentType(options_list=['--job-tags'], help='Space-separated list of tags to associate with the job. Replaces any existing tags.', nargs='+')
     job_params_type = CLIArgumentType(options_list=['--job-params'], help='Job parameters passed to the target as a list of key=value pairs, json string, or `@{file}` with json content.', action=JobParamsAction, nargs='+')
     target_capability_type = CLIArgumentType(options_list=['--target-capability'], help='Target-capability parameter passed to the compiler.')
     shots_type = CLIArgumentType(help='The number of times to run the program on the given target.')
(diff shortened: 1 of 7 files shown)


src/quantum/azext_quantum/operations/job.py
```diff
@@ -462,6 +462,34 @@ def delete(cmd, job_id, resource_group_name, workspace_name):
...
 
+def job_update(cmd, job_id, resource_group_name, workspace_name, job_name=None, job_priority=None, job_tags=None):
+    """
+    Update a submitted job's name, priority, and/or tags.
+    """
+    info = WorkspaceInfo(cmd, resource_group_name, workspace_name)
+    client = cf_jobs(cmd.cli_ctx, info.subscription, info.resource_group, info.name, info.endpoint)
+
+    update_options = {}
+    if job_name is not None:
```
rigidit: I think we should check also for empty and whitespace strings

src/quantum/azext_quantum/operations/job.py
```diff
@@ -462,6 +462,34 @@ def delete(cmd, job_id, resource_group_name, workspace_name):
...
+            update_options["priority"] = Priority(job_priority).value
+        except ValueError:
+            raise InvalidArgumentValueError(ERROR_MSG_INVALID_PRIORITY_ARGUMENT)
+    if job_tags is not None:
+        update_options["tags"] = job_tags
+
+    if not update_options:
+        raise RequiredArgumentMissingError("At least one of --job-name, --job-priority, or --job-tags must be specified.")
+
+    client.update(info.subscription, info.resource_group, info.name, job_id, update_options)
```
rigidit, line 490: is this client.update return updated job? if yes, we could skip client.get later

src/quantum/azext_quantum/operations/job.py
```diff
@@ -462,6 +462,34 @@ def delete(cmd, job_id, resource_group_name, workspace_name):
...
+
+    update_options = {}
+    if job_name is not None:
+        update_options["name"] = job_name
+    if job_priority is not None:
+        try:
+            update_options["priority"] = Priority(job_priority).value
+        except ValueError:
+            raise InvalidArgumentValueError(ERROR_MSG_INVALID_PRIORITY_ARGUMENT)
+    if job_tags is not None:
```
rigidit, line 480: let's maybe better validate tags, I think they should be comma separated strings right?

rigidit (changes requested):

src/quantum/azext_quantum/operations/job.py
v-elegacheva: no, client.update returns only the JobUpdateOptions that were sent, not the full job. That is why I follow it with client.get(...) and return that so the user sees the complete, current job state after the patch. I added a comment making this explicit.

src/quantum/azext_quantum/operations/job.py
v-elegacheva: it looks like --job-tags uses nargs='+', so it takes a space separated list. It looks like that is the standard Azure CLI convention for a list[str] and matches the SDK's tags:list[str]. I added a validation to trim each tag, drop blanks, and reject an all-empty/whitespce list. I can switch them to comma separated if prefered though

src/quantum/azext_quantum/operations/job.py
rigidit: I would use more pythonic way and put in one line - easier to read:
`if job_name and job_name.strip():`

src/quantum/azext_quantum/operations/job.py
```diff
@@ -462,6 +462,39 @@ def delete(cmd, job_id, resource_group_name, workspace_name):
...
+def job_update(cmd, job_id, resource_group_name, workspace_name, job_name=None, job_priority=None, job_tags=None):
+    """
+    Update a submitted job's name, priority, and/or tags.
+    """
+    info = WorkspaceInfo(cmd, resource_group_name, workspace_name)
+    client = cf_jobs(cmd.cli_ctx, info.subscription, info.resource_group, info.name, info.endpoint)
+
+    update_options = {}
+    if job_name is not None:
+        if not job_name.strip():
```
rigidit: I would use more pythonic way and put in one line - easier to read:
`if job_name and job_name.strip():`

src/quantum/azext_quantum/_params.py
```diff
@@ -39,6 +39,8 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
     job_name_type = CLIArgumentType(help='A friendly name to give to this run of the program.')
     job_name_filter_type = CLIArgumentType(help='Job name to be listed (search by prefix), example "My Job".')
     job_id_type = CLIArgumentType(options_list=['--job-id', '-j'], help='Job unique identifier in GUID format.')
+    job_priority_type = CLIArgumentType(options_list=['--job-priority'], help='The updated priority of the job. Allowed values: Standard, High.')
```
rigidit: Azure CLI authoring guidelines require enum-backed parameters to be declared with get_enum_type so that --help renders allowed values, tab-completion works, and the CLI core rejects bad input before the handler runs (giving a consistent error format)

src/quantum/azext_quantum/operations/job.py
```diff
@@ -462,6 +462,39 @@ def delete(cmd, job_id, resource_group_name, workspace_name):
     logger.warning("Deleted job %s.", job_id)
 
 
+def job_update(cmd, job_id, resource_group_name, workspace_name, job_name=None, job_priority=None, job_tags=None):
```
rigidit: can we use "update" name here instead of "job_update" to be consistent with other commands like delete cancel etc?

src/quantum/azext_quantum/operations/job.py
rigidit: let's keep space separated list, but I'm wondering if we want a way to clear tags?

src/quantum/azext_quantum/_params.py
```diff
@@ -130,6 +132,11 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
         c.argument('job_output_format', job_output_format_type)
         c.argument('entry_point', entry_point_type)
 
+    with self.argument_context('quantum job update') as c:
+        c.argument('job_name', options_list=['--job-name'], help='The updated friendly name of the job.')
```
rigidit: nit: other arguments in the file use a module-level CLIArgumentType. Add a job_name_update_type for consistency

rigidit (approved):

v-elegacheva:
hello @necusjz! Could you please review this PR when you have a chance? Thank you for your time :)

necusjz (approved):

PR #10186, opened 2026-08-06
{AKS} Add FlexNodes support

wenxuan0923:
### 🤖 PR Validation — ❌ Action needed

| Breaking Changes |
| :--: |
| ❌ 1 |

<details>
<summary>❌Azure CLI Extensions Breaking Change Test</summary>

><details>
> <summary>❌aks-preview</summary>
>
>|---|---|---|---|
>|❌ [1010 - ParaPropUpdate](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1010.md) |aks machine add|cmd `aks machine add` update parameter `spot_max_price`: updated property `default` from `nan` to `nan`|please change property `default` from `nan` to `nan` for parameter `spot_max_price` of cmd `aks machine add`|
>|⚠️ [1006 - ParaAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1006.md)|aks machine add|cmd `aks machine add` added parameter `labels`||
>|⚠️ [1006 - ParaAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1006.md)|aks machine add|cmd `aks machine add` added parameter `max_pods`||
>|⚠️ [1006 - ParaAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1006.md)|aks machine add|cmd `aks machine add` added parameter `node_taints`||
>|⚠️ [1006 - ParaAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1006.md)|aks machine update|cmd `aks machine update` added parameter `kubernetes_version`||
>|⚠️ [1001 - CmdAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1001.md)|aks nodepool get-bootstrap-data|cmd `aks nodepool get-bootstrap-data` added||
>
></details>
</details>

## Summary
- add FlexNodes node pool lifecycle support, including bootstrap data and Kubernetes upgrades
- add FlexNode Machine create and update support with minimal supported payloads
- validate operation-specific options while preserving existing regular Machines behavior
- redact bootstrap credentials in recordings and release aks-preview 21.0.0b14

## Testing
- FlexNodes live lifecycle in eastus2euap: passed
- refreshed cassette replay: passed
- affected unit tests: passed
- full aks-preview suite: 1177 passed, 119 skipped; 9 existing cassette transport failures reproduced on origin/main
- existing regular Machines recording tests: passed
- azdev style aks-preview: passed
- index tests: 7 passed, 2 skipped
- aks-preview 21.0.0b14 wheel build: passed


yonzhan:
AKS

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -3254,11 +3300,37 @@ def aks_machine_add(
...
+                raw_parameters,
+                {
+                    "kubernetes_version": "--kubernetes-version",
+                    "labels": "--labels",
+                    "max_pods": "--max-pods",
+                    "node_taints": "--node-taints",
+                },
+            )
+            return add_flexnode_machine(cmd, client, raw_parameters, no_wait)
+
```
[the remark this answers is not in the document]
wenxuan0923: Fixed in fb14a8ebf. Machine add now always resolves the parent pool type before choosing the payload path, so --vm-size on a FlexNodes pool is rejected by the FlexNodes option validator instead of falling through to the regular Machine payload. Added unit coverage and refreshed/replayed both affected regular Machines recordings.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
```diff
@@ -2800,6 +2800,185 @@ def test_aks_machines_nodepool(self, resource_group, resource_group_location):
             checks=[self.is_empty()],
         )
 
+    @AllowLargeResponse()
+    @AKSCustomResourceGroupPreparer(
+        random_name_length=17, name_prefix="clitest", location="eastus2euap"
+    )
+    def test_aks_flexnodes_supported_operations(self, resource_group, resource_group_location):
```
FumingZhang, line 2832: Queued [live test](https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=175407816&view=results) to validate the change.

src/aks-preview/azext_aks_preview/custom.py
```diff
@@ -2579,7 +2597,11 @@ def aks_agentpool_upgrade(cmd,
         instance.upgrade_settings.max_blocked_nodes = max_blocked_nodes
 
     # Blue-green upgrade settings
-    if not instance.upgrade_settings_blue_green:
+    blue_green_settings = (drain_batch_size, drain_timeout_bg, batch_soak_duration, final_soak_duration)
```
FumingZhang, line 2604: Is this still necessary? I believe the validation at line 2535 already rejects all parameters unsupported by FlexNodes.

src/aks-preview/azext_aks_preview/machine.py
```diff
@@ -23,6 +27,82 @@ def parse_key_value_list(pairs):
     return result
 
 
+def construct_flexnode_machine(cmd, raw_parameters, existed_machine=None):
```
FumingZhang, line 30: Since `machine_name` is optional, could we validate it here? Otherwise, the error is only raised when the request is sent.

FumingZhang (commented):
My overall concern is that this approach has become fragile. Since FlexNodes supports only a small subset of existing node pool and machine options, integrating it into the existing command groups requires extensive validation, payload filtering, and special-case branching to prevent unsupported parameters from being used.

Would a dedicated FlexNodes subcommand group provide a clearer and more maintainable interface? It could expose only supported operations and parameters, reduce the risk of accidentally sending incompatible fields, produce clearer help and error messages, and avoid repeatedly updating exclusion lists whenever existing commands gain new options.

I understand that reusing existing commands offers consistency and may preserve the expected CLI resource model, but I think the long-term maintenance and compatibility trade-offs are worth considering before expanding this pattern further.

src/aks-preview/azext_aks_preview/machine.py
```diff
@@ -23,6 +27,82 @@ def parse_key_value_list(pairs):
...
+            if raw_parameters.get("kubernetes_version") is not None
+            else existing_kubernetes.orchestrator_version
+            if existing_kubernetes is not None
+            else None
+        ),
+        max_pods=(
+            existing_kubernetes.max_pods
+            if existing_kubernetes is not None
+            else raw_parameters.get("max_pods")
+        ),
```
elvazhu521: max_pods here uses existing first, which is the opposite of the
orchestrator_version block right above it. But maxPods is an updatable field, so a
user-supplied value would be silently dropped on the update path
https://github.com/Azure/azure-rest-api-specs/blob/e1e13661cd75b44432074d3363952943f68df861/specification/containerservice/resource-manager/Microsoft.ContainerService/aks/preview/2026-05-02-preview/managedClusters.json#L10740-L10748 no readonly flag

src/aks-preview/azext_aks_preview/custom.py
wenxuan0923: Yes this guard is still necessary because validation only rejects explicitly supplied unsupported options. The generic path below **backfills** AgentPoolBlueGreenUpgradeSettings() whenever the GET response has no blue-green profile, which would otherwise add unsupported upgradeSettingsBlueGreen: {} to a FlexNodes PUT even when the user supplied no blue-green options.

src/aks-preview/azext_aks_preview/machine.py
wenxuan0923: Fixed in 472ff2ca7. construct_flexnode_machine now raises RequiredArgumentMissingError when machine_name is omitted, before constructing or sending the PUT payload. Added targeted unit coverage.

src/aks-preview/azext_aks_preview/tests/latest/test_aks_commands.py
wenxuan0923: The gate failed because its alternate subscription had AKSFlexNodePreview unregistered and overrode the scenario location to westus2. Fixed in 472ff2ca7: the scenario now idempotently registers and polls both required features, re-registers Microsoft.ContainerService to propagate them, and preserves eastus2euap. The updated live test and cassette replay both passed.

Queued a new live test here: https://dev.azure.com/msazure/CloudNativeCompute/_build/results?buildId=175510820&view=results

src/aks-preview/azext_aks_preview/machine.py
wenxuan0923: For FlexNode Machines, maxPods is intentionally create-only even though the shared Swagger model does not mark it readonly. (it is also the case for regular VMSS/VMs pool or machine, maxPods is immutable after create) `az aks machine update` does not register a `--max-pods` argument, and the update test verifies the existing value is preserved. I clarified this contract in 537f6d58d by splitting the create/update handling and documenting the preservation behavior.

wenxuan0923:
> My overall concern is that this approach has become fragile. Since FlexNodes supports only a small subset of existing node pool and machine options, integrating it into the existing command groups requires extensive validation, payload filtering, and special-case branching to prevent unsupported parameters from being used.
> 
> Would a dedicated FlexNodes subcommand group provide a clearer and more maintainable interface? It could expose only supported operations and parameters, reduce the risk of accidentally sending incompatible fields, produce clearer help and error messages, and avoid repeatedly updating exclusion lists whenever existing commands gain new options.
> 
> I understand that reusing existing commands offers consistency and may preserve the expected CLI resource model, but I think the long-term maintenance and compatibility trade-offs are worth considering before expanding this pattern further.

Thanks  @FumingZhang, That's a good point. We considered a nested FlexNodes group, but the current command shape follows the approved PRD to reuse the existing `az aks nodepool and `az aks machine` commands.

The supported surface is small today because this is the initial preview. We expect it to grow as we work toward feature parity, so a dedicated subgroup may become mostly duplicated commands and parameters over time.

The implementation is fail-closed: each operation has a positive allow-list, and payload builders emit only supported fields. New options are rejected automatically until we explicitly add them to the FlexNodes allow-list, so future features added to the existing commands are less likely to accidentally flow into FlexNodes requests.

I agree we should keep consolidating the internal routing rather than adding scattered special cases. Is the subgroup a blocker for this PR, or something we can revisit if the resource model diverges further?

FumingZhang:
> I agree we should keep consolidating the internal routing rather than adding scattered special cases. Is the subgroup a blocker for this PR, or something we can revisit if the resource model diverges further?

The subgroup isn’t a blocker—just my two cents. Since this is a preview extension, it’s also a good opportunity to gather customer feedback before deciding whether it should eventually move to a subgroup.

FumingZhang (dismissed):

wenxuan0923:
Hi @yanzhudd could you help merge this PR? Many thanks!

yanzhudd:
Hi @wenxuan0923 
could you please resolve the code conflicts?

wenxuan0923:
> Hi Thalia Wang (@wenxuan0923) could you please resolve the code conflicts?

Hi @yanzhudd!  Yes, I have resolved the conflicts, may I have review/approval please? Thank you!

FumingZhang (approved):

yanzhudd (approved):

PR #10194, opened 2026-08-07
Quantum Add 'az quantum suite-offers target list' to list provider-account targets and status

v-elegacheva:

| Breaking Changes |
| :--: |
| ⚠️ None |

<details>

><details>
>
>|---|---|---|---|
>|⚠️ [1011 - SubgroupAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1011.md)|quantum suite-offers|sub group `quantum suite-offers` added||
>
></details>
</details>

## Summary 
Adds a new command for admin/research-lead users who operate in a provider-account (suite offer) context rather than a workspace. 
`az quantum suite-offers target list` lists a provider account's targets and their status without requiring a workspace.

## What changed
- New command `az quantum suite-offers target list -p <providerId> [-l <location>]` in a new `suite-offers target` command group.
- `--location/-l` is optional; when omitted it is resolved automatically from the provider account's suite offer.
- Provider-account status is read from the V2 regional data plane (`.../suiteoffers/{providerId}/providerStatus`) by reusing the generated client's authenticated pipeline via `send_request` .
- Version bumped to `1.0.0b21` with a HISTORY.rst entry.

### DIFF
--- a/src/quantum/azext_quantum/_client_factory.py
+++ b/src/quantum/azext_quantum/_client_factory.py
@@ -6,11 +6,12 @@
 # pylint: disable=line-too-long,protected-access
 
 import os
+from azure.core.rest import HttpRequest
 from ._location_helper import normalize_location
 from .__init__ import CLI_REPORTED_VERSION
 from .vendored_sdks.azure_quantum_python._client import WorkspaceClient
 from .vendored_sdks.azure_mgmt_quantum import AzureQuantumMgmtClient
-from .vendored_sdks.azure_mgmt_quantum.operations import WorkspacesOperations, OfferingsOperations
+from .vendored_sdks.azure_mgmt_quantum.operations import WorkspacesOperations, OfferingsOperations, SuiteOffersOperations
 
 
 def is_env(name):
@@ -26,6 +27,16 @@ def base_url(location):
     if is_env('dogfood'):
         return f"https://{normalized_location}.quantum-test.azure.com/"
     return f"https://{normalized_location}.quantum.azure.com/"
+
+
+def base_url_v2(location):
+    # Suite offers are a V2 concept whose data plane is served from a
+    # '-v2'-suffixed regional host (for example, 'westus-v2.quantum.azure.com'). An explicit
+    # AZURE_QUANTUM_BASEURL override is honored verbatim.
+    url = base_url(location)
+    if 'AZURE_QUANTUM_BASEURL' in os.environ:
+        return url
+    return url.replace('.quantum', '-v2.quantum', 1)
 
 
 def _get_data_credentials(cli_ctx, subscription_id=None):
@@ -57,6 +68,10 @@ def cf_offerings(cli_ctx, *_) -> OfferingsOperations:
     return cf_quantum_mgmt(cli_ctx).offerings
 
 
+def cf_suite_offers(cli_ctx, *_) -> SuiteOffersOperations:
+    return cf_quantum_mgmt(cli_ctx).suite_offers
+
+
 # Data Plane clients
 
 def cf_quantum(cli_ctx, subscription: str, resource_group: str, ws_name: str, endpoint: str | None) -> WorkspaceClient:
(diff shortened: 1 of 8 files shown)


src/quantum/azext_quantum/operations/suiteoffers.py
```diff
@@ -0,0 +1,39 @@
...
+
+# pylint: disable=line-too-long
+
+from azure.cli.core.azclierror import InvalidArgumentValueError
+from .._client_factory import cf_suite_offers, cf_provider_account_status
+
+
+def list_targets(cmd, provider_id, location=None):
+    """
+    Get the list of targets and their status available through a provider account (suite offer),
```
04diiguyi: Can we just use suite offer instead of provider account since provider account is not public

src/quantum/azext_quantum/_client_factory.py
```diff
@@ -81,6 +96,23 @@ def cf_quotas(cli_ctx, subscription: str, resource_group: str, ws_name: str, end
     return cf_quantum(cli_ctx, subscription, resource_group, ws_name, endpoint).services.quotas
 
 
+def cf_provider_account_status(cli_ctx, subscription: str, location: str, provider_id: str):
+    """
+    Fetch the provider-account (suite offer) status from the regional data plane, using the
+    generated WorkspaceClient's authenticated pipeline via a custom request.
```
04diiguyi, line 102: Let's discuss this offline. this is an interesting edge case

src/quantum/azext_quantum/_client_factory.py
```diff
@@ -81,6 +96,23 @@ def cf_quotas(cli_ctx, subscription: str, resource_group: str, ws_name: str, end
...
+def cf_suite_offer_status(cli_ctx, subscription: str, location: str, provider_id: str):
+    """
+    Fetch the suite offer status from the regional data plane, using the
+    generated WorkspaceClient's authenticated pipeline via a custom request.
+    """
+    creds = _get_data_credentials(cli_ctx, subscription)
+    client = WorkspaceClient(base_url_v2(location), creds)
+    request = HttpRequest(
+        method="GET",
+        url=f"/subscriptions/{subscription}/providers/Microsoft.Quantum/suiteoffers/{provider_id}/providerStatus",
```
04diiguyi, line 108: the api version, can we get it from the client?

04diiguyi (approved):

src/quantum/azext_quantum/_client_factory.py
```diff
@@ -81,6 +96,23 @@ def cf_quotas(cli_ctx, subscription: str, resource_group: str, ws_name: str, end
...
 
 
+def cf_suite_offer_status(cli_ctx, subscription: str, location: str, provider_id: str):
+    """
+    Fetch the suite offer status from the regional data plane, using the
+    generated WorkspaceClient's authenticated pipeline via a custom request.
+    """
+    creds = _get_data_credentials(cli_ctx, subscription)
+    client = WorkspaceClient(base_url_v2(location), creds)
+    request = HttpRequest(
```
rigidit, line 106: is there a reason why we don't use cf_quantum_mgmt(cli_ctx).suite_offers here instead of construction http request by ourselves?

v-elegacheva: Well the suite_offers ( `SuiteOffersOperations` ) only generates `list_by_subscription` , which is an ARM call returning the static suite-offer catalog and there is no `providerStatus` operation in the mgmt SDK. `providerStatus` is a live data-plane endpoint served from  the -v2. regional host, not ARM, so the ARM-bound mgmt client can't reach it. I reuse the data-plane `WorkspaceClient` 's authenticated pipeline (correct regional endpoint + audience) and send the single ungenerated request. If/ when the service adds `providerStatus` to the Swagger, we can regenerate and drop the custom request.

rigidit: ok, so if you need to use providerStatus call to get targets availability it has different url that does not contain suiteOffers in it, see here https://github.com/Azure/azure-rest-api-specs/blob/5f39f28d8337b0a8021a7e104c9fb6994b4bcf49/specification/quantum/data-plane/Quantum/preview/2026-01-15-preview/quantum.json#L561

PR #10215, opened 2026-08-13
Quantum: add `az quantum workspace user list` command

kaverkiev:

| Breaking Changes |
| :--: |
| ⚠️ None |

<details>

><details>
>
>|---|---|---|---|
>|⚠️ [1001 - CmdAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1001.md)|quantum workspace user list|cmd `quantum workspace user list` added||
>
></details>
</details>

**Summary**

Adds az quantum workspace user list to list the users (user principals) with access to an Azure Quantum workspace. To match the Azure Quantum portal, only user principals are listed — groups and service principals are excluded. Supports --role (defaults to Quantum Workspace Data Contributor) and --include-inherited (to also include access inherited from the parent resource group/subscription).

**Details**

Lists Azure RBAC role assignments scoped to the workspace.
Supports --assignee, --assignee-object-id, and --role filters (reuses the existing workspace user params).
Adds an opt-in --include-inherited flag to also surface access inherited from the parent resource group and subscription (assignments granted above the workspace scope). 

**Testing**

New unit tests: scope construction, --include-inherited pass-through, and table transformer.
Added a live user list check to the test_workspace_user scenario.
Manually verified against a live workspace (default vs --include-inherited).

---

### General Guidelines

- [ ] Have you run `python scripts/ci/test_index.py -q` locally? (`pip install azdev` required)

For new extensions:

### About Extension Publish


src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +497,14 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+
+def list_users(cmd, resource_group_name=None, workspace_name=None, assignee=None, assignee_object_id=None, role=None, include_inherited=False):
+    """
+    List the users, groups, and service principals with access to an Azure Quantum workspace.
+    """
+    from azure.cli.command_modules.role.custom import list_role_assignments
+
+    info = WorkspaceInfo(cmd, resource_group_name, workspace_name)
+    scope = _get_workspace_resource_id(info)
+    return list_role_assignments(cmd, assignee=assignee, assignee_object_id=assignee_object_id, role=role, scope=scope, include_inherited=include_inherited)
```
rigidit: should we use only QUANTUM_WORKSPACE_DATA_CONTRIBUTOR_ROLE_ID for role here? I think we should not allow to pass role as an argument unless we have a good reason

kaverkiev: updated

src/quantum/azext_quantum/_params.py
```diff
@@ -70,8 +70,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
...
-    role_type = CLIArgumentType(options_list=['--role'], help="Role name or id. For 'create', the role granted to the user; for 'delete', the role assignment to remove. Defaults to the 'Quantum Workspace Data Contributor' role.")
+    role_type = CLIArgumentType(options_list=['--role'], help="Role name or id. For 'create', the role to grant; for 'delete', the assignment to remove; for 'list', the role to filter by. All default to the 'Quantum Workspace Data Contributor' role.")
     assignee_principal_type_type = CLIArgumentType(options_list=['--assignee-principal-type'], arg_type=get_enum_type(['User', 'Group', 'ServicePrincipal', 'ForeignGroup']), help="Use with '--assignee-object-id' to avoid errors caused by propagation latency in Microsoft Graph.")
+    include_inherited_type = CLIArgumentType(options_list=['--include-inherited'], help='If specified, also list role assignments inherited from the parent resource group and subscription scopes.')
```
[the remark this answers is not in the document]
kaverkiev: flag added

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +497,17 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+    List the users with access to an Azure Quantum workspace.
+    """
+    from azure.cli.command_modules.role.custom import list_role_assignments
+
+    info = WorkspaceInfo(cmd, resource_group_name, workspace_name)
+    scope = _get_workspace_resource_id(info)
+    role = role or QUANTUM_WORKSPACE_DATA_CONTRIBUTOR_ROLE_ID
+    assignments = list_role_assignments(cmd, assignee=assignee, assignee_object_id=assignee_object_id, role=role, scope=scope, include_inherited=include_inherited)
+    # Match the Quantum portal, which lists only user principals (not groups or service principals).
+    return [assignment for assignment in assignments if assignment.get("principalType") == "User"]
```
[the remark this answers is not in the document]
kaverkiev: Description updated

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +497,17 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
     scope = _get_workspace_resource_id(info)
     role = role or QUANTUM_WORKSPACE_DATA_CONTRIBUTOR_ROLE_ID
     return delete_role_assignments(cmd, role=role, scope=scope, assignee=assignee, assignee_object_id=assignee_object_id)
+
+
+def list_users(cmd, resource_group_name=None, workspace_name=None, assignee=None, assignee_object_id=None, role=None, include_inherited=False):
```
rigidit: I don't think we need assignee=None, assignee_object_id=None, role=None, include_inherited=False parameters, keep it simple

kaverkiev: Partially agree. 
We can drop --assignee, --assignee-object-id, and --role.
I believe we can keep `--include-inherited`: without it, list returns only users with a direct workspace-scope assignment, which is empty when access is inherited from the parent resource group or subscription.

kaverkiev: Updated

src/quantum/azext_quantum/_params.py
```diff
@@ -70,8 +70,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
...
     assignee_type = CLIArgumentType(options_list=['--assignee'], help='Represents a user, group, or service principal. Supported formats: object id, user sign-in name, or service principal name.')
     assignee_object_id_type = CLIArgumentType(options_list=['--assignee-object-id'], help="Use this parameter instead of '--assignee' to bypass Graph API invocation in case of insufficient privileges. This parameter only works with object ids for users, groups, service principals, and managed identities. For managed identities use the principal id. For service principals, use the object id and not the app id.")
-    role_type = CLIArgumentType(options_list=['--role'], help="Role name or id. For 'create', the role granted to the user; for 'delete', the role assignment to remove. Defaults to the 'Quantum Workspace Data Contributor' role.")
+    role_type = CLIArgumentType(options_list=['--role'], help="Role name or id. For 'create', the role to grant; for 'delete', the assignment to remove; for 'list', the role to filter by. All default to the 'Quantum Workspace Data Contributor' role.")
```
rigidit: I don't think this PR touches role_type right?

kaverkiev: It does touch, but only to update its help text because `user list` now uses --role as a filter. The argument’s behavior for create and delete is unchanged.

rigidit: But list_users has no role parameter, so c.argument('role', role_type) in the quantum workspace user context is a no-op for list. Users reading az quantum workspace user list --help won't see --role, yet az quantum workspace user --help-adjacent docs claim it exists.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +505,45 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+    """
+    List the users with access to an Azure Quantum workspace.
+    """
+    from azure.cli.command_modules.role.custom import list_role_assignments
+
+    info = WorkspaceInfo(cmd, resource_group_name, workspace_name)
+    scope = _get_workspace_resource_id(info)
+    assignments = []
+    for role_id in (QUANTUM_WORKSPACE_DATA_CONTRIBUTOR_ROLE_ID, QUANTUM_WORKSPACE_OWNER_ROLE_ID):
+        assignments += list_role_assignments(cmd, role=role_id, scope=scope, include_inherited=include_inherited)
```
rigidit: isn't that list_role_assignments call already return role definition name and principal name? please check that we don't make redundant Graph calls that can affect performance

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +505,45 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+    """
+    Enrich user role assignments with the display name and email resolved from Microsoft Graph,
+    matching the Name and Email columns shown in the Quantum portal. Best-effort: if the lookup
+    fails (for example, the caller cannot read the directory), the principal name is used instead.
+    """
+    principal_ids = {user["principalId"] for user in users if user.get("principalId")}
+    if not principal_ids:
+        return
+
+    from azure.cli.command_modules.role.custom import _graph_client_factory, _get_object_stubs
```
rigidit: can we use public Graph factory instead private one? azure.cli.command_modules.role.__init__ might already export public graph_client_factory

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +505,45 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+    try:
+        graph_client = _graph_client_factory(cmd.cli_ctx)
+        for obj in _get_object_stubs(graph_client, principal_ids):
+            directory_objects[obj.get("id")] = obj
+    except Exception as ex:  # pylint: disable=broad-except
+        logger.warning("Could not resolve user display names from Microsoft Graph: %s", ex)
+
+    for user in users:
+        obj = directory_objects.get(user.get("principalId"), {})
+        user["displayName"] = obj.get("displayName")
```
rigidit: should we add ` or user.get("principalName")` here?

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
```diff
@@ -324,6 +331,91 @@ def test_workspace_user(self):
...
+        assert rows[0]["Name"] == "Contoso User"
+        assert rows[0]["Email"] == "user@contoso.com"
+        assert rows[0]["Role"] == "Quantum Workspace Data Contributor"
+        assert rows[0]["Time Created"] == "2026-06-24T16:53:26.107178+00:00"
+
+        # Email falls back to the principal name when Graph did not return a mail address.
+        fallback = transform_users([{"principalName": "fallback@contoso.com"}])
+        assert fallback[0]["Email"] == "fallback@contoso.com"
+        assert fallback[0]["Name"] is None
+
```
rigidit: These are seems to be pure mock-based unit tests, but they live in QuantumWorkspacesScenarioTest(ScenarioTest). They inherit the VCR/recording and preparer machinery for no reason. Move them into a plain class QuantumWorkspaceUserListTest(unittest.TestCase) in the same file (the module already imports unittest) and use self.assertEqual rather than bare assert, matching the rest of the repo's mocked tests.

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: Done . switched to the public graph_client_factory

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: Good catch. Added fallback and test for it

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
kaverkiev: Done. Moved in separate QuantumWorkspaceUserListTest class

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: I set fill_principal_name=False on both calls since _fill_user_display_names already resolves name/email/UPN in one batched Graph call

src/quantum/azext_quantum/_params.py
```diff
@@ -73,8 +73,9 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
...
```
rigidit: Please revert that clause to remove list mentioning. list_users has no role parameter, so c.argument('role', role_type) in the quantum workspace user context is a no-op for list. Users reading az quantum workspace user list --help won't see --role, yet az quantum workspace user --help-adjacent docs claim it exists.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +503,46 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+    scope = _get_workspace_resource_id(info)
+    assignments = []
+    for role_id in (QUANTUM_WORKSPACE_DATA_CONTRIBUTOR_ROLE_ID, QUANTUM_WORKSPACE_OWNER_ROLE_ID):
+        # fill_principal_name=False avoids a per-call Microsoft Graph lookup that _fill_user_display_names already does in one batch.
+        assignments += list_role_assignments(cmd, role=role_id, scope=scope, include_inherited=include_inherited, fill_principal_name=False)
+    users = [assignment for assignment in assignments if assignment.get("principalType") == "User"]
+    _fill_user_display_names(cmd, users)
+    return users
+
+
```
rigidit, line 613: fill_principal_name=False removed the duplicate Graph calls, but workspace.py:515-521 still loops and calls list_role_assignments twice, and fill_role_definition_name still defaults to True. That enumerates every role definition visible at the workspace scope (hundreds of built-ins) — twice. azure-cli's own help calls this out: "Fill roleDefinitionName property in addition to roleDefinitionId. This operation is expensive."

A single call filtered client-side halves the ARM traffic, consider to changing to:
```
def list_users(cmd, resource_group_name=None, workspace_name=None, include_inherited=True):
    from azure.cli.command_modules.role.custom import list_role_assignments

    info = WorkspaceInfo(cmd, resource_group_name, workspace_name)
    scope = _get_workspace_resource_id(info)
    assignments = list_role_assignments(cmd, scope=scope, include_inherited=include_inherited)
    quantum_roles = (QUANTUM_WORKSPACE_DATA_CONTRIBUTOR_ROLE_ID, QUANTUM_WORKSPACE_OWNER_ROLE_ID)
    users = [a for a in assignments
             if a.get("principalType") == "User"
             and a.get("roleDefinitionId", "").rsplit("/", 1)[-1].lower() in quantum_roles]
    _fill_user_display_names(cmd, users)
    return users
    ```

File: src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py [hunk 12]
```diff
@@ -312,6 +314,11 @@ def test_workspace_user(self):
             self.check("ends_with(roleDefinitionId, 'c1410b24-3e69-4857-8f86-4d0a2e603250')", True)
         ])
 
+        # list users and verify the new assignment appears
+        self.cmd(f'az quantum workspace user list -g {test_resource_group} --workspace-name {test_workspace_temp} -o json', checks=[
+            self.check(f"length([?principalId=='{test_object_id}'])", 1)
```
rigidit: this asserts length([?principalId=='...']) == 1 while include_inherited defaults to True. If the test principal holds either Quantum role at RG or subscription scope, this returns 2+. Adding --include-inherited false fixes it and exercises the new flag in a live test.

src/quantum/azext_quantum/commands.py
```diff
@@ -56,6 +56,17 @@ def transform_jobs(results):
     return [transform_job(job) for job in results]
 
 
+def transform_users(results):
+    def one(result):
+        return OrderedDict([
+            ('Name', result.get('displayName')),
+            ('Email', result.get('mail') or result.get('principalName')),
+            ('Role', result.get('roleDefinitionName')),
+            ('Time Created', result.get('createdOn'))
```
rigidit: Can we change that to "Time Added"?

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +503,46 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+def _fill_user_display_names(cmd, users):
+    """
+    Enrich user role assignments with the display name and email resolved from Microsoft Graph
+    in a single batched lookup. This avoids a per-user Microsoft Graph lookup that would be slow and could hit throttling limits.
+    """
+    principal_ids = {user["principalId"] for user in users if user.get("principalId")}
+    if not principal_ids:
```
rigidit, line 620: there won't be any value of showing just roles and creation time without name/email any identifier of the user principal.
let's try to fetch name & email couple of time using retry and if still won't be able - just show error to the user asking to try later

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: 1. This change broke order (Data contributor first) that you are requested before. Single call returns ARM's order, so we'd need an explicit client-side sort to keep Contributor-first. Extra logic. Please confirm.
2. You dropped `fill_principal_name=False`. Without it you will re-introduce the Graph redundancy that I just removed
3.  Potentially unfiltered call returns all roles at scope before filtering. With a busy subscription scope the response is might be huge.

src/quantum/azext_quantum/commands.py
kaverkiev: What is the reason for this rename? It will break consistency with UI. Currently, we mirroring structure, just as we discussed.

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
kaverkiev: agree. Flag added

src/quantum/azext_quantum/_params.py
kaverkiev: updated

src/quantum/azext_quantum/_params.py
kaverkiev: Thanks for catching this. list doesn't support --role, so I reverted the help text change.

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: added 3 attempts + tests

src/quantum/azext_quantum/operations/workspace.py
rigidit: ok, so we can keep your implementation but make fill_role_definition_name set to false and that should solve the issue?

src/quantum/azext_quantum/commands.py
kaverkiev: resolved

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: Done. I kept the two filtered calls, disabled role-name lookup + set the known role names locally. This avoid  forr extra API calls.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +504,61 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+        return
+
+    from azure.cli.command_modules.role import graph_client_factory
+    from azure.cli.command_modules.role.custom import _get_object_stubs
+
+    directory_objects = {}
+    for attempt in range(MAX_RETRIES_USER_LOOKUP):
+        try:
+            graph_client = graph_client_factory(cmd.cli_ctx)
```
rigidit: can we create graph client once?

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +504,61 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+            directory_objects = {obj.get("id"): obj for obj in _get_object_stubs(graph_client, principal_ids)}
+            if principal_ids.issubset(directory_objects):
+                break
+        except Exception:  # pylint: disable=broad-except
+            directory_objects = {}
```
rigidit: broad except Exception issue: it retries programming errors and permanent authorization failures. Catch Graph/HTTP exceptions, inspect their status, and re-raise non-transient failures as an appropriate CLI error. Fail immediately on permanent errors such as 400, 401, and 403. Also log the final underlying exception with logger.debug(..., exc_info=True) so support can diagnose the generic end-user error.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +504,61 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+
+        if attempt < MAX_RETRIES_USER_LOOKUP - 1:
```
rigidit: check if we can honor `Retry-After` if provided, otherwise use exponential backoff, e.g. 2 ** attempt

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: Create client once before the retry loop.

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: Now honors Retry-After, falling back to exponential backoff.

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: Replaced the broad exception catch with specific Graph/HTTP errors.
retry only transient failures.

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
```diff
@@ -312,6 +315,11 @@ def test_workspace_user(self):
             self.check("ends_with(roleDefinitionId, 'c1410b24-3e69-4857-8f86-4d0a2e603250')", True)
         ])
 
+        # list users and verify the new assignment appears
+        self.cmd(f'az quantum workspace user list -g {test_resource_group} --workspace-name {test_workspace_temp} --include-inherited false -o json', checks=[
```
rigidit, line 322: This asserts the newly created assignment appears in user list, but the assignee created just above is a user-assigned managed identity, whose principalType is ServicePrincipal, but implementation filters Users only

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +505,80 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+            if attempt < MAX_RETRIES_USER_LOOKUP - 1:
+                retry_after = getattr(response, "headers", {}).get("Retry-After") if response else None
+                try:
+                    retry_delay = float(retry_after) if retry_after is not None else 2 ** attempt
+                except ValueError:
+                    retry_delay = 2 ** attempt
+                time.sleep(retry_delay)
+
+    logger.debug("Microsoft Graph user lookup failed after retries.", exc_info=last_error)
+    raise AzureInternalError("Could not reach Microsoft Graph to resolve user names and email addresses. Please try again later.") from last_error
```
rigidit: AzureInternalError means "Azure service-side internal error." A 429 (throttling) or a client-side network failure isn't that. AzureResponseError or ServiceError fits better — and this matters because azclierror class drives telemetry/exit-code classification.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +505,80 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+            status_code = getattr(ex, "status_code", None) or getattr(response, "status_code", None)
+            transient = status_code in (408, 429) or (status_code is not None and 500 <= status_code < 600)
+            if not transient:
+                error_type = {400: ClientRequestError, 401: UnauthorizedError, 403: ForbiddenError}.get(status_code, AzureResponseError)
+                raise error_type(str(ex)) from ex
+            last_error = ex
+            if attempt < MAX_RETRIES_USER_LOOKUP - 1:
+                retry_after = getattr(response, "headers", {}).get("Retry-After") if response else None
+                try:
+                    retry_delay = float(retry_after) if retry_after is not None else 2 ** attempt
```
rigidit: Retry-After is attacker/service-controlled; a large value hangs the CLI indefinitely. Clamp it (e.g. min(float(retry_after), 60)). Also, getattr(response, "headers", {}) returns None (not {}) if the attribute exists but is None, which would raise AttributeError — use (getattr(response, "headers", None) or {}).

kaverkiev: Done, 
Retry-After is capped at 60 seconds.
+ headers=None safely falls back to exponential backoff.

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: Good catch. Updated

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
kaverkiev: fixed

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +505,80 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+        principal_name = user.get("principalName")
+        user["displayName"] = obj.get("displayName") or obj.get("userPrincipalName") or principal_name or user.get("principalId")
+        user["mail"] = obj.get("mail") or obj.get("userPrincipalName") or principal_name
+
+
+def _resolve_directory_objects(cmd, principal_ids):
+    """
+    Resolve principal IDs to Microsoft Graph objects in one batched call, keyed by object id. Only
+    transient failures are retried; a principal the directory cannot resolve is simply absent from
+    the result, so the caller can fall back to the principal name rather than failing the command.
```
rigidit: this is not true now, right?

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -497,3 +505,80 @@ def remove_user(cmd, resource_group_name=None, workspace_name=None, assignee=Non
...
+
+    graph_client = graph_client_factory(cmd.cli_ctx)
+    last_error = None
+    for attempt in range(MAX_RETRIES_USER_LOOKUP):
+        try:
+            return {obj.get("id"): obj for obj in _get_object_stubs(graph_client, principal_ids)}
+        except (GraphError, HttpResponseError) as ex:
+            response = getattr(ex, "response", None)
+            status_code = getattr(ex, "status_code", None) or getattr(response, "status_code", None)
+            transient = status_code in (408, 429) or (status_code is not None and 500 <= status_code < 600)
```
rigidit: Given retry logic exists specifically for flakiness, treating status_code is None as transient would be more consistent:
transient = status_code is None or status_code in (408, 429) or 500 <= status_code < 600

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: Unresolved principals are still omitted, and the caller falls back to the principal name. I updated the wording to clarify that the batched Graph lookup may be retried on transient failures.”

src/quantum/azext_quantum/operations/workspace.py
kaverkiev: Agree. Updated + added test case

rigidit (approved):

yonzhan:
Please fix conflicting files.

kaverkiev:
> Please fix conflicting files.

Resolved!

rigidit (approved):

necusjz (approved):

PR #10233, opened 2026-08-19
[AKS] Fix load balancer options ignored when outbound type is not loadBalancer

Liunardy:

| Breaking Changes |
| :--: |
| ️✔️ None |

`az aks update --load-balancer-backend-pool-type nodeIP` was silently ignored on clusters whose outboundType is not loadBalancer (userDefinedRouting, NAT gateway, none, block). update_load_balancer_profile read the outbound type through get_outbound_type, which since 073b19f05 falls back to the value on the existing cluster. An unchanged outbound type therefore looked like a request to switch away from loadBalancer, and the whole profile was set to None. The request carried no loadBalancerProfile at all, so AKS applied nothing and returned success, leaving the cluster on nodeIPConfiguration.

Read outbound_type from the raw command parameters instead, so only an outbound type given on the command line counts as a switch, and clear just the five outbound fields rather than the whole profile. Inbound settings such as backendPoolType are valid for every outbound type and are now preserved, including when requested in the same command that changes --outbound-type. When no inbound settings remain the profile is still set to None, so the request body continues to omit loadBalancerProfile rather than sending an empty object.

Affected versions: 21.0.0b10 through 22.0.0b1.

### Testing

Verified live: on a userDefinedRouting cluster `--load-balancer-backend-pool-type nodeIP` now reaches AKS and applies. On a loadBalancer cluster `--outbound-type userDefinedRouting --load-balancer-backend-pool-type nodeIPConfiguration` applies both changes and clears managedOutboundIPs.

---

### Related command

az aks update --load-balancer-backend-pool-type

### General Guidelines

For new extensions:

### About Extension Publish

### DIFF
--- a/src/aks-preview/HISTORY.rst
+++ b/src/aks-preview/HISTORY.rst
@@ -12,6 +12,10 @@ To release a new version, please select a new version number (usually plus 1 to 
 Pending
 +++++++
 
+22.0.0b2
+++++++++
+* `az aks update`: Fix load balancer options such as `--load-balancer-backend-pool-type` being silently ignored on clusters whose `outboundType` is not `loadBalancer` (for example `userDefinedRouting` or a NAT gateway). Passing `--outbound-type` to switch away from `loadBalancer` now only drops the load balancer outbound settings instead of the whole load balancer profile, so options requested in the same command are still applied.
+
 22.0.0b1
 ++++++++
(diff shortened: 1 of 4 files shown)


yonzhan:
AKS

FumingZhang (approved):

yanzhudd (approved):

PR #10237, opened 2026-08-19
{AKS} AI Manager Grant the caller built-in roles on create and namespace add

PugDeveloper:

| Breaking Changes |
| :--: |
| ️✔️ None |

### Related command
`az aimanager create`, `az aimanager namespace add`

### Description
When an AI Manager or a namespace is created, the caller is now granted two built-in roles on the new resource's scope, so the creator can immediately manage and read what they created without a separate role-assignment step:

- `Azure AIManager Contributor` (ARM / management-plane) — `413f2675-4911-4010-be3b-c720b43a3c59`
- `Azure AIManager and namespace RBAC Reader` (Kubernetes RBAC) — `9c77f8a7-b0b9-4462-844c-de6e66add8ba`

Behavior:
- Runs at the end of `az aimanager create` (aiManager scope) and `az aimanager namespace add` (namespace scope), only after the create long-running operation succeeds.
- Best-effort: a successful create/add never fails because of the grant. If the caller is not an Owner or User Access Administrator (lacks `Microsoft.Authorization/roleAssignments/write`), a warning with the exact `az role assignment create` remediation is printed and the command continues.
- Idempotent (existing assignment skipped); resolves the caller as a user or service principal and retries the other type on a principal-type mismatch.
- Skipped with `--no-wait` (success cannot be confirmed) and with the new `--skip-role-assignments` opt-out flag.

Extension version: aimanager `1.4.1` -> `1.5.0`.

### General Guidelines
- [x] Have you run `azdev style aimanager` locally? (pylint + flake8 PASSED)
- [x] Have you run `python scripts/ci/test_index.py -q` locally? (OK)
- [x] My extension version conforms to the Extension version schema (minor bump, new feature + new `--skip-role-assignments` argument).

### About Extension Publish
Updated `setup.py` VERSION to `1.5.0` and added a `HISTORY.rst` entry; `src/index.json` not modified.

### DIFF
--- a/src/aimanager/azext_aimanager/_params.py
+++ b/src/aimanager/azext_aimanager/_params.py
@@ -45,6 +45,9 @@ def load_arguments(self, _):
     with self.argument_context('aimanager create') as c:
         c.argument('location', arg_type=get_location_type(self.cli_ctx),
                    validator=get_default_location_from_resource_group)
+        c.argument('skip_role_assignments', action='store_true',
+                   help="Do not grant the caller the built-in 'Azure AIManager Contributor' and "
+                        "'Azure AIManager and namespace RBAC Reader' roles on the new AI Manager.")
 
     with self.argument_context('aimanager list') as c:
         c.ignore('ai_manager_name')
@@ -101,6 +104,11 @@ def load_arguments(self, _):
                        help='Space-separated annotations (key=value) applied to the Kubernetes namespace.')
             c.argument('aks_custom_headers', options_list=['--aks-custom-headers'],
                        help='Comma-separated key=value pairs to specify custom headers.')
+
+    with self.argument_context('aimanager namespace add') as c:
+        c.argument('skip_role_assignments', action='store_true',
+                   help="Do not grant the caller the built-in 'Azure AIManager Contributor' and "
+                        "'Azure AIManager and namespace RBAC Reader' roles on the new namespace.")
 
     with self.argument_context('aimanager namespace get-credentials') as c:
         c.argument('path', options_list=['--file', '-f'],
--- a/src/aimanager/azext_aimanager/_params.py
+++ b/src/aimanager/azext_aimanager/_params.py
@@ -45,9 +45,6 @@ def load_arguments(self, _):
     with self.argument_context('aimanager create') as c:
         c.argument('location', arg_type=get_location_type(self.cli_ctx),
                    validator=get_default_location_from_resource_group)
-        c.argument('skip_role_assignments', action='store_true',
-                   help="Do not grant the caller the built-in 'Azure AIManager Contributor' and "
-                        "'Azure AIManager and namespace RBAC Reader' roles on the new AI Manager.")
 
     with self.argument_context('aimanager list') as c:
         c.ignore('ai_manager_name')
@@ -104,11 +101,6 @@ def load_arguments(self, _):
                        help='Space-separated annotations (key=value) applied to the Kubernetes namespace.')
             c.argument('aks_custom_headers', options_list=['--aks-custom-headers'],
                        help='Comma-separated key=value pairs to specify custom headers.')
-
-    with self.argument_context('aimanager namespace add') as c:
-        c.argument('skip_role_assignments', action='store_true',
-                   help="Do not grant the caller the built-in 'Azure AIManager Contributor' and "
-                        "'Azure AIManager and namespace RBAC Reader' roles on the new namespace.")
 
     with self.argument_context('aimanager namespace get-credentials') as c:
         c.argument('path', options_list=['--file', '-f'],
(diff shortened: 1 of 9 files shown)


src/aimanager/azext_aimanager/custom.py
```diff
@@ -30,6 +33,42 @@ def _get_model(cmd, name, operation_group):
...
+def _grant_caller_roles_on_success(cmd, poller, no_wait, skip_role_assignments, scope):
+    """Wait for the create to succeed, then grant the caller the built-in roles (best-effort).
+
+    Returns the value the command should return: the completed resource when we waited for it,
+    otherwise the poller. With --no-wait the grant is skipped, since the command returns before
+    the operation completes and success cannot be confirmed.
+    """
+    if skip_role_assignments:
+        return poller
+    if no_wait:
```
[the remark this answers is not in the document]
PugDeveloper: Fixed. `--skip-role-assignments` now only skips the grant; the command still waits for the create LRO and returns the resource, matching the `amg` create pattern. Only `--no-wait` returns the poller and skips the grant.

src/aimanager/azext_aimanager/tests/latest/test_aimanager.py
```diff
@@ -0,0 +1,131 @@
...
+    @patch.object(custom, "LongRunningOperation")
+    @patch(SUB_PATCH, return_value="sub")
+    @patch.object(custom, "assign_caller_roles")
+    @patch.object(custom, "_construct_aimanager", return_value=object())
+    def test_create_skips_roles_when_flag_set(self, _construct, mock_assign, _sub, mock_lro):
+        custom.create_aimanager(
+            self.cmd, self.client, "rg", "aim", location="eastus2", skip_role_assignments=True)
+
+        mock_assign.assert_not_called()
+        mock_lro.assert_not_called()
```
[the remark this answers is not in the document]
PugDeveloper: Fixed. Both skip-flag tests now assert the command still waits (`LongRunningOperation` called) while the grant is skipped, and added combined `--no-wait --skip-role-assignments` tests covering the new branch.

yonzhan:
AKS

src/aimanager/azext_aimanager/_params.py
```diff
@@ -45,6 +45,9 @@ def load_arguments(self, _):
     with self.argument_context('aimanager create') as c:
         c.argument('location', arg_type=get_location_type(self.cli_ctx),
                    validator=get_default_location_from_resource_group)
+        c.argument('skip_role_assignments', action='store_true',
```
xmzhao0822: Do we need to have this flag --skip-role-assignments? I think by default roles are assigned, user cannot skip

src/aimanager/azext_aimanager/_roleassignments.py
```diff
@@ -0,0 +1,114 @@
...
+        payload += '=' * (-len(payload) % 4)
+        claims = json.loads(base64.urlsafe_b64decode(payload))
+    except Exception:  # pylint: disable=broad-except
+        return None, None
+    # 'idtyp' == 'app' identifies an app-only (service principal) token; otherwise it is a user.
+    principal_type = 'ServicePrincipal' if claims.get('idtyp') == 'app' else 'User'
+    return claims.get('oid'), principal_type
+
+
+def assign_caller_roles(cmd, scope, role_definition_ids):
```
xmzhao0822, line 39: Totally I think we need to assign two roles. Just want to make sure when we assign the first role, if the role has been assigned by user themselves before the AIManager creation successfully, it will get something like "the role has existed". This returned message should not block the second role to be assigned.

xmzhao0822:
@techworldhello please help to review, thanks

src/aimanager/azext_aimanager/_roleassignments.py
PugDeveloper: added a test to confirm this won't be an issue

src/aimanager/azext_aimanager/_params.py
PugDeveloper: ah no we do not, just notice was referencing another feature that had admin access. Removed

xmzhao0822:
--skip-role-assignments flag is mentioned in the PR description, maybe forget to update the PR description?

src/aimanager/azext_aimanager/custom.py
```diff
@@ -30,6 +33,40 @@ def _get_model(cmd, name, operation_group):
...
+
+
+def _grant_caller_roles_on_success(cmd, poller, no_wait, scope):
+    """Wait for the create to succeed, then grant the caller the built-in roles (best-effort).
+
+    Returns the value the command should return: the completed resource when we waited for it,
+    otherwise the poller. With --no-wait the grant is skipped and the poller is returned, since
+    the command returns before the operation completes and success cannot be confirmed.
+    """
+    if no_wait:
```
xmzhao0822, line 56: If user entered --no-wait in the command line, the role assignment will not happen correct?

xmzhao0822: Maybe the current approach is correct, I will approve it. If there is anything that needs to improve the code experience, we can update that in the follow up PR.

xmzhao0822 (approved):

yanzhudd (approved):

PR #10239, opened 2026-08-20
{AKS} AIManager print manual grant commands when --no-wait skips role assignment

PugDeveloper:

| Breaking Changes |
| :--: |
| ️✔️ None |

### Related command
`az aimanager create`, `az aimanager namespace add`

### Description
Under `--no-wait`, the caller role grant is skipped because the command returns before creation is confirmed. The old warning suggested "Re-run without `--no-wait`" (blocked by the create existence guard) or "assign the roles manually" (no command given).

The warning now prints the exact `az role assignment create` commands for both roles, with the resolved caller object id, role names, and scope, so the user can copy-paste them once the resource is created. The command string is now built by a shared helper reused by the permission-denied warning.

No behavior change to the default (waited) path. Follow-up to #10237.

Extension version: aimanager `1.5.0` -> `1.5.1` (patch).

### General Guidelines
- [x] Have you run `azdev style aimanager` locally? (PASSED)
- [x] Have you run `python scripts/ci/test_index.py -q` locally? (OK)

### About Extension Publish
Bumped `setup.py` VERSION to `1.5.1` and added a `HISTORY.rst` entry.

### DIFF
--- a/src/aimanager/azext_aimanager/_roleassignments.py
+++ b/src/aimanager/azext_aimanager/_roleassignments.py
@@ -135,6 +135,6 @@ def warn_roles_skipped_no_wait(cmd, scope, role_definition_ids):
         "  " + _role_assignment_command(assignee, role_id, scope)
         for role_id in role_definition_ids)
     logger.warning(
-        "--no-wait was set, so the caller's role assignments on %s were skipped. Once the create "
-        "succeeds, grant them with (safe to run after creation completes; data-plane access can "
+        "--no-wait was set, so the caller's role assignments on %s were skipped. Grant them once "
+        "the create succeeds (requires Owner or User Access Administrator; data-plane access can "
         "take a few minutes to take effect):\n%s", scope, commands)
--- a/src/aimanager/azext_aimanager/_roleassignments.py
+++ b/src/aimanager/azext_aimanager/_roleassignments.py
@@ -105,10 +105,31 @@ def _assign_role(assignments_client, params_model, subscription_id, scope, role_
     return None
 
 
-def _warn_assignment_failed(scope, role_id, object_id):
+def _role_assignment_command(assignee, role_id, scope):
+    """Build the exact 'az role assignment create' command that grants one role to the caller."""
     role_name = AIMANAGER_ROLE_NAMES.get(role_id, role_id)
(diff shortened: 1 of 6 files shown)


src/aimanager/azext_aimanager/_roleassignments.py
```diff
@@ -105,10 +105,31 @@ def _assign_role(assignments_client, params_model, subscription_id, scope, role_
     return None
 
 
-def _warn_assignment_failed(scope, role_id, object_id):
+def _role_assignment_command(assignee, role_id, scope):
+    """Build the exact 'az role assignment create' command that grants one role to the caller."""
     role_name = AIMANAGER_ROLE_NAMES.get(role_id, role_id)
```
xmzhao0822: qq: it seems like this map AIMANAGER_ROLE_NAMES will try to get the role name based on the mapping on role_id, so the second role_id is just for default, which means if there is no either AIMANAGER_CONTRIBUTOR_ROLE_ID or AIMANAGER_RBAC_READER_ROLE_ID, but it assigns to role_name so sort of confused.

```
AIMANAGER_ROLE_NAMES = {
AIMANAGER_CONTRIBUTOR_ROLE_ID: "Azure AIManager Contributor",
AIMANAGER_RBAC_READER_ROLE_ID: "Azure AIManager and namespace RBAC Reader",
}
```

PugDeveloper: Yeah, can see the confusion. role_id being both the key and the default made that line read weird. Pulled it into a small _role_name() helper so it's obvious what's happening: if a role id isn't in the map it just falls back to the GUID, which az role assignment create --role takes fine anyway. It's only ever called with the two known roles so the fallback never actually kicks in, but the helper makes the intent clear now.

xmzhao0822: Actually I see now. `az role assignment create`  with `--role` parameter accepts either a role name or a role definition GUID - Azure CLI resolves a name to its GUID for you. That's a general Azure CLI capability.

xmzhao0822 (approved):
Looks good, approved.

xmzhao0822:
@yanzhudd Please help to merge the PR, thank you :)

yanzhudd (approved):

PR #10226, opened 2026-08-17
Quantum: Update user create/delete operation

kaverkiev:
### 🤖 PR Validation — ❌ Action needed

| Breaking Changes |
| :--: |
| ❌ 1 |

<details>
<summary>❌Azure CLI Extensions Breaking Change Test</summary>

><details>
> <summary>❌quantum</summary>
>
>|rule|cmd_name|rule_message|suggest_message|
>|---|---|---|---|
>|❌ [1007 - ParaRemove](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1007.md) |quantum workspace user create|cmd `quantum workspace user create` removed parameter `assignee_object_id`|please add back parameter `assignee_object_id` for cmd `quantum workspace user create`|
>|❌ [1007 - ParaRemove](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1007.md) |quantum workspace user create|cmd `quantum workspace user create` removed parameter `assignee_principal_type`|please add back parameter `assignee_principal_type` for cmd `quantum workspace user create`|
>|❌ [1007 - ParaRemove](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1007.md) |quantum workspace user create|cmd `quantum workspace user create` removed parameter `role`|please add back parameter `role` for cmd `quantum workspace user create`|
>|❌ [1007 - ParaRemove](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1007.md) |quantum workspace user delete|cmd `quantum workspace user delete` removed parameter `assignee_object_id`|please add back parameter `assignee_object_id` for cmd `quantum workspace user delete`|
>|❌ [1007 - ParaRemove](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1007.md) |quantum workspace user delete|cmd `quantum workspace user delete` removed parameter `role`|please add back parameter `role` for cmd `quantum workspace user delete`|
>|⚠️ [1001 - CmdAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1001.md)|quantum workspace user list|cmd `quantum workspace user list` added||
>
></details>
</details>

- Forces assignee_principal_type="User" when creating the role assignment, so only user principals can be granted access.
- Removed `--assignee-principal-type` entirely

---

This checklist is used to make sure that common guidelines for a pull request are followed.

### General Guidelines

[...]

PR #10282, opened 2026-08-31
Quantum: Update user create/delete operation

pavelbraginskiy:

| Breaking Changes |
| :--: |
| ⚠️ None |

<details>

><details>
>
>|---|---|---|---|
>|⚠️ [1015 - SubgroupPropUpdate](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1015.md)|quantum workspace user|sub group `quantum workspace user` updated property `commands` from `{'quantum workspace user create': {'name': 'quantum workspace user create', 'is_aaz': False, 'parameters': [{'name': '_change_reference', 'options': ['--change-reference']}, {'name': '_acquire_policy_token', 'options': ['--acquire-policy-token']}, {'name': 'resource_group_name', 'options': ['--resource-group', '-g'], 'id_part': 'resource_group', 'has_completer': True}, {'name': 'workspace_name', 'options': ['--workspace-name', '-w']}, {'name': 'assignee', 'options': ['--assignee']}, {'name': 'assignee_object_id', 'options': ['--assignee-object-id']}, {'name': 'assignee_principal_type', 'options': ['--assignee-principal-type'], 'choices': ['ForeignGroup', 'Group', 'ServicePrincipal', 'User']}, {'name': 'role', 'options': ['--role']}]}, 'quantum workspace user delete': {'name': 'quantum workspace user delete', 'is_aaz': False, 'confirmation': True, 'parameters': [{'name': '_change_reference', 'options': ['--change-reference']}, {'name': '_acquire_policy_token', 'options': ['--acquire-policy-token']}, {'name': 'resource_group_name', 'options': ['--resource-group', '-g'], 'id_part': 'resource_group', 'has_completer': True}, {'name': 'workspace_name', 'options': ['--workspace-name', '-w']}, {'name': 'assignee', 'options': ['--assignee']}, {'name': 'assignee_object_id', 'options': ['--assignee-object-id']}, {'name': 'role', 'options': ['--role']}, {'name': 'yes', 'options': ['--yes', '-y']}]}, 'quantum workspace user list': {'name': 'quantum workspace user list', 'is_aaz': False, 'parameters': [{'name': 'resource_group_name', 'options': ['--resource-group', '-g'], 'id_part': 'resource_group', 'has_completer': True}, {'name': 'workspace_name', 'options': ['--workspace-name', '-w']}, {'name': 'include_inherited', 'options': ['--include-inherited'], 'choices': ['false', 'true'], 'nargs': '?', 'default': True}]}}` to `{'quantum workspace user add': {'name': 'quantum workspace user add', 'is_aaz': False, 'parameters': [{'name': '_change_reference', 'options': ['--change-reference']}, {'name': '_acquire_policy_token', 'options': ['--acquire-policy-token']}, {'name': 'resource_group_name', 'options': ['--resource-group', '-g'], 'id_part': 'resource_group', 'has_completer': True}, {'name': 'workspace_name', 'options': ['--workspace-name', '-w']}, {'name': 'email', 'options': ['--email'], 'required': True}]}, 'quantum workspace user remove': {'name': 'quantum workspace user remove', 'is_aaz': False, 'confirmation': True, 'parameters': [{'name': '_change_reference', 'options': ['--change-reference']}, {'name': '_acquire_policy_token', 'options': ['--acquire-policy-token']}, {'name': 'resource_group_name', 'options': ['--resource-group', '-g'], 'id_part': 'resource_group', 'has_completer': True}, {'name': 'workspace_name', 'options': ['--workspace-name', '-w']}, {'name': 'email', 'options': ['--email'], 'required': True}, {'name': 'yes', 'options': ['--yes', '-y']}]}, 'quantum workspace user list': {'name': 'quantum workspace user list', 'is_aaz': False, 'parameters': [{'name': 'resource_group_name', 'options': ['--resource-group', '-g'], 'id_part': 'resource_group', 'has_completer': True}, {'name': 'workspace_name', 'options': ['--workspace-name', '-w']}, {'name': 'include_inherited', 'options': ['--include-inherited'], 'choices': ['false', 'true'], 'nargs': '?', 'default': True}]}}`||
>
></details>
</details>

Continuation of the work in #10226.
Replaces `az quantum workspace user create`/`delete` with `add`/`remove`, and makes these commands support only Users and not other principal types. 

This extension is in Preview, so no deprecation process is necessary for this breaking change.

---

### Related command

`az quantum workspace user add`

### General Guidelines

### About Extension Publish


pavelbraginskiy:
@microsoft-github-policy-service agree company="Akvelon Inc"

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
```diff
@@ -291,43 +291,37 @@ def test_workspace_keys(self):
...
 
-        # Create a user-assigned managed identity to use as the assignee. Using its
-        # object id (and '--assignee-object-id') keeps the test runnable under both
-        # user and service-principal logins, since no Microsoft Graph lookup is needed.
-        identity = self.cmd(f'az identity create -g {test_resource_group} -n {test_identity_name} -l {test_location} -o json').get_output_in_json()
-        test_object_id = identity["principalId"]
+        # Use the signed-in user because workspace access is user-only.
+        signed_in_user = self.cmd('az ad signed-in-user show -o json').get_output_in_json()
+        test_object_id = signed_in_user["id"]
+        test_user = signed_in_user["userPrincipalName"]
```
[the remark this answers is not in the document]
pavelbraginskiy: Now skips if no user is available.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -552,38 +551,42 @@ def _get_workspace_resource_id(info):
...
+    if not user:
+        raise RequiredArgumentMissingError("Please provide '--user' (the user's principal name or object ID).")
 
 
-def add_user(cmd, resource_group_name=None, workspace_name=None, assignee=None, assignee_object_id=None, assignee_principal_type=None, role=None):
+def _resolve_user_id(cmd, user):
+    from azure.cli.command_modules.role import graph_client_factory
+
+    _validate_user_arg(user)
+    return graph_client_factory(cmd.cli_ctx).user_get(user)["id"]
```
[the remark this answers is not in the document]
pavelbraginskiy: Since only emails are supported, graph permissions are required anyway to lookup the user id by email.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -52,6 +52,8 @@
 # Built-in "Quantum Workspace Owner" role.
 QUANTUM_WORKSPACE_OWNER_ROLE_ID = "30b3bcf2-670a-4bdc-8669-7e0ae0c0dfda"
 
+EMAIL_PATTERN = re.compile(r"^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$")
```
rigidit: Email regex seems rejects valid sign-in names — B2B guest UPNs (user_contoso.com#EXT#@fabrikam.onmicrosoft.com) and apostrophes (o'brien@contoso.com). Since the only real goal is rejecting object IDs, can we relax to a permissive [^@\s]+@[^@\s]+\.[^@\s]+ ? and drop the redundant ^/$ anchors if already using fullmatch.

src/quantum/azext_quantum/_params.py
```diff
@@ -196,10 +196,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
...
-    assignee_type = CLIArgumentType(options_list=['--assignee'], help='Represents a user, group, or service principal. Supported formats: object id, user sign-in name, or service principal name.')
-    assignee_object_id_type = CLIArgumentType(options_list=['--assignee-object-id'], help="Use this parameter instead of '--assignee' to bypass Graph API invocation in case of insufficient privileges. This parameter only works with object ids for users, groups, service principals, and managed identities. For managed identities use the principal id. For service principals, use the object id and not the app id.")
-    role_type = CLIArgumentType(options_list=['--role'], help="Role name or id. For 'create', the role granted to the user; for 'delete', the role assignment to remove. Defaults to the 'Quantum Workspace Data Contributor' role.")
-    assignee_principal_type_type = CLIArgumentType(options_list=['--assignee-principal-type'], arg_type=get_enum_type(['User', 'Group', 'ServicePrincipal', 'ForeignGroup']), help="Use with '--assignee-object-id' to avoid errors caused by propagation latency in Microsoft Graph.")
```
rigidit, line 202: could you remove stale linter exclusion (linter_exclusions.yml in repo's root) - still has a quantum workspace user create / assignee_principal_type entry for a command and parameter that no longer exist

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -552,38 +554,44 @@ def _get_workspace_resource_id(info):
...
+        raise RequiredArgumentMissingError("Please provide '--email' (the user's email address).")
+    if not EMAIL_PATTERN.fullmatch(email):
+        raise InvalidArgumentValueError(f"'{email}' is not a valid email address.")
+
+
+def _resolve_user_id(cmd, email):
+    from azure.cli.command_modules.role import graph_client_factory
+
+    _validate_email_arg(email)
+    return graph_client_factory(cmd.cli_ctx).user_get(email)["id"]
```
rigidit: can we return friendly error to the user if Graph returns 404 here?
raise ResourceNotFoundError(f"No user with the email address '{email}' was found in the directory.") from ex

src/quantum/azext_quantum/_params.py
```diff
@@ -215,12 +212,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
 
     with self.argument_context('quantum workspace user') as c:
         c.argument('workspace_name', workspace_name_type)
-        c.argument('assignee', assignee_type)
-        c.argument('assignee_object_id', assignee_object_id_type)
-        c.argument('role', role_type)
-
-    with self.argument_context('quantum workspace user create') as c:
-        c.argument('assignee_principal_type', assignee_principal_type_type)
+        c.argument('email', email_type)
```
rigidit: should we make `email` argument required for user add and user remove commands?

src/quantum/azext_quantum/_help.py
```diff
@@ -423,39 +423,27 @@
...
-                --assignee-object-id 00000000-0000-0000-0000-000000000000 --assignee-principal-type Group
+            az quantum workspace user add -g MyResourceGroup -w MyWorkspace \\
+                --email user@contoso.com
 """
 
-helps['quantum workspace user delete'] = """
+helps['quantum workspace user remove'] = """
     type: command
-    short-summary: Remove a user, group, or service principal's access to an Azure Quantum workspace.
+    short-summary: Remove a user's access to an Azure Quantum workspace.
```
rigidit, line 460: let's also mention what exactly role we are removing in long-summary similar to what we have in user add

src/quantum/azext_quantum/operations/workspace.py
pavelbraginskiy: Now uses the new regex

src/quantum/azext_quantum/operations/workspace.py
pavelbraginskiy: Yes, now raises ResourceNotFoundError

src/quantum/azext_quantum/_params.py
pavelbraginskiy: Now required

src/quantum/azext_quantum/_help.py
pavelbraginskiy: Done

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -552,38 +554,57 @@ def _get_workspace_resource_id(info):
...
+    try:
+        user = graph_client_factory(cmd.cli_ctx).user_get(email)
+    except GraphError as ex:
+        if getattr(ex.response, "status_code", None) == 404:
+            raise ResourceNotFoundError(
+                f"No user with the email address '{email}' was found in the directory."
+            ) from ex
+        raise
+
+    user_id = user.get("id") if isinstance(user, dict) else None
```
rigidit: nit: isinstance(user, dict) guard in _resolve_user_id is defensive beyond need -  Graph's GET /users/{id} always returns a JSON object containing id on a non-error response

rigidit (approved):

src/quantum/azext_quantum/operations/workspace.py
pavelbraginskiy: Removed this check

yonzhan:
Quantum

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -552,38 +559,81 @@ def _get_workspace_resource_id(info):
...
+    from azure.cli.command_modules.role import graph_client_factory
+    from azure.cli.command_modules.role.custom import GraphError
+
+    _validate_email_arg(email)
+    try:
+        user = graph_client_factory(cmd.cli_ctx).user_get(email)
+    except GraphError as ex:
+        if getattr(ex.response, "status_code", None) == 404:
+            raise ResourceNotFoundError(
+                f"No user with the email address '{email}' was found in the directory."
```
rigidit: maybe add additional information how they can resolve this issue? I'm not sure that this error message would be helpful to understand the issue and resolve it

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -552,38 +559,81 @@ def _get_workspace_resource_id(info):
...
 
-    _validate_assignee_args(assignee, assignee_object_id)
+    user_id = _resolve_user_id(cmd, email)
     info = WorkspaceInfo(cmd, resource_group_name, workspace_name)
     scope = _get_workspace_resource_id(info)
-    role = role or QUANTUM_WORKSPACE_DATA_CONTRIBUTOR_ROLE_ID
-    return create_role_assignment(cmd, role=role, scope=scope, assignee=assignee, assignee_object_id=assignee_object_id,
-                                  assignee_principal_type=assignee_principal_type)
+    if _list_user_workspace_role_assignments(cmd, user_id, scope):
+        raise ClientRequestError(f"User '{email}' already has access to this Azure Quantum workspace.")
```
rigidit: Maybe don't fail here? create_role_assignment is already idempotent, and CLI convention is that grant operations are idempotent.
Consider logger.warning(...) + returning the existing assignment

src/quantum/HISTORY.rst
```diff
@@ -3,6 +3,10 @@
 Release History
 ===============
 
+1.0.0b26
+++++++++++++++
+* Renamed ``az quantum workspace user create/delete`` to ``add/remove`` and limited both commands to users identified by email address.
```
rigidit: we should mention here that it is breaking changes. Suggested:
* [BREAKING CHANGE] Renamed ``az quantum workspace user create/delete`` to ``add/remove`` and limited both commands to users identified by email address. The ``--assignee``, ``--assignee-object-id``, ``--assignee-principal-type``, and ``--role`` parameters are replaced by ``--email``.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -552,38 +559,81 @@ def _get_workspace_resource_id(info):
...
-def _validate_assignee_args(assignee, assignee_object_id):
-    if not assignee and not assignee_object_id:
-        raise RequiredArgumentMissingError("Please provide either '--assignee' or '--assignee-object-id'.")
-    if assignee and assignee_object_id:
-        raise MutuallyExclusiveArgumentError("Only one of '--assignee' or '--assignee-object-id' can be specified.")
+def _validate_email_arg(email):
+    if not email:
+        raise RequiredArgumentMissingError("Please provide '--email' (the user's email address).")
+    if not EMAIL_PATTERN.fullmatch(email):
+        raise InvalidArgumentValueError(f"'{email}' is not a valid email address.")
```
rigidit: Argument validation belongs in _params.py.
CLI authoring guidelines put this in a validator= on the argument. Also, since required=True is set in _params.py:215-219, the RequiredArgumentMissingError branch is unreachable from the CLI

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -52,6 +52,13 @@
 # Built-in "Quantum Workspace Owner" role.
 QUANTUM_WORKSPACE_OWNER_ROLE_ID = "30b3bcf2-670a-4bdc-8669-7e0ae0c0dfda"
 
+QUANTUM_WORKSPACE_USER_ROLE_IDS = {
+    QUANTUM_WORKSPACE_DATA_CONTRIBUTOR_ROLE_ID,
+    QUANTUM_WORKSPACE_OWNER_ROLE_ID,
+}
+
+EMAIL_PATTERN = re.compile(r"[^@\s]+@[^@\s]+\.[^@\s]+")
```
rigidit: The email regex is too strict and too permissive.
Too strict: rejects valid UPNs with no dot in the domain (e.g. on-prem-synced admin@contoso). user_get accepts any UPN.
Too permissive: [^@\s] allows /, so a value like a/../groups@x.co fullmatches and is interpolated straight into the Graph URL path by _get_user_url (which only escapes #). Impact is negligible (the caller's own token, GET only), but excluding / and \ from the character classes costs nothing.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -552,38 +559,81 @@ def _get_workspace_resource_id(info):
...
+            logger.warning("Workspace-level access was removed for '%s', but inherited access from the resource "
+                           "group or subscription remains. Remove the inherited assignment at its scope to revoke "
+                           "access.", email)
+        return None
+    if assignments:
+        raise ClientRequestError(
+            f"User '{email}' has no workspace-level access to remove. Access inherited from the resource group or "
+            "subscription must be removed at that scope."
+        )
+    raise ClientRequestError(f"User '{email}' does not have access to this Azure Quantum workspace.")
```
rigidit: `if assignments: raise ClientRequestError...` reads like if there are assignments then we throw error that there are no assignments to remove.
Plus `ClientRequestError` seems the wrong type for both of these. It's documented in azclierror.py as "fallback of the request related errors — error occurs while attempting to make a request"; "there is nothing to remove" is `ResourceNotFoundError`.
Let's improve to something like:
```
    assignments = _list_user_workspace_role_assignments(cmd, user_id, scope)
    direct_assignments = [a for a in assignments if a["scope"].lower() == scope.lower()]
    inherited_assignments = [a for a in assignments if a["scope"].lower() != scope.lower()]

    if not direct_assignments:
        if inherited_assignments:
            raise ResourceNotFoundError(
                f"User '{email}' has no access assigned directly on this workspace. Their access is inherited "
                "from the resource group or subscription and must be removed at that scope."
            )
        raise ResourceNotFoundError(f"User '{email}' does not have access to this Azure Quantum workspace.")

    delete_role_assignments(cmd, ids=[a["id"] for a in direct_assignments])
    if inherited_assignments:
        logger.warning(...)
```

src/quantum/HISTORY.rst
rigidit: Worth one sentence in the PR description noting the extension is isPreview: true and we just want to release it without following the deprecation process.

src/quantum/azext_quantum/operations/workspace.py
pavelbraginskiy: Added a note to verify the user exists in the tenant and their email is spelled correctly.

src/quantum/azext_quantum/operations/workspace.py
pavelbraginskiy: Made the suggested change to the logic.

src/quantum/HISTORY.rst
pavelbraginskiy: Updated the history entry to that comment and added a note to the PR description.

src/quantum/azext_quantum/operations/workspace.py
pavelbraginskiy: Validation is now performed in `_validators.py`

src/quantum/azext_quantum/operations/workspace.py
pavelbraginskiy: Switched out the regex, now just requires a single @ and no `/` or `\`.

src/quantum/azext_quantum/operations/workspace.py
pavelbraginskiy: Switched to ResourceNotFoundError and refactored to the split direct/inherited assignments logic.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -552,38 +556,80 @@ def _get_workspace_resource_id(info):
...
 
-def add_user(cmd, resource_group_name=None, workspace_name=None, assignee=None, assignee_object_id=None, assignee_principal_type=None, role=None):
+    assignments = list_role_assignments(cmd, assignee_object_id=user_id, scope=scope,
+                                        include_inherited=True, fill_principal_name=False,
+                                        fill_role_definition_name=False)
+    return [assignment for assignment in assignments
+            if assignment["roleDefinitionId"].rsplit("/", 1)[-1].lower() in QUANTUM_WORKSPACE_USER_ROLE_IDS]
+
+
+def add_user(cmd, resource_group_name=None, workspace_name=None, email=None):
```
rigidit, line 615: add_user returns two different shapes. New assignment → single dict from create_role_assignment; already-has-access → the assignments list. A script doing az quantum workspace user add ... -o json | jq -r .id works on first run and breaks on re-run. Note that azure-cli's own create_role_assignment normalizes its idempotent path to list_role_assignments(...)[0] — a single object. Returning assignments[0] would match.

src/quantum/azext_quantum/_params.py
```diff
@@ -12,6 +12,14 @@
 from azure.cli.core.commands.parameters import get_enum_type, get_three_state_flag
 from azure.cli.core.util import shell_safe_json_parse
 
+# Accept UPNs with exactly one '@' and no dotted-domain requirement; reject whitespace and Graph URL path separators.
+EMAIL_PATTERN = re.compile(r"[^@/\\\s]+@[^@/\\\s]+")
+
+
+def validate_email(namespace):
+    if not EMAIL_PATTERN.fullmatch(namespace.email):
+        raise InvalidArgumentValueError(f"'{namespace.email}' is not a valid email address.")
```
rigidit: let's move this to _validators.py as every other validator in this extension is in _validators.py. No cycle today, but moving EMAIL_PATTERN + validate_email there and having _params.py import it (as commands.py already does) matches the layout.

src/quantum/azext_quantum/operations/workspace.py
pavelbraginskiy: The role assignments aren't given in a predictable order so just `[0]` doesn't guarantee a stable result, I added a helper to pick the "best" role assignment to show (the one closest to the role assignment that would have been created).

src/quantum/azext_quantum/_params.py
pavelbraginskiy: Moved.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -552,38 +556,97 @@ def _get_workspace_resource_id(info):
...
+        is_inherited = assignment_scope_parts != workspace_scope_parts
+        role_id = assignment["roleDefinitionId"].rsplit("/", 1)[-1].lower()
+        role_priority = 0 if role_id == QUANTUM_WORKSPACE_DATA_CONTRIBUTOR_ROLE_ID else 1
+        # Prefer the inherited assignment whose ARM scope is the nearest recognizable ancestor of the workspace.
+        scope_distance = (len(workspace_scope_parts) - len(assignment_scope_parts)
+                          if workspace_scope_parts[:len(assignment_scope_parts)] == assignment_scope_parts
+                          else len(workspace_scope_parts))
+        return is_inherited, role_priority, scope_distance, assignment["scope"].lower(), assignment["id"].lower()
+
+    return min(assignments, key=priority)
```
rigidit: _select_user_workspace_role_assignment is hard to follow — the 5-tuple sort key mixes three different concepts and the reader has to simulate min() to figure out the precedence. Two things would help:

is_inherited is redundant. scope_distance is 0 only when the assignment's scope parts equal the workspace's, which is exactly the not is_inherited case. So is_inherited never breaks a tie that scope_distance hasn't already broken — it just needs to move to the front of the key to preserve today's "direct beats inherited, regardless of role" precedence.

The scope comparison is duplicated with different semantics. This function normalizes via .strip("/").split("/"), but remove_user a few lines below uses raw assignment["scope"].lower() == scope.lower(). Same question, two answers — a trailing slash from ARM would make add and remove disagree about which assignments are direct.

Extracting both into helpers covers it:
```
def _scope_parts(scope):
    return scope.lower().strip("/").split("/")


def _is_workspace_scope(assignment, scope):
    return _scope_parts(assignment["scope"]) == _scope_parts(scope)


def _scope_distance(assignment_scope, workspace_scope):
    """ARM path segments between the workspace and an assignment's scope; unrelated scopes (e.g. management groups) sort last."""
    workspace_parts = _scope_parts(workspace_scope)
    assignment_parts = _scope_parts(assignment_scope)
    if workspace_parts[:len(assignment_parts)] != assignment_parts:
        return len(workspace_parts)
    return len(workspace_parts) - len(assignment_parts)


def _select_user_workspace_role_assignment(assignments, scope):
    # Nearest scope wins, then Data Contributor over Owner; scope/id break remaining ties so output is stable across runs.
    def sort_key(assignment):
        role_id = assignment["roleDefinitionId"].rsplit("/", 1)[-1].lower()
        return (_scope_distance(assignment["scope"], scope),
                0 if role_id == QUANTUM_WORKSPACE_DATA_CONTRIBUTOR_ROLE_ID else 1,
                assignment["scope"].lower(),
                assignment["id"].lower())

    return min(assignments, key=sort_key)
```

pavelbraginskiy: Implemented the suggestion, resolution should be clearer now.

rigidit (approved):

pavelbraginskiy:
Hi @necusjz, I believe this PR is ready now, please take a look when you can. Thank you.

rigidit:
Hi @necusjz , could you please help with releasing these changes? the commands for which we introducing breaking changes were introduced just month ago and not really used by customers, so we'd like to change them as soon as possible

necusjz (approved):

PR #10285, opened 2026-09-01
Quantum: Add 'az quantum suite-offer quotas' command

v-elegacheva:

| Breaking Changes |
| :--: |
| ⚠️ None |

<details>

><details>
>
>|---|---|---|---|
>|⚠️ [1001 - CmdAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1001.md)|quantum suite-offer quotas|cmd `quantum suite-offer quotas` added||
>|⚠️ [1011 - SubgroupAdd](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1011.md)|quantum suite-offer target|sub group `quantum suite-offer target` added||
>
></details>
</details>

## Summary

Adds suite-offer and V2 target-quota support to the `quantum` extension:

- **`az quantum suite-offer list`** lists the suite offers available to the subscription.
- **`az quantum suite-offer quotas --provider-id <id>`** lists per-target suite allocations merged with consumed Standard and High usage.
- **`az quantum suite-offer target list --provider-id <id>`** lists targets available through a suite offer without requiring a workspace.
- **`az quantum workspace quotas`** includes V2 target allocations and usages while preserving the existing V1 response format.
- **`az quantum workspace create/update --quota`** validates requested V2 target allocations against current workspace usage and subscription-level suite allocations.

Suite-offer data-plane commands derive the regional V2 endpoint from the suite offer's location.

### Suite offers

- Added subscription-level suite-offer listing.
- Added per-target quota allocation and usage output:
  - JSON retains the service contract's minute-based values.
  - Table output displays allocation and usage in hours.
  - Missing allocation or usage values are displayed as `0`.
  - Backend target ordering is preserved.
- Added suite-offer target status listing without requiring a workspace.
- Added overall, Standard-priority, and High-priority average queue times to target output.
- Removed the redundant provider column from the suite-offer target table.
- Consolidated suite-offer data-plane client creation into `cf_suite_offers_data_plane`.
  - Callers supply the suite offer location.
  - The factory constructs the corresponding regional V2 endpoint.
- Added compatibility for quota-usage responses returned as either a paged `{ "value": [...] }` envelope or a top-level array.

### Workspace quotas

- Preserved the existing flat V1 quota response fields.
- Added optional `targetId` to identify V2 target quota rows.
- Added separate `StandardMinutesLifetime` and `HighMinutesLifetime` rows for each V2 target.
- Combined target allocations from the ARM workspace resource with data-plane usage.
- Included usage-only targets so historical usage is not hidden when an allocation is absent.
- Used `workspace.properties.endpointUri` as the authoritative data-plane endpoint.
  - V1 workspaces call `/quotas`.
  - V2 workspaces skip `/quotas` and call only `/quotaUsages`.
- Returned missing V2 allocation and usage values as `0`.
- Reordered table columns to:
  - Scope
  - Provider ID
  - Target
  - Dimension
  - Limit
  - Utilization
  - Holds
  - Period

Before this change, V1 quota table output had no dedicated transformer and relied on the generic serialized field order. The shared transformer now provides consistent V1 and V2 table output.

### Workspace quota validation

Added always-on validation for V2 workspace target allocations on create and update.

For each target and priority, the requested absolute allocation must satisfy:

`current workspace usage <= requested workspace allocation <= subscription suite target allocation`

- Equality is allowed at both boundaries.
- Standard and High allocations are validated independently.
- Create validates against the subscription allocation, using zero as initial workspace usage.
- Update validates against live workspace usage and the subscription allocation.
- Priority values omitted during update are preserved and validated.
- Validation completes before the workspace write.
- A missing workspace usage row is treated as zero.
- A missing subscription target allocation is rejected with guidance to configure the allocation first.
- Separate actionable errors are returned when:
  - The subscription has no allocation for the target.
  - The requested allocation is below current workspace usage.
  - The requested allocation is above the subscription allocation.
- Error messages include the relevant command for viewing current workspace usage or subscription allocations.
- Other failures retrieving required validation data prevent the operation.
- No `quota add`, `--validate-only`, or validation bypass option was added.

### V2 workspace creation

- V2 workspace creation disables API-key authentication because API keys are not supported for V2 workspaces.
- V2 workspace creation does not auto-add legacy free-plan providers.
- Only explicitly selected V2 providers are included.
- Existing V1 auto-add and API-key behavior is preserved.

### Generated data-plane client

- Regenerated the vendored Python data-plane client from the public TypeSpec specification.
- Added optional Standard-priority and High-priority queue-time fields to `TargetStatus`.
- Updated quota usage model naming and extensible `Usage` mapping handling.
- Removed the explicit package-data entry for the private vendored `py.typed` marker.

### Public terminology

- Removed references to private product and implementation terminology from help, history, comments, docstrings, and error messages.
- Updated quota help to use “target quota allocation” and “requested allocation” rather than “final allocation.”

## Testing

- Full Quantum test suite: **76 passed, 14 skipped**.
- Focused workspace validation and table tests: **9 passed**.
- Focused suite-offer endpoint and ordering tests: **7 passed**.
- Suite-offer tests cover:
  - Request construction.
  - Paged and top-level array response handling.
  - Generated-model deserialization.
  - Table transformation.
  - Single-object provider-status handling.
  - Quota merge behavior.
  - Backend allocation ordering.
- Manual suite-offer validation:
  - List, quota, and target commands passed.
  - JSON and table output passed.
  - Overall, Standard, and High queue-time columns were verified.
- Manual V1 and V2 workspace quota output passed.
- Manual workspace create validation:
  - A valid Standard `1200` / High `120` allocation succeeded.
  - An allocation above the subscription limit was rejected.
  - An invalid create did not create an ARM resource.
  - The V2 workspace contained only the explicitly selected provider.
  - API keys were disabled.
- Manual workspace update validation:
  - A valid idempotent allocation update succeeded.
  - An allocation below fractional workspace usage was rejected.
  - An allocation above the subscription allocation was rejected.
  - An omitted High allocation was preserved.
  - Rejected updates did not modify existing allocations.
- A missing subscription target allocation was rejected with actionable guidance and did not modify the workspace.
- Pylint: **10.00/10**.
- Flake8, compileall, help rendering, VS Code diagnostics, pre-commit checks, wheel build, and `git diff --check`: passed.
- All temporary Azure resources were deleted after testing.


src/quantum/azext_quantum/operations/suite_offers.py
```diff
@@ -0,0 +1,124 @@
...
+def _merge_suite_offer_quotas(offer, usages, provider_id):
+    """
+    Combine the ARM quota allocations (limits) with the data-plane quota usages (consumed),
+    keyed by target id (None represents the subscription-level allocation).
+    """
+    props = offer.properties
+
+    # Allocations keyed by target id (None => subscription-level).
+    allocations = {}
+    if props.quotas is not None:
```
rigidit: let's use only targetQuotas for now

src/quantum/azext_quantum/tests/latest/test_quantum_suite_offers.py
```diff
@@ -0,0 +1,188 @@
...
+        self.assertEqual(row['Provider ID'], 'ionq')
+        self.assertEqual(row['Provider Name'], 'IonQ')
+        self.assertEqual(row['Company'], 'IonQ, Inc.')
+        self.assertEqual(row['Location'], 'eastus')
+
+    def test_transform_suite_offer_quotas(self):
+        quotas = [
+            {
+                'providerId': 'ionq',
+                'scope': 'Subscription',
```
rigidit: Use only "SubscriptionTarget" for suite offer quotas

src/quantum/azext_quantum/operations/suite_offers.py
```diff
@@ -0,0 +1,124 @@
...
+        usage = usage_by_key.get(key)
+
+        std_allocated = allocation.standard_minutes_lifetime if allocation is not None else None
+        high_allocated = allocation.high_minutes_lifetime if allocation is not None else None
+
+        usage_values = usage.usage if usage is not None else None
+        std_used = usage_values.standard_minutes_lifetime if usage_values is not None else None
+        high_used = usage_values.high_minutes_lifetime if usage_values is not None else None
+
+        row = OrderedDict()
```
rigidit, line 122: Did you consider creating a structure for return type to improve readability? and probably we can skip remaining and stick to initial structure like this:
{
    "providerId" : "atom-dev",
    "scope" : "SubscriptionTarget",
    "targetId" : "msft.sim.ac1000.physical",
    "allocation" : {
        "standardMinutesLifetime" : 600,
        "highMinutesLifetime" : 60 
    },
    "usage" : {
        "standardMinutesLifetime" : 120,
        "highMinutesLifetime" : 12
    }
}

src/quantum/azext_quantum/operations/suite_offers.py
v-elegacheva: Done!

src/quantum/azext_quantum/tests/latest/test_quantum_suite_offers.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/suite_offers.py
v-elegacheva: I restructured to match your proposed shape. Each row is now:
{
   "providerId": "...",
   "scope": "SubcriptionTarget",
   "targetId": "...",
   "allocation": { "standardMinutesLifetime": 0, "highMinutesLifetime": 0 },
   "usage": {"standardMinutesLifetime": 0, "highMinutesLifetime": 0}
}

`remaining` and `lastModifiedTime` are dropped. I added a small `_minutes()` helper to build the nested blocks. The table transformer and help text were also updated to match

yonzhan:
Quantum

src/quantum/azext_quantum/operations/suite_offers.py
```diff
@@ -0,0 +1,97 @@
...
+    Build one row per target quota allocation, attaching its matching data-plane usage.
+    Suite offer quotas are reported at the SubscriptionTarget scope only.
+    """
+    # Data-plane usages keyed by target id.
+    usage_by_target = {
+        usage.target_id: usage for usage in (usages or []) if usage.target_id is not None
+    }
+
+    rows = []
+    for target_quota in sorted(offer.properties.target_quotas or [], key=lambda q: q.target_id or ""):
```
rigidit, line 118: what is the reason of sorting target quotas here?

src/quantum/azext_quantum/operations/suite_offers.py
```diff
@@ -0,0 +1,97 @@
...
+        usage = usage_by_target.get(target_quota.target_id)
+        usage_values = usage.usage if usage is not None else None
+
+        row = OrderedDict()
+        row["providerId"] = provider_id
+        row["scope"] = "SubscriptionTarget"
```
rigidit: we can reuse usage.Scope here instead of magic string

src/quantum/azext_quantum/operations/suite_offers.py
```diff
@@ -0,0 +1,97 @@
...
+        row["targetId"] = target_quota.target_id
+        row["allocation"] = _minutes(
```
rigidit, line 126: I think in UI we show allocation and usage in hours? align with it

src/quantum/azext_quantum/operations/suite_offers.py
rigidit: thanks, but it still dict, do you think there will be benefit of creating a type with all of these fields and have dot access to the fields?

v-elegacheva: The return value of a CLI custom command is serialized straight to the user-facing output, and the CLI's `todict` serializer uses each object's raw attribute names (vars()). So a plain class/dataclass would emit snake_case keys (provider_id, standard_minutes_lifetime) instead of the providerId / standardMinutesLifetime contract, and a namedTuple serializes as a JSON array. OrderedDict gives exact control over the camelCase keys and ordering that define this command's output, and it is consistent with the rest of the quantum extension (all handlers/ transformers return dicts or SDK models). The dot-access benefit would only apply inside this ~ 15 line builder, which `_minutes()` already simplifies. If you'd like the shape documented in code, i can switch the row to a `TypedDict`. That gives type-checking + editor hints and still serializes correctly as a dict. A full dataclass would need custom camelCase serialization to avoid changing the output. Which way would you prefer?

src/quantum/azext_quantum/operations/suite_offers.py
v-elegacheva: It is just to give deterministic, stable output ordering. The service does not guarantee an order for targetQuotas (or the usages list), so sorting by targetId keeps the JSON/ table rows consistent across runs. Which also keeps diffs and the live test stable. I can drop it if you'd rather preserve the service's order :)

src/quantum/azext_quantum/operations/suite_offers.py
v-elegacheva: I agree! I did find something worth attention however: a target row can have no matching usage (usage is `None` for targets with no consumption), so usage.scope isn't always available to read. Since every target row is SubscriptionTarget - scoped by definition, I'll lift the literal into a named constant so it is not a magic string and stays independent of whether a usage row exists. If you would prefer, I can instead read usage.scope when present and fall back to the constant. Please let me know what you would prefer!

src/quantum/azext_quantum/operations/suite_offers.py
v-elegacheva: The underlying ARM / DP fields are standardMinutesLifetime / highMinutesLifetime. The values are  minutes by contract and the key names literally say "Minutes". The CLI mirrors the service payload, so converting to hours would make the value diagree with its own filed name and with ARM. My instinct was keeping the JSON in minutes (true to contract) and if it helps parity with the UI, adding hours to the table view only. If you would prefer to fully match the UI, we would need new hour-names fields (like standardHoursLifetime) rather than silently dividing the existing ones. Please let me know how you'd like to proceed on this one :)

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -444,12 +555,99 @@ def get(cmd, resource_group_name=None, workspace_name=None):
...
-    client = cf_quotas(cmd.cli_ctx, info.subscription, info.resource_group, info.name, info.endpoint)
-    response = client.list(info.subscription, info.resource_group, info.name)
-    return repack_response_json(response)
+
+    workspace = cf_workspaces(cmd.cli_ctx).get(info.resource_group, info.name)
+    properties = workspace.properties
+    providers = properties.providers if properties is not None else None
+
+    legacy_client = cf_quotas(
+        cmd.cli_ctx, info.subscription, info.resource_group, info.name, base_url(workspace.location))
```
rigidit: what is the reason of construction url instead of using info.endpoint here?
there is endpoint field in workspace resource property that contains url to DP, and if it's workspace V2 it will contain url to Data Plane v2

rigidit: and it is always requesting DPv1 even if workspace can be v2, which seems to be incorrect

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -444,12 +555,99 @@ def get(cmd, resource_group_name=None, workspace_name=None):
...
+    legacy_quotas = repack_response_json(
+        legacy_client.list(info.subscription, info.resource_group, info.name))
+
+    usages = []
+    workspace_kind = getattr(properties, 'workspace_kind', None) if properties is not None else None
+    if str(_enum_to_value(workspace_kind)).upper() == 'V2':
+        v2_client = cf_quotas(
+            cmd.cli_ctx, info.subscription, info.resource_group, info.name, base_url_v2(workspace.location))
```
rigidit: this seems to be making call to DPv2 to {{workspaceName}}/quotas endpoint which will return 404.
we should get target quota allocations from workspace resource itself

rigidit: oh, I see that it is used to list workspace usages only later. But anyway I think we could just reuse workspace's endpoint instead of constructing DPv2 url by ourselves

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: I agree. There was no need to reconstruct the dp url from the workspace location. I updated this to use `workspace.properties.endpoint_uri`, which contains the correct endpoint for the workspace's dp version.
 I also updated the quota flow to branch by workspace kind. v1 workspaces call the existing /quotas endpoint, while v2 workspaces skip that call and use only the v2 workspace usage API through endpointUri

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Fixed! For v2 workspaces, target quota allocations now come from workspace.properties.providers[].target_quotas on the ARM resource. The only dp request is /quotaUsages for current usage, so the v2 flow no longer calls /quotas. I also added tests that enforce the separate v1 and v2 routing paths

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -444,12 +555,100 @@ def get(cmd, resource_group_name=None, workspace_name=None):
...
+            legacy_client.list(info.subscription, info.resource_group, info.name))
+
+    return _merge_workspace_quotas(workspace, usages, legacy_quotas)
+
+
+_WORKSPACE_QUOTA_SCOPE = "Workspace"
+_WORKSPACE_QUOTA_PERIOD = "None"
+_TARGET_QUOTA_DIMENSIONS = (
+    ("StandardMinutesLifetime", "standard_minutes_lifetime"),
+    ("HighMinutesLifetime", "high_minutes_lifetime"),
```
rigidit: wondering if we could reuse values from autogenerated part for these

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +268,114 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
...
+    "high_minutes_lifetime": "highMinutesLifetime",
+}
+
+
+def _target_quota_usage_value(usage, attribute):
+    if usage is None:
+        return None
+    if hasattr(usage, "get"):
+        return usage.get(_TARGET_QUOTA_USAGE_FIELDS[attribute])
+    return getattr(usage, attribute, None)
```
rigidit: do we really need this? could we just use `usage.standard_minutes_lifetime`?

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +268,114 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
...
+            (item for item in suite_offer.properties.target_quotas or []
+             if item.target_id is not None and item.target_id.lower() == target_id),
+            None
+        )
+        if suite_target is None:
+            raise InvalidArgumentValueError(
+                f"Cannot validate --quota because target '{target_quota.target_id}' was not found in the "
+                f"suite offer for provider '{provider.provider_id}'. Run 'az quantum suite-offer quotas "
+                f"--provider-id {provider.provider_id}' to view available target allocations."
+            )
```
rigidit: if no corresponding target quota allocation found in suite offer, let's assume it is set to 0, so we show error to the user that target quota allocation should be set on subscription level first

src/quantum/azext_quantum/operations/suite_offers.py
rigidit: agreed, let's keep using dict

src/quantum/azext_quantum/operations/suite_offers.py
rigidit: for tests we can use other mechanisms to check
as I remember cli provides some built in tools to sort, isn't it?

src/quantum/azext_quantum/operations/suite_offers.py
rigidit: table view only is fine, thanks

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Yup! In the most recent commit, I updated the generated DimensionScope.WORKSPACE, MeterPeriod.NONE and Priority.STANDARD/Priority.HIGH values.  The CLI-specific MinutesLifetime suffix is still added when constructing the dimension names. JSON and table output remain unchanged

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: I checked the generated model- Usage is an extensible key-value model, so it does not define standard_minutes_lifetime or high_minutes_lifetime attributes. Direct attribute access raises AttributeError. The values are exposed through camelCase mapping keys. 
Updated in the most recent commit to remove the generic compatibility helper and read the generated model directly with usage.get("standardMinutesLifetime") and usage.get("highMinutesLifetime"). The tests now use the actual generated Usage model as well

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done! Updated in most recent commit so that a missing corresponding suite target allocation is treated as 0. 
A positive workspace allocation is rejected through the normal bonds validation, with an additional message telling the user to set the target quota allocation at the subscription level first. An explicit zero allocation remains valid when current usage is also zero.
I verified on PPE with a target that has a workspace allocation but no suire target allocation. The request was rejected with a suite allocation of 0, and the existing workspace allocation remained unchanged.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +269,104 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
...
+            current_usage = current_usage if current_usage is not None else 0
+            if final_allocation < current_usage or final_allocation > suite_allocation:
+                message = (
+                    f"The final {priority} allocation for provider '{provider.provider_id}', target "
+                    f"'{target_quota.target_id}' is {final_allocation} minutes. It must be between the current "
+                    f"workspace usage ({current_usage} minutes) and suite allocation ({suite_allocation} minutes), "
+                    "inclusive."
+                )
+                if suite_target is None:
+                    message += " Set the target quota allocation at the subscription level first."
```
rigidit: can we split three cases and come up with better message to user, somethinkg like:
```
if suite_target is None:
    raise InvalidArgumentValueError(
        f"--quota requests {final_allocation} minutes of {priority} time for provider "
        f"'{provider.provider_id}', target '{target_quota.target_id}', but the subscription has no "
        f"allocation for that target.\n"
        f"Allocate it at the subscription level first, then retry. To see current allocations run:\n"
        f"\taz quantum suite-offer quotas --provider-id {provider.provider_id}"
    )
if final_allocation < current_usage:
    raise InvalidArgumentValueError(
        f"--quota would set the {priority} allocation for provider '{provider.provider_id}', target "
        f"'{target_quota.target_id}' to {final_allocation} minutes, below the {current_usage} minutes "
        f"the workspace has already used. Specify at least {current_usage}.\n"
        f"To see current usage run:\n"
        f"\taz quantum workspace quotas -g {info.resource_group} -w {info.name}"
    )
if final_allocation > suite_allocation:
    raise InvalidArgumentValueError(
        f"--quota would set the {priority} allocation for provider '{provider.provider_id}', target "
        f"'{target_quota.target_id}' to {final_allocation} minutes, above the {suite_allocation} minutes "
        f"allocated to the subscription. Specify at most {suite_allocation}.\n"
        f"To see subscription allocations run:\n"
        f"\taz quantum suite-offer quotas --provider-id {provider.provider_id}"
    )
```

src/quantum/azext_quantum/_client_factory.py
```diff
@@ -28,6 +28,17 @@ def base_url(location):
     return f"https://{normalized_location}.quantum.azure.com/"
 
 
+def base_url_v2(location):
+    if 'AZURE_QUANTUM_BASEURL_V2' in os.environ:
+        return os.environ['AZURE_QUANTUM_BASEURL_V2']
+    if is_env('canary'):
+        return "https://eastus2euap-v2.quantum.azure.com/"
```
rigidit: could you check why we need that location hardcoded for canary? just using location might be enough

src/quantum/azext_quantum/operations/suite_offers.py
```diff
@@ -14,3 +28,114 @@ def list_suite_offers(cmd):
...
+    Build one row per target quota allocation, attaching its matching data-plane usage.
+    Suite offer quotas are reported at the SubscriptionTarget scope only.
+    """
+    # Data-plane usages keyed by target id.
+    usage_by_target = {
+        usage.target_id: usage for usage in (usages or []) if usage.target_id is not None
+    }
+
+    rows = []
+    for target_quota in sorted(offer.properties.target_quotas or [], key=lambda q: q.target_id or ""):
```
rigidit: we should not sort only for test purposes, consider to return result in backend returned order
and cli might have their own sorting mechanisms

src/quantum/azext_quantum/operations/suite_offers.py
```diff
@@ -5,7 +5,21 @@
...
-from .._client_factory import cf_suite_offers
+from collections import OrderedDict
+
+from azure.cli.core.azclierror import InvalidArgumentValueError, ResourceNotFoundError
+from azure.cli.core.commands.client_factory import get_subscription_id
+from azure.core.exceptions import ResourceNotFoundError as AzureResourceNotFoundError
+
+from .._client_factory import cf_suite_offers, cf_suite_offers_data_plane, base_url_v2
+
+# Suite offer quota allocations are always reported at the per-target scope.
```
rigidit: this can change in future, let's not make these 'always' statements

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -444,12 +549,100 @@ def get(cmd, resource_group_name=None, workspace_name=None):
...
+            legacy_client.list(info.subscription, info.resource_group, info.name))
+
+    return _merge_workspace_quotas(workspace, usages, legacy_quotas)
+
+
+_WORKSPACE_QUOTA_SCOPE = DimensionScope.WORKSPACE.value
+_WORKSPACE_QUOTA_PERIOD = MeterPeriod.NONE.value
+_TARGET_QUOTA_DIMENSIONS = (
+    (f"{Priority.STANDARD.value}MinutesLifetime", "standard_minutes_lifetime"),
+    (f"{Priority.HIGH.value}MinutesLifetime", "high_minutes_lifetime"),
```
rigidit: Let's not use Priority enum in building quota name

src/quantum/setup.py
```diff
@@ -49,5 +49,8 @@
     license='MIT',
     classifiers=CLASSIFIERS,
     packages=find_packages(),
-    package_data={'azext_quantum': ['azext_metadata.json', 'operations/templates/create-workspace-and-assign-role.json']},
+    package_data={
+        'azext_quantum': ['azext_metadata.json', 'operations/templates/create-workspace-and-assign-role.json'],
+        'azext_quantum.vendored_sdks.azure_quantum_python._client': ['py.typed'],
```
rigidit: let's remove this one. azext_quantum.vendored_sdks.azure_quantum_python._client is a private, underscore-prefixed, vendored copy inside a CLI extension. Type checking during development already works because Pylance reads the source tree, where the file exists regardless of what package_data ships.

src/quantum/azext_quantum/commands.py
```diff
@@ -103,6 +117,55 @@ def one(offer):
...
+
+        return OrderedDict([
+            ('Dimension', quota.get('dimension', '')),
+            ('Provider ID', quota.get('providerId', '')),
+            ('Scope', quota.get('scope', '')),
+            ('Target', quota.get('targetId', '')),
+            ('Limit', value('limit')),
+            ('Utilization', value('utilization')),
+            ('Holds', value('holds')),
+            ('Period', quota.get('period', '')),
```
rigidit: Can we change order to: Scope, Provider ID, Target, Dimension, Limit, Utilization, Holds, Period? I think it will look better for v2 workspaces.
But we didn't have this transform_workspace_quotas previously, how that worked for v1 workspaces?

src/quantum/azext_quantum/_params.py
```diff
@@ -184,7 +184,7 @@ def load_arguments(self, _):  # pylint: disable=too-many-locals
...
-    quota_type = CLIArgumentType(options_list=['--quota'], help='Target quota allocation as provider-id, target-id, standard-minutes-lifetime, and optional high-minutes-lifetime key=value pairs, a JSON object or array, or `@{file}` with JSON content. standard-minutes-lifetime is required for a new allocation. camelCase keys (providerId, targetId, ...) are also accepted. Repeat --quota once per target.', action=QuotaAction, nargs='+')
+    quota_type = CLIArgumentType(options_list=['--quota'], help='Final target quota allocation for a V2 workspace as provider-id, target-id, standard-minutes-lifetime, and optional high-minutes-lifetime key=value pairs, a JSON object or array, or `@{file}` with JSON content. Use --workspace-kind V2 when creating a workspace. Values are absolute and cannot exceed the suite target allocation or, when updating, be below current workspace usage. standard-minutes-lifetime is required for a new allocation. camelCase keys (providerId, targetId, ...) are also accepted. Repeat --quota once per target.', action=QuotaAction, nargs='+')
```
rigidit: 'Final' sounds not correct word here as we can always update target quota allocations any time right? let's just keep Target quota allocation ...

src/quantum/azext_quantum/_client_factory.py
```diff
@@ -85,6 +96,11 @@ def cf_quotas(cli_ctx, subscription: str, resource_group: str, ws_name: str, end
     return cf_quantum(cli_ctx, subscription, resource_group, ws_name, endpoint).services.quotas
 
 
+def cf_suite_offers_data_plane(cli_ctx, subscription: str, endpoint: str):
```
rigidit: consider requiring to supply location instead of endpoint, and constructing endpoint inside of the function here

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done! Validation now handles the three cases separately:
- no subscription allocation exists for the target
- The requested allocation is below current workspace usage
- The requested allocation exceeds the subscription allocation 
Each error now explains how to correct the value and includes the relevant az quantum workspace quotas or az quantum suite-offer quotas command. I verified all three messages manually against PPE

src/quantum/azext_quantum/_client_factory.py
v-elegacheva: That is a good point. The hardcoded eastus2euap override was added only for parity with the older v1 helper and was not required here. I removed it, so base_url_v2() now normalizes and uses the suplied location. The explicit AZURE_QUANTUM_BASEURL_V2 override remains available when a custom endpoint is required

src/quantum/azext_quantum/operations/suite_offers.py
v-elegacheva: I removed the CLI-side sorting, so suite target quota rows now preserve the order returned by the backend. I also added a test that verifies the backend allocation order is retained

src/quantum/azext_quantum/operations/suite_offers.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done! StandardMinutesLifetime and HighMinutesLifetime are now explicit dimension names, and the unused Priority import was removed

src/quantum/setup.py
v-elegacheva: Okay! I removed the explicit package_data entry for the private vendoed marker. Development type checking continues to use the source tree

src/quantum/azext_quantum/commands.py
v-elegacheva: Okay! Updated the table order to: Scope, Provider ID, Target, Dimension, Limit, Utilization, Holds, Period

Previously v1 workspace quotas had no dedicated table transformer and relied on the generic serialization field order. The transformer was introduced when v2 target rows were added so both v1 and v2 rows could share one stable shape. I verified the new order manually for both workspace versions.

src/quantum/azext_quantum/_params.py
v-elegacheva: Okay! The argument help now says "Target quota allocation" instead of "Final target quota allocation". I also changed the related public help wording from "final" to "requested" while keeping it clear that quota values are absolute

src/quantum/azext_quantum/_client_factory.py
v-elegacheva: Thank you! the cf_suite_offers_data_plnae() now requires the suite offer location and constructs the regional v2 endpoint internally. Both suite-offer quota and target operations now pass offer.properties.location

src/quantum/azext_quantum/operations/suite_offers.py
rigidit: agree to use const for now

src/quantum/azext_quantum/operations/suite_offers.py
```diff
@@ -14,3 +28,112 @@ def list_suite_offers(cmd):
...
+        (o for o in offers
+         if o.properties is not None
+         and o.properties.provider_id is not None
+         and o.properties.provider_id.lower() == provider_id.lower()),
+        None,
+    )
+    if offer is None:
+        raise InvalidArgumentValueError(
+            f"No suite offer was found for provider '{provider_id}' in subscription '{subscription_id}'."
+        )
```
rigidit: we could extract suite offer lookup into separate method - suite_offer_quotas and suite_offer_targets still carry the same ~12-line next(...) + if offer is None block verbatim.

src/quantum/azext_quantum/_help.py
```diff
@@ -233,6 +233,44 @@
             az quantum suite-offer list --query "[].{provider:properties.providerId, location:properties.location}" -o table
 """
 
+helps['quantum suite-offer quotas'] = """
+    type: command
+    short-summary: View quota allocations and their consumed usages for a suite offer in the current subscription.
+    long-summary: |
+        Returns the v2 quota allocations (limits) for each target of the suite offer together
+        with the consumed usages. Each entry reports the allocated and used standard and high priority
+        minutes. Missing usage values are returned as 0.
```
rigidit: should we mention that table view converts to hours?

src/quantum/HISTORY.rst
```diff
@@ -3,6 +3,18 @@
...
+++++++++++++++
+* Added the ``az quantum suite-offer target list`` command to list the targets, availability, and overall, Standard, and High average queue times available through a suite offer, without requiring a workspace.
+* Removed the redundant provider column from the ``az quantum suite-offer target list`` table output.
+* Updated ``az quantum suite-offer quotas`` to return ``0`` for missing Standard and High usage values.
+* Updated the ``az quantum workspace quotas`` command to include v2 target quota allocations and usages while preserving the existing response format for v1 providers.
+* Added always-on validation for V2 workspace target quota allocations on create and update, allowing requested Standard and High values between current workspace usage and suite target allocation, inclusive.
+
+1.0.0b26
+++++++++++++++
+* Added the ``az quantum suite-offer quotas`` command to view quota allocations merged with their consumed usages for a suite offer in the subscription.
```
rigidit: let's have only one version with all changes

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +269,120 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
...
 
+_TARGET_QUOTA_PRIORITIES = (
+    ("Standard", "standard_minutes_lifetime"),
+    ("High", "high_minutes_lifetime"),
+)
+_TARGET_QUOTA_USAGE_FIELDS = {
+    "standard_minutes_lifetime": "standardMinutesLifetime",
+    "high_minutes_lifetime": "highMinutesLifetime",
+}
+
```
rigidit: could you move this in contants section before class WorkspaceInfo?

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +269,120 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
...
+                f"Cannot validate --quota because no suite offer was found for provider '{provider.provider_id}'. "
+                "Run 'az quantum suite-offer list' to view available suite offers."
+            )
+        suite_target = next(
+            (item for item in suite_offer.properties.target_quotas or []
+             if item.target_id is not None and item.target_id.lower() == target_id),
+            None
+        )
+        for priority, attribute in _TARGET_QUOTA_PRIORITIES:
+            final_allocation = getattr(target_quota, attribute, None)
```
rigidit: rename to requested_allocation

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +269,120 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
         provider.target_quotas = target_quotas
 
 
+_TARGET_QUOTA_PRIORITIES = (
+    ("Standard", "standard_minutes_lifetime"),
+    ("High", "high_minutes_lifetime"),
+)
```
rigidit: priority part seems to be used only in error messages, let's skip it and only keep (standard_minutes_lifetime, high_minutes_lifetime)

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +269,120 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
...
+            raise InvalidArgumentValueError(
+                f"Cannot validate --quota because no suite offer was found for provider '{provider.provider_id}'. "
+                "Run 'az quantum suite-offer list' to view available suite offers."
+            )
+        suite_target = next(
+            (item for item in suite_offer.properties.target_quotas or []
+             if item.target_id is not None and item.target_id.lower() == target_id),
+            None
+        )
+        for priority, attribute in _TARGET_QUOTA_PRIORITIES:
```
rigidit: rename attribute to something more appropriate

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +269,120 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
...
+            try:
+                usages = usage_client.list_workspace_usages(
+                    info.subscription, info.resource_group, info.name, provider_id=provider.provider_id)
+            except AzureResourceNotFoundError:
+                usages = None
+            for usage in usages or []:
+                if usage.target_id is not None:
+                    usage_by_key[(provider_id, usage.target_id.lower())] = usage.usage
+
+    for provider_id, target_id in sorted(requested_keys):
```
rigidit: we have sorted this already above

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -444,12 +565,100 @@ def get(cmd, resource_group_name=None, workspace_name=None):
...
+
+    return _merge_workspace_quotas(workspace, usages, legacy_quotas)
+
+
+_WORKSPACE_QUOTA_SCOPE = DimensionScope.WORKSPACE.value
+_WORKSPACE_QUOTA_PERIOD = MeterPeriod.NONE.value
+_TARGET_QUOTA_DIMENSIONS = (
+    ("StandardMinutesLifetime", "standard_minutes_lifetime"),
+    ("HighMinutesLifetime", "high_minutes_lifetime"),
+)
```
rigidit: Move to const sections above class WorkspaceInfo

src/quantum/azext_quantum/_help.py
```diff
@@ -326,6 +364,9 @@
 helps['quantum workspace create'] = """
     type: command
     short-summary: Create a new Azure Quantum workspace.
+    long-summary: >-
+        Target quota values are absolute. For V2 workspaces, each requested Standard and High allocation is validated
+        against the provider's suite target allocation before the workspace is created.
```
rigidit: let's skip long summary for workspace create for now

src/quantum/azext_quantum/_help.py
```diff
@@ -400,6 +445,10 @@
 helps['quantum workspace update'] = """
     type: command
     short-summary: Update the given (or current) Azure Quantum workspace.
+    long-summary: >-
+        Target quota values are absolute. Each requested Standard and High allocation is validated against the current
+        workspace target usage and provider's suite target allocation, with equality allowed at both boundaries.
+        Priority values omitted from an existing target allocation are preserved and validated.
```
rigidit: let's skip long-summary for workspace update for now

src/quantum/azext_quantum/operations/suite_offers.py
v-elegacheva: Done! I extracted the case-insensitive lookup and not-found handling into _get_suite_offer(). Both suite_offer_quotas() and suite_offer_targets() now use the shared helper. I also added coverage for successful and missing-provider lookups

src/quantum/azext_quantum/_help.py
v-elegacheva: I added the mention. The help now states that the underlying values are reported in minutes and table output converts them to hours!

src/quantum/HISTORY.rst
v-elegacheva: All done! I consolidated all changes under 1.0.0b26

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done! Replaced _TARGET_QUOTA_PRIORITIES with _TARGET_QUOTA_ATTRIBUTES, containing only standard_minutes_lifetime and high_minutes_lifetime. The display name used in error messages is now derived from the attribute

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done! Renames it to quota_attribute in the quota validation logic

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done! requested_keys is sorted once when it is created and reused directly in both validation passes

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done!

src/quantum/azext_quantum/_help.py
v-elegacheva: Done!

src/quantum/azext_quantum/_help.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/suite_offers.py
```diff
@@ -5,7 +5,21 @@
...
+from azure.cli.core.commands.client_factory import get_subscription_id
+from azure.core.exceptions import ResourceNotFoundError as AzureResourceNotFoundError
+
+from .._client_factory import cf_suite_offers, cf_suite_offers_data_plane
+
+# Scope used for target quota rows returned by this command.
+_SUITE_OFFER_QUOTA_SCOPE = "SubscriptionTarget"
+
+
+def _quota_usage_value(usage, field):
```
rigidit: let's put private methods (that starts with '_') after public ones here

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -58,6 +60,21 @@
                                "https://azure.microsoft.com/support/legal/preview-supplemental-terms/\n\n" \
                                "Continue? (Y/N) "
 
+_TARGET_QUOTA_ATTRIBUTES = (
```
rigidit: we have 3 variables that contains standard_minutes_lifetime and high_minutes_lifetime in different forms, can we simplify this? maybe have only one variable, at least remove duplicated values

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +284,112 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
...
+        suite_target = next(
+            (item for item in suite_offer.properties.target_quotas or []
+             if item.target_id is not None and item.target_id.lower() == target_id),
+            None
+        )
+        for quota_attribute in _TARGET_QUOTA_ATTRIBUTES:
+            requested_allocation = getattr(target_quota, quota_attribute, None)
+            if requested_allocation is None:
+                continue
+            priority = quota_attribute.partition("_")[0].capitalize()
```
rigidit: not need to extract priority, standardMinutesLifetime/highMinutesLifetime is the quota type, and we can have other quota types in future not related to priority

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +284,112 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
...
+            if suite_target is None:
+                raise InvalidArgumentValueError(
+                    f"--quota requests {requested_allocation} minutes of {priority} time for provider "
```
rigidit: instead of `minutes of {priority} time' just name the type of the quota (standard minutes lifetime or high minutes lifetime)

rigidit: please update in all similar places

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -283,14 +406,18 @@ def create(cmd, resource_group_name, workspace_name, location, storage_account,
     if not info.resource_group:
         raise ResourceNotFoundError("Please run 'az quantum workspace set' first to select a default resource group.")
     quantum_workspace: QuantumWorkspace = _get_basic_quantum_workspace(location, info, storage_account)
+    workspace_kind_value = str(_enum_to_value(workspace_kind)).upper()
     if quota:
         _require_v2_workspace(workspace_kind)
+    if workspace_kind_value == 'V2':
```
rigidit: extract V1/V2 as const or reuse enum from autogenerated part if available

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -444,12 +572,92 @@ def get(cmd, resource_group_name=None, workspace_name=None):
...
+            cmd.cli_ctx, info.subscription, info.resource_group, info.name, endpoint)
+        for provider in providers or []:
+            try:
+                provider_usages = v2_client.list_workspace_usages(
+                    info.subscription, info.resource_group, info.name, provider_id=provider.provider_id)
+            except AzureResourceNotFoundError:
+                provider_usages = None
+            usages.extend(provider_usages or [])
+    else:
+        legacy_client = cf_quotas(
```
rigidit: client initialization code is the same, you can put initialization before if statement, and just name it client
rename legacy_quotas to v1_quotas

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -444,12 +572,92 @@ def get(cmd, resource_group_name=None, workspace_name=None):
...
+        "utilization": usage if usage is not None else 0,
+        "holds": 0.0,
+        "period": _WORKSPACE_QUOTA_PERIOD,
+        "targetId": target_id,
+    }
+
+
+def _merge_workspace_quotas(workspace, usages, legacy_quotas=None):
+    """
+    Preserve legacy quota rows and append one flat row per target and priority for v2 quotas.
```
rigidit: use v1 instead of legacy everywhere

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -444,12 +572,92 @@ def get(cmd, resource_group_name=None, workspace_name=None):
...
+        ((usage.provider_id or '').lower(), usage.target_id.lower()): usage
+        for usage in (usages or [])
+        if usage.target_id is not None
+    }
+
+    properties = workspace.properties
+    providers = properties.providers if properties is not None else None
+
+    rows = [row for row in (legacy_quotas or [])]
+    for provider in sorted(providers or [], key=lambda p: p.provider_id or ""):
```
rigidit: do we sort here only for predictability of output and tests?
if yes, let's align with decision in another thread that we just preserve order returned from backend

src/quantum/HISTORY.rst
```diff
@@ -3,6 +3,15 @@
 Release History
 ===============
 
+1.0.0b26
+++++++++++++++
+* Added the ``az quantum suite-offer quotas`` command to view quota allocations merged with their consumed usages for a suite offer in the subscription.
+* Added the ``az quantum suite-offer target list`` command to list the targets, availability, and overall, Standard, and High average queue times available through a suite offer, without requiring a workspace.
+* Removed the redundant provider column from the ``az quantum suite-offer target list`` table output.
```
rigidit: we can omit this as we just added suite-offer target list command

rigidit: same for 
Updated ``az quantum suite-offer quotas`` to return ``0`` for missing Standard and High usage values.

src/quantum/setup.py
```diff
@@ -49,5 +49,7 @@
     license='MIT',
     classifiers=CLASSIFIERS,
     packages=find_packages(),
-    package_data={'azext_quantum': ['azext_metadata.json', 'operations/templates/create-workspace-and-assign-role.json']},
+    package_data={
+        'azext_quantum': ['azext_metadata.json', 'operations/templates/create-workspace-and-assign-role.json'],
+    },
```
rigidit: we can revert to original package_data line

src/quantum/azext_quantum/operations/suite_offers.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done!

src/quantum/setup.py
v-elegacheva: Done!

src/quantum/HISTORY.rst
v-elegacheva: Done!

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done!

src/quantum/azext_quantum/operations/workspace.py
v-elegacheva: Done!

rigidit (approved):

yonzhan:
Quantum

v-elegacheva:
Hello @necusjz! Could you please review this PR when you get a chance? Thank you in advance!

src/quantum/azext_quantum/operations/suite_offers.py
```diff
@@ -14,3 +23,115 @@ def list_suite_offers(cmd):
...
+    subscription_id = get_subscription_id(cmd.cli_ctx)
+
+    # 1. Control-plane: locate the suite offer for the requested provider.
+    offer = _get_suite_offer(cmd, subscription_id, provider_id)
+    canonical_provider_id = offer.properties.provider_id
+
+    # 2. Data-plane (v2): fetch the consumed quota usages for that provider.
+    client = cf_suite_offers_data_plane(cmd.cli_ctx, subscription_id, offer.properties.location)
+    try:
+        usages = client.list_quota_usages(subscription_id, canonical_provider_id)
```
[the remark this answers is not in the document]
v-elegacheva: Fixed by materializing the generated ItemPaged inside the existing exception handler. Added regression coverage where the 404 is raised during iteration.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -267,6 +276,112 @@ def _apply_target_quotas(providers, quota, preserve_existing=False):
...
+
+    usage_by_key = {}
+    if include_usage:
+        usage_client = cf_quotas(
+            cmd.cli_ctx, info.subscription, info.resource_group, info.name, workspace.properties.endpoint_uri)
+        for provider_id in sorted({provider_id for provider_id, _ in requested_keys}):
+            provider = workspace_providers[provider_id]
+            try:
+                usages = usage_client.list_workspace_usages(
+                    info.subscription, info.resource_group, info.name, provider_id=provider.provider_id)
```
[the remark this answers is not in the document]
v-elegacheva: Fixed by materializing the generated ItemPaged inside the existing exception handler. Added regression coverage where the 404 is raised during iteration.

v-elegacheva:
Hello @necusjz ! Could you please review this PR when you have a chance? Thank you in advance!

necusjz (approved):

PR #10290, opened 2026-09-02
[MCC] Upgrade az mcc ent commands to the 2026-06-01 API version

dcwell:

| Breaking Changes |
| :--: |
| ⚠️ None |

<details>

><details>
> <summary>⚠️mcc</summary>
>
>|---|---|---|---|
>|⚠️ [1010 - ParaPropUpdate](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1010.md)|mcc ent node update|cmd `mcc ent node update` update parameter `auto_update_ring`: updated property `choices` from `['Fast', 'Preview', 'Slow']` to `['Beta', 'Stable']`||
>|⚠️ [1010 - ParaPropUpdate](https://github.com/Azure/azure-cli/blob/dev/doc/breaking_change_rules/1010.md)|mcc ent node update|cmd `mcc ent node update` update parameter `proxy`: updated property `choices` from `['Disabled', 'Enabled', 'None', 'Required']` to `['Disabled', 'Enabled']`||
>
></details>
</details>

---

### Related command

`az mcc ent node create|update|delete|show|list|get-deployment-details`
`az mcc ent resource create|delete|list|wait`

### General Guidelines

### About Extension Publish

---

## Summary

Upgrades every `az mcc ent` command from `2024-11-30-preview` to the `2026-06-01`
GA API version, and updates the customization layer to match the new API.

Command models: https://github.com/Azure/aaz/pull/1079

`mcc` `1.0.0b3` -> `1.0.0b4`.

## Breaking changes

Both are intentional and both are called out in `HISTORY.rst`.

**1. `--auto-update-ring` now accepts only `Stable` and `Beta`.**

`2026-06-01` renames the customer selectable update rings. Per the swagger, `Slow`,
`Fast` and `Preview` are "Legacy update ring retained for backward compatibility.
This value was used prior to 2026-06-01." `Stable` replaces `Slow` and `Beta`
replaces `Fast`. MCC is standardising on the new rings, so the legacy values are no
longer offered.

The service still *stores* the legacy value: sending `Stable` reads back as `Slow`,
and sending `Beta` reads back as `Fast`. Both pairings were confirmed against the
live service. The extension translates on output so a customer sees the same
vocabulary they typed.

**2. `--proxy` now accepts only `Enabled` and `Disabled`.**

This one restores behaviour rather than removing it. The generated enum carries the
service values (`None`/`Required`) from the `ProxyRequired` swagger enum, while
`pre_operations` has always rejected exactly those two values and told the user to
pass `Enabled`/`Disabled` instead. In `1.0.0b3` the generated enum happened to carry
all four values, so it worked. After regenerating against `2026-06-01` only the
service values remained, which left **every** value of `--proxy` failing — two at
argument parsing, two in validation. The extension now offers only the two customer
facing values and translates them to `Required`/`None` before the request is sent.

## Fixes

- **`--proxy` was completely unusable** after the API upgrade, as described above.
- **Proxy edits on a node that already had a proxy configured.** The instance state
  was compared against `Enabled`, but the service reports `Required`, so that branch
  was unreachable. Changing only `--proxy-host` or only `--proxy-port` failed, and
  disabling the proxy left the previous `proxyUrlConfiguration` on the resource.
- **Update ring scheduling.** The rule was written as `if ring == "Fast" ... else
  assume Slow`, which was true when only three rings existed. With five values, three
  fell through the `else`, so a node on `Stable`/`Beta`/`Preview` could be moved to a
  scheduled ring without ever being asked for `--auto-update-day/week/time`. The check
  is now written against the target ring: moving to `Stable` requires all three,
  moving to `Beta` rejects all three because that schedule is managed by Microsoft.
- **`--auto-update-day` accepted `0`.** The help said 1-7 and the swagger allows 0-7.
  It is now validated to 1-7, matching the existing `--auto-update-week` guard.
- **`show` and `list` leaked service vocabulary.** Proxy state and update ring are now
  reported with the same values the commands accept as input.

## Other changes

- Hid `--bgp-network-interface`, `--runtime-account-type` and the four
  `--open-firewall-port*` arguments. These arrived with `2026-06-01`, are reserved for
  an upcoming feature, and are not ready for customer use. They were only exposed on
  `update`; on `create` they sit under `cache_node`, which is already hidden.
- Corrected help for `--auto-update-ring`, `--auto-update-day`, `--auto-update-week`,
  `--auto-update-time`, `--cache-drive` and `--proxy` so it states the behaviour the
  extension actually enforces.
- Added examples to all nine registered commands. The previous examples lived only in
  the generated Python, so they were lost on the first regeneration; they are now part
  of the command model and survive future regenerations.

## Testing

`test_mcc` was re-recorded live against the `2026-06-01` API and passes in both live
and playback mode. It also now asserts that the proxy state round-trips as `Enabled`
rather than `Required`.

- `azdev test mcc` — passed (live re-record, then playback)
- `azdev style mcc` — pylint and flake8 passed
- `azdev linter --include-whl-extensions mcc` — 21/21 rules passed
- `python scripts/ci/test_index.py -q` — OK


yonzhan:
MCC

necusjz (approved):

PR #10292, opened 2026-09-03
[MCC] Fix validation errors that pointed users at a nonexistent argument

dcwell:

| Breaking Changes |
| :--: |
| ️✔️ None |

### Related command
`az mcc ent node update`, `az mcc ent node create`

### Description
Follow-up to #10290. Two user-facing error messages in the `mcc` extension pointed at the wrong thing. Both were found by an edge-case sweep of every validation path in the extension, run after #10290 merged.

#### 1. Proxy errors named an argument that does not exist

Eight proxy validation messages on `az mcc ent node update` told the user to set `--enable-proxy`:

> ValidationError: Parameter --enable-proxy is set to "Enabled", must provide --proxy-host and --proxy-port parameter.

`enable_proxy` is deliberately unregistered (`args_schema.enable_proxy._registered = False`); the argument the command exposes is `--proxy`. Following the guidance in the message produced:

> ERROR: unrecognized arguments: --enable-proxy Enabled

so the message led to a dead end. All eight now name `--proxy`. The internal argument name is unchanged.

#### 2. Create reported the cache node when it could not find the MCC resource

`MccEntNodeCreate.pre_operations` looks up the **MCC resource** to fill in `location`, but the failure message reported `args.cache_node_name`. A bare `except:` also discarded the underlying error, so a mistyped resource group surfaced as a missing MCC resource.

Before:
> ERROR: Cache node resource creation failed. CLI could not find the specified MCC resource with name 'MyCacheNode' that the cache node would be created on.

After:
> ERROR: Cache node resource creation failed. CLI could not find the MCC resource 'MyMccResource' in resource group 'MyResourceGroup' that the cache node would be created on. (ResourceGroupNotFound) Resource group 'MyResourceGroup' could not be found.

Both bare `except:` clauses are now `except Exception as ex` with `raise ... from ex`.

### Notes
- No behaviour change to validation itself — only the text of the errors and the preservation of the underlying exception.
- `1.0.0b4` has not been published to `index.json` yet, so these entries fold into the existing `1.0.0b4` block rather than adding a version.

### Testing
- `azdev linter --include-whl-extensions mcc` — PASSED
- `azdev style mcc` — pylint and flake8 PASSED
- `azdev test mcc` — 1 passed
- 105-command sweep (69 offline, 28 live against real Azure resources, 8 raw REST comparisons) re-run after the change.


yonzhan:
MCC

necusjz (approved):

PR #10329, opened 2026-09-11
Quantum: Don't use endpoint cache when workspace doesn't match

pavelbraginskiy:

| Breaking Changes |
| :--: |
| ️✔️ None |

Currently setting a default quantum workspace caches the workspace's endpoint, but then that endpoint is still used for all other workspaces, preventing them from being accessed.

This PR saves the workspace information together with the endpoint so that the cached endpoint value is not used for any other workspace.

---

### Related command

`az quantum workspace set`

### General Guidelines

### About Extension Publish

### DIFF
--- a/src/quantum/HISTORY.rst
+++ b/src/quantum/HISTORY.rst
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+1.0.0b26
+++++++++++
+* Scoped the workspace endpoint cache to its subscription, resource group, workspace, cloud, and tenant so it is not reused for other workspaces.
+* Moved the cache to ``quantum.workspace_endpoint_cache``. Re-run ``az quantum workspace set`` to populate the new cache.
+* Clarified the persistent flag defaults and endpoint caching performed by ``az quantum workspace set``. Automatic stale-endpoint invalidation is not included.
+
 1.0.0b25
 ++++++++++++++
 * Added the ``az quantum suite-offer list`` command to list the suite offers available to the subscription, including provider, location, and subscription-level quota allocations.
--- a/src/quantum/README.rst
+++ b/src/quantum/README.rst
@@ -87,23 +87,33 @@ Prepare to submit and manage jobs in Azure Quantum using the `az quantum` extens
 
 
 4. You can use `quantum workspace set` to select a default workspace you want to use to list and submit jobs.
-   Note that you also need to specify the resource group. If you set a default workspace by providing a resource group,
-   workspace name and location, you don't need to include those parameters in commands #5 to #8 below.
-   Anternatively, you can include them in each call.
+   This saves the resource group and workspace name as persistent Azure CLI flag defaults. The resource-group default
+   also applies to other Azure CLI commands. The default subscription and location are not changed. Alternatively,
+   you can specify the resource group and workspace name in each call without setting defaults.
 
    .. code-block::
 
-      az quantum workspace set -g MyResourceGroup -w MyWorkspace -l MyLocation -o table
+      az quantum workspace set -g MyResourceGroup -w MyWorkspace -o table
 
       Location     Name                               ResourceGroup
       -----------  ---------------------------------  --------------------------------
       westus       ws-yyyyyy                          rg-yyyyyyyyy
 
 
-.. note:
+.. note::
    Commands below assume that a default workspace has been set. If you prefer to specify it
    for each call, include the following parameters with commands below:
-   `-g MyResourceGroup -w MyWorkspace -l MyLocation`
+   `-g MyResourceGroup -w MyWorkspace`.
+
+   Explicit flags override their respective defaults independently. For example, specifying only `-w OtherWorkspace`
+   uses the saved resource group. A one-off job read does not change the saved defaults:
+
+   .. code-block::
+
+      az quantum job show --subscription OtherSubscription -g OtherGroup -w OtherWorkspace -j MyJobId
+
+   Use `az quantum workspace clear` to clear the resource-group and workspace-name defaults without changing
+   other settings.
 
 
 5. You can check the current default workspace with command `az quantum workspace show`.
--- a/src/quantum/azext_quantum/_help.py
+++ b/src/quantum/azext_quantum/_help.py
@@ -317,6 +317,8 @@ helps['quantum workspace'] = """
 helps['quantum workspace clear'] = """
     type: command
     short-summary: Clear the default Azure Quantum workspace.
+    long-summary: |
+        Clear the saved resource-group and workspace-name defaults and the workspace's cached endpoint. Other settings, including the default subscription, location, and target, are unchanged.
     examples:
       - name: Clear the default Azure Quantum workspace if previously set.
         text: |-
@@ -379,6 +381,12 @@ helps['quantum workspace quotas'] = """
 helps['quantum workspace set'] = """
     type: command
     short-summary: Select a default Azure Quantum workspace for future commands.
+    long-summary: |
+        Save the resource group and workspace name as persistent Azure CLI flag defaults, and save the workspace's data-plane endpoint in quantum.workspace_endpoint_cache. The resource-group default is shared with other Azure CLI commands. These settings persist across terminal sessions.
+
+        Explicit --resource-group and --workspace-name arguments override their respective defaults independently. The endpoint cache is reused only when the resolved subscription, resource group, workspace name, cloud, and tenant match the saved identity. Data-plane commands targeting another workspace discover its endpoint through Azure Resource Manager without changing the saved defaults.
+
+        This command does not change the default subscription or location. Cached endpoints are not automatically refreshed; run this command again to refresh the saved endpoint. Use 'az quantum workspace clear' to clear the saved flag defaults and Quantum endpoint cache.
     examples:
       - name: Set the default Azure Quantum workspace.
         text: |-
--- a/src/quantum/azext_quantum/operations/workspace.py
+++ b/src/quantum/azext_quantum/operations/workspace.py
@@ -60,6 +60,8 @@ C4A_TERMS_ACCEPTANCE_MESSAGE = "\nBy continuing you accept the Azure Quantum ter
 
 
 class WorkspaceInfo:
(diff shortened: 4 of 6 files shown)


yonzhan:
quantum

rigidit:
add Quantum prefix to PR title

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -84,7 +86,35 @@ def select_value(key, value):
         self.subscription = get_subscription_id(cmd.cli_ctx)
         self.resource_group = select_value('group', resource_group_name)
         self.name = select_value('workspace', workspace_name)
-        self.endpoint = select_value('endpoint', endpoint)
+        self.endpoint = endpoint if endpoint is not None else self._get_cached_endpoint(cmd)
+
+    def _endpoint_cache_identity(self, cmd):
```
rigidit, line 91: Let's skip:
- `arm_endpoint` — near-redundant. Subscription GUIDs don't repeat across clouds, so resource_id already discriminates. If you want cloud safety anyway, it's free (no Profile needed).
- `tenant_id` — unnecessary and the most expensive part. It forces Profile(cli_ctx).get_subscription(...) on every WorkspaceInfo construction, i.e. every quantum data-plane command, and it can raise if the subscription isn't in the local profile. A tenant move doesn't change a workspace's endpointUri, and auth is re-negotiated regardless — it guards nothing.
- `version` + JSON + isinstance checks — redundant. Any format change makes the stored value fail to match, which is already a cache miss and an ARM fallback. The not all((subscription, resource_group, name)) guard is also unneeded: a partially-resolved id can never equal a saved one.

Consider to implement simpler equivalent:
```
class WorkspaceInfo:
    _ENDPOINT_CACHE_KEY = 'workspace_endpoint_cache'

    def __init__(self, cmd, resource_group_name=None, workspace_name=None, endpoint=None):
        ...
        self.endpoint = endpoint if endpoint is not None else self._get_cached_endpoint(cmd)

    def _endpoint_cache_key(self):
        return _get_workspace_resource_id(self).lower()

    def _get_cached_endpoint(self, cmd):
        cached = cmd.cli_ctx.config.get('quantum', self._ENDPOINT_CACHE_KEY, None)
        if not cached:
            return None
        # "<workspace resource id> <endpoint>" - ignore the endpoint cached for a different workspace.
        resource_id, _, cached_endpoint = cached.partition(' ')
        return cached_endpoint if resource_id == self._endpoint_cache_key() and cached_endpoint else None

    def save(self, cmd, endpoint=''):
        from azure.cli.core.util import ConfiguredDefaultSetter
        with ConfiguredDefaultSetter(cmd.cli_ctx.config, False):
            cmd.cli_ctx.config.set_value(cmd.cli_ctx.config.defaults_section_name, 'group', self.resource_group)
            cmd.cli_ctx.config.set_value(cmd.cli_ctx.config.defaults_section_name, 'workspace', self.name)
            if endpoint:
                cmd.cli_ctx.config.set_value('quantum', self._ENDPOINT_CACHE_KEY,
                                             f"{self._endpoint_cache_key()} {endpoint}")
            else:
                cmd.cli_ctx.config.remove_option('quantum', self._ENDPOINT_CACHE_KEY)
```

rigidit:
Could you check if other places still hijacks defaults? For example, workspace update calls info.save(cmd, ws.properties.endpoint_uri), so az quantum workspace update -g other -w other silently repoints the user's saved group/workspace and the new cache at another workspace. Pre-existing, but it directly contradicts this PR's premise; either scope it to the case where no explicit -g/-w was given, or only refresh the cache when the identity already matches.

src/quantum/azext_quantum/tests/latest/test_workspace_endpoint_cache.py
```diff
@@ -0,0 +1,213 @@
+import json
```
rigidit, line 1: missing header

src/quantum/README.rst
```diff
@@ -87,23 +87,33 @@ Prepare to submit and manage jobs in Azure Quantum using the `az quantum` extens
 
 
 4. You can use `quantum workspace set` to select a default workspace you want to use to list and submit jobs.
-   Note that you also need to specify the resource group. If you set a default workspace by providing a resource group,
-   workspace name and location, you don't need to include those parameters in commands #5 to #8 below.
-   Anternatively, you can include them in each call.
+   This saves the resource group and workspace name as persistent Azure CLI flag defaults. The resource-group default
```
rigidit, line 90: this seems to be unrelated to the fix?

src/quantum/azext_quantum/_help.py
```diff
@@ -421,6 +423,12 @@
 helps['quantum workspace set'] = """
     type: command
     short-summary: Select a default Azure Quantum workspace for future commands.
+    long-summary: |
+        Save the resource group and workspace name as persistent Azure CLI flag defaults, and save the workspace's data-plane endpoint in quantum.workspace_endpoint_cache. The resource-group default is shared with other Azure CLI commands. These settings persist across terminal sessions.
```
rigidit, line 427: we should not include internal implementation detail into help to end users

src/quantum/HISTORY.rst
```diff
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+1.0.0b27
+++++++++++
+* Scoped the workspace endpoint cache to its subscription, resource group, workspace, cloud, and tenant so it is not reused for other workspaces.
```
rigidit, line 8: we should not include internal implementation details into history - customers won't parse that

rigidit (changes requested):

src/quantum/README.rst
rigidit: decided can be useful for additional user information

yonzhan:
Please fix CI issues

src/quantum/azext_quantum/operations/workspace.py
pavelbraginskiy: I implemented a simpler scheme that just records the resource id and endpoint. I kept it json, it's not any more or less complicated but I thought it would be less fragile to use a well-defined format.

src/quantum/HISTORY.rst
pavelbraginskiy: Fixed

src/quantum/azext_quantum/tests/latest/test_workspace_endpoint_cache.py
pavelbraginskiy: Added header

pavelbraginskiy:
> Could you check if other places still hijacks defaults? For example, workspace update calls info.save(cmd, ws.properties.endpoint_uri), so az quantum workspace update -g other -w other silently repoints the user's saved group/workspace and the new cache at another workspace. Pre-existing, but it directly contradicts this PR's premise; either scope it to the case where no explicit -g/-w was given, or only refresh the cache when the identity already matches.

I looked at `update` and `delete`. 

`update` was always setting the workspace updated as the set default workspace. This wasn't documented anywhere and is quite unintuitive, so it no longer affects the set workspace. 

`delete` tries to clear the set workspace if it's the one that was deleted. I improved this check to make sure that it would clear the correct workspace, previously it could get it wrong if the deleted workspace had the same name as an existing one but in a different subscription, or if the default workspace was overridden temporarily by the environment.

src/quantum/azext_quantum/operations/workspace.py
```diff
@@ -88,7 +90,51 @@ def select_value(key, value):
...
+        if (config.get(defaults_section, 'group', fallback='').lower() != self.resource_group.lower() or
+                config.get(defaults_section, 'workspace', fallback='').lower() != self.name.lower()):
+            return False
+        # Group/name defaults do not identify a subscription. The cached resource ID
+        # records the full identity selected by workspace set.
+        try:
+            cache = json.loads(config.get('quantum', self._ENDPOINT_CACHE_KEY, fallback='{}'))
+        except ValueError:
+            return False
+        return isinstance(cache, dict) and cache.get('resource_id') == self._normalized_resource_id()
```
rigidit: `is_global_default` (`workspace.py:111-137`) is now the heaviest part of the diff: ~27 lines, a raw `ConfigParser`, and a dependency on knack internals (`config.config_path`, `CONFIG_FILE_ENCODING`) to deliberately bypass the CLI's own config resolution. It buys two things:

1. **Subscription disambiguation** — `WorkspaceInfo(cmd).subscription` and `info.subscription` are always the same value inside one invocation, so `main`'s group+name comparison could wipe defaults when deleting a same-named workspace under `--subscription other`. Real, though narrow.
2. **Ignoring env/local-config overrides** so `AZURE_DEFAULTS_GROUP` can't cause the global file to be rewritten.

Point 1 doesn't need the ConfigParser at all — `self.endpoint` is *already* "the cached resource ID equals mine", which pins the subscription:

```python
client.begin_delete(info.resource_group, info.name, polling=False)
default_ws = WorkspaceInfo(cmd)
# endpoint resolves only when the cached identity matches, which pins the subscription too.
if (default_ws.resource_group, default_ws.name) == (info.resource_group, info.name) and info.endpoint:
    clear(cmd)
```

Point 2 is where I'd push back on the cost/benefit:

- It's inconsistent with the rest of the module. `az quantum workspace clear` under the same env overrides still wipes the *global* `group`/`workspace` unconditionally, so `clear` and `delete` now disagree about what "the default workspace" is. Either both should read persisted-only state or neither should.
- Elsewhere in the CLI, `AZURE_DEFAULTS_*` *is* the effective default; treating it as invisible is a surprise.
- It costs ~20 production lines plus the two most intricate tests in the file (`test_delete_environment_selected_workspace_preserves_global_settings`, `test_delete_uses_global_identity_with_local_config_enabled`, which has to build a merged local/global `CLIConfig` and patch `os.getcwd`).

My recommendation: drop `is_global_default` for the three-line version above, and if the env-override behavior is genuinely wanted, raise it as a separate change that applies to `clear` too.

pavelbraginskiy: Environment overrides are no longer handled.

rigidit (approved):

necusjz (approved):

PR #10350, opened 2026-09-16
Quantum: e2e test coverage

v-elegacheva:

| Breaking Changes |
| :--: |
| ️✔️ None |

## Summary

- Add live E2E coverage for V2 workspace quota create/update round trips and job file list/download.
- Strengthen suite-offer tests with non-empty assertions and dynamic provider/target discovery.
- Avoid hardcoded V2 provider and target identifiers in test source and recordings.
- Reliably clean up temporary Azure resources and CLI workspace defaults.
- Document which suite-offer and workspace quota commands apply to V2 providers and workspaces.
- Remove the obsolete long-name helper after migrating tests to create_random_name.
- Add offline coverage for discovery, quota polling, and cleanup behavior.

## Testing

- `python -m pytest src/quantum/azext_quantum/tests`
  - 113 passed
  - 14 skipped
  - 18 subtests passed
- `azdev style quantum`
- `azdev linter quantum`
- Live suite-offer scenarios: 3 passed
- Live V2 workspace quota create/update round trip: passed
- Live job file list/download round trip: passed


src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
```diff
@@ -115,13 +410,27 @@ def test_workspace_create_destroy(self):
...
+            def cleanup_workspaces():
+                cleanup_commands = [
+                    ('temporary workspace',
+                     f'az quantum workspace delete -g {test_resource_group} -w {workspace_name}')
+                    for workspace_name in reversed(workspaces_to_cleanup)
+                ]
+                cleanup_commands.append(('workspace defaults', 'az quantum workspace clear'))
+                run_cleanup_commands(self, cleanup_commands)
+
+            self.addCleanup(cleanup_workspaces)
```
rigidit, line 448: addCleanup silently defeats run_cleanup_commands' failure-preservation guard
utils.py detects an in-flight failure via sys.exc_info(). That works in a finally block, but unittest runs addCleanup callbacks after the test's exception has been captured by the outcome handler — I verified sys.exc_info() is (None, None, None) there. So in test_workspace_create_destroy, test_failed is always False, and a failing cleanup workspace delete re-raises and adds a second, misleading error on top of the real failure. Every other call site uses try/finally and works correctly. Either use try/finally here too, or make the flag an explicit parameter rather than inferring it.

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
```diff
@@ -194,9 +516,11 @@ def test_workspace_create_destroy(self):
                 self.check("name", test_workspace_temp),
                 self.check("properties.provisioningState", "Deleting")
             ])
+            workspaces_to_cleanup.remove(test_workspace_temp)
 
             # Create a workspace with a maximum length name, but make sure the deployment name was truncated to a valid length
-            test_workspace_temp = get_test_workspace_random_long_name()
+            test_workspace_temp = self.create_random_name(prefix='e2e-test-w', length=52)
```
rigidit: can you check if we lost a character here? 
get_test_workspace_random_long_name() seems produced exactly 53 chars (17 + 36); create_random_name(prefix='e2e-test-w', length=52) produces 52. That case exists specifically to test a maximum length name — use length=53.

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
```diff
@@ -76,8 +130,249 @@ def __init__(self, id, properties):
...
+        raise last_error
+
+    def test_workspace_v2_create_destroy(self):
+        test_location = get_test_workspace_location()
+        test_resource_group = get_test_resource_group()
+        test_storage_account = get_test_workspace_storage()
+        provider_id, target_id = self._get_v2_provider_and_target(test_location)
+        test_provider_sku_list = f'{provider_id}/default'
+        create_quota = f'provider-id={provider_id} target-id={target_id} standard-minutes-lifetime=0'
+        update_quota = f'provider-id={provider_id} target-id={target_id} standard-minutes-lifetime=1'
```
rigidit: standard-minutes-lifetime=0 makes the create half of the round trip near-vacuous. _target_quota_row emits "limit": allocation if allocation is not None else 0 — a limit of 0 is indistinguishable from "no allocation". Creating with 1 and updating to 2 would make both halves prove the value was actually applied

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
```diff
@@ -76,8 +130,249 @@ def __init__(self, id, properties):
...
+                ).get_output_in_json()
+                self._assert_workspace_quota_limit(
+                    quotas, provider_id, target_id, expected_limit
+                )
+                return
+            except AssertionError as error:
+                last_error = error
+                if attempt < attempts - 1:
+                    time.sleep(5)
+        raise last_error
```
rigidit, line 175: we may want to increase wait time, as I remeber there is a cache in backend that caches workspace resource, can you look what is the time?
I would recommend to increase total time to 1 minute, e.g. 6 iterations with 11 sec sleep

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
rigidit: and get_test_workspace_random_long_name now is dead, we can remove it

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
```diff
@@ -76,8 +130,249 @@ def __init__(self, id, properties):
...
+
+    def test_workspace_v2_create_destroy(self):
```
rigidit, line 177: shouldn't we mark tests that supposed to run in e2e pipeline with @live_only()?

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
v-elegacheva: Got it! `run_cleanup_commands` now accepts an explicit `test_failed` value, and `test_workspace_create_destroy` tracks whether the test completed successfully before its `addCleanup` callback runs. I also added unit coverage for both explicit failure states

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
v-elegacheva: Done! Restored the maximum-length test case to 53 characters

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
v-elegacheva: Done! Updated the round trip to create with 1 minute and update to 2 minutes. Dynamic discovery now requires a subscription allocation of at least 2 minutes, so both assertions prove that the requested values were applied.

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
v-elegacheva: Done! Increased polling to six attempts with 11-second intervals, providing approximately one minute for workspace quota cache propagation. The retry tests were updated accordingly

src/quantum/azext_quantum/tests/latest/test_quantum_workspace.py
v-elegacheva: This test inherits from `LiveScenarioTest`, which is itself decorated with `@live_only()`. Therefore all methods in this class are skipped unless `AZURE_TEST_RUN_LIVE` is enabled, while also bypassing VCR recording. I thought adding the decorator to the method would be redundant...

yonzhan:
Quantum

